PR #5, opened 2021-04-23
Navbar

tk-vincit:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[tailwind.config.js, line 7 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYy_eH2AaET_L2WoBb_:-MYy_eH31_OUCcO7iZrB:b-zcb679) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/tailwind.config.js#L7)):*
> ```JavaScript
>     extend: {
>       colors: {
>         brand: '#007AC9',
> ```

let's rather use the official naming of the colors: https://tyyliopas.hsl.fi/d/h8JR9dHeqfgd/braendi#/visuaalinen-ilme/vaerit
let's also put a link here to this site as a comment to remember the source
___
*[src/components/Navbar.tsx, line 5 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYycwymAulovcHIjVyN:-MYycwyn9N5b5oSns4pC:bwne5xe) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Navbar.tsx#L5)):*
> ```XML
> import { Link, NavLink } from 'react-router-dom';
> 
> export const Navbar: FunctionComponent = () => {
> ```

well, if the link is called NavLink with a big L, maybe this could be called NavBar with a big B :D but not this is def not the most important thing in the world...
___
*[src/components/Router.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYyb8ZN3_Y1724IBHyf:-MYyb8ZN3_Y1724IBHyg:bxwc200) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Router.tsx#L13)):*
> ```XML
>   return (
>     <BrowserRouter>
>       <Navbar />
> ```

I'd rather define the `routes` array in this component. Then, `Router` could pass it on to `Navbar` as a parameter that could draw the navlinks and also use it itself to (partially) define the `Route` mappings here below

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @pallost)

___
*[tailwind.config.js, line 7 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYy_eH2AaET_L2WoBb_:-MZBhlOV1FHEfRADyFYd:b-ly2ar9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/tailwind.config.js#L7)):*

let's rather use the official naming of the colors: https://tyyliopas.hsl.fi/d/h8JR9dHeqfgd/braendi#/visuaalinen-ilme/vaerit
let's also put a link here to this site as a comment to remember the source
</blockquote></details>

So do you want names like `hsl-sininen`, `hsl-raitiovaunun-vihreä` or `pikaratikka-turkoosi`?  :raised_eyebrow:   (Ok, don't know what would be good names for the latter ones but luckily those weren't usedyet.)

I'd also like to mention that behind that url was said that `HSL sininen on pääbrändiväri`, and according to that `brand` or `braind-primary` or `main` would be good names for this color...

But I can add link to the style guide here.
___
*[src/components/Navbar.tsx, line 5 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYycwymAulovcHIjVyN:-MZBj5Rj-0EOuVhpyg-t:bhblcu8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Navbar.tsx#L5)):*

well, if the link is called NavLink with a big L, maybe this could be called NavBar with a big B :D but not this is def not the most important thing in the world...
</blockquote></details>

`NavLink` comes from external Library, and I guess they named it like that because they also have `Link`.

I'll stick to this one.
___
*[src/components/Router.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYyb8ZN3_Y1724IBHyf:-MZBjYOx39h1Qyy8Ex7c:b-booc2a) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Router.tsx#L13)):*

I'd rather define the `routes` array in this component. Then, `Router` could pass it on to `Navbar` as a parameter that could draw the navlinks and also use it itself to (partially) define the `Route` mappings here below
</blockquote></details>

I kind of understand idea behind that, but when thinking a little bit more about it, I think that current solution is still better.

- If we define navbar's route listing here, we leak navbars responsibilies to router
- If we define navbar's routes here, we have to put ui strings in router. 
- What happens when we add routes like `/stops/:id`here? Clearly that kind of route won't go to Navbar. Should we then have `navbarRoutes` and `anotherRoutes` arrays here..?

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @pallost)

___
*[tailwind.config.js, line 7 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYy_eH2AaET_L2WoBb_:-MZBlJsmBJF2VKVq-uVf:b-ps04h) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/tailwind.config.js#L7)):*

So do you want names like `hsl-sininen`, `hsl-raitiovaunun-vihreä` or `pikaratikka-turkoosi`?  :raised_eyebrow:   (Ok, don't know what would be good names for the latter ones but luckily those weren't usedyet.)

I'd also like to mention that behind that url was said that `HSL sininen on pääbrändiväri`, and according to that `brand` or `braind-primary` or `main` would be good names for this color...

But I can add link to the style guide here.
</blockquote></details>

(but not sure even if link to style guide really belongs here or not. For example we already have `brand-darker` color definition here, but it's not listed in style guide. So is it a bit confusing to link to the style guide which covers only small subset of colors?

lassemaatta (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @pallost)

___
*[src/components/Router.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYyb8ZN3_Y1724IBHyf:-MZBzboz93MykMTFxQ6f:b-93gjxl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Router.tsx#L13)):*

I kind of understand idea behind that, but when thinking a little bit more about it, I think that current solution is still better.

- If we define navbar's route listing here, we leak navbars responsibilies to router
- If we define navbar's routes here, we have to put ui strings in router. 
- What happens when we add routes like `/stops/:id`here? Clearly that kind of route won't go to Navbar. Should we then have `navbarRoutes` and `anotherRoutes` arrays here..? 

</blockquote></details>

I'd move the whole route configuration to a separate place as data and make the `Navbar` and `Router` just parse the data. For example, we could specify which routes are included in the navbar (perhaps iterate the routes in order so the navbar link order is implicit).  Also, I think the `react-router-dom` supports specifying the view as component (e.g. https://reactrouter.com/web/example/route-config ). I'd rather not see `path="/some/random/path"` in multiple places in the code.

pallost (changes requested):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @lassemaatta, @pallost, and @tk-vincit)

___
*[tailwind.config.js, line 7 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYy_eH2AaET_L2WoBb_:-MZCcVtG6fUCAzmFR-gO:b8x5tpv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/tailwind.config.js#L7)):*

(but not sure even if link to style guide really belongs here or not. For example we already have `brand-darker` color definition here, but it's not listed in style guide. So is it a bit confusing to link to the style guide which covers only small subset of colors?
</blockquote></details>

ok, I did not notice that... well, let's leave this for now and return to it later when having more colors
___
*[src/components/Router.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYyb8ZN3_Y1724IBHyf:-MZCclMjBo0IsyoXDRjm:b-wy0a7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Router.tsx#L13)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

I'd move the whole route configuration to a separate place as data and make the `Navbar` and `Router` just parse the data. For example, we could specify which routes are included in the navbar (perhaps iterate the routes in order so the navbar link order is implicit).  Also, I think the `react-router-dom` supports specifying the view as component (e.g. https://reactrouter.com/web/example/route-config ). I'd rather not see `path="/some/random/path"` in multiple places in the code.
</blockquote></details>

I'm with Lasse :) That's what you get if someone with clojure background reviews javascript code :P

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[src/components/Router.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYyb8ZN3_Y1724IBHyf:-MZMHJRIDtKp0XP27HBY:b-rvlg34) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Router.tsx#L13)):*

I'm with Lasse :) That's what you get if someone with clojure background reviews javascript code :P
</blockquote></details>

I'm still not with this idea, or then I misundestrood something.
If we wouldn't have dynamic routes, e.g. `/stops/:id`, when things would be easy.
Then we could define e.g. enum with route names and import it where needed.

But e.g. the way from react-router docs:
```
const routes = [
  {
    path: "/sandwiches",
    component: Sandwiches
  },
  {
    path: "/tacos",
    component: Tacos,
    routes: [
      {
        path: "/tacos/bus",
        component: Bus
      },
      {
        path: "/tacos/cart",
        component: Cart
      }
    ]
  }
];
...
        <Switch>
          {routes.map((route, i) => (
            <RouteWithSubRoutes key={i} {...route} />
          ))}
        </Switch>
```
isn't very useful IMO.
Importing that kind of route array to e.g. navbar doesn't sound very elegant, as then we would have to expose which component each route is linked to, which isnt' something that navbar has to be aware of.
Also defining routes in an array instead of as nodes in Router component (as done right now) won't give any benefits here, it's just different syntax for defining exactly same thing, or in fact its inferior in this case because e.g. defining `exact` prop would be more verbose! And we haven't even gone to problems with dynamic routes yet... I agree that in theory it would indeed be nice to have routes defined only in one place, but in practice that easily turns into over-engineered solution which is less elegant as this kind of "basic" approach.

Did I misunderstood something? Do you have more practical tips or ideas about this?

lassemaatta (changes requested):
> Reviewed 5 of 13 files at r1, 1 of 2 files at r2, 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 13 of 15 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[src/components/Navbar.tsx, line 8 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZS23ib8vxuW74RO9Da:-MZS23icCvpfuDZpOipk:b-euaj18) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/31c2706ec1769cc5e51dd661c4c8bbb060d5104e/src/components/Navbar.tsx#L8)):*
> ```
> const routes = [
> ```

If you have the energy, I think you could simplify this further: Move the `title` to the route definition and iterate the original routes datastructure here. Note that currently you specify  `includeInNav?` in the route type, but don't use it anywhere(?). This would allow you to get rid of the need to refer to the individual routes (`routeDefinitions['/example']`).
___
*[src/components/Router.tsx, line 21 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZS3Bnp56P7wlanJmht:-MZS3Bnp56P7wlanJmhu:bre0oj6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/31c2706ec1769cc5e51dd661c4c8bbb060d5104e/src/components/Router.tsx#L21)):*
> ```
>         <Route path="*">
>           <p>Page not found</p>
>         </Route>
> ```

Could this be a "route" too?

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 11 of 16 files reviewed, 3 unresolved discussions (waiting on @lassemaatta and @pallost)

___
*[src/components/Navbar.tsx, line 8 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZS23ib8vxuW74RO9Da:-MZSUdpKB6jkAdrwDgjn:bkj739s) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/31c2706ec1769cc5e51dd661c4c8bbb060d5104e/src/components/Navbar.tsx#L8)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

> ```
> const routes = [
> ```

If you have the energy, I think you could simplify this further: Move the `title` to the route definition and iterate the original routes datastructure here. Note that currently you specify  `includeInNav?` in the route type, but don't use it anywhere(?). This would allow you to get rid of the need to refer to the individual routes (`routeDefinitions['/example']`).
</blockquote></details>

Good catch, I forgot `includeInNav` here as I decided to remove it because I was thinking that it might be more save to let navbar device itself which routes it wants to show and in which order, but... As you wish. 
___
*[src/components/Router.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MYyb8ZN3_Y1724IBHyf:-MZSV0YBDDCaum87GSYP:bf89uxs) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0502b675c73e320eef17b4b6ff74fa947342663d/src/components/Router.tsx#L13)):*

I'm still not with this idea, or then I misundestrood something.
If we wouldn't have dynamic routes, e.g. `/stops/:id`, when things would be easy.
Then we could define e.g. enum with route names and import it where needed.

But e.g. the way from react-router docs:
```
const routes = [
  {
    path: "/sandwiches",
    component: Sandwiches
  },
  {
    path: "/tacos",
    component: Tacos,
    routes: [
      {
        path: "/tacos/bus",
        component: Bus
      },
      {
        path: "/tacos/cart",
        component: Cart
      }
    ]
  }
];
...
        <Switch>
          {routes.map((route, i) => (
            <RouteWithSubRoutes key={i} {...route} />
          ))}
        </Switch>
```
isn't very useful IMO.
Importing that kind of route array to e.g. navbar doesn't sound very elegant, as then we would have to expose which component each route is linked to, which isnt' something that navbar has to be aware of.
Also defining routes in an array instead of as nodes in Router component (as done right now) won't give any benefits here, it's just different syntax for defining exactly same thing, or in fact its inferior in this case because e.g. defining `exact` prop would be more verbose! And we haven't even gone to problems with dynamic routes yet... I agree that in theory it would indeed be nice to have routes defined only in one place, but in practice that easily turns into over-engineered solution which is less elegant as this kind of "basic" approach.

Did I misunderstood something? Do you have more practical tips or ideas about this?
</blockquote></details>

Done in latest commit.
___
*[src/components/Router.tsx, line 21 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZS3Bnp56P7wlanJmht:-MZSV2IS-U-BKR-kHjF7:b-117imx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/31c2706ec1769cc5e51dd661c4c8bbb060d5104e/src/components/Router.tsx#L21)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

> ```
>         <Route path="*">
>           <p>Page not found</p>
>         </Route>
> ```

Could this be a "route" too?
</blockquote></details>

Technically yes. Not 100% sure if it is a good idea though, because this is "special" route and syntax as this matches any route and shows 404 page. But changed this as you wish, maybe it makes sense to go all in to"routes as a data structure" approach now when we have gone to that direction...

lassemaatta (approved):
Looks good to me.

> Reviewed 8 of 13 files at r1, 1 of 2 files at r2, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-:-MZWpUO54e3GbIEiDHdy:b9j9n43)* status: all files reviewed, 1 unresolved discussion (waiting on @pallost)

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: all files reviewed, 7 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[src/routes.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX3MB386oiheEFOxDY:-MZX3MB4A19HgaQPscEJ:b3odd7w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L13)):*
> ```XML
> 
> interface Route {
>   _routerRoute: string;
> ```

as you already defined paths as an enum, this could just use that type: `_routerRoute: Path`
___
*[src/routes.ts, line 14 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX3Zld0pRrjGqlINSs:-MZX3Zld0pRrjGqlINSt:b7fcvyi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L14)):*
> ```XML
> interface Route {
>   _routerRoute: string;
>   _exact?: boolean;
> ```

could you comment here what this `_exact` parameter is used for?
___
*[src/routes.ts, line 16 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX2YLWEi4OCuqX2NHk:-MZX2YLX5VAo52cz8g-q:b4zuff8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L16)):*
> ```XML
>   _exact?: boolean;
>   translationKey?: string;
>   getLink: (...args: any) => string; // eslint-disable-line @typescript-eslint/no-explicit-any
> ```

`getLink: (...args: (string | number)[]) => string;`? It's quite unlikely that this'll get other types of parameters than string or number
___
*[src/routes.ts, line 21 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX4HZu580cCKqaMOPp:-MZX4HZu580cCKqaMOPq:b-fbm1t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L21)):*
> ```XML
> }
> 
> function ExampleRoute() {
> ```

You could add typing to these for clarify:
`const ExampleRoute: FunctionComponent => () => {return React.createElement('h2', null, 'Example route');}`
___
*[src/routes.ts, line 39 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX4uEA5Pm1uBhf1Pno:-MZX4uEA5Pm1uBhf1Pnp:b-gfnsx2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L39)):*
> ```XML
> 
> export const routes: Record<Path, Route> = {
>   [Path.root]: {
> ```

As you are anyway always iterating through this with Object.values, I don't see a reason to have this as an object but could rather be an array. At least the order of the elements would be clear as it's not (always) guaranteed for an object.
___
*[src/components/Router.tsx, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX5snYEIchlGSa5D9h:-MZX5snYEIchlGSa5D9i:b-p8as1k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/components/Router.tsx#L9)):*
> ```XML
> export const Router: FunctionComponent = () => {
>   return (
>     <BrowserRouter>
> ```

is there a special reason that the Navbar is inside the Router? If not, I think it should rather be outside of it as the rendered Route and the Navbar don't actually strictly depend on each other. Or do you need it for e.g. highlighting an item in the Navbar if that's the active rendered Route?

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: all files reviewed, 7 unresolved discussions (waiting on @pallost)

___
*[src/routes.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX3MB386oiheEFOxDY:-MZXUt_Q9L70tSHdl2Ti:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L13)):*

as you already defined paths as an enum, this could just use that type: `_routerRoute: Path`
</blockquote></details>

Done.
___
*[src/routes.ts, line 14 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX3Zld0pRrjGqlINSs:-MZXV5r74FxhGpCTKMyU:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L14)):*

could you comment here what this `_exact` parameter is used for?
</blockquote></details>

Done.
___
*[src/routes.ts, line 16 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX2YLWEi4OCuqX2NHk:-MZXVDYq36mKD_d0NK-R:b-piuz86) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L16)):*

`getLink: (...args: (string | number)[]) => string;`? It's quite unlikely that this'll get other types of parameters than string or number
</blockquote></details>

Good idea.

Anyway, then line 57 gives following error:

```
(property) Route.getLink: (...args: (string | number)[]) => string
Type '(id: string) => string' is not assignable to type '(...args: (string | number)[]) => string'.
  Types of parameters 'id' and 'args' are incompatible.
    Type 'string | number' is not assignable to type 'string'.
      Type 'number' is not assignable to type 'string'.ts(2322)
routes.ts(16, 3): The expected type comes from property 'getLink' which is declared here on type 'Route'

```
Any ideas?
___
*[src/routes.ts, line 21 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX4HZu580cCKqaMOPp:-MZXWVprDyD4JhLFRB4V:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L21)):*

You could add typing to these for clarify:
`const ExampleRoute: FunctionComponent => () => {return React.createElement('h2', null, 'Example route');}`
</blockquote></details>

Done.
___
*[src/routes.ts, line 39 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX4uEA5Pm1uBhf1Pno:-MZXVxfY8lvUNBVxq5Hx:b71ggml) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L39)):*

As you are anyway always iterating through this with Object.values, I don't see a reason to have this as an object but could rather be an array. At least the order of the elements would be clear as it's not (always) guaranteed for an object.
</blockquote></details>

In the future we most likely have to access individual routes, and I'll rather write something like `routes['/resource'].getLink()` than `routes.find(item => item._routerLink === '/resource').getLink()`. Also note that TS compiler can check if the first one is correct, which is not the case in latter case.
___
*[src/components/Router.tsx, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX5snYEIchlGSa5D9h:-MZXYajm0E4QNic3trZ9:b-up3bas) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/components/Router.tsx#L9)):*

is there a special reason that the Navbar is inside the Router? If not, I think it should rather be outside of it as the rendered Route and the Navbar don't actually strictly depend on each other. Or do you need it for e.g. highlighting an item in the Navbar if that's the active rendered Route?
</blockquote></details>

Yes, there is special reason: :)
`Error: Invariant failed: You should not use <Link> outside a <Router>`
(I agree it would be better to have it outside)

lassemaatta (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 14 of 16 files reviewed, 7 unresolved discussions (waiting on @lassemaatta and @pallost)

___
*[src/routes.ts, line 39 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX4uEA5Pm1uBhf1Pno:-MZa_7eBBxwbDJgKxATc:br0ncjt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L39)):*

In the future we most likely have to access individual routes, and I'll rather write something like `routes['/resource'].getLink()` than `routes.find(item => item._routerLink === '/resource').getLink()`. Also note that TS compiler can check if the first one is correct, which is not the case in latter case.
</blockquote></details>

(I assume you meant something like `routes[Path.resource].getLink()` in your example, because `'/resource'` sort of defeats the whole purpose of this)

lassemaatta (commented):
> Reviewed 1 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 15 of 16 files reviewed, 7 unresolved discussions (waiting on @lassemaatta and @pallost)

___
*[src/routes.ts, line 14 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZaiGHp15RPvfm_C71Q:-MZaiGHp15RPvfm_C71R:b9j3vme) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/065ca76b6c51365cde89f46a096658c18d1657d2/src/routes.ts#L14)):*
> ```XML
> interface Route {
>   _routerRoute: Path;
>   _exact?: boolean; // When true, will only match if the path matches the location.pathname exactly.
> ```

Btw, what's the semantics of `_field` vs `field` here? E.g. why `_routerRoute` with underscore and `translationKey` without? 

(Also, super confusing the ts uses `?` in the field _name_ (`foo?: boolean`) to indicate optionality, instead of the type (`foo: boolean?`), especially as in other languages `foo?` is a valid field name)

lassemaatta (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 15 of 16 files reviewed, 7 unresolved discussions (waiting on @lassemaatta and @pallost)

___
*[src/routes.ts, line 16 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX2YLWEi4OCuqX2NHk:-MZkhGzQCkoGCD5qAU-n:b-7x8e8p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L16)):*

Good idea.

Anyway, then line 57 gives following error:

```
(property) Route.getLink: (...args: (string | number)[]) => string
Type '(id: string) => string' is not assignable to type '(...args: (string | number)[]) => string'.
  Types of parameters 'id' and 'args' are incompatible.
    Type 'string | number' is not assignable to type 'string'.
      Type 'number' is not assignable to type 'string'.ts(2322)
routes.ts(16, 3): The expected type comes from property 'getLink' which is declared here on type 'Route'

```
Any ideas?
</blockquote></details>

I know absolutely nothing about TypeScript, but how about:

`getLink: (...args: (string | number)[] | [string] | [number] ) => string;`

I found various solutions like above by googling for typescript function overloading.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: 13 of 18 files reviewed, 7 unresolved discussions (waiting on @lassemaatta and @pallost)

___
*[src/routes.ts, line 16 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX2YLWEi4OCuqX2NHk:-MZl3o54461k0StQBvQW:b-yx24g4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L16)):*
> getLink: (...args: (string | number)[] | [string] | [number] ) => string;
Tried also that, and still:
 ```Type '(id: string) => string' is not assignable to type '(...args: [string] | [number] | (string | number)[]) => string'.
  Types of parameters 'id' and 'args' are incompatible.
    Type '[string] | [number] | (string | number)[]' is not assignable to type '[id: string]'.
      Type '[number]' is not assignable to type '[id: string]'.
        Type at position 0 in source is not compatible with type at position 0 in target.
          Type 'number' is not assignable to type 'string'.ts(2322)
routes.ts(16, 3): The expected type comes from property 'getLink' which is declared here on type 'Route'```
Anyway, this problem is limited within scope of this file as we can still type individual `getLink` function definitions (see line 57). If you have solution ready I can use it, but I don't see ROI in spending more of my time trying to figure out how to do this... And yes, I already tried to google the solution when @pallost commented about this.
```

___
*[src/routes.ts, line 39 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX4uEA5Pm1uBhf1Pno:-MZl2h7kBVdSLc39y3b1:b-wkoky9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L39)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

(I assume you meant something like `routes[Path.resource].getLink()` in your example, because `'/resource'` sort of defeats the whole purpose of this)
</blockquote></details>

(Well, they both work and, TS compiler can check if  `'/resource'` is valid value for `Path` enum so its still type safe, but yes, maybe it makes more sense so use enums directly. The downside with that is that intellisense primarly suggests using string values...)

**Review by lassemaatta — APPROVED** [review]
Looks ok to me, but perhaps @pallost should check this out too.

> Reviewed 1 of 2 files at r5, 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-:-MZl4mdM3VUHGpe1sAmy:b-nojd0a)* status: all files reviewed, 8 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[src/components/navbar/Navbar.tsx, line 13 at r6](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZl3I2R-GI7AiO02_sn:-MZl3I2R-GI7AiO02_so:b-dw85zi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6d3ee5f671080a0228bb35786eabd93bc3937b2d/src/components/navbar/Navbar.tsx#L13)):*
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>

> ```
>       <Link to="/" aria-label={t('routes.root')}>
>                    src="/nav-logo.svg"
>           style={{ height: 40, marginLeft: 29 }}
>           alt=""
>           className="pr-5"
>         />
>       </Link>
> ```
</details>


I'd prefer to move this to a `NavIcon` component (there's probably a better name for it..) or somewhere, but this is a very minor issue.

```typescript
export const Navbar: FunctionComponent = () => {
  return (
    <nav className="flex flex-row bg-brand items-center">
      <NavIcon />
      <NavLinks />
      <LanguageDropdown />
    </nav>
  );
};

```

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: all files reviewed, 3 unresolved discussions (waiting on @lassemaatta and @tk-vincit)

___
*[src/routes.ts, line 16 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX2YLWEi4OCuqX2NHk:-MZlC1qSDB-SOnr61Ww6:bpxj67g) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L16)):*

> getLink: (...args: (string | number)[] | [string] | [number] ) => string;
Tried also that, and still:
 ```Type '(id: string) => string' is not assignable to type '(...args: [string] | [number] | (string | number)[]) => string'.
  Types of parameters 'id' and 'args' are incompatible.
    Type '[string] | [number] | (string | number)[]' is not assignable to type '[id: string]'.
      Type '[number]' is not assignable to type '[id: string]'.
        Type at position 0 in source is not compatible with type at position 0 in target.
          Type 'number' is not assignable to type 'string'.ts(2322)
routes.ts(16, 3): The expected type comes from property 'getLink' which is declared here on type 'Route'```
Anyway, this problem is limited within scope of this file as we can still type individual `getLink` function definitions (see line 57). If you have solution ready I can use it, but I don't see ROI in spending more of my time trying to figure out how to do this... And yes, I already tried to google the solution when @pallost commented about this.
</blockquote></details>

```
type RouteParam = string | number
(...args: RouteParam[] | RouteParam) => string
```
___
*[src/components/Router.tsx, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX5snYEIchlGSa5D9h:-MZlprpeBOZR4l28ogRH:b-u1wpk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/components/Router.tsx#L9)):*

Yes, there is special reason: :)
`Error: Invariant failed: You should not use <Link> outside a <Router>`
(I agree it would be better to have it outside)
</blockquote></details>

fair enough :)


pallost (commented):
I'll approve this PR. There are things that could be polished or done in a slightly different way, but none of them are actually blocking issues. If you have the energy to fix them, do so, otherwise let's finally go on to the next task because this PR has been sitting here forever :)

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-:-MZlq7yc0B1gR5VGgAKE:b-qln324)* status: all files reviewed, 3 unresolved discussions (waiting on @lassemaatta and @tk-vincit)

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/5)* status: all files reviewed, 3 unresolved discussions (waiting on @lassemaatta, @pallost, and @tk-vincit)

___
*[src/routes.ts, line 16 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZX2YLWEi4OCuqX2NHk:-MZqCqHR1IuATQuyVVks:bdbq9rg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/033e6474b0c5c839ff3bd7ed324dc8e9cfadf5e2/src/routes.ts#L16)):*

```
type RouteParam = string | number
(...args: RouteParam[] | RouteParam) => string
```
</blockquote></details>

`A rest parameter must be of an array type.`
___
*[src/routes.ts, line 14 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZaiGHp15RPvfm_C71Q:-MZqD5kb5cQSquawZ2jH:b-g0x3el) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/065ca76b6c51365cde89f46a096658c18d1657d2/src/routes.ts#L14)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

Btw, what's the semantics of `_field` vs `field` here? E.g. why `_routerRoute` with underscore and `translationKey` without? 

(Also, super confusing the ts uses `?` in the field _name_ (`foo?: boolean`) to indicate optionality, instead of the type (`foo: boolean?`), especially as in other languages `foo?` is a valid field name)
</blockquote></details>

underscore tries to indicate that those fields are meant to be use only internally.
___
*[src/components/navbar/Navbar.tsx, line 13 at r6](https://reviewable.io/reviews/hsldevcom/jore4-ui/5#-MZl3I2R-GI7AiO02_sn:-MZqF7Ka0EdeMYPXPNm3:b-32if43) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6d3ee5f671080a0228bb35786eabd93bc3937b2d/src/components/navbar/Navbar.tsx#L13)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

> ```
>       <Link to="/" aria-label={t('routes.root')}>
>                    src="/nav-logo.svg"
>           style={{ height: 40, marginLeft: 29 }}
>           alt=""
>           className="pr-5"
>         />
>       </Link>
> ```

I'd prefer to move this to a `NavIcon` component (there's probably a better name for it..) or somewhere, but this is a very minor issue.

```typescript
export const Navbar: FunctionComponent = () => {
  return (
    <nav className="flex flex-row bg-brand items-center">
      <NavIcon />
      <NavLinks />
      <LanguageDropdown />
    </nav>
  );
};

```
</blockquote></details>

Ok, will do that in next PR.

PR #7, opened 2021-05-04
Small fixes

tk-vincit:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/7)

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,5 @@
 {
   "singleQuote": true,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "twConfig": "tailwind.config.js"
 }
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ If you use VSCode, following plugins are recommended:
 
 - [esbenp.prettier-vscode](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) for formatting code automatically with `prettier`
 - [dbaeumer.vscode-eslint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) for showing linting errors within the code
+- [bradlc.vscode-tailwindcss](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss) for adding Tailwind suggestions to IntelliSense
 
 ## Docker image
 
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "jest": "^26.6.3",
     "postcss": "^8.2.10",
     "prettier": "2.2.1",
+    "prettier-plugin-tailwind": "^2.2.10",
     "tailwindcss": "^2.1.1",
     "ts-jest": "^26.5.4",
     "typescript": "^4.2.3"
--- a/src/components/navbar/BrandLogo.tsx
+++ b/src/components/navbar/BrandLogo.tsx
@@ -1,6 +1,7 @@
 import React, { FunctionComponent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
+import { Path, routes } from 'routes';
 
 interface Props {
   className?: string;
@@ -9,10 +10,12 @@ interface Props {
 
 export const BrandLogo: FunctionComponent<Props> = ({ className, style }) => {
   const { t } = useTranslation();
+  const target = routes[Path.root];
   return (
     <Link
-      to="/"
-      aria-label={t('routes.root')}
+      to={target.getLink()}
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      aria-label={t(target.translationKey!)}
       className={className}
       style={style}
     >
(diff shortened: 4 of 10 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/7)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[src/components/navbar/BrandLogo.tsx, line 14 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/7#-MZr3BBy0fXw8a1nLefL:-MZr3BBy0fXw8a1nLefM:b-frrpv9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b2230cdb61bfcb221a81fed699bea5267b49263b/src/components/navbar/BrandLogo.tsx#L14)):*
> ```XML
>   return (
>     <Link
>       to="/"
> ```

Here you could reuse the path and translationKey from `routes.ts`: `routes[Path.root]._routerRoute`, etc
___
*[src/components/navbar/Navbar.tsx, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/7#-MZr3fbpAVYlmoZDo5zH:-MZr3fbpAVYlmoZDo5zI:b-f7jvm9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b2230cdb61bfcb221a81fed699bea5267b49263b/src/components/navbar/Navbar.tsx#L9)):*
> ```XML
>   return (
>     <nav className="flex flex-row items-center bg-brand">
>       <BrandLogo className="pr-5" style={{ marginLeft: 29 }} />
> ```

well, if this class does not exist in tailwind, should we create some kind of `tailwind-custom.css` where we define our own custom rules?
```
ml-29 {
  margin-left: 29px
}
```
I'd rather use classNames everywhere instead of combining classes and styles to achieve the desired outlook

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/7)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @pallost)

___
*[src/components/navbar/BrandLogo.tsx, line 14 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/7#-MZr3BBy0fXw8a1nLefL:-MZrA69h2kP0lY_byx5M:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b2230cdb61bfcb221a81fed699bea5267b49263b/src/components/navbar/BrandLogo.tsx#L14)):*

Here you could reuse the path and translationKey from `routes.ts`: `routes[Path.root]._routerRoute`, etc
</blockquote></details>

Done.
___
*[src/components/navbar/Navbar.tsx, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/7#-MZr3fbpAVYlmoZDo5zH:-MZrA6x3-ET0KuJwtD_e:bfq43gk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b2230cdb61bfcb221a81fed699bea5267b49263b/src/components/navbar/Navbar.tsx#L9)):*

well, if this class does not exist in tailwind, should we create some kind of `tailwind-custom.css` where we define our own custom rules?
```
ml-29 {
  margin-left: 29px
}
```
I'd rather use classNames everywhere instead of combining classes and styles to achieve the desired outlook
</blockquote></details>

I don't see problem with this approach.
I think that in the future we can define like 99,x% of our styles with tailwind. Now when one sees "style={...}" definition, they hopefully also stop for a second and think about why there is need for that. If we hide inline styles behind tailwind look-a-like class name, then its more diffucult. For example in this case I wanted logo alignment to be pixel perfect to the one in hsl.fi. And by the way, this is currently the only one inline style in the whole codebase.

(Maybe we could add comments explaining why inline styles are used to each place where we are using those...)

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/7)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[src/components/navbar/Navbar.tsx, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/7#-MZr3fbpAVYlmoZDo5zH:-MZrBe6qAXvlyEQb8RjH:bebd0wi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b2230cdb61bfcb221a81fed699bea5267b49263b/src/components/navbar/Navbar.tsx#L9)):*

I don't see problem with this approach.
I think that in the future we can define like 99,x% of our styles with tailwind. Now when one sees "style={...}" definition, they hopefully also stop for a second and think about why there is need for that. If we hide inline styles behind tailwind look-a-like class name, then its more diffucult. For example in this case I wanted logo alignment to be pixel perfect to the one in hsl.fi. And by the way, this is currently the only one inline style in the whole codebase.

(Maybe we could add comments explaining why inline styles are used to each place where we are using those...)
</blockquote></details>

I think if we use the style attribute, it's self-explanatory that we can't find the same definition from tailwind

PR #8, opened 2021-05-05
Language dropdown improvements

tk-vincit:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/8)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/8)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @tk-vincit)

___
*[src/components/navbar/LanguageDropdown.tsx, line 15 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-WW9N8f2CwjdugVOE:-M_-WW9N8f2CwjdugVOF:b-ugd0ab) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L15)):*
> ```XML
> 
> export const LanguageDropdown: FunctionComponent<Props> = ({ className }) => {
>   useTranslation(); // This is needed in order to make component update itself when language is changed
> ```

actually, to get the `i18n` instance, it would be nicer to use the hook instead of importing it globally, like this:
`const { i18n } = useTranslation();`
___
*[src/components/navbar/LanguageDropdown.tsx, line 37 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Ws-9EVmCh753yCax:-M_-Ws-9EVmCh753yCay:bbnwc3k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L37)):*
> ```XML
>                                src="/icons/chevron-down.png"
>                 className={`ml-2 h-1.5 transform transition duration-150 ease-in-out ${
> ```

ouuu, such animation :D
___
*[src/components/navbar/LanguageDropdown.spec.tsx, line 4 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Wvxw0yqLSGvRORhT:-M_-Wvxw0yqLSGvRORhU:bijpzn3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.spec.tsx#L4)):*
> ```XML
> import React from 'react';
> import { render } from 'test-utils';
> import { i18n } from '../../i18n';
> ```

I wonder, can the useTranslate hook be used in tests instead of importing the i18n directly? It's not important, just came to mind...
___
*[src/components/navbar/LanguageDropdown.spec.tsx, line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-XCjZAyAQDtM1P9aj:-M_-XCjZAyAQDtM1P9ak:bmb3g7q) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.spec.tsx#L33)):*
> ```XML
>     expect(openDropdownButton).toHaveTextContent(finnishTitle);
> 
>     // click dropdown to open it:
> ```

could write a comment here that for some reason the dropdown does not get closed in tests after choosing the language

lassemaatta (commented):
> Reviewed 1 of 2 files at r1, 1 of 2 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/8)* status: all files reviewed, 5 unresolved discussions (waiting on @tk-vincit)

___
*[src/components/navbar/LanguageDropdown.tsx, line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Y33rE1GtPbtne1J3:-M_-Y33rE1GtPbtne1J4:b-kncbc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L42)):*
> ```XML
>                 alt=""
>               />
>             </Menu.Button>
> ```

This maybe more of a personal preference rather than an actual issue, but I'd like to see fewer instances of inlining stuff and more composition. At least for me it's really hard to see structure of the component when a) there's some much content in it and b) there's no context in the anonymous snippets of HTML. 

For example `` doesn't really tell me anything what that thing actually represents or what is its purpose. However, if it's extracted into a `<CollapseIcon />` or something, then it would have some semantic meaning and the details would be hidden in the particular component. 

(The same idea really applies to any situation, I don't like to see 1000+ line java methods either.)

(Naturally I'm assuming here that this is technically possible. I'm sure there's dozens of situations where this is impractical.)

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/8)* status: all files reviewed, 5 unresolved discussions (waiting on @lassemaatta and @pallost)

___
*[src/components/navbar/LanguageDropdown.tsx, line 15 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-WW9N8f2CwjdugVOE:-M_-zQRn4RqA65R0nw8p:b6zrrk4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L15)):*

actually, to get the `i18n` instance, it would be nicer to use the hook instead of importing it globally, like this:
`const { i18n } = useTranslation();`
</blockquote></details>

Good catch!
___
*[src/components/navbar/LanguageDropdown.tsx, line 37 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Ws-9EVmCh753yCax:-M_-zUwgDcqmBPhm5Evf:b-dafx0u) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L37)):*

ouuu, such animation :D
</blockquote></details>

same as in hsl.fi
___
*[src/components/navbar/LanguageDropdown.tsx, line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Y33rE1GtPbtne1J3:-M_0-Oh32tg98Gh-C_H6:bti1mkj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L42)):*
>  the component when a) there's some much content in it and b) there's no context in the anonymous snippets of HTML.
> For example  doesn't really tell me anything what that thing actually represents or what is its purpose. However, if it's extracted into a <CollapseIcon /> or something, then it would have some semantic
Well... E.g. in this case with the  tag, we would still have to define `className` things here. So yes, you could save 2 lines of code here, but then you would have to write 10 lines or so in another place...

Also it seems like there's still possibility that we'll find icon library which will provide icon components for us (please see discussion in slack), so its also possible that this will be changed in the near future. So probably not worth messing around with this right now.

And yeah, ` ` does not tell what's the purpose of the image, but in the other hand` ` is pretty self-explanatory at least IMO.
___
*[src/components/navbar/LanguageDropdown.spec.tsx, line 4 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Wvxw0yqLSGvRORhT:-M_-zk4qCv_SYfidAY_d:bhrueiu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.spec.tsx#L4)):*

I wonder, can the useTranslate hook be used in tests instead of importing the i18n directly? It's not important, just came to mind...
</blockquote></details>

Hooks can only be called inside of the body of a function component.
___
*[src/components/navbar/LanguageDropdown.spec.tsx, line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-XCjZAyAQDtM1P9aj:-M_-znhsEXwHpqGMBsaL:b9g8wiy) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.spec.tsx#L33)):*

could write a comment here that for some reason the dropdown does not get closed in tests after choosing the language
</blockquote></details>

Does it have any value for anyone? I think that would be just confusing.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/8)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @lassemaatta and @tk-vincit)

___
*[src/components/navbar/LanguageDropdown.tsx, line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Y33rE1GtPbtne1J3:-M_01g3h2jFCZ0e7MtWF:b-e4k4yw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L42)):*

>  the component when a) there's some much content in it and b) there's no context in the anonymous snippets of HTML.
> For example  doesn't really tell me anything what that thing actually represents or what is its purpose. However, if it's extracted into a <CollapseIcon /> or something, then it would have some semantic
Well... E.g. in this case with the  tag, we would still have to define `className` things here. So yes, you could save 2 lines of code here, but then you would have to write 10 lines or so in another place...

Also it seems like there's still possibility that we'll find icon library which will provide icon components for us (please see discussion in slack), so its also possible that this will be changed in the near future. So probably not worth messing around with this right now.

And yeah, ` ` does not tell what's the purpose of the image, but in the other hand` ` is pretty self-explanatory at least IMO.
</blockquote></details>

I don't see extracting this as a separate component important for now. However when we start to have more icons, it'll be useful to extract them into a common `<Icon type="chevron-down" />` to handle common icon behaviours easier.
___
*[src/components/navbar/LanguageDropdown.spec.tsx, line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-XCjZAyAQDtM1P9aj:-M_02CEO7Qe4D9cJFHrF:b4b95ol) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.spec.tsx#L33)):*

Does it have any value for anyone? I think that would be just confusing.
</blockquote></details>

well, the test would've contained some checks also that the dropdown opens after clicking the button and closes when choosing the language. However this is not what happens even though this would be the expected behaviour. It's good to document when something does not work as intended

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/8)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @lassemaatta and @pallost)

___
*[src/components/navbar/LanguageDropdown.tsx, line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Y33rE1GtPbtne1J3:-M_03EGKEw4pV_MFWAT6:blnu6k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L42)):*

I don't see extracting this as a separate component important for now. However when we start to have more icons, it'll be useful to extract them into a common `<Icon type="chevron-down" />` to handle common icon behaviours easier.
</blockquote></details>

Yeah, lets consider that after we get better insight from HSL if they are already using e.g. certain icon libary, or do we have to implement it from scratch. And does it make sense to implement it from scratch anyway.
___
*[src/components/navbar/LanguageDropdown.spec.tsx, line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-XCjZAyAQDtM1P9aj:-M_03f_T3ZHzOZhhNEIk:b-8ujee8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.spec.tsx#L33)):*

well, the test would've contained some checks also that the dropdown opens after clicking the button and closes when choosing the language. However this is not what happens even though this would be the expected behaviour. It's good to document when something does not work as intended
</blockquote></details>

Does it actually belong into scope of this test? Now test description says that we are testing language changing and this is pretty much minimal test for that.

lassemaatta (commented):
> Reviewed 2 of 2 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/8)* status: all files reviewed, 2 unresolved discussions (waiting on @lassemaatta, @pallost, and @tk-vincit)

___
*[src/components/navbar/LanguageDropdown.tsx, line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/8#-M_-Y33rE1GtPbtne1J3:-M_03ClL0lZodmYEq7IR:b-2ow4xo) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4ce6c844f7dae140c0abe1e7b36fed92bc7a23f1/src/components/navbar/LanguageDropdown.tsx#L42)):*

I don't see extracting this as a separate component important for now. However when we start to have more icons, it'll be useful to extract them into a common `<Icon type="chevron-down" />` to handle common icon behaviours easier.
</blockquote></details>

My comment wasn't really about this particular `img` icon, rather a more general idea about how we perhaps should split our components a bit more. I don't mean we have to promote every single thing to a separate file, but just having a few `const SomePrivateComponent: ...` might help structure the actual main component some file exports.

PR #12, opened 2021-06-06
User login and displaying of user permissions

jschummer:
Project level changes:
* redux and redux-thunk were added to the project
* the /api/auth/** paths are now routed to the backend on the
  development server

A Main component was created to perform tasks common to the whole
application. At this stage, it only fetches the user-info.

The user-info type needs to be adjusted when the type has been decided
on the backend side.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12)


jpeuralinna (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @jschummer)

___
*[src/components/navbar/Navbar.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mb_GsHl0mf5Hb6lPqoT:-Mb_GsHl0mf5Hb6lPqoU:b-34fc1p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a21ace5c16539e9540a2013fce34176bac49e2fd/src/components/navbar/Navbar.tsx#L13)):*
> ```XML
>       <NavLinks />
>       <UserDropdown className="ml-auto border-l border-black border-opacity-20" />
>       <LanguageDropdown className="border-l border-black border-opacity-20" />
> ```

I'm not sure where react elements get all their properties from, but just want to mention that we should have some unique id:s or class names for elements. So that they can be referenced easier in tests, since xpaths to classNames with long list of properties tend to break often.

tk-vincit (changes requested):
Generally looking good to me! 

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-:-Mbq2WUABfQwnG1y8Fdz:b-r3cl3u)* status: 0 of 27 files reviewed, 12 unresolved discussions (waiting on @jpeuralinna and @jschummer)

___
*[src/components/navbar/DropdownMenu.tsx, line 7 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbpy56h63QjyfpT63pe:-Mbpy56i5kz6x-zguLiO:b78oxhb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/DropdownMenu.tsx#L7)):*
> ```XML
>   buttonContent: ReactNode;
>   children: ReactNode;
>   // eslint-disable-next-line react/require-default-props
> ```

Note: I'll disable this rule globally in my next pr, so this line can be removed after this. No need to do anything at this point, this is mainly reminder for me.
___
*[src/components/navbar/DropdownMenu.tsx, line 12 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq1O3d5ktvh2rFzbf8:-Mbq1O3d5ktvh2rFzbf9:b10y10q) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/DropdownMenu.tsx#L12)):*
> ```XML
> 
> export const DropdownMenu = (props: Props) => {
>   const { buttonContent, children, testId } = props;
> ```

Not a blocker for me, but react convention (or at least convention that I'm most familiar with) would be to destructure these on previous row.
___
*[src/components/navbar/DropdownMenu.tsx, line 14 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-MbpykPA4Oh5PdkJWwfc:-MbpykPA4Oh5PdkJWwfd:b267h0b) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/DropdownMenu.tsx#L14)):*
> ```XML
>   const { buttonContent, children, testId } = props;
> 
>   const addClassName = (child: ReactElement, newClassName: string) => {
> ```

This should rather be in e.g. `.src/utils/index.ts` file. Anyway, this is OK for now, we can move this when this is used in multiple places.
___
*[src/components/navbar/Navbar.tsx, line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mb_GsHl0mf5Hb6lPqoT:-MbpzClK6rSdQadhDj1_:b-1g45qc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a21ace5c16539e9540a2013fce34176bac49e2fd/src/components/navbar/Navbar.tsx#L13)):*
<details><summary><i>Previously, jpeuralinna (Joonas Peuralinna) wrote…</i></summary><blockquote>

I'm not sure where react elements get all their properties from, but just want to mention that we should have some unique id:s or class names for elements. So that they can be referenced easier in tests, since xpaths to classNames with long list of properties tend to break often.
</blockquote></details>

We can add `data-testid` attribute for elements which need to be accessed from tests.
Anyway, in tests where we are accessing elements which are meaningful to the users, it might make more sense to query those by UI text.
___
*[src/components/navbar/Navbar.tsx, line 12 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq0jZXE-tuJxnEDL0r:-Mbq0jZXE-tuJxnEDL0s:bcm75u2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/Navbar.tsx#L12)):*
> ```XML
>       <BrandLogo className="pr-5" style={{ marginLeft: 29 }} />
>       <NavLinks />
>       <UserNavMenu className="ml-auto border-l border-black border-opacity-20" />
> ```

This has exactly same classNames as the item below.
You should either put it into constant and use it on both rows, or then save these components into array and then map through it and add className for both items with your neat `addClassName` function.
___
*[src/components/navbar/UserNavMenu.tsx, line 17 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbpzmal5h_JcpcxSfbn:-Mbpzmal5h_JcpcxSfbo:bu1pd3a) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/UserNavMenu.tsx#L17)):*
> ```XML
> 
>   const userImgElement = (
>     user
> ```

Alt text should be translated too. Or if image is not meaningful for the screen reader, then I think (but I'm not 100% sure) that you can just leave alt text out (=set it to empty string).
___
*[src/components/navbar/UserNavMenu.tsx, line 30 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq1nBW1eoh6BipyHEl:-Mbq1nBW1eoh6BipyHEm:bgw7baz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/UserNavMenu.tsx#L30)):*
> ```XML
>             login();
>           }}
>           to=""
> ```

Not 100% sure how this is supposed to work, but does it make sense to use NavLink if its target is empty? Should this be e.g. regular button or anchor tag?
___
*[src/context/UserContext.ts, line 17 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq-AjND3UKtF4R9-Rb:-Mbq-AjND3UKtF4R9-Rc:b-q12xgm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContext.ts#L17)):*
> ```XML
>   if (!context) {
>     throw new Error(
>       'user context not provided, please use UserContextProvider',
> ```

`'User context` or `'userContext`
___
*[src/context/UserContextProvider.tsx, line 31 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq2Itx1E0V3Aq-xerd:-Mbq2Itx1E0V3Aq-xere:b-fxhvlw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContextProvider.tsx#L31)):*
> ```XML
>   }, [getUserInfo]);
> 
>   const { children } = props;
> ```

This could be destructured already on the same line where this component is defined.
___
*[src/context/UserContextProvider.tsx, line 34 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq29gQBi9PkAiNM5Sv:-Mbq29gQBi9PkAiNM5Sw:blluioe) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContextProvider.tsx#L34)):*
> ```XML
> 
>   return (
>     <userContext.Provider
> ```

React elements usually start with capital letter. So I would call this `UserContext`
___
*[src/context/UserContextProvider.tsx, line 35 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq-r2dAuzOAEStrTb9:-Mbq-r2dAuzOAEStrTbA:b-2g7rgn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContextProvider.tsx#L35)):*
> ```XML
>   return (
>     <userContext.Provider
>       value={{
> ```

Do you want to limit which properties of the state are passed to the user context, or should you write this like `value={state}`?
___
*[src/types/UserInfo.ts, line 2 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq0E6n7lmi7L0-tiCR:-Mbq0E6n7lmi7L0-tiCS:bj2gi0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/types/UserInfo.ts#L2)):*
> ```XML
> /* eslint-disable  @typescript-eslint/no-explicit-any */
> export type UserInfo = any; // TODO: Change this to the actual UserInfo type once it has been decided in the backend
> ```

Optional: this could already be typed to include the fields that are used so far. Or if no fields from this type are used so far, then this should be `unknown` type.
Anyway kudos for extracting this to own type already. :+1:

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12)* status: 0 of 27 files reviewed, 13 unresolved discussions (waiting on @jpeuralinna and @tk-vincit)

___
*[src/components/navbar/DropdownMenu.tsx, line 7 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbpy56h63QjyfpT63pe:-MbuptQUDiCgTB-lIOIH:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/DropdownMenu.tsx#L7)):*

Note: I'll disable this rule globally in my next pr, so this line can be removed after this. No need to do anything at this point, this is mainly reminder for me.
</blockquote></details>

Done.
___
*[src/components/navbar/DropdownMenu.tsx, line 12 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq1O3d5ktvh2rFzbf8:-Mbuq3bNE8iH7EOR_zxV:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/DropdownMenu.tsx#L12)):*

Not a blocker for me, but react convention (or at least convention that I'm most familiar with) would be to destructure these on previous row.
</blockquote></details>

Done.
___
*[src/components/navbar/DropdownMenu.tsx, line 14 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-MbpykPA4Oh5PdkJWwfc:-Mbuq9ic9tP-cFwfsp35:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/DropdownMenu.tsx#L14)):*

This should rather be in e.g. `.src/utils/index.ts` file. Anyway, this is OK for now, we can move this when this is used in multiple places.
</blockquote></details>

Done.
___
*[src/components/navbar/Navbar.tsx, line 12 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq0jZXE-tuJxnEDL0r:-MburS1pBnQPoW_384VD:b-kbvv3v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/Navbar.tsx#L12)):*

This has exactly same classNames as the item below.
You should either put it into constant and use it on both rows, or then save these components into array and then map through it and add className for both items with your neat `addClassName` function.
</blockquote></details>

I decided to use the constant. I'd rather use addClassName -type of solutions only when the class names need to be generated dynamically.
___
*[src/components/navbar/UserNavMenu.tsx, line 17 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbpzmal5h_JcpcxSfbn:-MbuufYt-qa6Bmp2_bv5:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/UserNavMenu.tsx#L17)):*

Alt text should be translated too. Or if image is not meaningful for the screen reader, then I think (but I'm not 100% sure) that you can just leave alt text out (=set it to empty string).
</blockquote></details>

Done.
___
*[src/components/navbar/UserNavMenu.tsx, line 30 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq1nBW1eoh6BipyHEl:-Mbv1Ulh06b_JJ43hjde:b-egkakt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/UserNavMenu.tsx#L30)):*

Not 100% sure how this is supposed to work, but does it make sense to use NavLink if its target is empty? Should this be e.g. regular button or anchor tag?
</blockquote></details>

You're right, this apparently is a relic from earlier experiments. Changed it to an anchor now, that is the cleaner way in any case!
___
*[src/components/navbar/UserNavMenu.tsx, line 45 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbv1se46oCD6lUW-pmT:-Mbv1se46oCD6lUW-pmU:bh0040) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/UserNavMenu.tsx#L45)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>           <button
>             type="button"
>             onClick={() => {
>               logout();
>             }}
>           >
> ```
</details>

Similar to above, changed this to an anchor too
___
*[src/context/UserContext.ts, line 17 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq-AjND3UKtF4R9-Rb:-MburrAxDJvpEDssmVRH:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContext.ts#L17)):*

`'User context` or `'userContext`
</blockquote></details>

Done.
___
*[src/context/UserContextProvider.tsx, line 31 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq2Itx1E0V3Aq-xerd:-MbursOpCTCKMxg5mzjM:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContextProvider.tsx#L31)):*

This could be destructured already on the same line where this component is defined.
</blockquote></details>

Done.
___
*[src/context/UserContextProvider.tsx, line 34 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq29gQBi9PkAiNM5Sv:-Mbus9EfEcLok_jV9t7k:b-kxid2b) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContextProvider.tsx#L34)):*

React elements usually start with capital letter. So I would call this `UserContext`
</blockquote></details>

Yes, true. This probably got mixed up with the other `UserContext` defined above. I renamed that to `UserContextData` and this `userContext` became `UserContext`.
___
*[src/context/UserContextProvider.tsx, line 35 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq-r2dAuzOAEStrTb9:-MbusVSl5wRnjYGsXKCH:bfzqffr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/context/UserContextProvider.tsx#L35)):*

Do you want to limit which properties of the state are passed to the user context, or should you write this like `value={state}`?
</blockquote></details>

Some experiment rounds before, some stuff was actually added here. You're right that there's probably no reason to list the properties one-by-one now. Changed it.
___
*[src/types/UserInfo.ts, line 2 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbq0E6n7lmi7L0-tiCR:-Mbut8EI9-UfpqN1f8Zt:bcodkga) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/types/UserInfo.ts#L2)):*

Optional: this could already be typed to include the fields that are used so far. Or if no fields from this type are used so far, then this should be `unknown` type.
Anyway kudos for extracting this to own type already. :+1: 
</blockquote></details>

Well, the thing is that the fields come straight from the HSLID server. And the few ones we use in this (sort of demo) implementation have names that are not nice / even possible to list.

E.g. one key name is 'https://oneportal.trivore.com/claims/active_external_permissions' (yes, that is the key name...).

I'd prefer to go with any for now. It's definitely a must to define the type soon (in sync with the backend).

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12)* status: 0 of 27 files reviewed, 13 unresolved discussions (waiting on @jpeuralinna and @tk-vincit)

___
*[src/components/navbar/DropdownMenu.tsx, line 14 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-MbpykPA4Oh5PdkJWwfc:-Mbv2yTFBu4jxNLu4x2g:b-ezs7lv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/DropdownMenu.tsx#L14)):*

Done.
</blockquote></details>

There was actually a bug in this function causing the "undefined" class to be added in some cases... :-) Fixed it now too.

tk-vincit (commented):
Just two new, minor things about the code itself.

Also tested this locally and seems like User button in navbar doesn't do anything. Is that expected?

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-:-Mc8vHcgA3O8Rwcgh-5p:bjcg3kh)* status: 0 of 27 files reviewed, 6 unresolved discussions (waiting on @jpeuralinna, @jschummer, and @tk-vincit)

___
*[src/components/navbar/UserNavMenu.tsx, line 43 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mc8ryb-9Eyi1n3LR9OY:-Mc8ryb0E0W1pKMg5ALh:bdjx6q3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b35d50603e444b824f793e1906631fc92f967dfb/src/components/navbar/UserNavMenu.tsx#L43)):*
> ```XML
>           }
>         >
>           <a href={LOGOUT_URL} style={{ display: 'block' }}>
> ```

Why inline style and not `className="block"`?
___
*[src/context/UserContextProvider.tsx, line 15 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mc8sPuG9Qu55kebNKDj:-Mc8sPuG9Qu55kebNKDk:b-xpqkyu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b35d50603e444b824f793e1906631fc92f967dfb/src/context/UserContextProvider.tsx#L15)):*
> ```XML
> }
> 
> export const UserContextProvider = ({ children }: Props) => {
> ```

This component is not typed. Should be like this: `export const UserContextProvider = ({ children }: Props): JSX.Element  => {`

tk-vincit (commented):
(ah, reason why it didn't worked locally was that I didn't had auth backend up & running)

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-:-Mc8wFC23XxpCNsGmiAw:bm6kji0)* status: 0 of 27 files reviewed, 6 unresolved discussions (waiting on @jpeuralinna, @jschummer, and @tk-vincit)

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12)* status: 0 of 27 files reviewed, 6 unresolved discussions (waiting on @jpeuralinna and @tk-vincit)

___
*[src/components/navbar/UserNavMenu.tsx, line 43 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mc8ryb-9Eyi1n3LR9OY:-Mc91wCd4MzO2CJj5SO9:bmi2rfj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b35d50603e444b824f793e1906631fc92f967dfb/src/components/navbar/UserNavMenu.tsx#L43)):*

Why inline style and not `className="block"`?
</blockquote></details>

Pure incompetence :-)

I actually tried to ask google if there's a class I could use for this, but didn't find anything. That's why I concluded there might not be a class which applied the "display:block".

Corrected it now.

For the future: Where would I find documentation on the list of style classes?
___
*[src/context/UserContextProvider.tsx, line 15 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mc8sPuG9Qu55kebNKDj:-Mc94Kf5CnNG_Gz876-c:b3lyu9r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b35d50603e444b824f793e1906631fc92f967dfb/src/context/UserContextProvider.tsx#L15)):*

This component is not typed. Should be like this: `export const UserContextProvider = ({ children }: Props): JSX.Element  => {`
</blockquote></details>

Oh yes, fixed it. I actually noticed the same thing with `DropdownMenu` and changed that too.

Shouldn't / couldn't the linter say something about this btw?

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/12)* status: 0 of 27 files reviewed, 5 unresolved discussions (waiting on @jpeuralinna, @jschummer, and @tk-vincit)

___
*[src/components/navbar/UserNavMenu.tsx, line 45 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mbv1se46oCD6lUW-pmT:-Mc9K_P53oguIwmuEQAH:b-inpv78) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/22e2ead0bc4473e5c1b57d46ec7b7971b454ea9f/src/components/navbar/UserNavMenu.tsx#L45)):*

> ```
>           <button
>             type="button"
>             onClick={() => {
>               logout();
>             }}
>           >
> ```
Similar to above, changed this to an anchor too
</blockquote></details>

Didn't tested these locally, but if this should redirect user to different domain then this is OK.
___
*[src/components/navbar/UserNavMenu.tsx, line 43 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mc8ryb-9Eyi1n3LR9OY:-Mc9KO0sDKYyxHRSG_qZ:b-twwk0n) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b35d50603e444b824f793e1906631fc92f967dfb/src/components/navbar/UserNavMenu.tsx#L43)):*

Pure incompetence :-)

I actually tried to ask google if there's a class I could use for this, but didn't find anything. That's why I concluded there might not be a class which applied the "display:block".

Corrected it now.

For the future: Where would I find documentation on the list of style classes?
</blockquote></details>

We are using `tailwindcss` and I suggest using its official documentation: https://tailwindcss.com/docs/display
___
*[src/context/UserContextProvider.tsx, line 15 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/12#-Mc8sPuG9Qu55kebNKDj:-Mc9KqXoCXt5EZQQ3870:bvjebdj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b35d50603e444b824f793e1906631fc92f967dfb/src/context/UserContextProvider.tsx#L15)):*

Oh yes, fixed it. I actually noticed the same thing with `DropdownMenu` and changed that too.

Shouldn't / couldn't the linter say something about this btw?
</blockquote></details>

Linter doesn't complain about these right now, but if this becames bigger issue in the future then we could invest some minutes and find out if there is rule for that. :)

By the way, just noticed that this takes `children` prop. Then actually the correct way to type this would be:
`export const UserContextProvider: FunctionComponent = ({ children }) => {` (and then interface Props definition could be entirely dropper, because FunctionComponents include children prop implicitly  :roll_eyes: 

(and to make things even more messy, React has shorthand `React.FC` which is same as FunctionComponent...)

PR #13, opened 2021-06-10
Add apollo client to project

tk-vincit:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/13)


lassemaatta (changes requested):
> Reviewed 1 of 1 files at r1, 15 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/13)* status: 12 of 18 files reviewed, 1 unresolved discussion (waiting on @lassemaatta and @tk-vincit)

___
*[src/generated/graphql.tsx, line 1 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McPNEVMB388iEffNcGE:-McPNEVN8P9HOxS06sc-:b-itg301) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f79a69fea9ad70c1bb7c3e1d3f1a24fb1a471e59/src/generated/graphql.tsx#L1)):*
> ```XML
> import { gql } from '@apollo/client';
> ```

If this is a generated file, why is it added into git? Why can't this be generated in the build / CI process? (Didn't we already have this discussion earlier wrt. hasura?)

jschummer (commented):
Good change set, this will get us way ahead.

Some small comments only, no show stoppers.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-:-McSwGpt3r6JQSp01_f1:bf330md)* status: 12 of 18 files reviewed, 4 unresolved discussions (waiting on @lassemaatta and @tk-vincit)

___
*[.eslintrc.js, line 44 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McSsyIY6Ruvppm8pM1S:-McSsyIY6Ruvppm8pM1T:bg776u1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/.eslintrc.js#L44)):*
> ```
> react/require-default-props
> ```
:+1: Optional properties don't really made sense to me if you have to state that "this property is `undefined` unless it is explicitly defined" 🙄
___
*[package.json, line 18 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McSw17aB9EtDNfwsAY9:-McSw17aB9EtDNfwsAYA:b-2fnftz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/package.json#L18)):*
> ```
> ^3
> ```
Are you sure about the caret? Does dependabot take that into account?
___
*[src/generated/graphql.tsx, line 1 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McPNEVMB388iEffNcGE:-McSqCmQ98CCjrgQg8DY:b5q9iob) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f79a69fea9ad70c1bb7c3e1d3f1a24fb1a471e59/src/generated/graphql.tsx#L1)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

If this is a generated file, why is it added into git? Why can't this be generated in the build / CI process? (Didn't we already have this discussion earlier wrt. hasura?)
</blockquote></details>

If I understood correctly, adding this may be necessary if we want to be able to do frontend development without having hasura up and running locally. This is useful when doing e.g. auth-backend development.

On the other hand, I'm not sure we are going to have other backends and time soon, to which the frontend needs to talk to (other than hasura and auth-backend). Additionally, I'm not sure how easily a frontend dev env can be setup without hasura even though we include this.

In that sense it might an option to generate this on the fly in all cases.
___
*[src/graphql/index.ts, line 43 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McSsDa88zw7rgfl2mhg:-McSsDa88zw7rgfl2mhh:b-mbn88a) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/src/graphql/index.ts#L43)):*
> ```
> GQLClient = client;
> ```
Just out of curiosity: Is there any specific reason `client` is declared locally under a different name than under which it is exported?

Or would
```
export const GQLClient = new ApolloClient({
...
```
work as well?

Definitely no show stopper, just wondering and trying to learn :-)

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/13)* status: 12 of 18 files reviewed, 4 unresolved discussions (waiting on @jschummer, @lassemaatta, and @tk-vincit)

___
*[package.json, line 18 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McSw17aB9EtDNfwsAY9:-McUbptqAx42mH_Vbqic:b-n9bxy4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/package.json#L18)):*

> ```
> ^3
> ```
Are you sure about the caret? Does dependabot take that into account?
</blockquote></details>

That's default behaviour when we install dependencies with like `yarn install @apollo/client`. Also we already have plenty of deps with `^` and  `yarn.lock` should keep us safe - _unless_ dependabot breaks things. Which it shouldn't do as we should have tests to protect us...

So, I think this is fine for now, and if this isn't then we should check also other dependencies, and that is something that should be done in a separate PR.
___
*[src/generated/graphql.tsx, line 1 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McPNEVMB388iEffNcGE:-McUcjGV8eq6Nx73zej0:b17j85b) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f79a69fea9ad70c1bb7c3e1d3f1a24fb1a471e59/src/generated/graphql.tsx#L1)):*

If I understood correctly, adding this may be necessary if we want to be able to do frontend development without having hasura up and running locally. This is useful when doing e.g. auth-backend development.

On the other hand, I'm not sure we are going to have other backends and time soon, to which the frontend needs to talk to (other than hasura and auth-backend). Additionally, I'm not sure how easily a frontend dev env can be setup without hasura even though we include this.

In that sense it might an option to generate this on the fly in all cases.
</blockquote></details>

Yeah, without this frontend can't be even compiled without hasura, which is a bit tricky in e.g. current CI environment.

Also this serves as a documentation about what was the backend like when this particular version of UI was compiled, which _might_ be handy.

I think that when we discussed about this previously we ended up deciding that approach like this is good enough for now. If we decide otherwise, then we also have to decide how to do things in CI (e.g. are we using docker-compose to spin up dev env and are we ready to pay the price in dramatically increased compilation times? That kind of docker-compose might be needed also for local development, but we don't have that right now.)
___
*[src/graphql/index.ts, line 43 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McSsDa88zw7rgfl2mhg:-McUdfMe3QQACP3EjU91:b4flcn5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/src/graphql/index.ts#L43)):*

> ```
> GQLClient = client;
> ```
Just out of curiosity: Is there any specific reason `client` is declared locally under a different name than under which it is exported?

Or would
```
export const GQLClient = new ApolloClient({
...
```
work as well?

Definitely no show stopper, just wondering and trying to learn :-)
</blockquote></details>

I think the name "client" came from a documentation that I was reading when writing this. The way you suggested should work too, so no reasons to prefer this way.

lassemaatta (changes requested):
> Reviewed 1 of 1 files at r3, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/13)* status: all files reviewed, 5 unresolved discussions (waiting on @jschummer and @tk-vincit)

___
*[src/generated/graphql.tsx, line 1 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-McPNEVMB388iEffNcGE:-MchNkpA4SG2P95VlY69:b-837i3v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f79a69fea9ad70c1bb7c3e1d3f1a24fb1a471e59/src/generated/graphql.tsx#L1)):*

Yeah, without this frontend can't be even compiled without hasura, which is a bit tricky in e.g. current CI environment.

Also this serves as a documentation about what was the backend like when this particular version of UI was compiled, which _might_ be handy.

I think that when we discussed about this previously we ended up deciding that approach like this is good enough for now. If we decide otherwise, then we also have to decide how to do things in CI (e.g. are we using docker-compose to spin up dev env and are we ready to pay the price in dramatically increased compilation times? That kind of docker-compose might be needed also for local development, but we don't have that right now.)
</blockquote></details>

Okay.
___
*[src/graphql/index.ts, line 6 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-MchO2UjB9hOOBkSkiQQ:-MchO2UjB9hOOBkSkiQR:b-ykvkhu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/src/graphql/index.ts#L6)):*
> ```
> http://localhost:8080/v1/graphql
> ```

Should these URLs be defined in some centralized location?
___
*[src/graphql/subscriptions.ts, line 5 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-MchRGzD9CK4OSmzHE_G:-MchRGzD9CK4OSmzHE_H:b-wuqxqd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/src/graphql/subscriptions.ts#L5)):*
> ```
> subscrptions
> ```

`subscription`

jschummer (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/13)* status: all files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/13)* status: all files reviewed, 2 unresolved discussions (waiting on @lassemaatta)

___
*[src/graphql/index.ts, line 6 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-MchO2UjB9hOOBkSkiQQ:-MciLK_GDNKlm_YAWhnE:bzbot6m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/src/graphql/index.ts#L6)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

> ```
> http://localhost:8080/v1/graphql
> ```

Should these URLs be defined in some centralized location?
</blockquote></details>

Definitely, at least at some point. But perhaps that doesn't need to happen in this PR.
___
*[src/graphql/subscriptions.ts, line 5 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/13#-MchRGzD9CK4OSmzHE_G:-MciLBAw7cfHy4padpuT:b-cggg7p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/99488611210862d0c66e5be75dbb13c13874d819/src/graphql/subscriptions.ts#L5)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

> ```
> subscrptions
> ```

`subscription`
</blockquote></details>

There will be more than one subscription. At least until we decide to split those in multiple files.

PR #18, opened 2021-06-23
Avoid defining baseUrl in tsconfig

tk-vincit:
Originally baseUrl was set in order to be able to import `render` in
test files without relative paths.
Anyway, because it was defined in project-wide tsconfig it affected
the whole project and caused unexpected errors in development:
e.g. creating `src/leaflet.ts` file in project caused
`import * as L from 'leaflet';` style imports to break, as TS tried
to find those from our local file.

Technically we could define baseUrl in `tsconfig.test.json`, but then
doing project-wide TS type checking would have to be done in two steps
(one for test files, one for everything else) instead of all at once
(as we are doing right now). So for the sake of simplicity it might
be better to do this change now and consider other options if relative
imports become too annoying.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/18)

### DIFF
--- a/src/components/navbar/BrandLogo.tsx
+++ b/src/components/navbar/BrandLogo.tsx
@@ -1,7 +1,7 @@
 import React, { FunctionComponent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
-import { Path, routes } from 'routes';
+import { Path, routes } from '../../routes';
 
 interface Props {
   className?: string;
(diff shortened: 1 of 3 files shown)


jschummer (approved):
Makes sense and the solution looks good to me.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/18#-:-Mcx4ez28Ngac9gwP-sc:b-6utn2t)* status: 0 of 3 files reviewed, all discussions resolved

PR #19, opened 2021-07-02
Draw routes and stops on map with `leaflet.vectorgrid`

tk-vincit:
- Draw digiroad network + stops to map with `leaflet.vectorgrid`
- Add filters for showing/hiding routes and stops

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/19)


jschummer (changes requested):
Nice work!

Thanks for splitting the PR, it was a lot easier to see what is happening with this smaller version.

I found two places, on which I'd appreciate a comment or improvement.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-:-Mh2iZc88Y_5sbXKvrzK:b-lpse5w)* status: 0 of 46 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[package.json, line 7 at r17](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-Mh2fSLJ3F3KceAUw1bB:-Mh2fSLJ3F3KceAUw1bC:bfxxovu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ee7586f4f8b324445ca55c9bb27133183a21e8e8/package.json#L7)):*
> ```
>     "predev": "yarn generate-theme",
> ...
>     "prebuild": "yarn generate-theme",
> ```

Nice you took care of the automatic execution to keep the theme files in sync :+1: 
___
*[src/components/map/Controls.tsx, line 22 at r17](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-Mh2_1LlENvXJm3P1ji0:-Mh2_1Lm5Ji5s5JuVuvV:b8bxqcf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ee7586f4f8b324445ca55c9bb27133183a21e8e8/src/components/map/Controls.tsx#L22)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>

> ```
> enum PositionClass {
...
> }
> export type Position = 'topleft' | 'topright' | 'bottomleft' | 'bottomright';
> const mapPositionToPositionClass = (position: Position): PositionClass => {
...
> };
> ```
</details>


Is there a need for separate `Position` and `PositionClass` types, couldn't the `PositionClass` be used straight in all places which now use `Position`?

And if you don't want to have the style-class info too tightly coupled, you could maybe extract that into the `map`-function? E.g.

```
enum Position {
  TopLeft,
  TopRight,
  ...
}

const mapPositionToClass = (position: Position): ...
```

Also, would it be slightly better readable if the mapping would be realized using a dictionary?
___
*[src/uiComponents/FilterPanel.tsx, line 11 at r17](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-Mh2fqFM246PKtWU1xd1:-Mh2fqFM246PKtWU1xd2:b67flej) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ee7586f4f8b324445ca55c9bb27133183a21e8e8/src/uiComponents/FilterPanel.tsx#L11)):*
> ```
>   // eslint gives errors for "react/no-unused-prop-types" even though these are actually used..?
> ```

Could this be related to the properties being contained in an array which is part of another interface? If so, maybe you could include that info in the comment.

If not, do you have any other idea why eslint is treating this case so strangely?

jschummer (commented):
PS: "splitting" = "reorganizing". I actually first checked the github commits to see how the work was split and then came back here to write the comments. So less commits made this easier to review.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-:-Mh2jG8g1M4jXyOJMov0:b-tfyn1y)* status: 0 of 46 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

tk-vincit (commented):
Fixes posted as fixup, will squash it before merge if fixup is ok.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-:-MhCaXQp8RHE-qfVJkeJ:boclfub)* status: 0 of 46 files reviewed, 3 unresolved discussions (waiting on @jschummer and @tk-vincit)

___
*[src/components/map/Controls.tsx, line 22 at r17](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-Mh2_1LlENvXJm3P1ji0:-MhCUQsa7sOdUZr_2P7i:bc1ribk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ee7586f4f8b324445ca55c9bb27133183a21e8e8/src/components/map/Controls.tsx#L22)):*

> ```
> enum PositionClass {
...
> }
> export type Position = 'topleft' | 'topright' | 'bottomleft' | 'bottomright';
> const mapPositionToPositionClass = (position: Position): PositionClass => {
...
> };
> ```

Is there a need for separate `Position` and `PositionClass` types, couldn't the `PositionClass` be used straight in all places which now use `Position`?

And if you don't want to have the style-class info too tightly coupled, you could maybe extract that into the `map`-function? E.g.

```
enum Position {
  TopLeft,
  TopRight,
  ...
}

const mapPositionToClass = (position: Position): ...
```

Also, would it be slightly better readable if the mapping would be realized using a dictionary?
</blockquote></details>

The reason I crafted component interfaces to use "position strings" like `"topleft"` was that it is how positioning seems to be done with native leaflet components. Example documentation: https://react-leaflet.js.org/docs/api-components/#control-behavior
Giving position as string instead of enum is also more convenient for the user, as now there's no need to import `Position` enum among components that need positioning.

But I agree that switch-case is not elegant solution and this should rather be dictionary, don't understand why I just automatically always tend to write switch cases. 

Now when I look at this again I also started to wonder if PositionClass enum is needed at all (assuming that we want to keep this kind of component interface).

How about now?
___
*[src/uiComponents/FilterPanel.tsx, line 11 at r17](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-Mh2fqFM246PKtWU1xd1:-MhCaQ12B41WimeY1P-j:bdfoyna) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ee7586f4f8b324445ca55c9bb27133183a21e8e8/src/uiComponents/FilterPanel.tsx#L11)):*

> ```
>   // eslint gives errors for "react/no-unused-prop-types" even though these are actually used..?
> ```

Could this be related to the properties being contained in an array which is part of another interface? If so, maybe you could include that info in the comment.

If not, do you have any other idea why eslint is treating this case so strangely?
</blockquote></details>

Added a bit more documentation, how about now?

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/19)* status: 0 of 46 files reviewed, 2 unresolved discussions (waiting on @jschummer and @tk-vincit)

___
*[src/components/map/Controls.tsx, line 21 at r19](https://reviewable.io/reviews/hsldevcom/jore4-ui/19#-MhCnFpb289jmhbnWWHj:-MhDexTP2oLm6zkcCDvA:b8y3hi5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/087550ae87322a1bf30531b764b234fa9fd04ca3/src/components/map/Controls.tsx#L21)):*

> ```
> 'topleft'
> ```
Is it needed to write this as a string? Not a big deal, just generally wondering.
</blockquote></details>

Yes, I think so as this was defined like this: 
`export type Position = 'topleft' | 'topright' | 'bottomleft' | 'bottomright';`

PR #21, opened 2021-07-27
Hasura role-based authorization

jschummer:
- proxy Hasura requests through the dev server to have the browser
  attach the session cookie to each request
- the useQueryWithRole -hook is provided to request a specific role for
  a Hasura graphql query
- the installed authRoleMiddleware will pick up the requested role and
  attach it to the graphql request as a request header

Hasura can then be configured to use the jore4-auth webhook. Hasura will
forward both the session cookie and the requested role header to the
auth backend webhook, which will authorize the request if the user has a
permission matching the specified requested role.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/21)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -15,6 +15,11 @@ module.exports = withImages({
         destination: 'http://localhost:3001/api/:path*',
       },
       {
+        // Proxy graphql requests to hasura
+        source: '/api/hasura/:path*',
+        destination: 'http://localhost:8080/:path*',
+      },
+      {
         // Rewrite everything else to `pages/index`
         source: '/:any*',
         destination: '/',
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.20",
+    "@graphql-typed-document-node/core": "^3.1.0",
     "@headlessui/react": "^1.0.0",
     "@react-leaflet/core": "^1.1.0",
     "axios": "0.21.1",
--- a/src/graphql/auth.ts
+++ b/src/graphql/auth.ts
@@ -1,0 +1,36 @@
+import { ApolloLink, useQuery } from '@apollo/client';
+import { OperationVariables } from '@apollo/client/core';
+import { QueryHookOptions } from '@apollo/client/react/types/types';
+import { TypedDocumentNode } from '@graphql-typed-document-node/core';
+import { DocumentNode } from 'graphql';
+
+const REQUESTED_HASURA_ROLE_HEADER = 'x-hasura-role';
+
+export type Role = string;
+
+export const authRoleMiddleware = new ApolloLink((operation, forward) => {
+  const { role } = operation.variables;
+
+  // add the requested authorization role to the headers if it is specified
+  operation.setContext(({ headers = {} }) => ({
+    headers: {
+      ...headers,
+      ...(role && {
+        [REQUESTED_HASURA_ROLE_HEADER]: role,
+      }),
+    },
+  }));
+
+  return forward(operation);
+});
+
+export function useQueryWithRole<T>(
+  query: DocumentNode | TypedDocumentNode<T, OperationVariables>,
+  role: Role,
+  options?: QueryHookOptions<T>,
+) {
+  return useQuery(query, {
+    ...options,
+    variables: { ...options?.variables, role },
+  });
+}
--- a/src/graphql/index.ts
+++ b/src/graphql/index.ts
@@ -1,10 +1,19 @@
-import { ApolloClient, HttpLink, InMemoryCache, split } from '@apollo/client';
+import {
+  ApolloClient,
+  concat,
+  HttpLink,
+  InMemoryCache,
+  split,
+} from '@apollo/client';
 import { WebSocketLink } from '@apollo/client/link/ws';
 import { getMainDefinition } from '@apollo/client/utilities';
+import { authRoleMiddleware } from './auth';
 
 const httpLink = new HttpLink({
-  uri: 'http://localhost:8080/v1/graphql',
+  uri: 'http://localhost:3000/api/hasura/v1/graphql',
 });
+
+const apolloLink = concat(authRoleMiddleware, httpLink);
 
 // because next.js might run this on server-side and websockets aren't
 // supported there, we have to check if we are on browser before
@@ -29,9 +38,9 @@ const link = process.browser
       },
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       wsLink!,
-      httpLink,
+      apolloLink,
     )
-  : httpLink;
+  : apolloLink;
 
 const cache = new InMemoryCache({
   typePolicies: {
@@ -51,3 +60,5 @@ const client = new ApolloClient({
 });
 
 export const GQLClient = client;
+
+export * from './auth';
--- a/yarn.lock
+++ b/yarn.lock
@@ -1267,7 +1267,7 @@
     tslib "~2.2.0"
     value-or-promise "1.0.6"
 
-"@graphql-typed-document-node/core@^3.0.0":
+"@graphql-typed-document-node/core@^3.0.0", "@graphql-typed-document-node/core@^3.1.0":
   version "3.1.0"
   resolved "https://registry.yarnpkg.com/@graphql-typed-document-node/core/-/core-3.1.0.tgz#0eee6373e11418bfe0b5638f654df7a4ca6a3950"
   integrity sha512-wYn6r8zVZyQJ6rQaALBEln5B1pzxb9shV5Ef97kTvn6yVGrqyXVnDqnU24MXnFubR+rZjBY9NWuxX3FB2sTsjg==


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/21)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @jschummer)

___
*[next.config.js, line 18 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5L2lgEvGKGx6GU1B_:-Mg5L2lgEvGKGx6GU1Ba:bmi7i3i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/next.config.js#L18)):*
> ```JavaScript
>       },
>       {
>         // Proxy graphql requests to hasura
> ```

Just realised that these rewrites won't work on server builds (https://nextjs.org/docs/api-reference/next.config.js/introduction) - have you already considered how to do this kind of redirects in production?
___
*[src/graphql/index.ts, line 1 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5O9WS0WwY8iN-vIdO:-Mg5O9WS0WwY8iN-vIdP:b-gd21vz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L1)):*
> ```XML
> import {
> ```

Maybe we should create `src/graphql/auth.ts` file and put these auth-related things there. 

If we still want to expose those from this file (for prettier imports, `import { xxx } from './graphql'` instead of `import { xxx } from './graphql/auth'` we could still add `export * from './auth'` to this file.
___
*[src/graphql/index.ts, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5LjJ1-zTN-87erHu-:-Mg5LjJ1-zTN-87erHu0:b39eowg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L20)):*
> ```XML
> 
> // eslint-disable-next-line  @typescript-eslint/no-explicit-any
> export function useQueryWithRole<TData = any>(
> ```

Does `TData` name give any additional documentation when compared to e.g. just `T`?
___
*[src/graphql/index.ts, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5LotC6SDWkcwX3GmX:-Mg5LotC6SDWkcwX3GmY:b7r6fzc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L20)):*
> ```XML
> 
> // eslint-disable-next-line  @typescript-eslint/no-explicit-any
> export function useQueryWithRole<TData = any>(
> ```

Question: do we want to set default type to `any` here, or would it be better to call this explicitly as `useQueryWithRole<any>` if needed?

The better choice will depend on the fact how often this is is actually called with <any>. If we most likely call this with type param, then its best to omit the defautl `any` value as it makes it easier to forget giving type parameter to this function. But if this is called often with `any`, then default value might make sense as then we don't have to disable linter in multiple places.
___
*[src/graphql/index.ts, line 22 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5MgXX3ryrlP3QDIX8:-Mg5MgXX3ryrlP3QDIX9:b-jgkr2h) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L22)):*
> ```XML
> export function useQueryWithRole<TData = any>(
>   query: DocumentNode | TypedDocumentNode<TData, OperationVariables>,
>   role: string,
> ```

Should we have separate type for role? In the beginning it could be just something like `type Role = string;`, but in the near future it would probably look something like `type Role = 'admin' | 'regular';` (or more likely be enum value).
___
*[src/graphql/index.ts, line 36 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5OqNz1Ma2a7tn992X:-Mg5OqNz1Ma2a7tn992Y:b61cfx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L36)):*
> ```XML
>     headers: {
>       ...headers,
>       ...(operation.variables.role && {
> ```

This expression is difficult to read.
At least it could be simplified like this, but its still far from ideal:
```
const { role } = operation.variables;

...

...{  role && { [REQUESTED_HASURA_ROLE_HEADER]: role } } 
```
With `immer` this could be simpler (especially after you learn how to use it), something like this:

```
const { role } = operation.variables;

const updatedHeaders = produce(headers, draft => {
  if (role) {
    draft[REQUESTED_HASURA_ROLE_HEADER]: role
  }
})

```

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/21)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @tk-vincit)

___
*[next.config.js, line 18 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5L2lgEvGKGx6GU1B_:-Mg_5VKB9_hucNhF37yl:b-yx6rx9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/next.config.js#L18)):*

Just realised that these rewrites won't work on server builds (https://nextjs.org/docs/api-reference/next.config.js/introduction) - have you already considered how to do this kind of redirects in production?
</blockquote></details>

Yes, in production, API gateway will handle this (IIRC we already talked about this in some meeting some weeks ago).

So as you noticed, this part is only to get the whole thing to work at all in the development environment.
___
*[src/graphql/index.ts, line 1 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5O9WS0WwY8iN-vIdO:-Mg_bmjJ6kMGT66mOjM-:b-nhzms1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L1)):*

Maybe we should create `src/graphql/auth.ts` file and put these auth-related things there. 

If we still want to expose those from this file (for prettier imports, `import { xxx } from './graphql'` instead of `import { xxx } from './graphql/auth'` we could still add `export * from './auth'` to this file.
</blockquote></details>

True, I now put `useQueryWithRole` in `auth.ts` and exported the way you suggested. Probably other similar functions will be created there in the future (for mutations at least), so this was a good idea :+1: 
___
*[src/graphql/index.ts, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5LjJ1-zTN-87erHu-:-Mg_670eDzKEHnut0yDZ:blp1vp6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L20)):*

Does `TData` name give any additional documentation when compared to e.g. just `T`?
</blockquote></details>

The thing is that I tried to do this as generic as the encapsuled `useQuery` function and also used the same type parameters where possible. The original `useQuery` uses also `TOptions`, that's why the `TData` is separate.

If you think that the reference to the `useQuery`'s type parameter is not needed, this can be of course renamed to `T`. I don't have any strong opinion on this, I changed it to `T` for now.
___
*[src/graphql/index.ts, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5LotC6SDWkcwX3GmX:-Mg_7JVF4qzL4wRuhk1d:b-ws758h) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L20)):*

Question: do we want to set default type to `any` here, or would it be better to call this explicitly as `useQueryWithRole<any>` if needed?

The better choice will depend on the fact how often this is is actually called with <any>. If we most likely call this with type param, then its best to omit the defautl `any` value as it makes it easier to forget giving type parameter to this function. But if this is called often with `any`, then default value might make sense as then we don't have to disable linter in multiple places.
</blockquote></details>

The same as above goes, tried to make this as much `useQuery`-looking as I could. Took out the `any` out of there now.
___
*[src/graphql/index.ts, line 22 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5MgXX3ryrlP3QDIX8:-Mg_7TaB5DtmuRFo4dff:b-ffqldo) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L22)):*

Should we have separate type for role? In the beginning it could be just something like `type Role = string;`, but in the near future it would probably look something like `type Role = 'admin' | 'regular';` (or more likely be enum value).
</blockquote></details>

This is a good point, changed this in `auth.ts`!
___
*[src/graphql/index.ts, line 36 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5OqNz1Ma2a7tn992X:-Mg_XPeP6Vj0jZlBXalp:b-u51z7x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L36)):*

This expression is difficult to read.
At least it could be simplified like this, but its still far from ideal:
```
const { role } = operation.variables;

...

...{  role && { [REQUESTED_HASURA_ROLE_HEADER]: role } } 
```
With `immer` this could be simpler (especially after you learn how to use it), something like this:

```
const { role } = operation.variables;

const updatedHeaders = produce(headers, draft => {
  if (role) {
    draft[REQUESTED_HASURA_ROLE_HEADER]: role
  }
})

```
</blockquote></details>

After reading about `immer`, I agree that it would be nice to use it here.

However, when I tried it, I ran into this TS issue: https://github.com/immerjs/immer/issues/710

So for now, I only decomposed the `variables` as you suggested above.

tk-vincit (approved):
One new question. LGTM.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-:-MgezSDEDnF1r1JLqC9i:b486n37)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @jschummer and @tk-vincit)

___
*[next.config.js, line 18 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5L2lgEvGKGx6GU1B_:-MgexKgo94wOD73_y3j4:b2ol6p7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/next.config.js#L18)):*

Yes, in production, API gateway will handle this (IIRC we already talked about this in some meeting some weeks ago).

So as you noticed, this part is only to get the whole thing to work at all in the development environment.
</blockquote></details>

 👍
___
*[src/graphql/index.ts, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5LjJ1-zTN-87erHu-:-MgeyO5w4YvzCC64774X:bnwm555) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L20)):*

The thing is that I tried to do this as generic as the encapsuled `useQuery` function and also used the same type parameters where possible. The original `useQuery` uses also `TOptions`, that's why the `TData` is separate.

If you think that the reference to the `useQuery`'s type parameter is not needed, this can be of course renamed to `T`. I don't have any strong opinion on this, I changed it to `T` for now.
</blockquote></details>

Ok. Didin't looked into typings of `useQuery` when commenting about this. :) I think both ways are OK.
___
*[src/graphql/index.ts, line 22 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5MgXX3ryrlP3QDIX8:-MgeycSa0hdn2gZcKZkR:btojbfd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L22)):*

This is a good point, changed this in `auth.ts`!
</blockquote></details>

Great!
___
*[src/graphql/index.ts, line 36 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mg5OqNz1Ma2a7tn992X:-MgezGNE3XRBVpPxWJ_j:b-d9m9ek) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1c18edcbf18bde64ac497b8515ad324094dba257/src/graphql/index.ts#L36)):*

After reading about `immer`, I agree that it would be nice to use it here.

However, when I tried it, I ran into this TS issue: https://github.com/immerjs/immer/issues/710

So for now, I only decomposed the `variables` as you suggested above. 
</blockquote></details>

Ok. We can investigate that immer issue later. 
___
*[src/graphql/index.ts, line 14 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mgez0tO8_cx6wNC1Rrj:-Mgez0tP7nTL0OCFfS50:bb5r7kg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5ef3c6e1559c09868f0c4849d45c4f7c2d829a9e/src/graphql/index.ts#L14)):*
> ```XML
> const REQUESTED_HASURA_ROLE_HEADER = 'x-hasura-role';
> 
> const authRoleMiddleware = new ApolloLink((operation, forward) => {
> ```

Dunno if this would also belong to `auth.ts` file? Both ways are fine for me.

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/21)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[src/graphql/index.ts, line 14 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/21#-Mgez0tO8_cx6wNC1Rrj:-Mgf96bQ-G3Vvw4vFZjS:b4k3cl4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5ef3c6e1559c09868f0c4849d45c4f7c2d829a9e/src/graphql/index.ts#L14)):*

Dunno if this would also belong to `auth.ts` file? Both ways are fine for me.
</blockquote></details>

I think you have a point, it is cleaner if this is also put in `auth.ts`. After all, it sort of is the "opposite" of `useQueryWithRole`. Changed it now.

PR #27, opened 2021-09-03
Route listing

tk-vincit:
Resolves https://github.com/HSLdevcom/jore4/issues/443

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/27)


jschummer (changes requested):
Nice work!

I appreciate that you put the whitespace change in its own commit (even though I had to go over to github to actually notice...).

I'm still wondering if we can somehow prevent the "route" confusion. But I guess the word has a strong background in both transportation and web contexts, so there might not be any good solution.

If you could check that one commit message and comment on the state keeping, we'll get this one through soon :+1: 

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-:-Miu0RI54s0wzPNcddD2:bp1g777)* status: 0 of 25 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[-- commits, line 22 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-MitxxEyEuT_8iJC3R6t):*
> ```
> Prevent
> ```
If I saw correctly, the change mentioned here makes it unnecessary to set the width explicitly (by using `inline-block`).

Is "prevent" really the right word here? Maybe "Do not set width ... explicitly" would be better?
___
*[src/routes.ts, line 51 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-Mityfil2nuc--wuMT6f:-Mityfil2nuc--wuMT6g:bqwxsya) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9a5b3d1b142c0dbadd61be05cacad7e5063b6c62/src/routes.ts#L51)):*
> ```
> routes.routes
> ```
😬 Not that I'd be able to come up with a better term. I guess we just have to live with this duplicate meaning of the word "route" :smile: 
___
*[src/components/mapgl-map/Map.tsx, line 33 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-MitzY0a-b2xxscILp3L:-MitzY0a-b2xxscILp3M:b6a0v5i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9a5b3d1b142c0dbadd61be05cacad7e5063b6c62/src/components/mapgl-map/Map.tsx#L33)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>   const [showDynamicInfraLinks, setShowDynamicInfraLinks] = useState(
>     !routeSelected,
>   );
>   const [showInfraLinks, setShowInfraLinks] = useState(!routeSelected);
>   const [showRoute, setShowRoute] = useState(routeSelected);
>   const [showStops, setShowStops] = useState(!routeSelected);
> ```
</details>

Do we want to store this state also when the user
1. has a route selected
2. "unselects" the route
3. selects another route?

Might well be that this is wanted behaviour, I'm just not sure if it is.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/27)* status: 0 of 25 files reviewed, 2 unresolved discussions (waiting on @jschummer)

___
*[-- commits, line 22 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-MitxxEyEuT_8iJC3R6t):*

> ```
> Prevent
> ```
If I saw correctly, the change mentioned here makes it unnecessary to set the width explicitly (by using `inline-block`).

Is "prevent" really the right word here? Maybe "Do not set width ... explicitly" would be better?
</blockquote></details>

You are 100% correct. I guess I should have said "Avoid"...
___
*[src/routes.ts, line 51 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-Mityfil2nuc--wuMT6f:-MiuW-2z9tWetJYNmVFx:b-hnnuq5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9a5b3d1b142c0dbadd61be05cacad7e5063b6c62/src/routes.ts#L51)):*

> ```
> routes.routes
> ```
😬 Not that I'd be able to come up with a better term. I guess we just have to live with this duplicate meaning of the word "route" :smile: 
</blockquote></details>

Yeah... This is not perfect, but I don't know better naming conventions for these either. :sweat_smile: 
___
*[src/components/mapgl-map/Map.tsx, line 33 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-MitzY0a-b2xxscILp3L:-MiuWheT2U7nWtyWaW2F:b-b87g16) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9a5b3d1b142c0dbadd61be05cacad7e5063b6c62/src/components/mapgl-map/Map.tsx#L33)):*

> ```
>   const [showDynamicInfraLinks, setShowDynamicInfraLinks] = useState(
>     !routeSelected,
>   );
>   const [showInfraLinks, setShowInfraLinks] = useState(!routeSelected);
>   const [showRoute, setShowRoute] = useState(routeSelected);
>   const [showStops, setShowStops] = useState(!routeSelected);
> ```
Do we want to store this state also when the user
1. has a route selected
2. "unselects" the route
3. selects another route?

Might well be that this is wanted behaviour, I'm just not sure if it is.
</blockquote></details>

Those only set _default_ state of those toggles.

I guess there's no single "correct" way to do this, but I think that current solution is quite logical:
- if user selected route to show, don't show digiroad network / stops by default (so that they won't hide route)
- if user hasn't route to show, show digiroad network and stops by default
- in any case, let user toggle visibility of digiroad network and/or stops and/or route (if available) as she wants

(this might change in the future, but now we only have example data available anyway so its probably not worth adjusting this too much at this point.)

tk-vincit (commented):
Changed "Prevent" to "Avoid" already. :wink: 

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/27#-:-Mj-1QpZ6Ayi7KbPYkbt:b-9km7u2)* status: 0 of 25 files reviewed, 1 unresolved discussion (waiting on @jschummer)

PR #39, opened 2021-10-14
Check generated schema

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/39)

### DIFF
--- a/.github/workflows/check-generated-resources.yml
+++ b/.github/workflows/check-generated-resources.yml
@@ -1,0 +1,39 @@
+name: Check generated resources
+
+on:
+  pull_request:
+
+jobs:
+  check_generated_graphql_schema:
+    name: Check if the generated graphql schema is up to date
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup caching for node_modules
+        uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Set up docker dependencies (in the background)
+        run: ./start-dependencies.sh &
+
+      - name: Install node_modules
+        # making sure that we do have the proper set of node_modules installed regardless of what's cached
+        run: yarn install --frozen-lockfile
+
+      - name: Wait for Hasura to be available
+        uses: HSLdevcom/jore4-tools/github-actions/healthcheck@healthcheck-v1
+        with:
+          command: 'curl --fail http://localhost:3201/healthz --output /dev/null --silent'
+          # hasura takes time to start up as it depends on the database
+          retries: 50
+
+      - name: Regenerate schemas
+        run: yarn gql:generate-types
+
+      - name: Verify that generate command made no changes, so schema is up to date
+        run: '[ -z "$(git status --porcelain)" ]'
--- a/docker/docker-compose.custom.yml
+++ b/docker/docker-compose.custom.yml
@@ -1,5 +1,7 @@
 services:
   jore4-hasura:
+    # locking hasura image so that we can develop the UI against a static graphql API
+    image: 'hsldevcom/jore4-hasura:main--e4f5c5ad2dd1530207300c2d487de70fccb70431'
     environment:
       HASURA_GRAPHQL_ENABLE_CONSOLE: 'true'
 
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -5413,6 +5413,29 @@
         "possibleTypes": null
       },
       {
+        "kind": "INPUT_OBJECT",
+        "name": "infrastructure_network_resolve_point_to_closest_link_args",
+        "description": null,
+        "fields": null,
+        "inputFields": [
+          {
+            "name": "geog",
+            "description": null,
+            "type": {
+              "kind": "SCALAR",
+              "name": "geography",
+              "ofType": null
+            },
+            "defaultValue": null,
+            "isDeprecated": false,
+            "deprecationReason": null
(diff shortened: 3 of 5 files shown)


tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/39)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[.github/workflows/check-generated-resources.yml, line 26 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/39#-MoDoTF3-2tkpnB9l3Tk:-MoDoTF3-2tkpnB9l3Tl:byfkl3t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1b467fa6f10af4c2762c5103b1ae773c19437ccd/.github/workflows/check-generated-resources.yml#L26)):*
> ```YAML
> 
>       - name: Install node_modules
>         if: steps.yarn-cache.outputs.cache-hit != 'true'
> ```

So this runs yarn install only if cache is found on certain runner? What if cache is empty, do we install dependencies at all?

Also I'm not sure when the cache is invalidated. Anyway, it might be safe to never invalidate cache and always run `yarn install` and always update cache. Yarn install removes unneeded packages if needed, and majos changes to dependencies happen quite rarely. And all this is based on assumption that github's cache is significantly faster than plain `yarn install`...
___
*[docker/docker-compose.custom.yml, line 3 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/39#-MoDntO51k0_cWDINf1r:-MoDntO51k0_cWDINf1s:b-6b2x6t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1b467fa6f10af4c2762c5103b1ae773c19437ccd/docker/docker-compose.custom.yml#L3)):*
> ```YAML
> services:
>   jore4-hasura:
>     image: 'hsldevcom/jore4-hasura:main--e4f5c5ad2dd1530207300c2d487de70fccb70431'
> ```

Should you have documented why we use certain image here? ("Lock hasura version to prevent UI from breaking at arbitrariry moment if hasura's schema happens to change" or something like that)

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/39)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[.github/workflows/check-generated-resources.yml, line 26 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/39#-MoDoTF3-2tkpnB9l3Tk:-MoDszUq0jWZdg0uj6nO:b-e293x2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1b467fa6f10af4c2762c5103b1ae773c19437ccd/.github/workflows/check-generated-resources.yml#L26)):*

So this runs yarn install only if cache is found on certain runner? What if cache is empty, do we install dependencies at all?

Also I'm not sure when the cache is invalidated. Anyway, it might be safe to never invalidate cache and always run `yarn install` and always update cache. Yarn install removes unneeded packages if needed, and majos changes to dependencies happen quite rarely. And all this is based on assumption that github's cache is significantly faster than plain `yarn install`...
</blockquote></details>

cache does not seem to be dependent on the runner. If cache is empty then it's not "hit" so the yarn install would trigger

to be safe, I removed the check and do a yarn install nevertheless of the cache state. I had read on forums that this triggers a full reinstall of node_modules, but I guess they just didn't use the frozen-lockfile parameter... I just tested and the yarn install only takes a single second so it does in fact recognize that all the dependencies are up to date
___
*[docker/docker-compose.custom.yml, line 3 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/39#-MoDntO51k0_cWDINf1r:-MoDsOQy0395xmS-8qld:bb7f2mf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1b467fa6f10af4c2762c5103b1ae773c19437ccd/docker/docker-compose.custom.yml#L3)):*

Should you have documented why we use certain image here? ("Lock hasura version to prevent UI from breaking at arbitrariry moment if hasura's schema happens to change" or something like that)
</blockquote></details>

done

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/39)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @pallost and @tk-vincit)

___
*[.github/workflows/check-generated-resources.yml, line 26 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/39#-MoDoTF3-2tkpnB9l3Tk:-MoDyqip4QSk1c75F59f:b4mve9k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1b467fa6f10af4c2762c5103b1ae773c19437ccd/.github/workflows/check-generated-resources.yml#L26)):*

cache does not seem to be dependent on the runner. If cache is empty then it's not "hit" so the yarn install would trigger

to be safe, I removed the check and do a yarn install nevertheless of the cache state. I had read on forums that this triggers a full reinstall of node_modules, but I guess they just didn't use the frozen-lockfile parameter... I just tested and the yarn install only takes a single second so it does in fact recognize that all the dependencies are up to date
</blockquote></details>

`yarn install` utilizes previously installed  `node_modules`, and `--frozen-lockfile` is not related to that any way. `--forzen-lockfile` just checks that lockfile matches package.json and full install can be done according to lockfile.


"I had read on forums that this triggers a full reinstall of node_modules, ..." This is false. You can try this locally:
- rm node_modules
- yarn install (makes full install, takes some time)
- yarn install (just checkes that everythnig is up-to-date, runs pretty quickly)

PR #42, opened 2021-11-05
Use `react-hook-form` + `zod` instead for `formik` + `yup`

tk-vincit:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/42)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -55,6 +55,15 @@ module.exports = {
       'error',
       { extensions: ['.tsx', '.ts'] },
     ],
+    'react/jsx-props-no-spreading': [
+      'error',
+      // allow spreading props in...
+      // - form-related fields as it seems to be convention
+      //   with `react-hook-form`'s `register` method: https://react-hook-form.com/get-started
+      // - in certain `react-map-gl` related components as it seems to be
+      //   convention also with those: https://visgl.github.io/react-map-gl/docs/api-reference/layer
+      { exceptions: ['input', 'Layer'] },
+    ],
     'import/extensions': [
       'error',
       'ignorePackages',
--- a/package.json
+++ b/package.json
@@ -21,10 +21,10 @@
     "@apollo/client": "^3.4.11",
     "@graphql-typed-document-node/core": "^3.1.0",
     "@headlessui/react": "^1.0.0",
+    "@hookform/resolvers": "^2.8.3",
     "@seznam/compose-react-refs": "^1.0.6",
     "@turf/area": "^6.5.0",
     "axios": "0.21.1",
-    "formik": "^2.2.9",
     "graphql": "^15.5.3",
     "i18next": "^20.2.1",
     "immer": "^9.0.5",
@@ -34,13 +34,14 @@
     "qs": "^6.10.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "react-hook-form": "^7.18.1",
     "react-i18next": "^11.8.12",
     "react-icons": "^4.3.1",
     "react-map-gl": "^6.1.16",
     "react-map-gl-draw": "0.21.1",
     "react-router-dom": "^5.2.0",
     "react-spinners": "^0.11.0",
-    "yup": "^0.32.11"
+    "zod": "^3.11.6"
   },
   "devDependencies": {
     "@graphql-codegen/cli": "^1.21.5",
--- a/src/components/forms/StopForm.tsx
+++ b/src/components/forms/StopForm.tsx
@@ -10,14 +10,19 @@ const FormSchema = yup.object().shape({
   lng: yup.number().min(-180).max(180).required(),
 });
 
-const initialValues = { finnishName: '', lat: '', lng: '' };
+export type FormState = yup.InferType<typeof FormSchema>;
 
 interface Props {
   className?: string;
-  onChange: (values: typeof initialValues) => void;
+  initialValues: FormState;
+  onChange: (values: FormState) => void;
 }
 
-export const StopForm = ({ className, onChange }: Props): JSX.Element => {
+export const StopForm = ({
+  className,
+  initialValues,
+  onChange,
+}: Props): JSX.Element => {
   const { t } = useTranslation();
   return (
     <Formik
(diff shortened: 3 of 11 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/42)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[-- commits, line 16 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/42#-MnyDBoP5DWxRjxxY5NL):*
good! I've always wanted to try `zod`...
___
*[src/components/forms/StopForm.tsx, line 14 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/42#-MnyDZNh4fTFows578gR:-MnyDZNh4fTFows578gS:bx3ahm0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6f46efcf1f77276cbf55645ccfb09789c8664ba6/src/components/forms/StopForm.tsx#L14)):*
> ```XML
> });
> 
> export type FormState = z.infer<typeof schema>;
> ```

I don't think you need to export this as it's only used internally. However if you do want to use this FormState outside, let's give it some more unique naming, like StopFormState
___
*[src/components/forms/StopForm.tsx, line 35 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/42#-MnyE1-PBPWtWkJDQGZM:-MnyE1-PBPWtWkJDQGZN:b-nw6vqa) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6f46efcf1f77276cbf55645ccfb09789c8664ba6/src/components/forms/StopForm.tsx#L35)):*
> ```XML
>   });
> 
>   const onSubmit = (data: FormState) =>
> ```

if it takes the values from outside, it should also outsource the output callbacks as well (like `onSubmit`, `onCancel`, etc). So either all form inputs and outputs are handled inside or all are handled outside.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/42)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[src/components/forms/StopForm.tsx, line 14 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/42#-MnyDZNh4fTFows578gR:-Mnyr1LY9VZ3FLSGsH7t:b-59gz2o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6f46efcf1f77276cbf55645ccfb09789c8664ba6/src/components/forms/StopForm.tsx#L14)):*

I don't think you need to export this as it's only used internally. However if you do want to use this FormState outside, let's give it some more unique naming, like StopFormState
</blockquote></details>

This is used also externally. Personally I prefer this kind of naming, as it makes easier to keep form implementations consistent (less likely to make errors if copypasting existing form definition as a base for new form). Also name collisions are unlikely as usually there is only one form in one view component, but if needed you can still rename export in the import phase: `import { FormState as StopFormState, StopForm } from '../forms/StopForm';`
___
*[src/components/forms/StopForm.tsx, line 35 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/42#-MnyE1-PBPWtWkJDQGZM:-MnyqX-h6pqeLqNRaFP7:bgak48q) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6f46efcf1f77276cbf55645ccfb09789c8664ba6/src/components/forms/StopForm.tsx#L35)):*

if it takes the values from outside, it should also outsource the output callbacks as well (like `onSubmit`, `onCancel`, etc). So either all form inputs and outputs are handled inside or all are handled outside.
</blockquote></details>

Not sure if I understand this comment completely. This form handles its own state completely. Just the initial values are passed from outside.

This kind of architechture (form handles all) seems to be the way to go with `react-hook-form`.

How would you want to get initial values here if not passed from outside?

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/42)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @pallost and @tk-vincit)

___
*[src/components/forms/StopForm.tsx, line 35 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/42#-MnyE1-PBPWtWkJDQGZM:-MnyzD_L15TeuqOQSMpM:b-hxlwyl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6f46efcf1f77276cbf55645ccfb09789c8664ba6/src/components/forms/StopForm.tsx#L35)):*

Not sure if I understand this comment completely. This form handles its own state completely. Just the initial values are passed from outside.

This kind of architechture (form handles all) seems to be the way to go with `react-hook-form`.

How would you want to get initial values here if not passed from outside?
</blockquote></details>

if not passed from outside, the values could be initialized by fetching them directly from the app store.
But well, the boundaries are not that strict... also I just remembered that we are using graphql in this project to the boundaries of responsibilities of components is blurred even more...

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/42)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[src/components/forms/StopForm.tsx, line 35 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/42#-MnyE1-PBPWtWkJDQGZM:-Mnz38PS0BzPnzLPtmKe:bwkzuid) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6f46efcf1f77276cbf55645ccfb09789c8664ba6/src/components/forms/StopForm.tsx#L35)):*

if not passed from outside, the values could be initialized by fetching them directly from the app store.
But well, the boundaries are not that strict... also I just remembered that we are using graphql in this project to the boundaries of responsibilities of components is blurred even more... 
</blockquote></details>

At least in this case stop locations in map are not stored in global app state. Let's see how things are going to evolve in the future, but right now this is the way to go.

tk-vincit:
E2e tests are broken for the reasons unrelated to this PR, so let's merge this now.

PR #55, opened 2021-12-27
Route creation improvements

jschummer:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55)


jarkkoka (changes requested):
It seemed to me that stop direction is not yet taken into account. Is this correct observation?

> Reviewed 1 of 5 files at r1, 3 of 3 files at r2, 3 of 4 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, 1 of 2 files at r6, 1 of 1 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-:-MsoWyjD8U6QG4MGVDEH:b-deklrs)* status: all files reviewed, 3 unresolved discussions (waiting on @jschummer)

___
*[src/components/map/DrawRouteLayer.tsx, line 119 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoVPt92Z2Z0dKTeOCL:-MsoVPtA8QqU8iR_sya-:b4r3hfr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L119)):*
> ```XML
>           (item: ExplicitAny, index: number) => ({
>             infrastructure_link_id: item.infrastructure_link_id,
>             is_traversal_forwards:
> ```

According to usual code styling conventions the object field names/keys should be like "infrastructureLinkId" (as opposed to "inrastructure_link_id").



_Code quote:_
```XML
            infrastructure_link_id: item.infrastructure_link_id,
            is_traversal_forwards:
              routeResponse.routes[0]?.paths[index]?.traversalForwards,
          }),
```
___
*[src/components/map/DrawRouteLayer.tsx, line 136 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoW53a04D0O0hIXzHf:-MsoW53a04D0O0hIXzHg:bbwf3jn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L136)):*
> ```XML
>             infraLink.infrastructure_link_id ===
>             stop.located_on_infrastructure_link_id,
>         );
> ```

It seems to me that stops are not filtered by their traffic flow direction at this point. Stops should be matched with the direction of infrastructure link that is extracted from the `routeResponse`.



_Code quote:_
```XML
      const stopsResponse = await fetchStopsByInfraLinkIds({
        infraLinkIds: infraLinks.map(
          (link: ExplicitAny) => link.infrastructure_link_id,
        ),
      });
      const findStopIndexByInfraLink = (
        stop: ServicePatternScheduledStopPoint,
      ) =>
        infraLinks.findIndex(
          (infraLink: InfrastructureNetworkInfrastructureLink) =>
            infraLink.infrastructure_link_id ===
            stop.located_on_infrastructure_link_id,
        );
      // Sort the returned stops by the infrastructure link order found in the route. This is needed, because the
      // stops are returned in arbitrary order by fetchStopsByInfraLinkIds.
      const stopsWithinRoute =
        // @ts-expect-error problem with generated types?
        [...stopsResponse.data.service_pattern_scheduled_stop_point].sort(
          (
            stop1: ServicePatternScheduledStopPoint,
            stop2: ServicePatternScheduledStopPoint,
          ) =>
            findStopIndexByInfraLink(stop1) - findStopIndexByInfraLink(stop2),
        );
```
___
*[src/components/map/ModalMap.tsx, line 50 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoW_oA6NoFdGpWm60w:-MsoW_oA6NoFdGpWm60x:b-f7u10o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/ModalMap.tsx#L50)):*
> ```XML
>     ) {
>       // TODO: user should be able to select starting stop and final stop from some kind of UI.
>       // TODO: for now, just use the first and last stops found.
> ```

I think this is the only correct solution when selecting terminus stops.



_Code quote:_
```XML
TODO: for now, just use the first and last stops found.
```

jschummer (commented):
It actually was correct, yes.

This observation of yours also made me realise that we didn't have any constraint to check this yet. Luckily, it was not too difficult to add this to the constraint, which was already checking the other route <-> journey pattern references. Also, I did add the relevant front end code here, so the stops "on the wrong side of the road" are filtered out now.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-:-Mt1eisC44yifJHHPZSF:bcn8bke)* status: 10 of 15 files reviewed, 3 unresolved discussions (waiting on @jarkkoka)

___
*[src/components/map/DrawRouteLayer.tsx, line 119 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoVPt92Z2Z0dKTeOCL:-Mt1fMpF9ikUdtQwgqMo:bgoj5ab) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L119)):*

According to usual code styling conventions the object field names/keys should be like "infrastructureLinkId" (as opposed to "inrastructure_link_id").
</blockquote></details>

This was actually utilizing some graphql-generated type below, even though it wasn't declared here. That's where the snake case originated from.

But you do have a good point in that it anyway wasn't good style utilizing that type indirectly (the object created here is not sent via graphql), so I created an own intermediate type with the better camel case, like you suggested.
___
*[src/components/map/DrawRouteLayer.tsx, line 136 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoW53a04D0O0hIXzHf:-Mt1gVLm2bJoaABA6Wb8:ba73ojw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L136)):*

It seems to me that stops are not filtered by their traffic flow direction at this point. Stops should be matched with the direction of infrastructure link that is extracted from the `routeResponse`.
</blockquote></details>

Yes, this was true, but it fixed now. Also, the route <-> journey pattern constraint (other PR in hasura repo) is checking this now.
___
*[src/components/map/ModalMap.tsx, line 50 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoW_oA6NoFdGpWm60w:-Mt1gfbW7wrf1KZDyyLu:b9w5pu7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/ModalMap.tsx#L50)):*

I think this is the only correct solution when selecting terminus stops.
</blockquote></details>

Yes, but in theory these don't have to be the terminus stops of the journey pattern. The route could start at a certain stop, which is not part of the journey pattern (no passengers can enter). Not sure if this will find a real application later on, but in theory it's possible.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55)* status: 10 of 15 files reviewed, 3 unresolved discussions (waiting on @jarkkoka)

___
*[src/components/map/DrawRouteLayer.tsx, line 119 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoVPt92Z2Z0dKTeOCL:-Mt6rB8N2y_s_LNF4VvL:bym88hl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L119)):*

This was actually utilizing some graphql-generated type below, even though it wasn't declared here. That's where the snake case originated from.

But you do have a good point in that it anyway wasn't good style utilizing that type indirectly (the object created here is not sent via graphql), so I created an own intermediate type with the better camel case, like you suggested.
</blockquote></details>

I do see a reason to do mapping between the graphql entities and the UI entities. The main reason however is the possible incompatibility between the graphql generated type and the actual intended type (e.g. string vs UUID).

Instead of bloating the display components with the API-related mapping functions, let's rather have these in the `graphql` folder. So e.g. you could create an `graphql/infrastructureNetwork.ts` file and add a `mapGQLInfrastructureLink()` function there. It'll also allow us to use the same mapping function in other components as well. For now, I'd also have the UI type definition in this same file, e.g.:
```
export type InfrastructureLink {
        infrastructureLinkId: string;
        isTraversalForwards: boolean;
      };
```

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55)* status: 10 of 15 files reviewed, 5 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[src/components/map/DrawRouteLayer.tsx, line 112 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6sRDm-Fi7JxBSzBt8:-Mt6sRDm-Fi7JxBSzBt9:b-blzfiw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/components/map/DrawRouteLayer.tsx#L112)):*
> ```XML
>         externalLinkIds,
>       });
>       const infraLinks: {
> ```

let's not use inline types much as they complicate the readability. Just create an own type for this, e.g. with the name `InfrastructureLink`
___
*[src/components/map/DrawRouteLayer.tsx, line 129 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6sfamCq6eMnAx0KCl:-Mt6sfamCq6eMnAx0KCm:b2x5fss) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/components/map/DrawRouteLayer.tsx#L129)):*
> ```XML
>         infraLinkIds: infraLinks.map((link) => link.infrastructureLinkId),
>       });
>       const findStopIndexByInfraLink = (
> ```

I see this rather as a utility function... as such it does not belong to a drawing component layer. You could rafactor it to a separate file and take parameters (infraLinks: InfrastructureLink[], stop: ServicePatternScheduledStopPoint)

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55)* status: 10 of 15 files reviewed, 6 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[src/components/map/DrawRouteLayer.tsx, line 150 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6tA4t-_R5r-ztCm4f:-Mt6tA4t-_R5r-ztCm4g:b-80u24o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/components/map/DrawRouteLayer.tsx#L150)):*
> ```XML
>             );
> 
>             return (
> ```

Isn't this easier to read?
```
const isForwardStop = ['forward', 'bidirectional'].includes(stop.direction)
const isBackwardStop = ['backward', bidirectional'].includes(stop.direction)
return link && (link.isTraversalForwards && isForwardStop || !link.isTraversalForwards && isBackwardStop)
```

But I also suspect that as a link must always have a direction defined, you don't even need to take the `bidirectional` direction into consideration in this boolean expression

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55)* status: 10 of 15 files reviewed, 8 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[src/components/map/ModalMap.tsx, line 65 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6uogA6jMLc_SCVSb3:-Mt6uogBCF4OP_Q_KpbB:b-zg6r7a) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/components/map/ModalMap.tsx#L65)):*
> ```XML
>         infrastructure_links_along_route: {
>           data: infraLinksAlongRoute.map((link, index) => ({
>             infrastructure_link_id: link.infrastructureLinkId,
> ```

let's refactor this mapping function to be in the previously mentioned graphql/infrastructureNetwork.ts file as a separate, reusable function
___
*[src/context/MapEditorReducer.ts, line 16 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6yQMy6ipQ3bB6ZwBE:-Mt6yQMy6ipQ3bB6ZwBF:b-rvf87f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/context/MapEditorReducer.ts#L16)):*
> ```XML
>   drawingMode: Mode | undefined;
>   busRoute?: BusRouteResponse;
>   stopsWithinRoute?: ExplicitAny[]; // TODO: correct typings?
> ```

I'd suspect that if we have a query that finds stops within a route then this type would be generated as well. Worth a try...?

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55)* status: 10 of 16 files reviewed, 8 unresolved discussions (waiting on @jarkkoka, @jschummer, and @pallost)

___
*[src/components/map/DrawRouteLayer.tsx, line 119 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoVPt92Z2Z0dKTeOCL:-MtD7m-87DjvxlY2PRCV:b-lv17ic) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L119)):*

I do see a reason to do mapping between the graphql entities and the UI entities. The main reason however is the possible incompatibility between the graphql generated type and the actual intended type (e.g. string vs UUID).

Instead of bloating the display components with the API-related mapping functions, let's rather have these in the `graphql` folder. So e.g. you could create an `graphql/infrastructureNetwork.ts` file and add a `mapGQLInfrastructureLink()` function there. It'll also allow us to use the same mapping function in other components as well. For now, I'd also have the UI type definition in this same file, e.g.:
```
export type InfrastructureLink {
        infrastructureLinkId: string;
        isTraversalForwards: boolean;
      };
```
</blockquote></details>

I agree about the type, but I don't think it makes much sense to put a 

`const mapGQLInfrastructureLink = (infraLinkResponse, routeResponse) => ...`

in there, because ordering the links according to the routeResponse seems to be quite specific to this use case.

I've not put the type in the `graphql/infrastructureNetwork.ts` file, but suggest to leave the mapping in here unless we do find another place doing this specific mapping.
___
*[src/components/map/DrawRouteLayer.tsx, line 129 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6sfamCq6eMnAx0KCl:-MtD8fzT5K04k8NRcUHe:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/components/map/DrawRouteLayer.tsx#L129)):*

I see this rather as a utility function... as such it does not belong to a drawing component layer. You could rafactor it to a separate file and take parameters (infraLinks: InfrastructureLink[], stop: ServicePatternScheduledStopPoint)
</blockquote></details>

Done.
___
*[src/components/map/DrawRouteLayer.tsx, line 150 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6tA4t-_R5r-ztCm4f:-MtD8kj15vW6oxEy0gQw:b-ecn31m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/components/map/DrawRouteLayer.tsx#L150)):*

Isn't this easier to read?
```
const isForwardStop = ['forward', 'bidirectional'].includes(stop.direction)
const isBackwardStop = ['backward', bidirectional'].includes(stop.direction)
return link && (link.isTraversalForwards && isForwardStop || !link.isTraversalForwards && isBackwardStop)
```

But I also suspect that as a link must always have a direction defined, you don't even need to take the `bidirectional` direction into consideration in this boolean expression
</blockquote></details>

I'm not sure I understand your comment. Stops, whose direction is 'bidirectional' can be traversed from either side and this needs to be taken into account. I did refactor this now and it looks much simpler.
___
*[src/components/map/ModalMap.tsx, line 65 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6uogA6jMLc_SCVSb3:-MtDBkAm3Zb_ubYOn7Ae:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/components/map/ModalMap.tsx#L65)):*

let's refactor this mapping function to be in the previously mentioned graphql/infrastructureNetwork.ts file as a separate, reusable function
</blockquote></details>

Done.
___
*[src/context/MapEditorReducer.ts, line 16 at r11](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-Mt6yQMy6ipQ3bB6ZwBE:-MtDBoyL-XQqYN1cC1Ac:bf6rw3l) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c3e27763dbcab4d8bdfc7816c614b86ff39ce2ae/src/context/MapEditorReducer.ts#L16)):*

I'd suspect that if we have a query that finds stops within a route then this type would be generated as well. Worth a try...?
</blockquote></details>

Yes, but I think this is not part of this PR.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55)* status: 10 of 16 files reviewed, 3 unresolved discussions (waiting on @jarkkoka and @pallost)

___
*[src/components/map/DrawRouteLayer.tsx, line 119 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoVPt92Z2Z0dKTeOCL:-MtH-WCc5x8Zvo-8uWyL:b-9gz4a0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L119)):*

I agree about the type, but I don't think it makes much sense to put a 

`const mapGQLInfrastructureLink = (infraLinkResponse, routeResponse) => ...`

in there, because ordering the links according to the routeResponse seems to be quite specific to this use case.

I've not put the type in the `graphql/infrastructureNetwork.ts` file, but suggest to leave the mapping in here unless we do find another place doing this specific mapping.
</blockquote></details>

OK, let's leave it like this for now

jarkkoka (approved):
:lgtm: 

> Reviewed 3 of 3 files at r10, 1 of 3 files at r14, 3 of 3 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-:-MtIJPRQC079h7DTo7nJ:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @jschummer)

___
*[src/components/map/DrawRouteLayer.tsx, line 136 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoW53a04D0O0hIXzHf:-MtIHB9z2wHpVq9PVVcm:biwo6ll) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/DrawRouteLayer.tsx#L136)):*

Yes, this was true, but it fixed now. Also, the route <-> journey pattern constraint (other PR in hasura repo) is checking this now.
</blockquote></details>

Ok, looks good! :+1: 
___
*[src/components/map/ModalMap.tsx, line 50 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/55#-MsoW_oA6NoFdGpWm60w:-MtIHjQ0A9hkOFWWlEaF:b-ch62cx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/341c4687ae01febb105e87a45fcdda8264a869cd/src/components/map/ModalMap.tsx#L50)):*

Yes, but in theory these don't have to be the terminus stops of the journey pattern. The route could start at a certain stop, which is not part of the journey pattern (no passengers can enter). Not sure if this will find a real application later on, but in theory it's possible.
</blockquote></details>

Hmmm...maybe such stops (on which passengers are not allowed to board) should still be part of the journey pattern. I think there should be an enum property or boolean properties defined for "stop point on journey pattern" table/class telling whether boarding/alighting is allowed.

PR #64, opened 2022-01-21
Utils

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/64)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/64)* status: 0 of 38 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[-- commits, line 6 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwYEgH23DpdVaDOxlj):*
Maybe mention what you actually did:

"Added pen icon" (or something like that, its a bit difficult read from the diffs what was actually done)
___
*[-- commits, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwXaUEA6EhSM_Enuzn):*
Maybe you could also have mentioned why it was removed (probably something like this):

"In commit 964786e83de3d9d4edb045d92a66b3f9aab77b76 font called 'fontello' was
commited to repository by mistake. 'fontello' font is same as 'hsl-icons' and we need only
one of those". 

(Didn't checked commit id, but I assume you did as you just mentioned it to me :) )
___
*[src/time.spec.ts, line 5 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwVuFl92H0RjEVFw-n:-MtwVuFl92H0RjEVFw-o:bn2mpyv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/915fc9e0a9949f5c2535418ff833f66d23f4fae7/src/time.spec.ts#L5)):*
> ```XML
> import { formatDate } from './time';
> 
> describe(`formatDate()`, () => {
> ```

Once again excellent opportunity to write this like `describe(`${formatDate.name}()`, () => {` :wink: 
Then test description stays correct even if function is renamed at some point.
___
*[src/time.spec.ts, line 12 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwW3sQADoVsZMiBDnu:-MtwW3sQADoVsZMiBDnv:b-g8lcxg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/915fc9e0a9949f5c2535418ff833f66d23f4fae7/src/time.spec.ts#L12)):*
> ```XML
>   test('Should be able to format dates from ISO string inputs', () => {
>     const output = formatDate(DateTime.DATETIME_FULL, isoDate);
>     expect(output).toBe('20. huhtikuuta 2017 klo 14.32 UTC+3');
> ```

No need to change anything and maybe you already considered this, but anyway: as a general protip, these look like excellent opportunities to use `toMatchInlineSnapshot()`. 

Of if result of first and second test case dhould be exactly the same, then it might make sense to have it as own string variable (const) and then write expects so that they would use it. Same for test cases 3 and 4.

Anyway, this is also fine by me and no need to change anything :+1:

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/64)* status: 0 of 38 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[-- commits, line 6 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwYEgH23DpdVaDOxlj):*

Maybe mention what you actually did:

"Added pen icon" (or something like that, its a bit difficult read from the diffs what was actually done)
</blockquote></details>

Could also mention that it isn't used yet:
"Added pen icon that will be needed in the near future"

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/64)* status: 0 of 38 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[-- commits, line 6 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwYEgH23DpdVaDOxlj):*

Could also mention that it isn't used yet:
"Added pen icon that will be needed in the near future"
</blockquote></details>

Done.
___
*[-- commits, line 9 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwXaUEA6EhSM_Enuzn):*

Maybe you could also have mentioned why it was removed (probably something like this):

"In commit 964786e83de3d9d4edb045d92a66b3f9aab77b76 font called 'fontello' was
commited to repository by mistake. 'fontello' font is same as 'hsl-icons' and we need only
one of those". 

(Didn't checked commit id, but I assume you did as you just mentioned it to me :) )
</blockquote></details>

Done.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/64)* status: 0 of 38 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[src/time.spec.ts, line 5 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwVuFl92H0RjEVFw-n:-Mtwa29IEHZddY-VGRSs:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/915fc9e0a9949f5c2535418ff833f66d23f4fae7/src/time.spec.ts#L5)):*

Once again excellent opportunity to write this like `describe(`${formatDate.name}()`, () => {` :wink: 
Then test description stays correct even if function is renamed at some point.
</blockquote></details>

Done.
___
*[src/time.spec.ts, line 12 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwW3sQADoVsZMiBDnu:-MtwaETDAJCC3dTO1cAj:b-bp8rkv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/915fc9e0a9949f5c2535418ff833f66d23f4fae7/src/time.spec.ts#L12)):*

No need to change anything and maybe you already considered this, but anyway: as a general protip, these look like excellent opportunities to use `toMatchInlineSnapshot()`. 

Of if result of first and second test case dhould be exactly the same, then it might make sense to have it as own string variable (const) and then write expects so that they would use it. Same for test cases 3 and 4.

Anyway, this is also fine by me and no need to change anything :+1: 
</blockquote></details>

inline snapshotting does not work in the current environment for some reason. Could you check it?

also, snapshotting is not really meant for this where we are checking for scalar values

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/64)* status: 0 of 38 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[src/time.spec.ts, line 12 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/64#-MtwW3sQADoVsZMiBDnu:-Mtwc1UQ3UzKDIqsaW_X:bo8p3fz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/915fc9e0a9949f5c2535418ff833f66d23f4fae7/src/time.spec.ts#L12)):*

inline snapshotting does not work in the current environment for some reason. Could you check it?

also, snapshotting is not really meant for this where we are checking for scalar values
</blockquote></details>

Will check inline snapshots :+1:

PR #67, opened 2022-01-24
Line details page

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/67)

### DIFF
--- a/src/components/RoutesAndLinesPage.tsx
+++ b/src/components/RoutesAndLinesPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
 import {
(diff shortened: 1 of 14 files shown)


pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/67)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[src/components/routes-and-lines/AdditionalInformation.tsx, line 41 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/67#-MuFuPB2EluFfxq7WIxq:-MuGPGdb7px3fZAuCofS:b-2syqha) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ccae2261c2cec17e411fdbfa80f2ec7803d546e8/src/components/routes-and-lines/AdditionalInformation.tsx#L41)):*

Do we want to localize these? Same also on similar row below.
</blockquote></details>

No, because these values should come from the database. Here I put the "!" mark there even that these are hardcoded values for now

PR #85, opened 2022-02-09
Allow enabling/disabling docker volume for jore4-testdb

pallost:
Having the volume enabled by default ended up causing lots of issues when switching between hasura versions resulting in incompatible migration states.
This compose yaml section has only been commented out however because in some cases it is actually desired to save the state between development sessions.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/85)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,7 @@ yarn-error.log*
 .vscode
 docker/*
 !docker/docker-compose.custom.yml
-!docker/2524600800000_seed-data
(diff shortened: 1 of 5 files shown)


tk-vincit (approved):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/85)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pallost)

PR #93, opened 2022-02-14
Revert "Add regex validation to validity end"

tk-vincit:
This reverts commit 52bffa18a630419815023499fcb1b96e51f5bb5f.

Reverted change actually introduced a bug where form validation
started to fail if form was set to "indefinte" and "validityEnd"
date was missing.

Proper way to tackle the issue would be introducing
some kind of conditional validation e.g. by using
`partial()` and `refine()` methods from `zod`
as instucted here: https://github.com/colinhacks/zod/issues/61

Example implementation would look something like this:

```
export const schema = z
  .object({
    priority: z.nativeEnum(Priority),
    validityStart: z
      .string()
      .min(1)
      .regex(/[0-9]{4}-[0-9]{2}-[0-9]{2}/),
  })
  .merge(
    z
      .object({
        validityEnd: z.string().regex(/[0-9]{4}-[0-9]{2}-[0-9]{2}/),
        indefinite: z.boolean(),
      })
      .partial()
      .refine((data) => !!data.indefinite || !!data.validityEnd),
  );

```

The problem is that zod won't let us to `merge()` schemas after
`refine()` has been called (https://giters.com/colinhacks/zod/issues/597),
and causes problems in all places where ConfirmationForm has been used.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/93)

### DIFF
--- a/src/components/forms/ConfirmSaveForm.tsx
+++ b/src/components/forms/ConfirmSaveForm.tsx
@@ -12,6 +12,12 @@ export const schema = z.object({
     .string()
     .min(1)
     .regex(/[0-9]{4}-[0-9]{2}-[0-9]{2}/),
+  // TODO: also validityEnd could/should be validated against regex
+  // but only when "indefinite" is set to false. Anyway, seems like zod
+  // schemas start to cause TS errors if merged with each other after
+  // .partial() and .refine() methods have been called, so validation
+  // is left out for now. See message of commit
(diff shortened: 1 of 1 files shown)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/93)* status: 0 of 1 files reviewed, all discussions resolved

PR #127, opened 2022-03-28
Implement template route selection in route creation

eleksi:
- User can copy existing route's geometry when creating a new route
- Routes can be filtered by date and priority

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@graphql-tools/load": "^7.5.1",
     "@graphql-tools/merge": "^8.2.2",
     "@graphql-typed-document-node/core": "^3.1.0",
-    "@headlessui/react": "^1.0.0",
+    "@headlessui/react": "^1.5.0",
     "@hookform/resolvers": "^2.8.3",
     "@nebula.gl/edit-modes": "0.23.8",
     "@reduxjs/toolkit": "^1.8.0",
--- a/src/components/forms/ChooseRouteDropdown.spec.tsx
+++ b/src/components/forms/ChooseRouteDropdown.spec.tsx
@@ -1,0 +1,205 @@
+import { MockedProvider } from '@apollo/client/testing';
+import { act, fireEvent, screen } from '@testing-library/react';
+import { DateTime } from 'luxon';
+import {
+  GetRouteDetailsByLabelWildcardDocument,
+  useGetRouteDetailsByLabelWildcardQuery,
+} from '../../generated/graphql';
+import { Priority } from '../../types/Priority';
+import { render, sleep } from '../../utils/test-utils';
+import { ChooseRouteDropdown } from './ChooseRouteDropdown';
+
+describe(`<${ChooseRouteDropdown.name} />`, () => {
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const onChange = () => {};
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const onBlur = () => {};
+  const value = '';
+  const testId = 'ChooseRouteDropdown1';
+  const buttonTestId = `${testId}-button`;
+  const inputTestId = `${testId}-input`;
+
+  const mocks = [
+    {
+      request: {
+        query: GetRouteDetailsByLabelWildcardDocument,
+        variables: {
+          label: '%',
+          date: '2022-03-23T00:00:00.000+02:00',
+          priorities: [10],
+        },
+      },
+      result: {
+        data: {
+          route_route: [
+            {
+              __typename: 'route_route',
+              route_id: 'route1',
+              description_i18n: 'Route 1',
+              validity_start: DateTime.fromISO('2022-03-20T22:00:00+00:00'),
+              validity_end: DateTime.fromISO('2022-03-29T20:59:59.999+00:00'),
+              priority: 10,
+              label: '123',
+              direction: 'outbound',
+              route_shape: null,
+              starts_from_scheduled_stop_point_id: 'start',
+              ends_at_scheduled_stop_point_id: 'end',
+              on_line_id: 'line_id',
+            },
+            {
+              __typename: 'route_route',
+              route_id: 'route2',
+              description_i18n: 'Route 2',
+              validity_start: DateTime.fromISO('2021-01-01T22:00:00+00:00'),
+              validity_end: DateTime.fromISO('2022-12-31T21:59:59.999+00:00'),
+              priority: 10,
+              label: '456',
+              direction: 'outbound',
+              route_shape: null,
+              starts_from_scheduled_stop_point_id: 'start',
+              ends_at_scheduled_stop_point_id: 'end',
+              on_line_id: 'line_id',
+            },
+          ],
+        },
+      } as ReturnType<typeof useGetRouteDetailsByLabelWildcardQuery>,
+    },
+    {
+      request: {
+        query: GetRouteDetailsByLabelWildcardDocument,
+        variables: {
+          label: '1%',
+          date: '2022-03-23T00:00:00.000+02:00',
+          priorities: [10],
+        },
+      },
+      result: {
+        data: {
+          route_route: [
+            {
+              __typename: 'route_route',
+              route_id: 'route1',
+              description_i18n: 'Route 1',
+              validity_start: DateTime.fromISO('2022-03-20T22:00:00+00:00'),
+              validity_end: DateTime.fromISO('2022-03-29T20:59:59.999+00:00'),
+              priority: 10,
+              label: '123',
+              direction: 'outbound',
+              route_shape: null,
+              starts_from_scheduled_stop_point_id: 'start',
+              ends_at_scheduled_stop_point_id: 'end',
+              on_line_id: 'line_id',
+            },
+          ],
+        },
+      } as ReturnType<typeof useGetRouteDetailsByLabelWildcardQuery>,
+    },
+  ];
+
+  test('Opens dropdown when clicked and shows all routes', async () => {
+    const { container, asFragment } = render(
+      <MockedProvider mocks={mocks} addTypename={false}>
+        <ChooseRouteDropdown
+          testId={testId}
+          value={value}
+          onChange={onChange}
+          onBlur={onBlur}
+          date={DateTime.fromISO('2022-03-23T00:00:00.000+02:00')}
+          priorities={[Priority.Standard]}
+        />
+      </MockedProvider>,
+    );
+
+    // dropdown is collapsed, value is not yet bound to dropdown, 'Choose route' text shows
+    expect(screen.getByTestId(buttonTestId)).toHaveTextContent(
+      'Valitse reitti',
+    );
+    expect(container.querySelector('li')).toBeNull();
+    expect(asFragment()).toMatchSnapshot();
+
+    // wait for the graphql call to execute
+    await act(() => sleep(0));
+
+    // dropdown is collapsed, data is loaded and 'Choose route' text still shows in button
+    expect(screen.getByTestId(buttonTestId)).toHaveTextContent(
+      'Valitse reitti',
+    );
+    expect(container.querySelector('li')).toBeNull();
+    expect(asFragment()).toMatchSnapshot();
+
+    // click dropdown to open it:
+    const openDropdownButton = screen.getByTestId(buttonTestId);
+    fireEvent.click(openDropdownButton);
+
+    // dropdown is open, both lines show
+    const items = container.querySelectorAll('li');
+
+    expect(items[0].querySelectorAll('div > div')[0].textContent).toBe(
+      `123 | Route 1`,
+    );
+    expect(items[0].querySelectorAll('div > div')[1].textContent).toBe(
+      `21.3.2022 - 29.3.2022`,
+    );
+    expect(items[1].querySelectorAll('div > div')[0].textContent).toBe(
+      `456 | Route 2`,
+    );
+    expect(items[1].querySelectorAll('div > div')[1].textContent).toBe(
+      `2.1.2021 - 31.12.2022`,
+    );
+    expect(items.length).toBe(2);
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test('Filters shown routes when query is set', async () => {
+    const { container, asFragment } = render(
+      <MockedProvider mocks={mocks} addTypename={false}>
+        <ChooseRouteDropdown
+          testId={testId}
+          value={value}
+          onChange={onChange}
+          onBlur={onBlur}
+          date={DateTime.fromISO('2022-03-23T00:00:00.000+02:00')}
+          priorities={[Priority.Standard]}
+        />
+      </MockedProvider>,
+    );
+
+    // dropdown is collapsed, value is not yet bound to dropdown, 'Choose route' text shows
+    expect(screen.getByTestId(buttonTestId)).toHaveTextContent(
+      'Valitse reitti',
+    );
+    expect(container.querySelector('li')).toBeNull();
+    expect(asFragment()).toMatchSnapshot();
+
+    // wait for the graphql call to execute
+    await act(() => sleep(0));
+
+    // dropdown is collapsed, data is loaded and 'Choose route' text still shows in button
+    expect(screen.getByTestId(buttonTestId)).toHaveTextContent(
+      'Valitse reitti',
+    );
+    expect(container.querySelector('li')).toBeNull();
+    expect(asFragment()).toMatchSnapshot();
+
+    // write query
+    const input = screen.getByTestId(inputTestId);
+    fireEvent.change(input, { target: { value: '1' } });
+
+    // wait for the graphql call to execute
+    await act(() => sleep(0));
+
+    // dropdown is open, first line shows
+    const items = container.querySelectorAll('li');
+
+    expect(items[0].querySelectorAll('div > div')[0].textContent).toBe(
+      `123 | Route 1`,
+    );
+    expect(items[0].querySelectorAll('div > div')[1].textContent).toBe(
+      `21.3.2022 - 29.3.2022`,
+    );
+    expect(items.length).toBe(1);
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+});
--- a/src/components/forms/ChooseRouteDropdown.tsx
+++ b/src/components/forms/ChooseRouteDropdown.tsx
@@ -1,0 +1,96 @@
+import { DateTime } from 'luxon';
+import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import {
+  RouteRoute,
+  useGetRouteDetailsByLabelWildcardQuery,
+} from '../../generated/graphql';
+import { mapRoutesDetailsResult } from '../../graphql';
+import { MAX_DATE, MIN_DATE } from '../../time';
+import { Priority } from '../../types/Priority';
+import { Combobox, ComboboxInputProps } from '../../uiComponents';
+import { mapToVariables } from '../../utils';
+import { DateRange } from '../routes-and-lines/DateRange';
+
+interface Props extends ComboboxInputProps {
+  testId?: string;
+  date: DateTime;
+  priorities: Priority[];
+}
+
+const mapToOptionContent = (item: RouteRoute) => (
+  <div className="flex flex-col">
+    <div>
+      <span className="font-bold">{item.label}</span>
+      {` | ${item.description_i18n}`}
+    </div>
+    <div className="text-sm">
+      <DateRange
+        startDate={item.validity_start || MIN_DATE}
+        endDate={item.validity_end || MAX_DATE}
+      />
+    </div>
+  </div>
+);
+
+const mapToOption = (item: RouteRoute) => ({
+  key: item.route_id,
+  value: item.route_id,
+  render: () => mapToOptionContent(item),
+});
+
+export const ChooseRouteDropdown = ({
+  testId,
+  value,
+  onChange,
+  onBlur,
+  date,
+  priorities,
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+
+  const [query, setQuery] = useState('');
+
+  const routesResult = useGetRouteDetailsByLabelWildcardQuery(
+    mapToVariables({
+      label: `${query}%`,
+      date: date.toISO(),
+      priorities,
+    }),
+  );
+
+  const routes = mapRoutesDetailsResult(routesResult);
+
+  const options = routes?.map(mapToOption) || [];
+
+  const selectedRoute = routes?.find(
+    (item) => item.route_id === value,
+  ) as RouteRoute;
+
+  // If no route is selected, show "Choose route"
+  const mapToButtonContent = (displayedRoute?: RouteRoute) => {
+    // Headless UI Combobox component takes care of displaying query
+    if (query && !displayedRoute) return null;
+
+    return (
+      <div className="w-full">
+        {displayedRoute
+          ? `${displayedRoute.label} (${displayedRoute.description_i18n})`
+          : t('routes.chooseRoute')}
+      </div>
+    );
+  };
+
+  return (
+    <Combobox
+      id="choose-route-combobox"
+      testId={testId}
+      buttonContent={mapToButtonContent(selectedRoute)}
+      options={options}
+      value={value}
+      onChange={onChange}
+      onBlur={onBlur}
+      onQueryChange={setQuery}
+    />
+  );
+};
--- a/src/components/forms/RoutePropertiesForm.tsx
+++ b/src/components/forms/RoutePropertiesForm.tsx
@@ -2,42 +2,28 @@ import { zodResolver } from '@hookform/resolvers/zod';
 import React from 'react';
 import { Controller, FormProvider, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
-import { z } from 'zod';
 import { Column, Row } from '../../layoutComponents';
 import { ChooseLineDropdown } from './ChooseLineDropdown';
-import {
-  ConfirmSaveForm,
-  FormState as ConfirmSaveFormState,
-  schema as confirmSaveFormSchema,
-} from './ConfirmSaveForm';
+import { ConfirmSaveForm } from './ConfirmSaveForm';
+import { routeFormSchema, RouteFormState } from './RoutePropertiesForm.types';
 
-interface Props {
+export interface RouteFormProps {
   id?: string;
   routeLabel?: string | null;
   className?: string;
-  defaultValues: Partial<FormState>;
-  onSubmit: (state: FormState) => void;
+  defaultValues: Partial<RouteFormState>;
+  onSubmit: (state: RouteFormState) => void;
 }
 
-export const schema = z
-  .object({
-    label: z.string().min(1),
-    description_i18n: z.string().min(1),
-    on_line_id: z.string().uuid().min(1),
-  })
-  .merge(confirmSaveFormSchema);
-
-export type FormState = z.infer<typeof schema> & ConfirmSaveFormState;
-
 const RoutePropertiesFormComponent = (
-  { id, routeLabel, className, defaultValues, onSubmit }: Props,
+  { id, routeLabel, className, defaultValues, onSubmit }: RouteFormProps,
   ref: ExplicitAny,
 ): JSX.Element => {
   const { t } = useTranslation();
 
-  const methods = useForm<FormState>({
+  const methods = useForm<RouteFormState>({
     defaultValues,
-    resolver: zodResolver(schema),
+    resolver: zodResolver(routeFormSchema),
   });
 
   const {
--- a/src/components/forms/RoutePropertiesForm.tsx
+++ b/src/components/forms/RoutePropertiesForm.tsx
@@ -1,8 +1,12 @@
+import { Switch as HuiSwitch } from '@headlessui/react';
 import { zodResolver } from '@hookform/resolvers/zod';
-import React from 'react';
+import React, { useContext, useState } from 'react';
 import { Controller, FormProvider, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
+import { MapEditorContext } from '../../context/MapEditor';
 import { Column, Row } from '../../layoutComponents';
+import { Switch, SwitchLabel } from '../../uiComponents';
+import { TemplateRouteSelector } from '../routes-and-lines/TemplateRouteSelector';
 import { ChooseLineDropdown } from './ChooseLineDropdown';
 import { ConfirmSaveForm } from './ConfirmSaveForm';
 import { routeFormSchema, RouteFormState } from './RoutePropertiesForm.types';
@@ -20,11 +24,18 @@ const RoutePropertiesFormComponent = (
   ref: ExplicitAny,
 ): JSX.Element => {
   const { t } = useTranslation();
+  const {
+    state: { editedRouteData, creatingNewRoute },
+    dispatch,
+  } = useContext(MapEditorContext);
 
   const methods = useForm<RouteFormState>({
     defaultValues,
     resolver: zodResolver(routeFormSchema),
   });
+
+  const [showTemplateRouteSelector, setShowTemplateRouteSelector] =
+    useState(false);
 
   const {
     register,
@@ -32,6 +43,18 @@ const RoutePropertiesFormComponent = (
     formState: { errors },
     control,
   } = methods;
+
+  const setTemplateRoute = (uuid?: UUID) => {
+    dispatch({
+      type: 'setState',
+      payload: {
+        editedRouteData: {
+          ...editedRouteData,
+          templateRouteId: uuid,
+        },
+      },
+    });
+  };
 
   return (
     // eslint-disable-next-line react/jsx-props-no-spreading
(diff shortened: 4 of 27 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)* status: 0 of 26 files reviewed, 7 unresolved discussions (waiting on @eleksi and @pallost)

___
*[src/components/forms/ChooseRouteDropdown.tsx, line 27 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFIsEW3NP75vZwfHq5:-MzFIsEW3NP75vZwfHq6:b-wcjst4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.tsx#L27)):*
> ```XML
>     </div>
>     <div className="text-sm">
>       {`${mapToShortDate(item.validity_start || MIN_DATE)} - ${mapToShortDate(
> ```

this is a bit long to write inline, could you refactor it to its own component?
___
*[src/components/forms/ChooseRouteDropdown.tsx, line 70 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFKdJ-5rsrD5Ajy4WI:-MzFKdJ-5rsrD5Ajy4WJ:b5o8bez) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.tsx#L70)):*
> ```XML
>   // if no route is selected, show "Choose route"
>   const mapToButtonContent = (displayedRoute?: RouteRoute) => {
>     if (query && !displayedRoute) return null;
> ```

I don't think this works correctly. If you didn't write any search term, the query variable's value will stay an empty string ('') which is falsey. This means that if you don't write anything to the combobox but select an item from all routes, the button content won't be refreshed.

As a fix, I think you just need to check for the existence of 'selectedRoute', no need to check 'query'
___
*[src/components/forms/ChooseRouteDropdown.spec.tsx, line 5 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFJYjPAE8KuK4enZwF:-MzFJYjPAE8KuK4enZwG:bo3u22h) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.spec.tsx#L5)):*
> ```XML
> import { DateTime } from 'luxon';
> import {
>   GetRouteDetailsByLabelWildcardDocument,
> ```

you even created a test for this! Nice!
___
*[src/components/forms/ChooseRouteDropdown.spec.tsx, line 12 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFJdy-AdRLSbRP83E0:-MzFJdy-AdRLSbRP83E1:b86b9at) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.spec.tsx#L12)):*
> ```XML
> import { ChooseRouteDropdown } from './ChooseRouteDropdown';
> 
> describe('<ChooseRouteDropdown />', () => {
> ```

could build the test suite name dynamically here, so that if we rename the component, the test suite description will also follow it:

````
describe(`<${ChooseLineDropdown.name} />`, () => {...
```
___
*[src/components/forms/ChooseRouteDropdown.spec.tsx, line 203 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFKO0T2fWfOtGqe01u:-MzFKO0T2fWfOtGqe01v:b-ch9z5u) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.spec.tsx#L203)):*
> ```XML
>     expect(items.length).toBe(1);
> 
>     expect(asFragment()).toMatchSnapshot();
> ```

good tests!
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 7 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMGVmBw1CbO-pD7p_:-MzFMGVmBw1CbO-pD7pa:bwpwqtn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L7)):*
> ```XML
> } from './ConfirmSaveForm';
> 
> export interface RouteFormProps {
> ```

these types belong to RoutePropertiesForm, let's not move them to their own file but keep in RoutePropertiesForm.tsx
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMN8aEh3yIJ9LzN5d:-MzFMN8aEh3yIJ9LzN5e:b-1164k0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L13)):*
> ```XML
>   defaultValues: Partial<RouteFormState>;
>   onSubmit: (state: RouteFormState) => void;
>   creatingNewRoute?: boolean;
> ```

it's a good naming convention to start these boolean flags with 'is': `isCreatingNewRoute`


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)* status: 0 of 26 files reviewed, 12 unresolved discussions (waiting on @eleksi and @pallost)

___
*[src/components/forms/RoutePropertiesForm.tsx, line 40 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMehUEn3FKAYb9NS2:-MzFMehUEn3FKAYb9NS3:bycbw08) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L40)):*
> ```XML
>   });
> 
>   const [basedOnTemplateRoute, setBasedOnTemplateRoute] = useState(false);
> ```

you already store this information in MapEditorContext. You could just use the `mapEditorContextState.editedRouteData.templateRouteId` variable instead of creating an own state variable for this
___
*[src/components/forms/RoutePropertiesForm.tsx, line 50 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFQCfr1DzTKD3kT2Ih:-MzFQCfr1DzTKD3kT2Ii:b3u6jrw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L50)):*
> ```XML
> 
>   const setTemplateRoute = (uuid?: UUID) => {
>     dispatch({
> ```

you're going to cause yourself an awful lot of hassle if you only keep the templateRouteId here for reference. As you can see in DrawRouteLayer already, you need to check for the existence of "templateRouteId" basically everywhere.

It would be much simpler that when the templateRouteId is set/changed, you retrieve the template route's metadata + geometry data and write it directly to the editedRouteData object (ofc without the route_id field)
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMN8aEh3yIJ9LzN5d:-MzFN1ERAKWiuFegCpkX:b-9ulbf5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L13)):*

it's a good naming convention to start these boolean flags with 'is': `isCreatingNewRoute`
</blockquote></details>

actually, you should already know whether this is a new route or an edited route by checking (mapEditorContextState.editedRouteData.id === undefined), let's not create an own property for this in vein
___
*[src/components/forms/TemplateRouteForm.tsx, line 17 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFOI5A5XyhZSZN2Bov:-MzFOI5A5XyhZSZN2Bow:b-p634xu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/TemplateRouteForm.tsx#L17)):*
> ```XML
> 
>   const [priority, setPriority] = useState(Priority.Standard);
> ```

if this is a form, it should behave like one! Not having own state variables and stuff but rather as a partial form that can be integrated into a main form. See example: ConfirmSaveForm

also, currently the change detection works incorrectly: the onChange function is only called (thus this "form's" state is sent to the caller) when you choose a route from the dropdown. If you only set the priority or the observation date, the change is not sent to the caller
___
*[src/components/forms/TemplateRouteForm.tsx, line 28 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFO5_66BNojGlT9pI7:-MzFO5_66BNojGlT9pI8:b-qymkm4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/TemplateRouteForm.tsx#L28)):*
> ```XML
>           <label>{t('priority.label')}</label>
>           <Row className="space-x-1">
>             <SimpleButton
> ```

I start to see this priority button list copied everywhere... let's extract it finally to its own component. Not necessarily part of this PR but should be done soon-ish
___
*[src/components/map/DrawRouteLayer.tsx, line 110 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFPPZZEx6acSz_2svP:-MzFPPZZEx6acSz_2svQ:b-4zjm7j) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/map/DrawRouteLayer.tsx#L110)):*
> ```XML
>   }, [mode]);
> 
>   const baseGeometryRouteId = editedRouteData.id || templateRouteId;
> ```

should write a comment here that you are trying to fetch an existing route's geometry in case you are in edit mode and the template route's id in case you are creating a route based on a template. However you are not handling the case too nicely when you are trying to create a brand new route, without a base template, in which case 'baseGeometryRouteId' is undefined.

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)* status: 0 of 26 files reviewed, 12 unresolved discussions (waiting on @pallost)

___
*[src/components/forms/ChooseRouteDropdown.tsx, line 27 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFIsEW3NP75vZwfHq5:-MzJP2Ej0B6lLdY5H9Ts:b-qwc4te) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.tsx#L27)):*

this is a bit long to write inline, could you refactor it to its own component?
</blockquote></details>

I can do that.
___
*[src/components/forms/ChooseRouteDropdown.tsx, line 70 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFKdJ-5rsrD5Ajy4WI:-MzJP4EB0WY7pVqDXpKk:b-b3oepn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.tsx#L70)):*

I don't think this works correctly. If you didn't write any search term, the query variable's value will stay an empty string ('') which is falsey. This means that if you don't write anything to the combobox but select an item from all routes, the button content won't be refreshed.

As a fix, I think you just need to check for the existence of 'selectedRoute', no need to check 'query'
</blockquote></details>

It seems cryptic, maybe lacks some comments. The HUI combobox component itself handles part of this logic, for that reason we have to render empty here in some cases so we don't show multiple texts. This is only the content for the button, and the input has a value as well. Testing this yourself could probably reveal if it really works correctly or not.
___
*[src/components/forms/ChooseRouteDropdown.spec.tsx, line 5 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFJYjPAE8KuK4enZwF:-MzJPR5t0gbe5Zo1c4zk:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.spec.tsx#L5)):*

you even created a test for this! Nice!
</blockquote></details>

Done.
___
*[src/components/forms/ChooseRouteDropdown.spec.tsx, line 12 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFJdy-AdRLSbRP83E0:-MzJPcGE3OvA-ZcVRS0w:btfrn50) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.spec.tsx#L12)):*

could build the test suite name dynamically here, so that if we rename the component, the test suite description will also follow it:

````
describe(`<${ChooseLineDropdown.name} />`, () => {...
```
</blockquote></details>

I will do that.
___
*[src/components/forms/ChooseRouteDropdown.spec.tsx, line 203 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFKO0T2fWfOtGqe01u:-MzJPeW62-F0suBaNef9:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.spec.tsx#L203)):*

good tests!
</blockquote></details>

Done.
___
*[src/components/forms/RoutePropertiesForm.tsx, line 40 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMehUEn3FKAYb9NS2:-MzJPgOg75-QLyoz4hUV:b-e64c1q) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L40)):*

you already store this information in MapEditorContext. You could just use the `mapEditorContextState.editedRouteData.templateRouteId` variable instead of creating an own state variable for this
</blockquote></details>

This is state for the toggle component, whether to show the template route selection form or not.
___
*[src/components/forms/RoutePropertiesForm.tsx, line 50 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFQCfr1DzTKD3kT2Ih:-MzJPni63tSmf-QVDi7f:b49eyg2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L50)):*

you're going to cause yourself an awful lot of hassle if you only keep the templateRouteId here for reference. As you can see in DrawRouteLayer already, you need to check for the existence of "templateRouteId" basically everywhere.

It would be much simpler that when the templateRouteId is set/changed, you retrieve the template route's metadata + geometry data and write it directly to the editedRouteData object (ofc without the route_id field)
</blockquote></details>

We do not want to take the metadata, only geometry. 

But yes, this is a good point, I also was considering it. But I think this current solution makes it actually easier, as using template route is pretty much the same case as editing existing route. The logic for editing existing route is to get the infra links and stops in draw route layer. We only need to check for templateRouteId's existence in one place and it is in draw route layer.
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 7 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMGVmBw1CbO-pD7p_:-MzJQnf4DeK268jUJKH2:b98nbbm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L7)):*

these types belong to RoutePropertiesForm, let's not move them to their own file but keep in RoutePropertiesForm.tsx
</blockquote></details>

Are you sure? I found this change very useful, as we were using these types in million different places and the exported type's name was renamed differently in every place.

Also, the alternative solution is to keep them in their place and add dependency cycle ignore comments. We use these types in context and will use context in this form.
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMN8aEh3yIJ9LzN5d:-MzJRVFc8f_iFdsgPiVB:b-5eawb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L13)):*

actually, you should already know whether this is a new route or an edited route by checking (mapEditorContextState.editedRouteData.id === undefined), let's not create an own property for this in vein
</blockquote></details>

That is true, but I wanted to keep context out of this form component. I don't think this component should know anything about mapEditorContext, as it is also used in the list view.

I can rename this if it is kept this way.
___
*[src/components/forms/TemplateRouteForm.tsx, line 17 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFOI5A5XyhZSZN2Bov:-MzJSOhV4a9ZIZx56mLM:bo52vww) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/TemplateRouteForm.tsx#L17)):*

if this is a form, it should behave like one! Not having own state variables and stuff but rather as a partial form that can be integrated into a main form. See example: ConfirmSaveForm

also, currently the change detection works incorrectly: the onChange function is only called (thus this "form's" state is sent to the caller) when you choose a route from the dropdown. If you only set the priority or the observation date, the change is not sent to the caller
</blockquote></details>

I have to disagree here, and I expect the purpose if this form is not clear so I will explain here.

There is only one value that we are interested in here and it is the templateRouteId. The other fields are there only to filter routes that are displayed in the dropdown from which the template route is selected from. Therefore we are not interested in the values of any other form fields outside this component, so onChange does not care about them. Maybe its name could be something different to make this more clear.

RoutePropertiesForm is for route's metadata. Every field it contains will go to backend. In this case we only have this form inside the route properties form because the UI requires it. It does not really have anything to do with the route properties form technically, unlike e.g. ConfirmSaveForm. I also considered making this part of the route properties form but I rejected that idea quickly as I realized it would just make everything harder. We would first have to alter the type of the form schema to contain things outside of metadata. Then when we save it we would have to take templateRouteId out of there as it is not to be send to the back end. We would have to extract templateRouteId from there and use it to fetch template route's data.
___
*[src/components/forms/TemplateRouteForm.tsx, line 28 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFO5_66BNojGlT9pI7:-MzJVu8d7jHGaJ2n5Vne:b-ovwaw7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/TemplateRouteForm.tsx#L28)):*

I start to see this priority button list copied everywhere... let's extract it finally to its own component. Not necessarily part of this PR but should be done soon-ish
</blockquote></details>

Good idea.
___
*[src/components/map/DrawRouteLayer.tsx, line 110 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFPPZZEx6acSz_2svP:-MzJW6pX28nAldLyV28y:b-yry66a) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/map/DrawRouteLayer.tsx#L110)):*

should write a comment here that you are trying to fetch an existing route's geometry in case you are in edit mode and the template route's id in case you are creating a route based on a template. However you are not handling the case too nicely when you are trying to create a brand new route, without a base template, in which case 'baseGeometryRouteId' is undefined.
</blockquote></details>

I will comment this. If you refer with the `baseGeometryRouteId` undefined comment to the lines below these, I agree it is not pretty how it is handled. Did you mean something like this?



_Code snippet:_
```XML
const routesResult = useGetRoutesWithInfrastructureLinksQuery(
    mapToVariables({ route_ids: baseGeometryRouteId ? [baseGeometryRouteId] : [] }),
  );
```


eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)* status: 0 of 26 files reviewed, 12 unresolved discussions (waiting on @pallost)

___
*[src/components/forms/RoutePropertiesForm.types.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMN8aEh3yIJ9LzN5d:-MzJYtV94h1IlrIMNL0q:b-cton4d) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L13)):*

That is true, but I wanted to keep context out of this form component. I don't think this component should know anything about mapEditorContext, as it is also used in the list view.

I can rename this if it is kept this way.
</blockquote></details>

I actually ended up bringing context to route properties after this decision was made, which was not great. So the purpose of creatingNewRoute prop is kind of gone now. So I can actually delete this prop and get straight from context I guess. I hope map editor context is cleared after map modal is closed so there is never anything in editedRouteData.id if we are in the list view.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)* status: 0 of 26 files reviewed, 7 unresolved discussions (waiting on @eleksi, @pallost, and @tk-vincit)

___
*[src/components/forms/ChooseRouteDropdown.tsx, line 70 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFKdJ-5rsrD5Ajy4WI:-MzOSg8F3yfpL_lKf456:bk5mhxn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.tsx#L70)):*

It seems cryptic, maybe lacks some comments. The HUI combobox component itself handles part of this logic, for that reason we have to render empty here in some cases so we don't show multiple texts. This is only the content for the button, and the input has a value as well. Testing this yourself could probably reveal if it really works correctly or not.
</blockquote></details>

OK, I was not aware of this :thumbsup: Could you write a line of comment about this here?
___
*[src/components/forms/RoutePropertiesForm.tsx, line 40 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMehUEn3FKAYb9NS2:-MzONm-bD1p2PTST_yhw:b-60j2u9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L40)):*

This is state for the toggle component, whether to show the template route selection form or not.
</blockquote></details>

ok :+1:  let's name it to "showTemplateRouteSelector" or something similar then that the name resembles better of what you're trying to do here
___
*[src/components/forms/RoutePropertiesForm.tsx, line 50 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFQCfr1DzTKD3kT2Ih:-MzONCso72_rz10eut-f:b-53ik3c) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L50)):*

We do not want to take the metadata, only geometry. 

But yes, this is a good point, I also was considering it. But I think this current solution makes it actually easier, as using template route is pretty much the same case as editing existing route. The logic for editing existing route is to get the infra links and stops in draw route layer. We only need to check for templateRouteId's existence in one place and it is in draw route layer.
</blockquote></details>

I disagree on this: you clearly needed to involve `templateRouteId` check all around the DrawRouteLayer and most probably it will affect other places as well in the future. You could of course only fill in the route geometry data for the editedRouteData if that's the only thing you are using for now. But quite likely you'll need at least part of the metadata as well sometime in the future
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 7 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMGVmBw1CbO-pD7p_:-MzOPRWBBGBX-wLc8z6V:be4gx74) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L7)):*

Are you sure? I found this change very useful, as we were using these types in million different places and the exported type's name was renamed differently in every place.

Also, the alternative solution is to keep them in their place and add dependency cycle ignore comments. We use these types in context and will use context in this form.
</blockquote></details>

it still feels quite detached to me if this form schema is actually apart from the form but could ask @tk-vincit about this. It's annoying to have dependency cycles warnings but in this case I don't think they are a real issue.
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMN8aEh3yIJ9LzN5d:-MzOPK7q509QeoXAdyPp:bg2teah) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L13)):*

I actually ended up bringing context to route properties after this decision was made, which was not great. So the purpose of creatingNewRoute prop is kind of gone now. So I can actually delete this prop and get straight from context I guess. I hope map editor context is cleared after map modal is closed so there is never anything in editedRouteData.id if we are in the list view.
</blockquote></details>

I don't see these changes, so I guess this is still work in progress...?
___
*[src/components/forms/TemplateRouteForm.tsx, line 17 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFOI5A5XyhZSZN2Bov:-MzOQHFd6Ay02go2vUDV:bgdvnr7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/TemplateRouteForm.tsx#L17)):*

I have to disagree here, and I expect the purpose if this form is not clear so I will explain here.

There is only one value that we are interested in here and it is the templateRouteId. The other fields are there only to filter routes that are displayed in the dropdown from which the template route is selected from. Therefore we are not interested in the values of any other form fields outside this component, so onChange does not care about them. Maybe its name could be something different to make this more clear.

RoutePropertiesForm is for route's metadata. Every field it contains will go to backend. In this case we only have this form inside the route properties form because the UI requires it. It does not really have anything to do with the route properties form technically, unlike e.g. ConfirmSaveForm. I also considered making this part of the route properties form but I rejected that idea quickly as I realized it would just make everything harder. We would first have to alter the type of the form schema to contain things outside of metadata. Then when we save it we would have to take templateRouteId out of there as it is not to be send to the back end. We would have to extract templateRouteId from there and use it to fetch template route's data.
</blockquote></details>

I see. So this is not a form then, but some kind of extended form input component with built-in search capability? Let's then name it some different way, like "TemplateRouteSelector" or something similar and move it out of the forms directory.
___
*[src/components/map/DrawRouteLayer.tsx, line 110 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFPPZZEx6acSz_2svP:-MzOR0Qu0yrKiARXsB6d:b-y97c2z) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/map/DrawRouteLayer.tsx#L110)):*

I will comment this. If you refer with the `baseGeometryRouteId` undefined comment to the lines below these, I agree it is not pretty how it is handled. Did you mean something like this?
</blockquote></details>

well, in this case you technically shouldn't even place that backend call with the hook if no edited or template route are selected...
```
  const routesResult = useGetRoutesWithInfrastructureLinksQuery({
    skip: !baseGeometryRouteId,
    variables: {route_ids: [baseGeometryRouteId]}
  );
```

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)* status: 0 of 31 files reviewed, 7 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[src/components/forms/ChooseRouteDropdown.tsx, line 70 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFKdJ-5rsrD5Ajy4WI:-MzOoUOuEW8E1C73PSWv:b-mjez8r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/ChooseRouteDropdown.tsx#L70)):*

OK, I was not aware of this :thumbsup: Could you write a line of comment about this here?
</blockquote></details>

Yes I will add comments!
___
*[src/components/forms/RoutePropertiesForm.tsx, line 40 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMehUEn3FKAYb9NS2:-MzOoYECEafaV5HJ9Kfu:btfrn50) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L40)):*

ok :+1:  let's name it to "showTemplateRouteSelector" or something similar then that the name resembles better of what you're trying to do here
</blockquote></details>

I will do that.
___
*[src/components/forms/RoutePropertiesForm.tsx, line 50 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFQCfr1DzTKD3kT2Ih:-MzPNpKC-DMFpP6BGjei:bbo6jop) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.tsx#L50)):*

I disagree on this: you clearly needed to involve `templateRouteId` check all around the DrawRouteLayer and most probably it will affect other places as well in the future. You could of course only fill in the route geometry data for the editedRouteData if that's the only thing you are using for now. But quite likely you'll need at least part of the metadata as well sometime in the future
</blockquote></details>

We agreed in chat that the current solution works better at least with the current implementation of DrawRouteLayer.
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 7 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMGVmBw1CbO-pD7p_:-MzPO74DDyJIVcSyjCMG:b1y1t28) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L7)):*

it still feels quite detached to me if this form schema is actually apart from the form but could ask @tk-vincit about this. It's annoying to have dependency cycles warnings but in this case I don't think they are a real issue.
</blockquote></details>

I ended up leaving some types here and moved `RouteFormProps` back to `RoutePropertiesForm.ts`
___
*[src/components/forms/RoutePropertiesForm.types.ts, line 13 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFMN8aEh3yIJ9LzN5d:-MzPXFv82wUHsNUwCn6R:b35f05) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/RoutePropertiesForm.types.ts#L13)):*

I don't see these changes, so I guess this is still work in progress...?
</blockquote></details>

creatingNewRoute prop has now been deleted.
___
*[src/components/forms/TemplateRouteForm.tsx, line 17 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFOI5A5XyhZSZN2Bov:-MzOq-nu9swVDGgzwaHd:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/forms/TemplateRouteForm.tsx#L17)):*

I see. So this is not a form then, but some kind of extended form input component with built-in search capability? Let's then name it some different way, like "TemplateRouteSelector" or something similar and move it out of the forms directory.
</blockquote></details>

Done.
___
*[src/components/map/DrawRouteLayer.tsx, line 110 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/127#-MzFPPZZEx6acSz_2svP:-MzPWfWa2xcd3Yj02Ndr:b-myhacw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/79a45e7151e672a9be5ea7614cbcf4cf6cc65a3c/src/components/map/DrawRouteLayer.tsx#L110)):*

well, in this case you technically shouldn't even place that backend call with the hook if no edited or template route are selected...
```
  const routesResult = useGetRoutesWithInfrastructureLinksQuery({
    skip: !baseGeometryRouteId,
    variables: {route_ids: [baseGeometryRouteId]}
  );
```
</blockquote></details>

Thanks, I didn't know about this skip property!

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/127)* status: 0 of 31 files reviewed, 1 unresolved discussion (waiting on @pallost and @tk-vincit)

PR #138, opened 2022-04-05
Add route direction selection

eleksi:
Resolves HSLdevcom/jore4#653

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/138)

### DIFF
--- a/src/components/forms/DirectionDropdown.tsx
+++ b/src/components/forms/DirectionDropdown.tsx
@@ -1,6 +1,6 @@
 import { useTranslation } from 'react-i18next';
-import { ReusableComponentsVehicleModeEnum } from '../../generated/graphql';
-import { mapVehicleModeToUiName } from '../../i18n/uiNameMappings';
+import { mapDirectionToUiName } from '../../i18n/uiNameMappings';
+import { RouteDirection } from '../../types/RouteDirection';
 import { FormInputProps } from '../../uiComponents';
 import { EnumDropdown } from './EnumDropdown';
 
@@ -8,7 +8,7 @@ interface Props extends FormInputProps {
   testId?: string;
 }
 
-export const VehicleModeDropdown = ({
+export const DirectionDropdown = ({
   testId,
   value,
   onChange,
@@ -17,14 +17,14 @@ export const VehicleModeDropdown = ({
   const { t } = useTranslation();
 
   return (
-    <EnumDropdown<ReusableComponentsVehicleModeEnum>
+    <EnumDropdown
       testId={testId}
       value={value}
       onChange={onChange}
       onBlur={onBlur}
-      enumType={ReusableComponentsVehicleModeEnum}
-      placeholder={t('lines.chooseVehicleMode')}
-      uiNameMapper={mapVehicleModeToUiName}
+      enumType={RouteDirection}
+      placeholder={t('routes.chooseDirection')}
+      uiNameMapper={mapDirectionToUiName}
     />
   );
 };
--- a/src/components/forms/LineTypeDropdown.tsx
+++ b/src/components/forms/LineTypeDropdown.tsx
@@ -8,13 +8,20 @@ interface Props extends FormInputProps {
   testId?: string;
 }
 
-export const LineTypeDropdown = (props: Props): JSX.Element => {
+export const LineTypeDropdown = ({
+  testId,
+  value,
+  onChange,
+  onBlur,
+}: Props): JSX.Element => {
   const { t } = useTranslation();
 
   return (
     <EnumDropdown<RouteTypeOfLineEnum>
-      // eslint-disable-next-line react/jsx-props-no-spreading
-      {...props}
+      testId={testId}
+      value={value}
+      onChange={onChange}
+      onBlur={onBlur}
       placeholder={t('lines.chooseLinesType')}
       enumType={RouteTypeOfLineEnum}
       uiNameMapper={mapLineTypeToUiName}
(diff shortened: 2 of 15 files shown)


HenrikHartiala (changes requested):
> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/138)* status: 10 of 14 files reviewed, 3 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[src/components/forms/DirectionDropdown.tsx, line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/138#-N-Wy44m1sdLJ9XJcdz7:-N-Wy44m1sdLJ9XJcdz8:b-trspv1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2f15325b689d2fddb501bb35ffc9da6bacb556d1/src/components/forms/DirectionDropdown.tsx#L17)):*
> ```XML
>     <EnumDropdown
>       // eslint-disable-next-line react/jsx-props-no-spreading
>       {...props}
> ```

I would rather have these props not spread here, because it is a bit more readable that what props are received by the component especially when the hover over interfaces is not showing any information. Though I do see that this same style has been used earlier already (LineTypeDropdown and VehicleModeDropdown), but I would rather have them explicit as well at some point. 

And in addition, you get rid of the eslint-disable comment

What do you think?
___
*[src/components/forms/RoutePropertiesForm.tsx, line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/138#-N-X0hhPDo57ef20bu2X:-N-X0hhPDo57ef20bu2Y:b-9sxpm1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2f15325b689d2fddb501bb35ffc9da6bacb556d1/src/components/forms/RoutePropertiesForm.tsx#L77)):*
> ```XML
>         )}
>         <Row className="mb-5 flex-wrap gap-2">
>           <Column className="w-80 flex-auto">
> ```

Is this explicit `w-80` needed?
___
*[src/components/forms/RoutePropertiesForm.tsx, line 97 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/138#-N-X2zO-8gSKKKn0rccM:-N-X2zO0DvmMsrnKHyiR:bwey7es) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2f15325b689d2fddb501bb35ffc9da6bacb556d1/src/components/forms/RoutePropertiesForm.tsx#L97)):*
> ```XML
>             </p>
>           </Column>
>           <Column className="w-1/6 flex-auto">
> ```

I think you could remove all the explicit widths from this file, and the form will be nicely scaled when window is resized. Now It breaks a little bit when the window is resized. At least on this direction column the width breaks the layout.

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/138)* status: 8 of 16 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[src/components/forms/DirectionDropdown.tsx, line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/138#-N-Wy44m1sdLJ9XJcdz7:-N-Y04fF5cu7Vg5bAsS8:b-zaljme) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2f15325b689d2fddb501bb35ffc9da6bacb556d1/src/components/forms/DirectionDropdown.tsx#L17)):*

I would rather have these props not spread here, because it is a bit more readable that what props are received by the component especially when the hover over interfaces is not showing any information. Though I do see that this same style has been used earlier already (LineTypeDropdown and VehicleModeDropdown), but I would rather have them explicit as well at some point. 

And in addition, you get rid of the eslint-disable comment

What do you think?

</blockquote></details>

Yes why not. This implementation was just a copy paste from the other dropdowns, but I can fix this to them as well.
___
*[src/components/forms/RoutePropertiesForm.tsx, line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/138#-N-X0hhPDo57ef20bu2X:-N-Y-nkg5UJFd3Y4kGfF:b-r9ag33) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2f15325b689d2fddb501bb35ffc9da6bacb556d1/src/components/forms/RoutePropertiesForm.tsx#L77)):*

Is this explicit `w-80` needed?
</blockquote></details>

It is needed, so that in route creation form in map view this field takes up the entire view width.
___
*[src/components/forms/RoutePropertiesForm.tsx, line 97 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/138#-N-X2zO-8gSKKKn0rccM:-N-Y-L_e5PTqMWi2TorF:blej02o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2f15325b689d2fddb501bb35ffc9da6bacb556d1/src/components/forms/RoutePropertiesForm.tsx#L97)):*

I think you could remove all the explicit widths from this file, and the form will be nicely scaled when window is resized. Now It breaks a little bit when the window is resized. At least on this direction column the width breaks the layout.
</blockquote></details>

They are required to make the form look nice in both places it is used, in the route edit page where the form is very wide and in the route creation popup in map view which is narrow.

There could be some more elegant solution to this other than hard coding minimum width, but for now this does the trick.

HenrikHartiala (approved):
Great!

> Reviewed 3 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/138#-:-N-Y3xW_9Bzzgb1JYt_O:btojbfd)* status: 11 of 16 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

PR #141, opened 2022-04-06
Validate that a stop cannot be on the wrong side of a one-way street

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/141)

Closes https://github.com/HSLdevcom/jore4/issues/695

### DIFF
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -2345,8 +2345,8 @@
               "kind": "NON_NULL",
               "name": null,
               "ofType": {
-                "kind": "SCALAR",
-                "name": "String",
+                "kind": "ENUM",
+                "name": "infrastructure_network_direction_enum",
                 "ofType": null
               }
             },
@@ -27196,8 +27196,8 @@
               "kind": "NON_NULL",
               "name": null,
               "ofType": {
-                "kind": "SCALAR",
-                "name": "String",
+                "kind": "ENUM",
+                "name": "infrastructure_network_direction_enum",
                 "ofType": null
               }
             },
@@ -28456,8 +28456,8 @@
               "kind": "NON_NULL",
               "name": null,
               "ofType": {
-                "kind": "SCALAR",
-                "name": "String",
+                "kind": "ENUM",
+                "name": "infrastructure_network_direction_enum",
                 "ofType": null
               }
             },
@@ -29360,8 +29360,8 @@
             "name": "direction",
             "description": "The direction(s) of traffic with respect to the digitization, i.e. the direction of the specified line string.",
             "type": {
-              "kind": "SCALAR",
-              "name": "String",
+              "kind": "ENUM",
+              "name": "infrastructure_network_direction_enum",
               "ofType": null
             },
             "defaultValue": null,
--- a/schemaLoader.ts
+++ b/schemaLoader.ts
@@ -33,7 +33,7 @@ const hasuraOverrideTypeDefs = gql`
     scheduled_stop_point_id: uuid!
     measured_location: geography_point!
     located_on_infrastructure_link_id: uuid!
-    direction: String!
+    direction: infrastructure_network_direction_enum!
     label: String!
     priority: Int!
   }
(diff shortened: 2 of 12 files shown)


HenrikHartiala (changes requested):
> Reviewed 4 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/141)* status: 4 of 10 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala and @pallost)

___
*[src/hooks/stops/useEditStop.ts, line 107 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/141#-MzzA8go64xghgQf6ZO-:-MzzA8go64xghgQf6ZO0:b-7uonm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/eb86ef3797b6b92cf2dd3b77d8556fe4e5019151/src/hooks/stops/useEditStop.ts#L107)):*
> ```XML
>       mapGetStopWithRouteGraphDataByIdResult(stopRoutesResult);
> 
>     let deleteStopFromJourneyPatterns: JourneyPatternJourneyPattern[] = [];
> ```

I think this `let deleteStopFromJourneyPatterns...` could / should be just `const deleteStopFromJourneyPatterns = getJourneyPatternsToDeleteStopFrom(...)`

Same thing for the `let deleteStopFromRoutes...`

Not sure if this is because of the typing, but in that case the typing could be added to the method called.



_Code quote:_
```XML
getJourneyPatternsToDeleteStopFrom
```
___
*[src/hooks/stops/useGetStopLinkAndDirection.ts, line 40 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/141#-MzzBv_XCbsUHsLIS92G:-MzzBv_XCbsUHsLIS92H:bdxiply) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/eb86ef3797b6b92cf2dd3b77d8556fe4e5019151/src/hooks/stops/useGetStopLinkAndDirection.ts#L40)):*
> ```XML
>   // based on internal_service_pattern.check_scheduled_stop_point_infrastructure_link_direction()
>   const areDirectionsCompatible = (
>     stopDirection: string,
> ```

Should we have an enum for the stopDirection as well?

HenrikHartiala (approved):
> Reviewed 6 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/141)* status: 7 of 12 files reviewed, all discussions resolved (waiting on @HenrikHartiala)

___
*[src/hooks/stops/useEditStop.ts, line 107 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/141#-MzzA8go64xghgQf6ZO-:-N-2tmSf3Ya0X8Bay4rX:b-fovh7e) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/eb86ef3797b6b92cf2dd3b77d8556fe4e5019151/src/hooks/stops/useEditStop.ts#L107)):*

I think this `let deleteStopFromJourneyPatterns...` could / should be just `const deleteStopFromJourneyPatterns = getJourneyPatternsToDeleteStopFrom(...)`

Same thing for the `let deleteStopFromRoutes...`

Not sure if this is because of the typing, but in that case the typing could be added to the method called.
</blockquote></details>

Ah, did not notice at first that the assignment of these variables are on different scope. This could be refactored with helper functions so that we wouldn't need to use the `let`s. But it's okay this way as well :+1:

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/141)* status: 7 of 12 files reviewed, all discussions resolved (waiting on @HenrikHartiala)

___
*[src/hooks/stops/useEditStop.ts, line 107 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/141#-MzzA8go64xghgQf6ZO-:-N-2Vhep0F0l_AIeA5z1:b-4b11x6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/eb86ef3797b6b92cf2dd3b77d8556fe4e5019151/src/hooks/stops/useEditStop.ts#L107)):*

Ah, did not notice at first that the assignment of these variables are on different scope. This could be refactored with helper functions so that we wouldn't need to use the `let`s. But it's okay this way as well :+1: 
</blockquote></details>

I've started refactoring it, but it's a bigger change and not really in the scope of this PR. We're also building on top of this function in later commits... I'll do this later, another time, OK?

PR #147, opened 2022-04-11
Add line details observation by date

HenrikHartiala:
HSLdevcom/jore4#493

Add observation date functionality to line details page.

When navigating to the line details page for the first time, it will use **todays date** as observation date, if the line is active **today**. If the line **is not active** today, it will use the lines **first active date** as observation date.

With the date picker, the user can clear the date to see the raw details of the line.

Technical note:
The query will filter out the draft-versions and also choose only the line, routes, stops which are active on the given date. But because graphQL seems not to have groupBy, the correct line_routes are filtered in the code (to show only the highest priority route per label).

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)

### DIFF
--- a/src/components/routes-and-lines/line-details/ActionsRow.tsx
+++ b/src/components/routes-and-lines/line-details/ActionsRow.tsx
@@ -1,0 +1,61 @@
+import produce from 'immer';
+import { DateTime } from 'luxon';
+import qs from 'qs';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useHistory } from 'react-router';
+import { useUrlQuery } from '../../../hooks';
+import { useGetLineDetails } from '../../../hooks/line-details/useGetLineDetails';
+import { Column, Container, Row } from '../../../layoutComponents';
+import { SimpleButton } from '../../../uiComponents';
+
+export const ActionsRow = ({ className = '' }: { className?: string }) => {
+  const { t } = useTranslation();
+  const history = useHistory();
+  const queryParams = useUrlQuery();
+
+  const { observationDate } = useGetLineDetails();
+
+  const onDateChange = (date: DateTime) => {
+    const updatedUrlQuery = produce(queryParams, (draft) => {
+      if (date.isValid) {
+        draft.selectedDate = date.toISODate();
+        delete draft.showAll;
+      } else {
+        draft.showAll = true.toString();
+        delete draft.selectedDate;
+      }
+    });
+
+    const queryString = qs.stringify(updatedUrlQuery);
+    history.push({
+      search: `?${queryString}`,
+    });
+  };
+
+  return (
+    <Container className={className}>
+      <Row>
+        <h2 className="text-sm font-bold">{t('filters.observationDate')}</h2>
+      </Row>
+      <Row>
+        <Column className="w-1/4">
+          <input
+            type="date"
+            value={observationDate?.toISODate() || ''}
+            onChange={(e) => onDateChange(DateTime.fromISO(e.target.value))}
+            className="flex-1"
+          />
+        </Column>
+        <SimpleButton
+          className="ml-auto"
+          inverted
+          disabled
+          onClick={() => console.log('TODO')} // eslint-disable-line no-console
+        >
+          {t('lines.showDrafts')}
+        </SimpleButton>
+      </Row>
+    </Container>
+  );
+};
--- a/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
+++ b/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
@@ -2,12 +2,12 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { AiFillPlusCircle } from 'react-icons/ai';
 import { useParams } from 'react-router-dom';
+import { RouteLine } from '../../../generated/graphql';
 import {
-  RouteLine,
-  useGetLineDetailsWithRoutesByIdQuery,
-} from '../../../generated/graphql';
-import { mapLineDetailsWithRoutesResult } from '../../../graphql';
-import { useAppDispatch, useMapUrlQuery } from '../../../hooks';
+  useAppDispatch,
+  useGetLineDetails,
+  useMapUrlQuery,
+} from '../../../hooks';
 import { Column, Container, Row } from '../../../layoutComponents';
 import {
   resetMapEditorStateAction,
@@ -16,8 +16,8 @@ import {
 } from '../../../redux';
 import { DateLike, mapToShortDate } from '../../../time';
 import { SimpleButton } from '../../../uiComponents';
-import { mapToVariables } from '../../../utils';
 import { PageHeader } from '../common/PageHeader';
+import { ActionsRow } from './ActionsRow';
 import { AdditionalInformation } from './AdditionalInformation';
 import { MapPreview } from './MapPreview';
 import { RouteStopsTable } from './RouteStopsTable';
@@ -91,13 +91,11 @@ const LineTitle: React.FC<LineTitleProps> = ({
 export const LineDetailsPage = (): JSX.Element => {
   const { t } = useTranslation();
   const { id } = useParams<{ id: string }>();
+
   const dispatch = useAppDispatch();
   const { addMapOpenQueryParameter } = useMapUrlQuery();
 
-  const lineDetailsResult = useGetLineDetailsWithRoutesByIdQuery(
-    mapToVariables({ line_id: id }),
-  );
-  const line = mapLineDetailsWithRoutesResult(lineDetailsResult);
+  const { line } = useGetLineDetails();
 
   const buildValidityPeriod = (
     validityStart?: DateLike | null,
@@ -131,8 +129,9 @@ export const LineDetailsPage = (): JSX.Element => {
           </Column>
         </Row>
       </PageHeader>
+      <ActionsRow className="!pt-4 !pb-0" />
       {line && (
-        <Container>
+        <Container className="pt-10">
           <Column>
             <Row>
               <AdditionalInformation className="w-2/3" line={line} />
--- a/src/generated/graphql.tsx
+++ b/src/generated/graphql.tsx
@@ -6701,6 +6701,15 @@ export type GetLineDetailsWithRoutesByIdQueryVariables = Exact<{
 
 export type GetLineDetailsWithRoutesByIdQuery = { __typename?: 'query_root', route_line_by_pk?: { __typename?: 'route_line', line_id: UUID, name_i18n: string, short_name_i18n?: string | null | undefined, primary_vehicle_mode: ReusableComponentsVehicleModeEnum, type_of_line: RouteTypeOfLineEnum, transport_target: HslRouteTransportTargetEnum, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string, line_routes: Array<{ __typename?: 'route_route', route_id: UUID, description_i18n?: string | null | undefined, starts_from_scheduled_stop_point_id: UUID, ends_at_scheduled_stop_point_id: UUID, route_shape?: GeoJSON.LineString | null | undefined, on_line_id: UUID, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string, direction: string, infrastructure_links_along_route: Array<{ __typename?: 'route_infrastructure_link_along_route', infrastructure_link: { __typename?: 'infrastructure_network_infrastructure_link', scheduled_stop_point_located_on_infrastructure_link: Array<{ __typename?: 'service_pattern_scheduled_stop_point', scheduled_stop_point_id: UUID, label: string, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, scheduled_stop_point_in_journey_patterns: Array<{ __typename?: 'journey_pattern_scheduled_stop_point_in_journey_pattern', journey_pattern_id: UUID, scheduled_stop_point_id: UUID, scheduled_stop_point_sequence: number, is_timing_point: boolean, is_via_point: boolean, journey_pattern: { __typename?: 'journey_pattern_journey_pattern', on_route_id: UUID } }> }> } }>, starts_from_scheduled_stop_point?: { __typename?: 'service_pattern_scheduled_stop_point', scheduled_stop_point_id: UUID, label: string, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined } | null | undefined, ends_at_scheduled_stop_point?: { __typename?: 'service_pattern_scheduled_stop_point', scheduled_stop_point_id: UUID, label: string, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined } | null | undefined }> } | null | undefined };
 
+export type GetHighestPriorityLineDetailsWithRoutesQueryVariables = Exact<{
+  lineFilters?: Maybe<RouteLineBoolExp>;
+  lineRouteFilters?: Maybe<RouteRouteBoolExp>;
+  routeStopFilters?: Maybe<ServicePatternScheduledStopPointBoolExp>;
+}>;
+
+
+export type GetHighestPriorityLineDetailsWithRoutesQuery = { __typename?: 'query_root', route_line: Array<{ __typename?: 'route_line', line_id: UUID, name_i18n: string, short_name_i18n?: string | null | undefined, primary_vehicle_mode: ReusableComponentsVehicleModeEnum, type_of_line: RouteTypeOfLineEnum, transport_target: HslRouteTransportTargetEnum, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string, line_routes: Array<{ __typename?: 'route_route', route_id: UUID, description_i18n?: string | null | undefined, starts_from_scheduled_stop_point_id: UUID, ends_at_scheduled_stop_point_id: UUID, route_shape?: GeoJSON.LineString | null | undefined, on_line_id: UUID, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string, direction: string, infrastructure_links_along_route: Array<{ __typename?: 'route_infrastructure_link_along_route', infrastructure_link: { __typename?: 'infrastructure_network_infrastructure_link', scheduled_stop_point_located_on_infrastructure_link: Array<{ __typename?: 'service_pattern_scheduled_stop_point', scheduled_stop_point_id: UUID, label: string, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, scheduled_stop_point_in_journey_patterns: Array<{ __typename?: 'journey_pattern_scheduled_stop_point_in_journey_pattern', journey_pattern_id: UUID, scheduled_stop_point_id: UUID, scheduled_stop_point_sequence: number, is_timing_point: boolean, is_via_point: boolean, journey_pattern: { __typename?: 'journey_pattern_journey_pattern', on_route_id: UUID } }> }> } }>, starts_from_scheduled_stop_point?: { __typename?: 'service_pattern_scheduled_stop_point', scheduled_stop_point_id: UUID, label: string, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined } | null | undefined, ends_at_scheduled_stop_point?: { __typename?: 'service_pattern_scheduled_stop_point', scheduled_stop_point_id: UUID, label: string, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined } | null | undefined }> }> };
+
 export type GetRouteDetailsByIdsQueryVariables = Exact<{
   route_ids?: Maybe<Array<Scalars['uuid']> | Scalars['uuid']>;
 }>;
@@ -7456,6 +7465,62 @@ export function useGetLineDetailsWithRoutesByIdLazyQuery(baseOptions?: Apollo.La
 export type GetLineDetailsWithRoutesByIdQueryHookResult = ReturnType<typeof useGetLineDetailsWithRoutesByIdQuery>;
 export type GetLineDetailsWithRoutesByIdLazyQueryHookResult = ReturnType<typeof useGetLineDetailsWithRoutesByIdLazyQuery>;
 export type GetLineDetailsWithRoutesByIdQueryResult = Apollo.QueryResult<GetLineDetailsWithRoutesByIdQuery, GetLineDetailsWithRoutesByIdQueryVariables>;
+export const GetHighestPriorityLineDetailsWithRoutesDocument = gql`
+    query GetHighestPriorityLineDetailsWithRoutes($lineFilters: route_line_bool_exp, $lineRouteFilters: route_route_bool_exp, $routeStopFilters: service_pattern_scheduled_stop_point_bool_exp) {
+  route_line(where: $lineFilters, order_by: {priority: desc}, limit: 1) {
+    ...line_all_fields
+    line_routes(where: $lineRouteFilters) {
+      ...route_with_stops
+      infrastructure_links_along_route {
+        infrastructure_link {
+          scheduled_stop_point_located_on_infrastructure_link(where: $routeStopFilters) {
+            ...scheduled_stop_point_default_fields
+            scheduled_stop_point_in_journey_patterns {
+              ...scheduled_stop_point_in_journey_pattern_default_fields
+              journey_pattern {
+                on_route_id
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
+    ${LineAllFieldsFragmentDoc}
+${RouteWithStopsFragmentDoc}
+${ScheduledStopPointDefaultFieldsFragmentDoc}
+${ScheduledStopPointInJourneyPatternDefaultFieldsFragmentDoc}`;
+
+/**
+ * __useGetHighestPriorityLineDetailsWithRoutesQuery__
+ *
+ * To run a query within a React component, call `useGetHighestPriorityLineDetailsWithRoutesQuery` and pass it any options that fit your needs.
+ * When your component renders, `useGetHighestPriorityLineDetailsWithRoutesQuery` returns an object from Apollo Client that contains loading, error, and data properties
+ * you can use to render your UI.
+ *
+ * @param baseOptions options that will be passed into the query, supported options are listed on: https://www.apollographql.com/docs/react/api/react-hooks/#options;
+ *
+ * @example
+ * const { data, loading, error } = useGetHighestPriorityLineDetailsWithRoutesQuery({
+ *   variables: {
+ *      lineFilters: // value for 'lineFilters'
+ *      lineRouteFilters: // value for 'lineRouteFilters'
+ *      routeStopFilters: // value for 'routeStopFilters'
+ *   },
+ * });
+ */
+export function useGetHighestPriorityLineDetailsWithRoutesQuery(baseOptions?: Apollo.QueryHookOptions<GetHighestPriorityLineDetailsWithRoutesQuery, GetHighestPriorityLineDetailsWithRoutesQueryVariables>) {
+        const options = {...defaultOptions, ...baseOptions}
+        return Apollo.useQuery<GetHighestPriorityLineDetailsWithRoutesQuery, GetHighestPriorityLineDetailsWithRoutesQueryVariables>(GetHighestPriorityLineDetailsWithRoutesDocument, options);
+      }
+export function useGetHighestPriorityLineDetailsWithRoutesLazyQuery(baseOptions?: Apollo.LazyQueryHookOptions<GetHighestPriorityLineDetailsWithRoutesQuery, GetHighestPriorityLineDetailsWithRoutesQueryVariables>) {
+          const options = {...defaultOptions, ...baseOptions}
+          return Apollo.useLazyQuery<GetHighestPriorityLineDetailsWithRoutesQuery, GetHighestPriorityLineDetailsWithRoutesQueryVariables>(GetHighestPriorityLineDetailsWithRoutesDocument, options);
+        }
+export type GetHighestPriorityLineDetailsWithRoutesQueryHookResult = ReturnType<typeof useGetHighestPriorityLineDetailsWithRoutesQuery>;
+export type GetHighestPriorityLineDetailsWithRoutesLazyQueryHookResult = ReturnType<typeof useGetHighestPriorityLineDetailsWithRoutesLazyQuery>;
+export type GetHighestPriorityLineDetailsWithRoutesQueryResult = Apollo.QueryResult<GetHighestPriorityLineDetailsWithRoutesQuery, GetHighestPriorityLineDetailsWithRoutesQueryVariables>;
 export const GetRouteDetailsByIdsDocument = gql`
     query GetRouteDetailsByIds($route_ids: [uuid!]) {
   route_route(where: {route_id: {_in: $route_ids}}) {
@@ -8263,6 +8328,10 @@ export function useGetLineDetailsWithRoutesByIdAsyncQuery() {
           return useAsyncQuery<GetLineDetailsWithRoutesByIdQuery, GetLineDetailsWithRoutesByIdQueryVariables>(GetLineDetailsWithRoutesByIdDocument);
         }
 export type GetLineDetailsWithRoutesByIdAsyncQueryHookResult = ReturnType<typeof useGetLineDetailsWithRoutesByIdAsyncQuery>;
+export function useGetHighestPriorityLineDetailsWithRoutesAsyncQuery() {
+          return useAsyncQuery<GetHighestPriorityLineDetailsWithRoutesQuery, GetHighestPriorityLineDetailsWithRoutesQueryVariables>(GetHighestPriorityLineDetailsWithRoutesDocument);
+        }
+export type GetHighestPriorityLineDetailsWithRoutesAsyncQueryHookResult = ReturnType<typeof useGetHighestPriorityLineDetailsWithRoutesAsyncQuery>;
 export function useGetRouteDetailsByIdsAsyncQuery() {
           return useAsyncQuery<GetRouteDetailsByIdsQuery, GetRouteDetailsByIdsQueryVariables>(GetRouteDetailsByIdsDocument);
         }
--- a/src/graphql/route.ts
+++ b/src/graphql/route.ts
@@ -6,6 +6,7 @@ import {
   InsertLineOneMutation,
   RouteLine,
   RouteRoute,
+  useGetHighestPriorityLineDetailsWithRoutesQuery,
   useGetLineDetailsByIdQuery,
   useGetRouteDetailsByIdsQuery,
   useGetRouteDetailsByLabelWildcardQuery,
@@ -201,9 +202,48 @@ const GET_LINE_DETAILS_WITH_ROUTES_BY_ID = gql`
     }
   }
 `;
+
 export const mapLineDetailsWithRoutesResult = (
   result: GqlQueryResult<GetLineDetailsWithRoutesByIdQuery>,
 ) => result.data?.route_line_by_pk as RouteLine | undefined;
+
+const GET_HIGHEST_PRIORITY_LINE_DETAILS_WITH_ROUTES = gql`
+  query GetHighestPriorityLineDetailsWithRoutes(
+    $lineFilters: route_line_bool_exp
+    $lineRouteFilters: route_route_bool_exp
+    $routeStopFilters: service_pattern_scheduled_stop_point_bool_exp
+  ) {
+    route_line(where: $lineFilters, order_by: { priority: desc }, limit: 1) {
+      ...line_all_fields
+      line_routes(where: $lineRouteFilters) {
+        ...route_with_stops
+        infrastructure_links_along_route {
+          infrastructure_link {
+            scheduled_stop_point_located_on_infrastructure_link(
+              where: $routeStopFilters
+            ) {
+              ...scheduled_stop_point_default_fields
+              scheduled_stop_point_in_journey_patterns {
+                ...scheduled_stop_point_in_journey_pattern_default_fields
+                journey_pattern {
+                  on_route_id
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+// The used query has limit: 1 for the result so it can't have more than 1 route_line in result array.
+export const mapHighestPriorityLineDetailsWithRoutesResult = (
+  result: ReturnType<typeof useGetHighestPriorityLineDetailsWithRoutesQuery>,
+) =>
+  result.data?.route_line.length
+    ? (result.data?.route_line[0] as RouteLine)
+    : undefined;
 
 const GET_ROUTE_DETAILS_BY_IDS = gql`
   query GetRouteDetailsByIds($route_ids: [uuid!]) {
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -1,3 +1,4 @@
+export * from './line-details/useGetLineDetails';
 export * from './redux';
 export * from './routes';
 export * from './search/useSearch';
--- a/src/hooks/line-details/useGetLineDetails.ts
+++ b/src/hooks/line-details/useGetLineDetails.ts
@@ -1,0 +1,149 @@
+import produce from 'immer';
+import { groupBy } from 'lodash';
+import { DateTime } from 'luxon';
+import { useParams } from 'react-router-dom';
+import {
+  RouteDirectionEnum,
+  RouteLine,
+  RouteRoute,
+  useGetHighestPriorityLineDetailsWithRoutesQuery,
+  useGetLineDetailsWithRoutesByIdQuery,
+} from '../../generated/graphql';
+import {
+  mapHighestPriorityLineDetailsWithRoutesResult,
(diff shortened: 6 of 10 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)* status: 0 of 8 files reviewed, 17 unresolved discussions (waiting on @eleksi, @HenrikHartiala, and @pallost)

___
*[src/components/routes-and-lines/lineDetails/TopActionsRow.tsx, line 9 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WUkiNAAA8jgZx7YIZ:-N-WUkiNAAA8jgZx7YI_:b-tvk4w5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/components/routes-and-lines/lineDetails/TopActionsRow.tsx#L9)):*
> ```XML
> import { SimpleButton } from '../../../uiComponents';
> 
> export const TopActionsRow = ({ className }: { className?: string }) => {
> ```

What if this is not positioned to "Top" in some context? Maybe this should be called just "ActionsRow" or something like that.
___
*[src/components/routes-and-lines/lineDetails/TopActionsRow.tsx, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WV0La9VuIMPaPnn-S:-N-WV0La9VuIMPaPnn-T:b-y8rebt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/components/routes-and-lines/lineDetails/TopActionsRow.tsx#L20)):*
> ```XML
> 
>     history.push({
>       search: `?${queryString}`,
> ```

This won't work if we already have query in the url. I recommend handling query strings with library such as `qs` (https://www.npmjs.com/package/qs) that we are already using.
We have `src/hooks/useUrlQuery.ts` hook for parsing current query, then you could modify that accordingly.
___
*[src/graphql/route.ts, line 210 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-Wc6m539-MJVTTBDqR:-N-Wc6m539-MJVTTBDqS:b-b8jz2v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/graphql/route.ts#L210)):*
> ```XML
> 
> const GET_ACTIVE_LINE_DETAILS_WITH_ROUTES_BY_DATE = gql`
>   query GetActiveLineDetailsWithRoutesByDate(
> ```

Just wondering if label, date and filterPriority could be given with `BoolExp` filter. That could simplify this a bit. See `src/hooks/useCheckValidityAndPriorityConflicts.ts` as an example.
___
*[src/graphql/route.ts, line 218 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WcYbb-nHZGTnf5ego:-N-WcYbb-nHZGTnf5egp:b-8hu5yg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/graphql/route.ts#L218)):*
> ```XML
>       where: {
>         label: { _eq: $label }
>         validity_start: { _lte: $date }
> ```

At least stops might have validity_start `null`. How about lines?
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 2 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WWCti5AFRAKll6DLe:-N-WWCti5AFRAKll6DLf:bcayup7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L2)):*
> ```XML
> import produce from 'immer';
> import _ from 'lodash';
> ```

Don't import the whole lodash library. That increases bundle size for nothing as you're not using all of its features anyway.
You can import individual features from lodash like this:
`import merge from 'lodash/merge';`
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 32 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WWoqy-deHhP02On26:-N-WWoqy-deHhP02On27:b-4lcr8c) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L32)):*
> ```XML
> 
>   const filteredLine = produce(line, (draft) => {
>     if (draft) {
> ```

Can this really be undefined?
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 43 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WX-bLEQG7vaHvhDwb:-N-WX-bLEQG7vaHvhDwc:b-9pn7nk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L43)):*
> ```XML
> const getInitialDate = (
>   showAll: boolean,
>   selectedDate: string,
> ```

Maybe you could call this `selectedISODate` or something like that. Now the reader wonders what kind of date string this is.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 47 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-Wbjq71bL_pxQmdKTR:-N-Wbjq71bL_pxQmdKTS:bn47ln6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L47)):*
> ```XML
>   validityEnd?: DateTime,
> ) => {
>   if (showAll) {
> ```

Does it make sense to get this parameter as input to this function if we're just returning undefined if it is true? Interface if this function would be clearer if it would be handled before this function is even called.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 52 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WXVpyAhrjTJXW80R0:-N-WXVpyAhrjTJXW80R1:bxds6w4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L52)):*
> ```XML
> 
>   if (selectedDate) {
>     return DateTime.fromISO(selectedDate);
> ```

Should probably use `parseDate` from `src/time.ts`
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 60 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WaUU26oM8tGHILABM:-N-WaUU26oM8tGHILABN:b8znfwe) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L60)):*
> ```XML
>     (!validityEnd || validityEnd >= DateTime.now());
> 
>   if (!isActiveToday) {
> ```

This code would be easier to read if written like this: 
```
if (isActiveToday) {
  return DateTime.now()
}
return validityStart;
```
Then we could avoid negation from if clause and keep `isActiveToday` close to the logical clause coupled to it ("active today -> return today")
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 64 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WXdL76oPJAiN4ztuQ:-N-WXdL76oPJAiN4ztuR:b-fkiic1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L64)):*
> ```XML
>   }
> 
>   return DateTime.fromISO(DateTime.now().toISODate());
> ```

We're casting DateTime.now() to ISO date and then initializing DateTime from ISO date? That doesn't make sense. `DateTime.now()` probably gives the same result already?
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 81 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZ8c95OU9ke2pRaVc:-N-WZ8c95OU9ke2pRaVd:bei2gzn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L81)):*
> ```XML
>   );
> 
>   const line = mapLineDetailsWithRoutesResult(lineDetailsResult) as RouteLine;
> ```

`mapLineDetailsWithRoutesResult` already casts value to `RouteLine | undefined`. Now we're effectively doing non-null assesrion with `as` operator, which sounds fishy.
Remove `as` casting from here and make non-null assertion properly if needed, please. :)
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 94 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZceL-1SSdhWEzH-u:-N-WZceL-1SSdhWEzH-v:b-6k5zhe) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L94)):*
> ```XML
>       date: observationDate,
>       // If observing non-draft line, filter out everything that are linked to drafts
>       filterPriority: line?.priority !== 30 ? [30] : [],
> ```

What are these magic numbers? I think we have enum for these...
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 100 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZmfPAmKYNKzQRd_7:-N-WZmfPAmKYNKzQRd_8:b9c4iwb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L100)):*
> ```XML
>   const lineByDate = mapActiveLineDetailsByDateResult(
>     lineByDateResult,
>   ) as RouteLine[];
> ```

Same thing here, we're making non-null assertion with `as` operator and it doesn't make sense.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 102 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-W_p8u8vJ935A_jsex:-N-W_p8v1Fqzl4kNrxTQ:b-q8st9b) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L102)):*
> ```XML
>   ) as RouteLine[];
> 
>   const filteredLine = lineByDate
> ```

`lineByDate` -> `lineByDate?.length`to avoid exception if array is empty.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 103 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZuASCXK264SST8Y1:-N-WZuAT3Ql9OKVoh2QE:biyp1ba) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L103)):*
> ```XML
> 
>   const filteredLine = lineByDate
>     ? filterLineDetailsByDate(lineByDate[0])
> ```

Why we are taking first element from array? What if there are multiple elements in array?
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 107 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-W_AsX9-2RjLNRUKlf:-N-W_AsX9-2RjLNRUKlg:b-plgu39) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L107)):*
> ```XML
> 
>   return {
>     line: showAll ? line : filteredLine,
> ```

You're calculating filtered line but throwing it away if `showAll` is true? If the operation is not expensive I'm fine with this, but you could also get rid of unneeded calculation somehow like this: `const filteredLine = (!showAll && lineByDate) ? filterLineDetailsByDate(...) : undefined`.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)* status: 0 of 8 files reviewed, 17 unresolved discussions (waiting on @eleksi, @pallost, and @tk-vincit)

___
*[src/components/routes-and-lines/lineDetails/TopActionsRow.tsx, line 9 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WUkiNAAA8jgZx7YIZ:-N-XSAuf6fZF6gWVyjuu:bfeatro) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/components/routes-and-lines/lineDetails/TopActionsRow.tsx#L9)):*

What if this is not positioned to "Top" in some context? Maybe this should be called just "ActionsRow" or something like that.
</blockquote></details>

True, I was already debating with myself that this might not be the best naming.
___
*[src/components/routes-and-lines/lineDetails/TopActionsRow.tsx, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WV0La9VuIMPaPnn-S:-N-X__Sb1a-v7HBFJ5EZ:bfskl1k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/components/routes-and-lines/lineDetails/TopActionsRow.tsx#L20)):*

This won't work if we already have query in the url. I recommend handling query strings with library such as `qs` (https://www.npmjs.com/package/qs) that we are already using.
We have `src/hooks/useUrlQuery.ts` hook for parsing current query, then you could modify that accordingly.
</blockquote></details>

Thanks, I'll made changes to use qs and immer. Not sure if this solution is what you were thinking, but now it won't overwrite existing queryString, only the parts related to this.
___
*[src/graphql/route.ts, line 210 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-Wc6m539-MJVTTBDqR:-N-Y9F4o-8Yq9S_I7qOa:b-xsmed3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/graphql/route.ts#L210)):*

Just wondering if label, date and filterPriority could be given with `BoolExp` filter. That could simplify this a bit. See `src/hooks/useCheckValidityAndPriorityConflicts.ts` as an example.
</blockquote></details>

Im not sure if this is the best case (or possible) to be compiled on code side properly, because the where conditions differ a bit from eachother. (line, routes and stops)

With Tamas we did try to reduce repetition on this by trying to get this part
```
validity_start: { _lte: $date }
                _or: [
                  { validity_end: { _gte: $date } }
                  { validity_end: { _is_null: true } }
                ]
```
as fragmentation, but that did not work out. For this first iteration I'll leave this as it is and check tomorrow if there is a nice way to change this nicer.
___
*[src/graphql/route.ts, line 218 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WcYbb-nHZGTnf5ego:-N-Y9o1D8GFQchAW7mfW:bbrqgmm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/graphql/route.ts#L218)):*

At least stops might have validity_start `null`. How about lines?
</blockquote></details>

On slack thread petri said that at least from jore3 the stops will always have validity_start. But I can add the support for null's for now for all of these. Let's see how we decide on the nullabilities of validity_start
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 2 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WWCti5AFRAKll6DLe:-N-XpCcP7Eyx-FgVX8cq:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L2)):*

Don't import the whole lodash library. That increases bundle size for nothing as you're not using all of its features anyway.
You can import individual features from lodash like this:
`import merge from 'lodash/merge';`
</blockquote></details>

:+1: 
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 32 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WWoqy-deHhP02On26:-N-XpVj578vnO31DZem7:bgke0ul) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L32)):*

Can this really be undefined?
</blockquote></details>

Hmm, no. Good one
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 43 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WX-bLEQG7vaHvhDwb:-N-Xp_gT1Q8hcmM-88rl:bb70w76) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L43)):*

Maybe you could call this `selectedISODate` or something like that. Now the reader wonders what kind of date string this is.
</blockquote></details>

Nice
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 47 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-Wbjq71bL_pxQmdKTR:-N-XpwTG7BE3w5t-Nzey:bpa65ca) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L47)):*

Does it make sense to get this parameter as input to this function if we're just returning undefined if it is true? Interface if this function would be clearer if it would be handled before this function is even called.
</blockquote></details>

Good point, makes more sense.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 52 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WXVpyAhrjTJXW80R0:-N-XqWwP2wPDoAHXfMpX:bd5wsvf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L52)):*

Should probably use `parseDate` from `src/time.ts`
</blockquote></details>

Ahh, I was wondering if we had some library for these(or if luxon had some methods), but did not think of checking if we had own methods for these. Nice. Was trying to find the `isDateInRange` method earlier.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 60 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WaUU26oM8tGHILABM:-N-XrErO7IOlzNpLI2Rg:b-80bnvp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L60)):*

This code would be easier to read if written like this: 
```
if (isActiveToday) {
  return DateTime.now()
}
return validityStart;
```
Then we could avoid negation from if clause and keep `isActiveToday` close to the logical clause coupled to it ("active today -> return today")
</blockquote></details>

True.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 64 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WXdL76oPJAiN4ztuQ:-N-XrXeADiMLroXbGW5d:b-pf7q1n) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L64)):*

We're casting DateTime.now() to ISO date and then initializing DateTime from ISO date? That doesn't make sense. `DateTime.now()` probably gives the same result already?
</blockquote></details>

I'm not completely sure why, but the DateTime.now() is not working, because it keeps updating and makes infinite re-renders. Not sure where it goes wrong.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 81 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZ8c95OU9ke2pRaVc:-N-XuLhJ-Ua0hQVB9tP0:b-ep7msk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L81)):*

`mapLineDetailsWithRoutesResult` already casts value to `RouteLine | undefined`. Now we're effectively doing non-null assesrion with `as` operator, which sounds fishy.
Remove `as` casting from here and make non-null assertion properly if needed, please. :)
</blockquote></details>

Hmm, not sure what I've been thinking here. I'll fix it
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 94 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZceL-1SSdhWEzH-u:-N-XvKuUC1NxwjshQd0v:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L94)):*

What are these magic numbers? I think we have enum for these...
</blockquote></details>

:+1: 
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 100 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZmfPAmKYNKzQRd_7:-N-XvcaO2H_MDhXSj1BF:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L100)):*

Same thing here, we're making non-null assertion with `as` operator and it doesn't make sense.
</blockquote></details>

:+1: 
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 102 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-W_p8u8vJ935A_jsex:-N-Xve1DCuLFtXMLe5fN:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L102)):*

`lineByDate` -> `lineByDate?.length`to avoid exception if array is empty.
</blockquote></details>

:+1: 
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 103 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZuASCXK264SST8Y1:-N-XwRPB2OMH5KJc-90R:b-hb95gi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L103)):*

Why we are taking first element from array? What if there are multiple elements in array?
</blockquote></details>

There can only be one result, either draft or primary line. But I can't find how to "FirstOrDefault()" in GraphQL.

Ah, but I can move the [0] in the grapqhl file though. Not sure if this is how it should be done..?
```
export const mapActiveLineDetailsByDateResult = (
  result: ReturnType<typeof useGetActiveLineDetailsWithRoutesByDateQuery>,
) => result.data?.route_line[0] as RouteLine | undefined;
```
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 107 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-W_AsX9-2RjLNRUKlf:-N-XzTglBmMGSpl5iUE5:b-orumjy) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L107)):*

You're calculating filtered line but throwing it away if `showAll` is true? If the operation is not expensive I'm fine with this, but you could also get rid of unneeded calculation somehow like this: `const filteredLine = (!showAll && lineByDate) ? filterLineDetailsByDate(...) : undefined`.
</blockquote></details>

Good point

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @eleksi, @HenrikHartiala, and @pallost)

___
*[src/graphql/route.ts, line 210 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-Wc6m539-MJVTTBDqR:-N05FrHU0wkwyeRwOuqV:b32w1t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/graphql/route.ts#L210)):*

Im not sure if this is the best case (or possible) to be compiled on code side properly, because the where conditions differ a bit from eachother. (line, routes and stops)

With Tamas we did try to reduce repetition on this by trying to get this part
```
validity_start: { _lte: $date }
                _or: [
                  { validity_end: { _gte: $date } }
                  { validity_end: { _is_null: true } }
                ]
```
as fragmentation, but that did not work out. For this first iteration I'll leave this as it is and check tomorrow if there is a nice way to change this nicer.
</blockquote></details>

Ok
___
*[src/graphql/route.ts, line 218 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WcYbb-nHZGTnf5ego:-N05GBH23rrcCcTh_71k:bum5cdy) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/graphql/route.ts#L218)):*

On slack thread petri said that at least from jore3 the stops will always have validity_start. But I can add the support for null's for now for all of these. Let's see how we decide on the nullabilities of validity_start
</blockquote></details>

I'm a bit worried about having these here, as we already have very similar date range queries in `src/hooks/useCheckValidityAndPriorityConflicts.ts`. Maintaining these will be nightmare if we realise that we have a bug in some date range query and then have to find out all places where same bug might be. 

Let's discuss about this issue in Slack.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 64 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WXdL76oPJAiN4ztuQ:-N05H57R3uttT_hjnlqa:bqdso7r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L64)):*

I'm not completely sure why, but the DateTime.now() is not working, because it keeps updating and makes infinite re-renders. Not sure where it goes wrong.
</blockquote></details>

Ok. Sounds very weird, but as it is at least commented now then I think this is OK for now.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 102 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-W_p8u8vJ935A_jsex:-N05HQjYCyS94aZSdf6e:bgorblt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L102)):*

:+1: 
</blockquote></details>

Seems like this wasn't changed, but is the reason that lineByDate is not array anymore or what..? Its a bit difficult to parse all the changes in my head :)
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 103 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZuASCXK264SST8Y1:-N05Hg0DE978ns96YR6H:b-wid7wv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L103)):*

There can only be one result, either draft or primary line. But I can't find how to "FirstOrDefault()" in GraphQL.

Ah, but I can move the [0] in the grapqhl file though. Not sure if this is how it should be done..?
```
export const mapActiveLineDetailsByDateResult = (
  result: ReturnType<typeof useGetActiveLineDetailsWithRoutesByDateQuery>,
) => result.data?.route_line[0] as RouteLine | undefined;
```
</blockquote></details>

Well, that kind of just moves the problem to another place. Maybe better place (as the problem originates from graphql), but still...

I'm not 100% familiar with the situation here, so lets start from these questions:
- can the result have more than 1 items at some situation? If yes, is that error or can we just dismiss other items? If you're unsure, should we then console.log/warn something? If we already know that is an error, should we `throw` something then?
- Can the result have only 0 items? If that happens, we are implicitly throwing error. Is that what we want?

I'm not the expert here, I'm asking silly questions here because if we have logical error here, it is really difficult to spot afterwards. :)
___
*[src/components/routes-and-lines/lineDetails/TopActionsRow.tsx, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WV0La9VuIMPaPnn-S:-N05EDlT-hnuNruECUcV:b5cevr5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/components/routes-and-lines/lineDetails/TopActionsRow.tsx#L20)):*

Thanks, I'll made changes to use qs and immer. Not sure if this solution is what you were thinking, but now it won't overwrite existing queryString, only the parts related to this.
</blockquote></details>

Might be a bit easier to use if produce part wasn't nested to `qs.stringify()` (even though we have to definie additional variable then:
```

const onDateChange = (date: DateTime) => {
  const updatedUrlQuery = produce(queryParams, (draft) => {
    if (date.isValid) {
      draft.selectedDate = date.toISODate();
      draft.showAll = undefined;
    } else {
      draft.showAll = true.toString();
      draft.selectedDate = undefined;
    }
  });
  const queryString = qs.stringify(updatedUrlQuery);
...

```
I'd also double check if setting values to undefined actually removes those from query object, or do you have to remove those like this:
`delete draft.showAll`

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @eleksi, @pallost, and @tk-vincit)

___
*[src/graphql/route.ts, line 218 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WcYbb-nHZGTnf5ego:-N0FwPa0Dlr1MbYAXvgG:b64z8nn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/graphql/route.ts#L218)):*

I'm a bit worried about having these here, as we already have very similar date range queries in `src/hooks/useCheckValidityAndPriorityConflicts.ts`. Maintaining these will be nightmare if we realise that we have a bug in some date range query and then have to find out all places where same bug might be. 

Let's discuss about this issue in Slack.
</blockquote></details>

Yes, lets!
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 64 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WXdL76oPJAiN4ztuQ:-N0Fr-2_8qFfMTQ3TWBI:b-op3wj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L64)):*

Ok. Sounds very weird, but as it is at least commented now then I think this is OK for now.
</blockquote></details>

Yeah, maybe we can check this out at some day at the office with two or more pairs of eyes.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 102 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-W_p8u8vJ935A_jsex:-N0FtiWVBB0HfjUpLz8C:bspugwz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L102)):*

Seems like this wasn't changed, but is the reason that lineByDate is not array anymore or what..? Its a bit difficult to parse all the changes in my head :)
</blockquote></details>

Yes you are correct. I think I made those changes after acknowledging this and then forgot to comment here. So it is no longer array and will just be undefined if the lineByDate has no result.
___
*[src/hooks/lineDetails/useGetLineDetails.ts, line 103 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WZuASCXK264SST8Y1:-N0FrFbI43lE1IZjYHe2:bbcsojg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/hooks/lineDetails/useGetLineDetails.ts#L103)):*

Well, that kind of just moves the problem to another place. Maybe better place (as the problem originates from graphql), but still...

I'm not 100% familiar with the situation here, so lets start from these questions:
- can the result have more than 1 items at some situation? If yes, is that error or can we just dismiss other items? If you're unsure, should we then console.log/warn something? If we already know that is an error, should we `throw` something then?
- Can the result have only 0 items? If that happens, we are implicitly throwing error. Is that what we want?

I'm not the expert here, I'm asking silly questions here because if we have logical error here, it is really difficult to spot afterwards. :)
</blockquote></details>

Very good points! Yep, I do know that the problem is in another place now, but this same is at least used in
```
export const mapRouteDetailsResult = (
  result: ReturnType<typeof useGetRouteDetailsByIdsQuery>,
) => result.data?.route_route[0] as RouteRoute | undefined;
```

But to answer the other questions:
* This can't ever have more than one because it has this `limit: 1` in the query. I do not know why the result is still provided in an array though. 
* Yes the result can have 0 items if the date is chosen so that there is no line active with the given label

Neither am I so this kind of comments are very welcome. I would assume that there is some other way of having the result as not an array, since there only can be 0 or 1 results (because of the limit), but haven't figured out how (other than the result[0]).
___
*[src/components/routes-and-lines/lineDetails/TopActionsRow.tsx, line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N-WV0La9VuIMPaPnn-S:-N0FqSFY9fnR_09ZvQ39:b-mmskv2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d22c170a0b58d11e55b7e42953fd9f3b305a1feb/src/components/routes-and-lines/lineDetails/TopActionsRow.tsx#L20)):*

Might be a bit easier to use if produce part wasn't nested to `qs.stringify()` (even though we have to definie additional variable then:
```

const onDateChange = (date: DateTime) => {
  const updatedUrlQuery = produce(queryParams, (draft) => {
    if (date.isValid) {
      draft.selectedDate = date.toISODate();
      draft.showAll = undefined;
    } else {
      draft.showAll = true.toString();
      draft.selectedDate = undefined;
    }
  });
  const queryString = qs.stringify(updatedUrlQuery);
...

```
I'd also double check if setting values to undefined actually removes those from query object, or do you have to remove those like this:
`delete draft.showAll`


</blockquote></details>

Good ones! Your way is more readable and I do like `delete` a lot better than assigning undefined.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @eleksi, @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`src/graphql/route.ts` line 242 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N0_k1Do4v7FbmuDZxsr:-N0_k1Do4v7FbmuDZxss:b-jciub) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de0a2657d1e66cab92bbb99f217ecf013c5e9c92/src/graphql/route.ts#L242)):*
> ```XML
> export const mapActiveLineDetailsByDateResult = (
>   result: ReturnType<typeof useGetActiveLineDetailsWithRoutesByDateQuery>,
> ) => result.data?.route_line[0] as RouteLine | undefined;
> ```

Mentioned this also in previous thread, but if `result.data.route_line.lenght === 0`, we are implicitly throwing error here - is that something we want to do or should that be handled somehow?  E.g. `result.data?.route_line.lenght ? result.data?.route_line[0] as RouteLine : undefined`

Could also consider commething why we are taking just first item: `// gql query should have limit: 1 so we should never have more than 1 items` 
or then if-block:
```
if (result.data?.route_line.lenght > 1) {
throw new Error('Expected to receive only 1 item')
}
```  
(error message could maybe be improved, that's just an example)
```

___
*[`src/utils/gql.ts` line 7 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N0_jHl52CU41tAAkgqk:-N0_jHl52CU41tAAkgql:brmzxsi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de0a2657d1e66cab92bbb99f217ecf013c5e9c92/src/utils/gql.ts#L7)):*
> ```XML
>  * results which are not active on the given date
>  */
> export const constructActiveDateFilterGql = (date?: DateTime | null) => ({
> ```

` ...dateFilterQql` -> `...dateQqlFilter`? (same for the rest)
___
*[`src/components/routes-and-lines/line-details/ActionsRow.tsx` line 37 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N0_iQffCb6F0Tlp0YsD:-N0_iQffCb6F0Tlp0YsE:bibbsfu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de0a2657d1e66cab92bbb99f217ecf013c5e9c92/src/components/routes-and-lines/line-details/ActionsRow.tsx#L37)):*
> ```XML
> 
>   return (
>     <Container className={`pt-4 pb-0 ${className}`}>
> ```

Seems like padding is used for adjusting layout of this component here, so it should be passed from outside.
pb-0 is default, so it's not needed at all.
`${className}`  -> `${className || ''}` or set default value for className.


**Review by HenrikHartiala — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @eleksi, @pallost, and @tk-vincit)

___
*[`src/graphql/route.ts` line 242 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N0_k1Do4v7FbmuDZxsr:-N0aAKjq-oYeINBqBnQ-:b-1xuawn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de0a2657d1e66cab92bbb99f217ecf013c5e9c92/src/graphql/route.ts#L242)):*

Mentioned this also in previous thread, but if `result.data.route_line.lenght === 0`, we are implicitly throwing error here - is that something we want to do or should that be handled somehow?  E.g. `result.data?.route_line.lenght ? result.data?.route_line[0] as RouteLine : undefined`

Could also consider commething why we are taking just first item: `// gql query should have limit: 1 so we should never have more than 1 items` 
or then if-block:
```
if (result.data?.route_line.lenght > 1) {
throw new Error('Expected to receive only 1 item')
}
```  
(error message could maybe be improved, that's just an example)
```

</blockquote></details>

Ah right, now I understood this a bit better. I'll add the length check and then add comment about the `limit: 1`. Not sure if the error is needed since it should not be possible to happen because of the limit. 

Although, I changed this query to have the filters set in code, so this query should now be named differently.
___
*[`src/utils/gql.ts` line 7 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N0_jHl52CU41tAAkgqk:-N0de90c6rAbwHVBWZiQ:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de0a2657d1e66cab92bbb99f217ecf013c5e9c92/src/utils/gql.ts#L7)):*

` ...dateFilterQql` -> `...dateQqlFilter`? (same for the rest)
</blockquote></details>

:+1: 
___
*[`src/components/routes-and-lines/line-details/ActionsRow.tsx` line 37 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N0_iQffCb6F0Tlp0YsD:-N0a9Cmx4FaX2Kuho8yt:b-q82zhp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de0a2657d1e66cab92bbb99f217ecf013c5e9c92/src/components/routes-and-lines/line-details/ActionsRow.tsx#L37)):*

Seems like padding is used for adjusting layout of this component here, so it should be passed from outside.
pb-0 is default, so it's not needed at all.
`${className}`  -> `${className || ''}` or set default value for className.
</blockquote></details>

Thanks, will fix this. Although the <Container> does have `py-20`, so the pb-0 is needed.


HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/147)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @eleksi, @pallost, and @tk-vincit)

___
*[`src/components/routes-and-lines/line-details/ActionsRow.tsx` line 37 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/147#-N0_iQffCb6F0Tlp0YsD:-N0dx4CI2BqDxjGH5LkA:bbwzgf3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de0a2657d1e66cab92bbb99f217ecf013c5e9c92/src/components/routes-and-lines/line-details/ActionsRow.tsx#L37)):*

Tip: if `pb-0` is used for overwriting styles of parent component, it is safer to do it with !important definition: `!pb-0`.
</blockquote></details>

Right, okay. I'll add it as well. (Need to rerun the gql generates anways still)

PR #150, opened 2022-04-19
Use redux to handle map editor state

eleksi:
- Replace MapEditorContext with redux store

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150)


tk-vincit (changes requested):
Nice to have this refactoring done.

Sorry about many commets that might expand scope of this PR, but I think the whole map editor state has started to grow maybe too much and at least it is pretty difficult to understand already so this might be good time to improve things, at least a little bit.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-:-N04dzqpAzNP0u5aQI3M:b-66ltt8)* status: 0 of 27 files reviewed, 12 unresolved discussions (waiting on @eleksi and @tk-vincit)

___
*[src/components/forms/RoutePropertiesForm.tsx, line 50 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04fqhk7RSWVHlXADH_:-N04fqhk7RSWVHlXADHa:b4ue6m9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/forms/RoutePropertiesForm.tsx#L50)):*
> ```XML
>   const setTemplateRoute = (uuid?: UUID) => {
>     dispatch(
>       setStateAction({
> ```

Shold we rather have `setTemplateRouteId` action for this?
___
*[src/components/map/Map.tsx, line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04UBNt0Iqcpcd42JYG:-N04UBNt0Iqcpcd42JYH:bdfwd2o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/Map.tsx#L17)):*
> ```XML
> import { DynamicInfraLinksVectorLayer, InfraLinksVectorLayer } from './network';
> import { ObservationDateOverlay } from './ObservationDateOverlay';
> import { DrawRouteLayer, RouteLayer, Routes } from './routes/index';
> ```

`/index` not needed unless this somehow avoids some kind of dependency cycle
___
*[src/components/map/ModalMap.tsx, line 99 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04VRKm7pSznZ7zFri4:-N04VRKm7pSznZ7zFri5:b3hkhen) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/ModalMap.tsx#L99)):*
> ```XML
>     mapRef?.current?.onDeleteDrawnRoute();
>     dispatch(
>       drawingMode === Mode.Draw
> ```

We are reading `drawingMode` from redux's state to this component only to be able to adjust redux's state according to its value. That leaks things that redux could handle by itself to this component. How about if we just dispatch e.g. `toggleDrawRouteAction()` from here and do this kind of `if else` things on redux slice?

Same for `onEditRoute`.
___
*[src/components/map/ModalMap.tsx, line 117 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Wubf9HC6tcX0wFok:-N04Wubf9HC6tcX0wFol:bm3nw27) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/ModalMap.tsx#L117)):*
> ```XML
>     mapRef?.current?.onDeleteDrawnRoute();
> 
>     dispatch(stopDrawRouteAction());
> ```

Regarding my previous comment: ok, seems like `stopDrawRouteAction` is used also individually. I'm just wondering if it could be replaced with "reset map editor state" action or does it have to be its own action? Of course suggested `toggleDrawRouteAction` could also take boolean parameter, but that makes it more difficult to understand and kind of defeats the purpose.
___
*[src/components/map/RouteStopsOverlay.tsx, line 33 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04g2bM4LWPyVI04Fyf:-N04g2bM4LWPyVI04Fyg:bmru0i3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/RouteStopsOverlay.tsx#L33)):*
> ```XML
>   const setOnRoute = (belongsToRoute: boolean) => {
>     dispatch(
>       setStateAction({
> ```

Should we rather have `setDraftRouteStops` (or something like that) actio for this?
___
*[src/components/map/routes/Routes.tsx, line 45 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04gJOr35JpAP3U37D_:-N04gJOr35JpAP3U37Da:b-4fqo34) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/routes/Routes.tsx#L45)):*
> ```XML
>           templateRouteId: editedRouteData.templateRouteId,
>         },
>         drawingMode: editedRouteData.templateRouteId ? Mode.Edit : Mode.Draw,
> ```

This looks like something that redux could figure out by itself
___
*[src/hooks/useShowRoutesOnModal.ts, line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04ayve2Smb5FEtJEi2:-N04ayve2Smb5FEtJEi3:btvs72) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/hooks/useShowRoutesOnModal.ts#L17)):*
> ```XML
>   const showRoutesOnModalById = (routeIds: UUID[]) => {
>     dispatch(resetAction());
>     dispatch(
> ```

Wondering if we should have only single action for this instead of first resetting state in first action and then setting initiallyDisplayedRouteId's in second action.
Unluckily seems like we are setting initiallyDisplayedRouteId's also in other place but without clearing state there, but not sure if these could be unified anyway.
___
*[src/redux/slices/mapEditor.ts, line 6 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04a3-L0kYxgOF1XyBz:-N04a3-L0kYxgOF1XyC-:byf9lsx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L6)):*
> ```XML
> 
> interface IState {
>   hasRoute: boolean;
> ```

I think this should nowadays be called something like "hasDraftRoute" (hasUnsavedRoute?) or something like that. Now its difficult to understand meaning of this.
___
*[src/redux/slices/mapEditor.ts, line 56 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Op3DBTxEsZI564Ql:-N04Op3E9fHy-wVsJwER:b-2gptpc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L56)):*
> ```XML
>     startDrawRoute: (state) => {
>       state.drawingMode = Mode.Draw;
>       state.creatingNewRoute = true;
> ```

I don't quite understand meaning of this variable. Seems like this is set to `true` if `drawingMode === Mode.Draw` and `false` otherwise. So, should this actually be selector or am I missing something?
___
*[src/redux/slices/mapEditor.ts, line 60 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04cpLEDoRvUswTK-GI:-N04cpLEDoRvUswTK-GJ:bqtmjaq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L60)):*
> ```XML
>     },
>     stopDrawRoute: (state) => {
>       state.drawingMode = undefined;
> ```

If intention is to reset these to the initial state, it would probably be better to set these to `= initialState.drawingMode` etc. as has been done two rows below. 
___
*[src/redux/slices/mapEditor.ts, line 75 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Z6koENQ1vU_cjooj:-N04Z6koENQ1vU_cjook:b6nzh4h) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L75)):*
> ```XML
>       };
>     },
>     stopEditRoute: (state) => {
> ```

`startEditRoute` action changes three different things and this just sets drawingMode to undefined. Is that correct? At least this is confusing.
Is this action even needed at all or could resetAction used instead?
___
*[src/redux/slices/mapEditor.ts, line 82 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Q3r1-k899QK2rPOf:-N04Q3r1-k899QK2rPOg:bl12wy6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L82)):*
> ```XML
> 
> export const {
>   reset: resetAction,
> ```

I have discussed about this kind of naming of actions before and the decision was that names like this are too generic. This should be rather called something like `resetMapEditorStateAction`. Same for the `setState` action.

eleksi (commented):
I added a new commit addressing the refactoring itself. There were a lot of changes to do and it made this PR bigger, but I hope it can still be handled.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-:-N0AbFLh7OVK7oizU54X:bb1td18)* status: 0 of 46 files reviewed, 9 unresolved discussions (waiting on @tk-vincit)

___
*[src/components/forms/RoutePropertiesForm.tsx, line 50 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04fqhk7RSWVHlXADH_:-N04pEPiCN-wTmmRbRya:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/forms/RoutePropertiesForm.tsx#L50)):*

Shold we rather have `setTemplateRouteId` action for this?
</blockquote></details>

Done.
___
*[src/components/map/Map.tsx, line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04UBNt0Iqcpcd42JYG:-N04pQzO4kehn9JDRYre:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/Map.tsx#L17)):*

`/index` not needed unless this somehow avoids some kind of dependency cycle
</blockquote></details>

Done.
___
*[src/components/map/ModalMap.tsx, line 99 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04VRKm7pSznZ7zFri4:-N05D_Ta-J_oqvJQgh9c:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/ModalMap.tsx#L99)):*

We are reading `drawingMode` from redux's state to this component only to be able to adjust redux's state according to its value. That leaks things that redux could handle by itself to this component. How about if we just dispatch e.g. `toggleDrawRouteAction()` from here and do this kind of `if else` things on redux slice?

Same for `onEditRoute`.
</blockquote></details>

Done.
___
*[src/components/map/ModalMap.tsx, line 117 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Wubf9HC6tcX0wFok:-N04zrbK8COH4fplPMom:b-8n0zie) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/ModalMap.tsx#L117)):*

Regarding my previous comment: ok, seems like `stopDrawRouteAction` is used also individually. I'm just wondering if it could be replaced with "reset map editor state" action or does it have to be its own action? Of course suggested `toggleDrawRouteAction` could also take boolean parameter, but that makes it more difficult to understand and kind of defeats the purpose.
</blockquote></details>

We do not want to reset the entire state. For example, if we have selected to see route 65x, then click "Draw route" then cancel, route 65x would disappear, since displayed routes are also kept in map editor state. We can question whether it is the right place for those to be stored, but for now we cannot reset the state here.
___
*[src/components/map/RouteStopsOverlay.tsx, line 33 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04g2bM4LWPyVI04Fyf:-N054UDE5P8praN5Q1kA:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/RouteStopsOverlay.tsx#L33)):*

Should we rather have `setDraftRouteStops` (or something like that) actio for this?
</blockquote></details>

Done.
___
*[src/components/map/routes/Routes.tsx, line 45 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04gJOr35JpAP3U37D_:-N055uvBEEPWaanXxIaW:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/routes/Routes.tsx#L45)):*

This looks like something that redux could figure out by itself
</blockquote></details>

Done.
___
*[src/redux/slices/mapEditor.ts, line 6 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04a3-L0kYxgOF1XyBz:-N05KhhIA9jIcD3FIXnX:b-vlrtvh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L6)):*

I think this should nowadays be called something like "hasDraftRoute" (hasUnsavedRoute?) or something like that. Now its difficult to understand meaning of this.
</blockquote></details>

Done. This is now a selector instead of state field.
___
*[src/redux/slices/mapEditor.ts, line 56 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Op3DBTxEsZI564Ql:-N05BCV_7sjKw0m3Gxpk:b-kopwgq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L56)):*

I don't quite understand meaning of this variable. Seems like this is set to `true` if `drawingMode === Mode.Draw` and `false` otherwise. So, should this actually be selector or am I missing something?
</blockquote></details>

We could still have route creation in progress even if mode is edit, as we can edit the drawn route geometry before saving after drawing it.
___
*[src/redux/slices/mapEditor.ts, line 75 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Z6koENQ1vU_cjooj:-N05CFZ65sSSc0dH9caE:bqyrc9p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L75)):*

`startEditRoute` action changes three different things and this just sets drawingMode to undefined. Is that correct? At least this is confusing.
Is this action even needed at all or could resetAction used instead?
</blockquote></details>

Well, it kind of depends on how we want this to work. If we want the edit mode to toggled on and off, we cannot reset those other fields. If we want the only way to exit edit mode to be saving changes or discarding them, that is another story. I guess these things have to be checked with design, what we want the buttons to do etc. For now I will not touch this.
___
*[src/redux/slices/mapEditor.ts, line 82 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Q3r1-k899QK2rPOf:-N05BgcK3Y66bRsRqpMv:b-1kvdlz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L82)):*

I have discussed about this kind of naming of actions before and the decision was that names like this are too generic. This should be rather called something like `resetMapEditorStateAction`. Same for the `setState` action.
</blockquote></details>

Sure, it is too generic indeed. Fixed

tk-vincit (changes requested):
Will have to continue the review when I have a bit more time, but starting to look good!

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-:-N0ArIV44UXqD8It101J:bde6q0g)* status: 0 of 30 files reviewed, 6 unresolved discussions (waiting on @eleksi and @tk-vincit)

___
*[src/redux/slices/mapEditor.ts, line 165 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0AqybtBNx2_ZbUnPKq:-N0AqybtBNx2_ZbUnPKr:b-9vflg4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6780905ed1f8187878bd9b9062314c22dbeedb56/src/redux/slices/mapEditor.ts#L165)):*
> ```XML
> export const {
>   reset: resetMapEditorStateAction,
>   startDrawRoute: startDrawRouteAction,
> ```

Do we still want to export these as we have now those toggle* actions? (Ok, seems like at least stopDrawRoute is used, how about the rest?



_Code quote:_
```XML
  startDrawRoute: startDrawRouteAction,
  stopDrawRoute: stopDrawRouteAction,
  startEditRoute: startEditRouteAction,
  stopEditRoute: stopEditRouteAction,
```

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150)* status: 0 of 30 files reviewed, 5 unresolved discussions (waiting on @eleksi and @tk-vincit)

___
*[`src/components/map/ModalMap.tsx` line 117 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Wubf9HC6tcX0wFok:-N0UuZK5DHn-84ok0iTA:bf4q5e9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/ModalMap.tsx#L117)):*

We do not want to reset the entire state. For example, if we have selected to see route 65x, then click "Draw route" then cancel, route 65x would disappear, since displayed routes are also kept in map editor state. We can question whether it is the right place for those to be stored, but for now we cannot reset the state here.
</blockquote></details>

Ok, maybe this is good enough for now
___
*[`src/components/map/RouteStopsOverlay.tsx` line 33 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04g2bM4LWPyVI04Fyf:-N0UujzV3c7q5t38kAGj:bb0qia8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/components/map/RouteStopsOverlay.tsx#L33)):*

Done.
</blockquote></details>

Much cleaner interface now! :+1: 
___
*[`src/redux/slices/mapEditor.ts` line 56 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Op3DBTxEsZI564Ql:-N0VHe7L-9bIVFb4l3Xy:b32w1t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L56)):*

We could still have route creation in progress even if mode is edit, as we can edit the drawn route geometry before saving after drawing it.
</blockquote></details>

Ok
___
*[`src/redux/slices/mapEditor.ts` line 75 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N04Z6koENQ1vU_cjooj:-N0VHcJM0q2lIwUThEQ8:b32w1t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bac040c22c5b3bbd708851d784f978c1494ecf53/src/redux/slices/mapEditor.ts#L75)):*

Well, it kind of depends on how we want this to work. If we want the edit mode to toggled on and off, we cannot reset those other fields. If we want the only way to exit edit mode to be saving changes or discarding them, that is another story. I guess these things have to be checked with design, what we want the buttons to do etc. For now I will not touch this.
</blockquote></details>

Ok
___
*[`src/redux/slices/mapEditor.ts` line 90 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0VG7nU0iMMKWBxiOfe:-N0VG7nU0iMMKWBxiOff:bpyevdr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/283c953a4059af03e49258a62e66e4077e979628/src/redux/slices/mapEditor.ts#L90)):*
> ```XML
>       };
>     },
>     populateNewRouteForm: (
> ```

How does this action actually populate anything? Any reasons why this shouldn't be called something like `setEditerRouteMetaData`? (Which also sounds weird as we are setting form state to "metaData" field. This is confusing...)
___
*[`src/redux/slices/mapEditor.ts` line 98 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0VGmFq9zerLNQch000:-N0VGmFq9zerLNQch001:b97zoy6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/283c953a4059af03e49258a62e66e4077e979628/src/redux/slices/mapEditor.ts#L98)):*
> ```XML
>     saveNewRouteForm: (
>       state,
>       action: PayloadAction<Partial<RouteFormState>>,
> ```

`RouteFormState` seems to be saved twice, in this action and in `populateNewRouteForm`. Why whe have two actions for saving same data?
___
*[`src/redux/slices/mapEditor.ts` line 147 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0UwE2v-EoPifej0ik5:-N0UwE2v-EoPifej0ik6:byjh6nb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/283c953a4059af03e49258a62e66e4077e979628/src/redux/slices/mapEditor.ts#L147)):*
> ```XML
>   void,
>   void,
>   { state: RootState }
> ```

Slices should be independent from other slices (as authot of redux-toolit has stated himself: https://stackoverflow.com/a/62452227 ), so they shouldn't have RootState dependency.  Ok, seems to be only for typings so its not that bad, but still something we should avoid if possible.

Does this stackoverflow question help or do we have to go with this? https://stackoverflow.com/questions/65277731/redux-toolkit-accessing-state-from-thunk

Other alternative would be extracting these out of this slice to e.g. `redux/thunks/mapEditor.ts` file (to be created)

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150)* status: 0 of 34 files reviewed, 5 unresolved discussions (waiting on @tk-vincit)

___
*[`src/redux/slices/mapEditor.ts` line 165 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0AqybtBNx2_ZbUnPKq:-N0VK0kv4KLvbLTIGRFi:bf68zst) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/6780905ed1f8187878bd9b9062314c22dbeedb56/src/redux/slices/mapEditor.ts#L165)):*

Do we still want to export these as we have now those toggle* actions? (Ok, seems like at least stopDrawRoute is used, how about the rest?
</blockquote></details>

Oh, I forgot to answer. Initially I removed unnecessary exported actions, but after moving thunks to a different file, they all are needed outside this file.
___
*[`src/redux/slices/mapEditor.ts` line 90 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0VG7nU0iMMKWBxiOfe:-N0VJUZN-G93goeFL3wS:b-uz3gik) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/283c953a4059af03e49258a62e66e4077e979628/src/redux/slices/mapEditor.ts#L90)):*

How does this action actually populate anything? Any reasons why this shouldn't be called something like `setEditerRouteMetaData`? (Which also sounds weird as we are setting form state to "metaData" field. This is confusing...)
</blockquote></details>

This is used when we for example click + icon on line page (to create a new route to this line) and want the line field to be populated when we open route creation form. This is different from saving the form, as we do not want to do any of the drawing mode changes etc here.
___
*[`src/redux/slices/mapEditor.ts` line 98 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0VGmFq9zerLNQch000:-N0VJwqL9ULzM3q3mYY1:bv1p8l5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/283c953a4059af03e49258a62e66e4077e979628/src/redux/slices/mapEditor.ts#L98)):*

`RouteFormState` seems to be saved twice, in this action and in `populateNewRouteForm`. Why whe have two actions for saving same data?
</blockquote></details>

I explained this in previous comment.
___
*[`src/redux/slices/mapEditor.ts` line 147 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0UwE2v-EoPifej0ik5:-N0VKR7kAMCs0qJJVhhj:b-x0vtpy) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/283c953a4059af03e49258a62e66e4077e979628/src/redux/slices/mapEditor.ts#L147)):*

Slices should be independent from other slices (as authot of redux-toolit has stated himself: https://stackoverflow.com/a/62452227 ), so they shouldn't have RootState dependency.  Ok, seems to be only for typings so its not that bad, but still something we should avoid if possible.

Does this stackoverflow question help or do we have to go with this? https://stackoverflow.com/questions/65277731/redux-toolkit-accessing-state-from-thunk

Other alternative would be extracting these out of this slice to e.g. `redux/thunks/mapEditor.ts` file (to be created)
</blockquote></details>

The RootState is here just for typing. https://stackoverflow.com/a/67284522

I can move thunks outside this file to avoid dependency cycle, though.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150)* status: 0 of 34 files reviewed, 4 unresolved discussions (waiting on @eleksi and @tk-vincit)

___
*[`src/redux/slices/mapEditor.ts` line 108 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0ZbJeO1CVvJ65jKtWs:-N0ZbJeO1CVvJ65jKtWt:bchmg7q) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4d577d88e21ab16c98816d75f8901e924570c6e9/src/redux/slices/mapEditor.ts#L108)):*
> ```XML
>         : Mode.Draw;
>     },
>     setInitiallyDisplayedRouteIds: (state, action: PayloadAction<UUID[]>) => {
> ```

Now this action says that it just sets initiallyDisplayedRouteIds, but as a side effect it seems to other fields in the state to the initial values. This either needs better name, or if this is logical error, return block could be replaced with `state.initiallyDisplayedRouteIds = action.payload`

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150)* status: 0 of 34 files reviewed, 4 unresolved discussions (waiting on @tk-vincit)

___
*[`src/redux/slices/mapEditor.ts` line 90 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0VG7nU0iMMKWBxiOfe:-N0ZgrJD3tqOQ8VTv37u:b-y561et) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/283c953a4059af03e49258a62e66e4077e979628/src/redux/slices/mapEditor.ts#L90)):*

This is used when we for example click + icon on line page (to create a new route to this line) and want the line field to be populated when we open route creation form. This is different from saving the form, as we do not want to do any of the drawing mode changes etc here.
</blockquote></details>

setRouteMetadata
___
*[`src/redux/slices/mapEditor.ts` line 108 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/150#-N0ZbJeO1CVvJ65jKtWs:-N0_8U0OEBrxyEzlmD0B:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4d577d88e21ab16c98816d75f8901e924570c6e9/src/redux/slices/mapEditor.ts#L108)):*

Now this action says that it just sets initiallyDisplayedRouteIds, but as a side effect it seems to other fields in the state to the initial values. This either needs better name, or if this is logical error, return block could be replaced with `state.initiallyDisplayedRouteIds = action.payload`
</blockquote></details>

Done.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/150)* status: 0 of 34 files reviewed, all discussions resolved (waiting on @tk-vincit)

PR #158, opened 2022-04-26
Move route save logic out of ModalMap component

eleksi:


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/158)* status: 0 of 10 files reviewed, 8 unresolved discussions (waiting on @eleksi and @tk-vincit)

___
*[`src/components/map/ModalMap.tsx` line 59 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0dxhnkBb3Xi2w6F9WL:-N0dxhnkBb3Xi2w6F9WM:brml3i6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d036d1383ab846cd0ae2f9742653dd4bfbdce8ff/src/components/map/ModalMap.tsx#L59)):*
> ```XML
>   const {
>     saveRoute,
>     defaultErrorHandler: defaultSaveErrorHandler,
> ```

defaultSaveRouteErrorHandler :wink: Anyway, as we should refactor this out here soon anyway this is OK for now.
___
*[`src/components/map/ModalMap.tsx` line 102 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0eAljX5BS66kpXXa9v:-N0eAljX5BS66kpXXa9w:bogwf5m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d036d1383ab846cd0ae2f9742653dd4bfbdce8ff/src/components/map/ModalMap.tsx#L102)):*
> ```XML
>     try {
>       // Save variable value since saveRoute resets the value in state
>       const newRouteCreated = creatingNewRoute;
> ```

Wtf? This sounds very hacky. Also this won't work, here we are saving just a reference to the object so if original object is modified so is this.

I think we rather want to receive `newRoute` as a return value from `saveRoute` and use it here if needed. Altough seems like we don't need this at all, if `saveRoute` doesn't throw then we are showing success toast anyway, so I guess its safe to delete drawn route also then.



_Suggestion:_
```XML
      // Save variable value since saveRoute resets the value in s
      const newRouteCreated = creatingNewRoute;
```
___
*[`src/hooks/routes/useEditRouteGeometry.ts` line 26 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0e9D7a5Q0abSVJhNxC:-N0e9D7a5Q0abSVJhNxD:b-2quuzc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f5ff8cbe47cba4a6ca4b0a05bb06fe79ca5860b9/src/hooks/routes/useEditRouteGeometry.ts#L26)):*
> ```XML
> 
> export const useEditRouteGeometry = () => {
>   const [updateRouteGeometryMutation] = useUpdateRouteGeometryMutation();
> ```

Should this be at the same file with mapping function that was just removed from here? And if yes, then this hook needs a new name.
___
*[`src/hooks/routes/useSaveRoute.ts` line 151 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0eDDa76DSOlaZ8BmWV:-N0eDDa76DSOlaZ8BmWW:b-moppi2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d036d1383ab846cd0ae2f9742653dd4bfbdce8ff/src/hooks/routes/useSaveRoute.ts#L151)):*
> ```XML
>   } = useContext(MapFilterContext);
> 
>   const prepareCreate = useCallback(
> ```

Wondering why we need useCallbacks here, as other similar hooks wont' need those and this wasn't needed previously.
___
*[`src/hooks/routes/useSaveRoute.ts` line 179 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0eBjkI8Ax5UwDigoYi:-N0eBjkI8Ax5UwDigoYj:b-ipg2tp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d036d1383ab846cd0ae2f9742653dd4bfbdce8ff/src/hooks/routes/useSaveRoute.ts#L179)):*
> ```XML
>   };
> 
>   const saveRoute = useCallback(async () => {
> ```

This function is long and difficult to read.
I think we should have separate functions 
a) creating new route
b) editing existing route

Then this function would call either one depending if we have `routeId` present.
(Also I think it should be called `routeId` and not `editingRouteId`, as we are not editing id of the route...)
___
*[`src/hooks/routes/useSaveRoute.ts` line 200 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0eC2H007skeTE5ztrZ:-N0eC2H007skeTE5ztr_:bcf2h3i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d036d1383ab846cd0ae2f9742653dd4bfbdce8ff/src/hooks/routes/useSaveRoute.ts#L200)):*
> ```XML
>       await updateRouteGeometryMutation(variables);
>     } else {
>       if (!routeDetails) {
> ```

This is a bit confusing, when this can happen? Maybe add comment here.
___
*[`src/hooks/routes/useSaveRoute.ts` line 221 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0eCsTf3yQ_EXeqUsED:-N0eCsTf3yQ_EXeqUsEE:b-w0jhlk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d036d1383ab846cd0ae2f9742653dd4bfbdce8ff/src/hooks/routes/useSaveRoute.ts#L221)):*
> ```XML
>       const newRoute = response.data!.insert_route_route_one!;
> 
>       dispatch(initializeMapEditorWithRoutesAction([newRoute.route_id]));
> ```

I'd rather separate these to own function. Adjusting map editor state based on graphql response isn't really related to this operation (saving route). You can add that function to this file and call it from here if you want, its step to the right direction anyway.



_Code quote:_
```XML
      dispatch(initializeMapEditorWithRoutesAction([newRoute.route_id]));

      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
      const validityStart = newRoute.validity_start!;

      if (
        !isDateInRange(observationDate, validityStart, newRoute?.validity_end)
      ) {
        mapFilterDispatch(setObservationDate(validityStart));
        showSuccessToast(t('filters.observationDateAdjusted'));
      }
```
___
*[`src/locales/en-US/common.json` line 29 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/158#-N0e42E92hwmHCX7ihRm:-N0e42E92hwmHCX7ihRn:b-r31ovv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d036d1383ab846cd0ae2f9742653dd4bfbdce8ff/src/locales/en-US/common.json#L29)):*
> ```JSON
>     "direction": "Direction",
>     "chooseDirection": "Choose direction",
>     "tooFewStopsOnRoute": "There must be at least two stops on route!"
> ```

I assume users would be happier if we wouldn't shout to them with "!". Same for finnish version.

PR #193, opened 2022-05-24
Small fixes and fine tuning

eleksi:
Multiple small fixes and fine tuning all around

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/193)

### DIFF
--- a/src/components/map/routes/RouteEditor.tsx
+++ b/src/components/map/routes/RouteEditor.tsx
@@ -185,11 +185,11 @@ const RouteEditorComponent = (
   };
 
   const onDeleteConfirm = async () => {
-    try {
-      if (!editedRouteId) {
-        return;
-      }
+    if (!editedRouteId) {
+      return;
+    }
 
+    try {
       // delete the route from the backend
       await deleteRoute(editedRouteId);
       showSuccessToast(t('routes.deleteSuccess'));
--- a/src/components/map/stops/EditStopModal.tsx
+++ b/src/components/map/stops/EditStopModal.tsx
@@ -30,6 +30,7 @@ export const EditStopModal = ({
   };
   return (
     <HTMLOverlay
+      style={{ zIndex: 10 }}
       redraw={() => (
         <div className="flex max-h-full py-5 pl-5">
           <Modal
--- a/src/generated/graphql.tsx
+++ b/src/generated/graphql.tsx
@@ -9259,7 +9259,6 @@ export const GetStopsByLocationDocument = gql`
     query GetStopsByLocation($measured_location_filter: geography_comparison_exp) {
   service_pattern_scheduled_stop_point(
     where: {measured_location: $measured_location_filter}
-    limit: 300
   ) {
     ...scheduled_stop_point_all_fields
   }
--- a/src/graphql/servicePattern.ts
+++ b/src/graphql/servicePattern.ts
@@ -78,7 +78,6 @@ const QUERY_GET_STOPS_BY_LOCATION = gql`
   ) {
     service_pattern_scheduled_stop_point(
       where: { measured_location: $measured_location_filter }
-      limit: 300
     ) {
       ...scheduled_stop_point_all_fields
     }


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/193)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @eleksi and @pallost)

___
*[`src/components/map/stops/EditStopModal.tsx` line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/193#-N2uIDH4B18a8gpV-xID:-N2uIDH4B18a8gpV-xIE:b-c19bm4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b738fa7283b49887a7fb3ce3f49dddfc1225e5ba/src/components/map/stops/EditStopModal.tsx#L33)):*
> ```XML
>   return (
>     <HTMLOverlay
>       style={{ zIndex: 10 }}
> ```

Instead of using z-index, wouldn't it be a possibility to just instantiate the layers in proper order?
Anyhow, even if using z-index, I'd prefer to see a list of these numbers in a separate file to know that which z-index numbers are used and in what order. E.g.
```
const zIndexes = {
  editStopModal: 10,
  routeEditor: 20,
  [etc. etc]
}
```

This would allow us to globally rearrange the z-indexes in case necessary without having to hunt them down from components scattered around

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/193)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[`src/components/map/stops/EditStopModal.tsx` line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/193#-N2uIDH4B18a8gpV-xID:-N32sx6lAuyEkMJK63Rg:b-c127ct) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b738fa7283b49887a7fb3ce3f49dddfc1225e5ba/src/components/map/stops/EditStopModal.tsx#L33)):*

Instead of using z-index, wouldn't it be a possibility to just instantiate the layers in proper order?
Anyhow, even if using z-index, I'd prefer to see a list of these numbers in a separate file to know that which z-index numbers are used and in what order. E.g.
```
const zIndexes = {
  editStopModal: 10,
  routeEditor: 20,
  [etc. etc]
}
```

This would allow us to globally rearrange the z-indexes in case necessary without having to hunt them down from components scattered around
</blockquote></details>

We have ordered the layers by instantiating them in proper order. However, in this case the problem is that we have one big Stops layer that contains both `Stop` markers and `EditStopLater` with `EditStopModal` inside this layer. Stop markers we want to show underneath everything else in the map, so it needs to be the bottommost layer. But we want to show the contents of `EditStopLayer` over many other layers in the map, therefore the z-index.

So, there is only 1 z-index we need to define right now. At some point we could refactor Stop markers and stop info edit form to be complete separate layers, so even the z-index wouldn't be necessary.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/193)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @pallost)

___
*[`src/components/map/stops/EditStopModal.tsx` line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/193#-N2uIDH4B18a8gpV-xID:-N3J9sTmAQYE0leQVoRJ:bytw31s) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9571d635af105e78b065682d5ffcb0c5cac6af8e/src/components/map/stops/EditStopModal.tsx#L33)):*

We have ordered the layers by instantiating them in proper order. However, in this case the problem is that we have one big Stops layer that contains both `Stop` markers and `EditStopLater` with `EditStopModal` inside this layer. Stop markers we want to show underneath everything else in the map, so it needs to be the bottommost layer. But we want to show the contents of `EditStopLayer` over many other layers in the map, therefore the z-index.

So, there is only 1 z-index we need to define right now. At some point we could refactor Stop markers and stop info edit form to be complete separate layers, so even the z-index wouldn't be necessary.
</blockquote></details>

OK, it would require a major refactoring to get the EditStopModal out of the Stops layer, so let's keep it like this for now

PR #200, opened 2022-05-25
Add hover styles to buttons

eleksi:
- Add hover effect to button: on hover display wider border.
- To make button not move on hover, add container element to button and adjust button margin on hover

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/200)

### DIFF
--- a/src/components/forms/line/LineForm.tsx
+++ b/src/components/forms/line/LineForm.tsx
@@ -66,7 +66,7 @@ export const LineForm = ({ defaultValues, onSubmit }: Props): JSX.Element => {
         </Row>
         <Row className="mt-8 space-x-5">
           <SimpleButton
-            className="ml-auto"
+            containerClassName="ml-auto"
             onClick={onCancel}
             inverted
             testId={testIds.cancelButton}
--- a/src/components/map/MapFooter.tsx
+++ b/src/components/map/MapFooter.tsx
@@ -71,7 +71,7 @@ export const MapFooter: React.FC<Props> = ({
         {t('map.addStop')}
       </SimpleButton>
       <SimpleButton
-        className="!px-3"
+        className="h-full !px-3"
         onClick={onDeleteRoute}
         disabled={!hasChangesInProgress}
       >
@@ -79,7 +79,7 @@ export const MapFooter: React.FC<Props> = ({
       </SimpleButton>
       <Visible visible={hasChangesInProgress}>
         <SimpleButton
-          className="!ml-auto"
+          containerClassName="!ml-auto"
           onClick={onCancel}
           disabled={!hasChangesInProgress}
           inverted
--- a/src/components/map/Modal.tsx
+++ b/src/components/map/Modal.tsx
@@ -37,7 +37,7 @@ const ModalFooter = ({ onCancel, onSave }: FooterProps): JSX.Element => {
   return (
     <HeaderFooterContainer>
       <Row className="space-x-4">
-        <SimpleButton className="ml-auto" onClick={onCancel} inverted>
+        <SimpleButton containerClassName="ml-auto" onClick={onCancel} inverted>
           {t('cancel')}
         </SimpleButton>
         <SimpleButton onClick={onSave}>{t('save')}</SimpleButton>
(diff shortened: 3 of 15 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/200)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @eleksi and @pallost)

___
*[`src/uiComponents/SimpleButton.tsx` line 46 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/200#-N2umjVWDUzqPo6O4-Dv:-N2umjVWDUzqPo6O4-Dw:bq0mq95) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d3328c23a97b620cd3159750104d2b6afe80e962/src/uiComponents/SimpleButton.tsx#L46)):*
> ```XML
>   if ((props as ButtonProps).onClick) {
>     return (
>       <span className={containerClassName}>
> ```

Umm, why did you need another wrapper? The <button/> itself is already a wrapper around some other component. Apparently you are also applying the hover style to this outer wrapper, which I think will result in some really funky looks in the end as the <span/> makes a rectangular wrapper around the <button/>, which in case might be round in some case

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/200)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[`src/uiComponents/SimpleButton.tsx` line 46 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/200#-N2umjVWDUzqPo6O4-Dv:-N33RYi1CgJasuhw3L7f:b-g4wiow) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d3328c23a97b620cd3159750104d2b6afe80e962/src/uiComponents/SimpleButton.tsx#L46)):*

Umm, why did you need another wrapper? The <button/> itself is already a wrapper around some other component. Apparently you are also applying the hover style to this outer wrapper, which I think will result in some really funky looks in the end as the <span/> makes a rectangular wrapper around the <button/>, which in case might be round in some case
</blockquote></details>

As described in the commit description, we need a wrapper, because hover style has wider border than normal style and to keep the button from moving on hover we need to play with the margin of the element inside. When we have 1px border, we add 1px margin to the inner element. When we have 2px border, we have 0px margin in the inner element. We do not want to play with the outer element's margin, because we don't want to interfere with margins set to this component from outside.

The outer element is not supposed to have any hover styles, only the one inside. The containerClassName is for margins etc. that are applied to this component from other components using it.

This change does make the component more complex than a simple button should be, but if we want to implement the hover styles design has given to this component, I don't know other way to do it.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/200)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @eleksi and @pallost)

___
*[`src/uiComponents/SimpleButton.tsx` line 46 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/200#-N2umjVWDUzqPo6O4-Dv:-N3JXMVE2EUUTPXT_Lop:b-yfev9n) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d3328c23a97b620cd3159750104d2b6afe80e962/src/uiComponents/SimpleButton.tsx#L46)):*

As described in the commit description, we need a wrapper, because hover style has wider border than normal style and to keep the button from moving on hover we need to play with the margin of the element inside. When we have 1px border, we add 1px margin to the inner element. When we have 2px border, we have 0px margin in the inner element. We do not want to play with the outer element's margin, because we don't want to interfere with margins set to this component from outside.

The outer element is not supposed to have any hover styles, only the one inside. The containerClassName is for margins etc. that are applied to this component from other components using it.

This change does make the component more complex than a simple button should be, but if we want to implement the hover styles design has given to this component, I don't know other way to do it.
</blockquote></details>

Yeah, it makes sense... There are other ways of doing so, but this is fine. I checked it and it also renders properly.

PR #203, opened 2022-05-30
Use label to refer stops from journey pattern

eleksi:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/203)


eleksi:
The broken test passes in this run:
https://github.com/HSLdevcom/jore4-robot/actions/runs/2446540373

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/203)* status: 0 of 32 files reviewed, 11 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`docker/docker-compose.custom.yml` line 6 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3rrJ7o3V6-flfG0O3s:-N3rrJ7o3V6-flfG0O3t:b-pcql2i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/docker/docker-compose.custom.yml#L6)):*
> ```YAML
>     # Link to available jore4-hasura images in Docker Hub:
>     # https://hub.docker.com/r/hsldevcom/jore4-hasura/tags?page=1&ordering=last_updated
>     image: 'hsldevcom/jore4-hasura:seed-drop-route-start-end-stop-points--20220602-65c2fab9b0058007f66b242dc6ff7556482c4564'
> ```

In the future, please do these hasura version updates + graphql schema generations in a separate commit
___
*[`src/components/forms/stop/StopForm.tsx` line 124 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3rrtV71NFpGcFj_PxE:-N3rrtV71NFpGcFj_PxF:b-qmfkki) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/components/forms/stop/StopForm.tsx#L124)):*
> ```XML
> 
>   const onCreate = async (state: FormState) => {
>     const changes = await prepareCreate({
> ```

well, if you move the try-catch away from here to onSubmit, please also remove is from onEdit, otherwise you'll show those error messages twice ;)
___
*[`src/components/map/RouteStopsOverlay.tsx` line 128 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3rt3oYC__8EYiBcLpc:-N3rt3oYC__8EYiBcLpd:b-ddc6k3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/components/map/RouteStopsOverlay.tsx#L128)):*
> ```XML
>       {routeStops?.map((routeStop) => (
>         <StopRow
>           key={routeStop.label}
> ```

not the scope of this PR, but this component could just take the whole ScheduledStopPoint instance instead of its individual attributes
___
*[`src/components/map/routes/DrawRouteLayer.tsx` line 156 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3ryj4Y3imTlkvED8pm:-N3ryj4Y3imTlkvED8pn:b-f3sj8u) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/components/map/routes/DrawRouteLayer.tsx#L156)):*
> ```XML
>       });
> 
>       const stopsResult = await getStopsByIds({
> ```

Umm, why not just fix extractScheduledStopPointIds to actually return you the labels instead of ids? :)
___
*[`src/components/map/stops/Stops.tsx` line 52 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s1k2U0w8s8bJDaHTt:-N3s1k2U0w8s8bJDaHTu:b-hki122) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/components/map/stops/Stops.tsx#L52)):*
> ```XML
> 
>   const unfilteredStops = mapGetStopsResult(stopsResult);
>   const stops = filter(unfilteredStops || []);
> ```

I just noticed this now... not part of this PR, but this stop filter should actually produce filters for the graphql query and not filter the already fetched stops! This will make the map view super slow if we have lots of stops to display. I'll create a ticket about this.
___
*[`src/graphql/route.ts` line 438 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s3CF1ESavQ0TFN-pD:-N3s3CF1ESavQ0TFN-pE:b-shmg2f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/graphql/route.ts#L438)):*
> ```XML
> 
>     insert_journey_pattern_journey_pattern_one(object: $new_journey_pattern) {
>       on_route_id
> ```

I assume this is removed from here as all the metadata updating happens through the PatchRoute mutation, right...?
___
*[`src/hooks/useExtractRouteFromFeature.ts` line 39 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s4Ez_6hO4bhJbktC4:-N3s4Ez_6hO4bhJbktC5:b-vl9uhe) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/useExtractRouteFromFeature.ts#L39)):*
> ```XML
>   return stopLabels.map((item) => ({
>     label: item,
>     belongsToRoute: !removedStopLabels?.includes(item),
> ```

this is a major issue to think about: currently you could choose to only show past shops from the map and draw a route using these "ghost stops". I'm not sure if there's any easy solution for this, but maybe at least create a ticket to the backlog to consider what would be a proper fix for this.
Maybe could filter the displayed stops on the map to only show the ones whose validity period intersects with the edited route's validity period, I don't know...
___
*[`src/hooks/routes/useEditRouteGeometry.ts` line 72 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s5MyY9cOuFkkJ20yx:-N3s5MyY9cOuFkkJ20yy:b-fe082b) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/routes/useEditRouteGeometry.ts#L72)):*
> ```XML
>     variables: UpdateRouteGeometryMutationVariables,
>   ) => {
>     await mutateFunction({
> ```

does this behave properly? So you edit the route geometry then remove the whole route metadata from cache. I guess to update the `route_shape` field...? I don't think apollo is smart enough to refetch the route's new metadata. Doesn't the UI explode after you edit the route metadata, doesn't e.g. the route geometry get hidden?
I think what you need here is actually just modify the mutation to return `...route_all_fields` everywhere instead of just `route_id`, thus the cache will update automatically.
___
*[`src/hooks/routes/useValidateRoute.ts` line 18 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s6KqG0hhDduuzLlDD:-N3s6KqG0hhDduuzLlDE:bc8hjky) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/daa199f0850d7cd0f2302f89bc295f19367a0667/src/hooks/routes/useValidateRoute.ts#L18)):*
> ```XML
>   mapDateInputToValidityStart,
> } from '../../utils';
> import { extractFirstAndLastStopFromStops } from './utils';
> ```

let's also delete the utility function itself
___
*[`src/hooks/stops/useCreateStop.ts` line 71 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s9-L28MJO7FgdYX4o:-N3s9-L28MJO7FgdYX4p:b-rurz62) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/stops/useCreateStop.ts#L71)):*
> ```XML
> 
>     // if added stop conflicts with existing routes, warn user.
>     // for example, if a stop with same label has already been added to a route,
> ```

I'm not really sure how this could happen... don't we have some constraint that prevents from having multiple stops with the same validity period having the same label? But maybe this is some different edge case. Anyway, good that it's checked
___
*[`src/hooks/stops/useMapStops.ts` line 67 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s0XKv1CZhp5QoFI_a:-N3s0XKw-LBdVFuS7tIX:b-hg9jtt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/stops/useMapStops.ts#L67)):*
> ```XML
> 
>   const getStopHighlighted = useCallback(
>     (id: UUID, label: string): boolean => {
> ```

I checked and with a relatively small amount of work, you could switch from selectedStopId to selectedStopLabel. It would allow you to get rid of this id parameter that you only use for this comparison

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/203)* status: 0 of 32 files reviewed, 13 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`src/hooks/stops/useEditStop.ts` line 73 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3sDqqwEQ_6CLKJhKD1:-N3sDqqwEQ_6CLKJhKD2:b-5h5z2u) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/stops/useEditStop.ts#L73)):*
> ```XML
> 
>     // if a stop is moved away from the route geometry, remove it from its journey patterns
>     const brokenRoutesResult = await getBrokenRoutes({
> ```

come on, I had refactored this function to look so nice...! Please move all these "brokenRoutes" stuff at least to their own function! And try to keep the logic clean... here you throw lots of code inline instead of keeping them in logical functions with logical responsibilities!
___
*[`src/locales/en-US/common.json` line 99 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3sF7u3A7-ycFSlAj2Y:-N3sF7u3A7-ycFSlAj2Z:b-8zax85) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/locales/en-US/common.json#L99)):*
> ```JSON
>     "cannotEditTerminalStops": "Editing the start and end stops of a route is not allowed",
>     "addToRoute": "Add to route",
>     "stopBreaksRoutes": "The route to be added is not compatible with the following routes: ({{ routeLabels }})"
> ```

The *stop* to be added...?

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/203)* status: 0 of 33 files reviewed, 5 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`docker/docker-compose.custom.yml` line 6 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3rrJ7o3V6-flfG0O3s:-N41S-PU3frFPoAVgjtw:b-gunkje) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/docker/docker-compose.custom.yml#L6)):*

I would have done this, but in this case pretty much everything would break when hasura is updated. Therefore, only updating hasura version and schema would leave the repository in broken state. If it is ok that the repository is broken between those two commits, I could have separated them here as well.
</blockquote></details>

True. That's fine then!
___
*[`src/hooks/routes/useEditRouteGeometry.ts` line 72 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s5MyY9cOuFkkJ20yx:-N41UVWa-F2dMsgIMYlI:byjmd2p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/routes/useEditRouteGeometry.ts#L72)):*

It does behave properly. Route details are fetched after removing them from cache by other queries.

I cannot add `...route_all_fields` to the query, because there is no route data manipulated directly in the query. We only use `route_infrastructure_link_along_route` and `journey_pattern_journey_pattern` to manipulate route's data. 

We could also add explicit route data fetching to this function after the mutation or fetch them inside the mutation if it sounds better? I tried to add `keyFields` for `route_infrastructure_link_along_route` and `journey_pattern_journey_pattern` to cache policies, but still the update did not work..
</blockquote></details>

yeah, could try to add a refetch here after the mutation instead of emptying the cache. Let's not expect that other components/queries will take care of the refetching for you.
___
*[`src/hooks/routes/useValidateRoute.ts` line 18 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s6KqG0hhDduuzLlDD:-N41UvQB78kryBcO0cmw:bdstxiu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/daa199f0850d7cd0f2302f89bc295f19367a0667/src/hooks/routes/useValidateRoute.ts#L18)):*

Hmm it is already deleted. Could it be you had some old version?
</blockquote></details>

weird, I did see the function in my own checkout... well, good that it's deleted!
___
*[`src/hooks/stops/useEditStop.ts` line 73 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3sDqqwEQ_6CLKJhKD1:-N41VSR57tqEU_xasjCP:b-ucfknh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/stops/useEditStop.ts#L73)):*

Moved to separate function.
</blockquote></details>

Thanks!
___
*[`src/hooks/stops/useMapStops.ts` line 67 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/203#-N3s0XKv1CZhp5QoFI_a:-N41SoOX2zxFMtM3Dess:b-mzf1vp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5305a035d47fd9d4ef85eeca54ecd58a79e8ecde/src/hooks/stops/useMapStops.ts#L67)):*

Do we want to do that? We could have multiple instances of the same stop visible at the same time. We probably still want to be able to select and edit them one by one.
</blockquote></details>

That's true. Let's keep it as ID.
But this function should not take the label also as a parameter. I see you need the label otherwise because in your abstraction layer you only only have labels, in `stopLabelsOnEditedRoute`. Instead of directly using `mapRouteStopsToStopLabels` to fetch too specific data, you could just fetch only the RouteStops first and then the labels of these stops later where you need it

eleksi:
Tests pass in custom run:
https://github.com/HSLdevcom/jore4-robot/actions/runs/2466967361

PR #204, opened 2022-05-30
Coding style

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/204)

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,3 +187,72 @@ Maps in UI has four to five toggles depending on situation:
 - **1. c** show/hide route (Not available if no route is selected)
 - **2. a** show/hide stops served from jore4-mbtiles-server
 - **2. b** show/hide stops served from martin
+
+## Coding style
+
+### Inline components
+
+**What?** Avoid using inline components unless it's a must.
+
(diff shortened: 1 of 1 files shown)


eleksi (approved):
It is good to have these documented somewhere, but especially the import issue is something that happens more out of IDE auto import and inattention, rather than not knowing how it should be done. It would be great if some linter rule would warn about it, or even better the IDE would import the components correctly. But this definitely is better than nothing, and responsibility goes also to the developers to pay more attention to these things.

There were also some other concerns related to code style e.g. inline functions in JSX etc, you are probably making separate PR about those? At least the inline function issue could probably be taken care of with linter rule `react/no-bind` https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/204#-:-N3JixOs4HaXT8_0cyFC:b-cx8by)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pallost)

PR #206, opened 2022-06-01
Form validation

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/206)

The E2E tests fail because the selectors have changed for some form elements. The fix can be found from this PR: https://github.com/HSLdevcom/jore4-robot/pull/31

Closes https://github.com/HSLdevcom/jore4/issues/815


tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/206)* status: 0 of 20 files reviewed, 2 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`src/components/forms/common/ControlledElement.tsx` line 15 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/206#-N3Tf2CO0lVDvbMGMNYG:-N3Tf2CPElshGS7g4-am:b-ynixwv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8d9877a653c7deeef795259205372af88604d6e3/src/components/forms/common/ControlledElement.tsx#L15)):*
> ```XML
> interface Props<FormState extends FieldValues> {
>   className?: string;
>   id: string;
> ```

These seem to be repeating in multiple interfaces and thus could be extracted to own interface that could be used to with `extends` keyword in these cases.



_Code quote:_
```XML
  className?: string;
  id: string;
  fieldPath: Path<FormState>;
  testId: string;
```
___
*[`src/components/forms/common/InputField.tsx` line 28 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/206#-N3TdhGX2xlAcWpuiKhN:-N3TdhGX2xlAcWpuiKhO:buzvaw2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8d9877a653c7deeef795259205372af88604d6e3/src/components/forms/common/InputField.tsx#L28)):*
> ```XML
> interface HTMLInputProps {
>   inputElementRenderer?: never;
>   type: HTMLInputTypeAttribute | undefined;
> ```

Same as `type?: HTMLInputTypeAttribute`, right? Same for InputElement.

Began to wonder if this should actually be required. OK, seems like "text" is default valut for type prop (https://www.w3schools.com/tags/att_input_type.asp) , but in the other hand seems like it was explicitly given at least in `src/components/forms/route/TerminusNameInputs.tsx` anyway...
___
*[`src/components/forms/common/ValidationErrorList.tsx` line 37 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/206#-N3Tfo_ME1zdxJa0ZUed:-N3Tfo_ME1zdxJa0ZUee:b-5hhz3c) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8d9877a653c7deeef795259205372af88604d6e3/src/components/forms/common/ValidationErrorList.tsx#L37)):*
> ```XML
>   } = useFormContext<FormState>();
> 
>   // TODO check how this behaves if there are multiple errors for the same field
> ```

Just making sure that this is here on purpose and not forgotten here?

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/206)* status: 0 of 20 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`src/components/forms/common/InputField.tsx` line 57 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/206#-N3TjhfW7KBkHI4wCz9U:-N3TjhfX-SPlbOzXujk7:b-7kv63k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8d9877a653c7deeef795259205372af88604d6e3/src/components/forms/common/InputField.tsx#L57)):*
> ```XML
>           inputElementRenderer={inputElementRenderer}
>           fieldPath={fieldPath}
>           id={`${translationPrefix}.${fieldPath}`}
> ```

Note to self: instead of hardcoding the value like this, let's maybe resolve this in the props level so it could be overwritten if needed:
```
{..., id = `${translationPrefix}.${fieldPath}`, ...}: Props
```

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/206)* status: 0 of 20 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[`src/components/forms/common/ControlledElement.tsx` line 15 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/206#-N3Tf2CO0lVDvbMGMNYG:-N3Y2k6H0Dp_ZpkwJcVG:bdcv6kv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8d9877a653c7deeef795259205372af88604d6e3/src/components/forms/common/ControlledElement.tsx#L15)):*

These seem to be repeating in multiple interfaces and thus could be extracted to own interface that could be used to with `extends` keyword in these cases.
</blockquote></details>

Good point!
___
*[`src/components/forms/common/InputField.tsx` line 28 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/206#-N3TdhGX2xlAcWpuiKhN:-N3Y2uZR4gZUHxyqt8Rb:b-coh25u) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8d9877a653c7deeef795259205372af88604d6e3/src/components/forms/common/InputField.tsx#L28)):*

Same as `type?: HTMLInputTypeAttribute`, right? Same for InputElement.

Began to wonder if this should actually be required. OK, seems like "text" is default valut for type prop (https://www.w3schools.com/tags/att_input_type.asp) , but in the other hand seems like it was explicitly given at least in `src/components/forms/route/TerminusNameInputs.tsx` anyway...
</blockquote></details>

I know that the default for `<input>`s is "text", but here I actually wanted the caller to explicitly say what kind of input we are talking about
___
*[`src/components/forms/common/ValidationErrorList.tsx` line 37 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/206#-N3Tfo_ME1zdxJa0ZUed:-N3Y5So56SrSjET69hbj:bvel70m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8d9877a653c7deeef795259205372af88604d6e3/src/components/forms/common/ValidationErrorList.tsx#L37)):*

Just making sure that this is here on purpose and not forgotten here?
</blockquote></details>

yeah, so officially if a field has multiple errors, then the react-form-hook should return with all the errors for the field. However I tried to reproduce it and I couldn't... maybe it's just the way how zod works and it stops at the first validation rule...?

PR #228, opened 2022-06-16
Fix search feature to search routes correctly

HenrikHartiala:
Resolves HSLdevcom/jore4#753

This PR:
* Changes the search logic to query for lines and routes independently (not only lines and then use the routes within)
* `_like` operator for `label` changed to `_ilike` for case insensitive searches
* Adds default hardcoded sorting of the results by `label` and `validity_start` **ascending**

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/228)

### DIFF
--- a/src/generated/graphql.tsx
+++ b/src/generated/graphql.tsx
@@ -6973,12 +6973,15 @@ export type ListAllLinesQueryVariables = Exact<{ [key: string]: never; }>;
 
 export type ListAllLinesQuery = { __typename?: 'query_root', route_line: Array<{ __typename?: 'route_line', line_id: UUID, label: string, name_i18n: LocalizedString, short_name_i18n: LocalizedString, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined }> };
 
-export type SearchAllLinesQueryVariables = Exact<{
-  filter?: Maybe<RouteLineBoolExp>;
+export type SearchLinesAndRoutesQueryVariables = Exact<{
+  lineFilter?: Maybe<RouteLineBoolExp>;
+  routeFilter?: Maybe<RouteRouteBoolExp>;
+  lineOrderBy?: Maybe<Array<RouteLineOrderBy> | RouteLineOrderBy>;
+  routeOrderBy?: Maybe<Array<RouteRouteOrderBy> | RouteRouteOrderBy>;
 }>;
 
 
-export type SearchAllLinesQuery = { __typename?: 'query_root', route_line: Array<{ __typename?: 'route_line', line_id: UUID, label: string, name_i18n: LocalizedString, short_name_i18n: LocalizedString, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, line_routes: Array<{ __typename?: 'route_route', route_id: UUID, name_i18n: LocalizedString, description_i18n?: LocalizedString | null | undefined, origin_name_i18n: LocalizedString, origin_short_name_i18n: LocalizedString, destination_name_i18n: LocalizedString, destination_short_name_i18n: LocalizedString, route_shape?: GeoJSON.LineString | null | undefined, on_line_id: UUID, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string, direction: RouteDirectionEnum }> }> };
+export type SearchLinesAndRoutesQuery = { __typename?: 'query_root', route_line: Array<{ __typename?: 'route_line', line_id: UUID, name_i18n: LocalizedString, short_name_i18n: LocalizedString, primary_vehicle_mode: ReusableComponentsVehicleModeEnum, type_of_line: RouteTypeOfLineEnum, transport_target: HslRouteTransportTargetEnum, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string }>, route_route: Array<{ __typename?: 'route_route', route_id: UUID, name_i18n: LocalizedString, description_i18n?: LocalizedString | null | undefined, origin_name_i18n: LocalizedString, origin_short_name_i18n: LocalizedString, destination_name_i18n: LocalizedString, destination_short_name_i18n: LocalizedString, route_shape?: GeoJSON.LineString | null | undefined, on_line_id: UUID, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string, direction: RouteDirectionEnum }> };
 
 export type ListOwnLinesQueryVariables = Exact<{
   limit?: Maybe<Scalars['Int']>;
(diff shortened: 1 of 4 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/228)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @eleksi, @HenrikHartiala, and @tk-vincit)

___
*[`src/graphql/route.ts` line 120 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gJOwU4GkU0j8pqWSH:-N4gJOwU4GkU0j8pqWSI:bc9hjnr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L120)):*
> ```XML
> 
> const SEARCH_ALL_LINES_AND_ROUTES = gql`
>   query SearchAllLinesAndRoutes(
> ```

naming suggestion: SearchLinesAndRoutes
___
*[`src/graphql/route.ts` line 127 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gMMvH2t60gnjsX4Yv:-N4gMMvH2t60gnjsX4Yw:b-2zqsu3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L127)):*
> ```XML
>   ) {
>     route_line(where: $lineFilter, order_by: $lineOrderBy) {
>       ...line_default_fields
> ```

good idea to combine these into a single query! On the other hand I can't find the place where do you only search for lines or only for routes if the user has decided so? Currently the same filter is applied to both routes and lines, so we always execute both searches even though we might not even need one another... 

You might need to add the @skip directive here https://hasura.io/docs/latest/graphql/core/databases/postgres/queries/variables-aliases-fragments-directives/#skipif-boolean which allows you not to search 
___
*[`src/graphql/route.ts` line 137 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gJW5E38iF0NBP7q5U:-N4gJW5E38iF0NBP7q5V:be48o75) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L137)):*
> ```XML
>   result: ReturnType<typeof useSearchAllLinesAndRoutesQuery>,
> ) => ({
>   lines: result.data?.route_line as RouteLine[],
> ```

before the response arrives, the data object might be undefined. You could however default to empty array here:
```
{
lines: (result.data?.route_line || []) as RouteLine[],
routes: (result.data?.route_route || []) as RouteRoute[],
}
```
___
*[`src/utils/search.ts` line 76 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gM20R51V8vBOuEw4r:-N4gM20R51V8vBOuEw4s:bxl7t5s) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/utils/search.ts#L76)):*
> ```XML
>   });
> 
>   const filter = result.reduce((next, curr) => ({ ...curr, ...next }));
> ```

I can't follow 100% what's happening here, are you trying to flatten this previous array? Wouldn't it work to just call .flatMap((entry) => {...}) instead of the current .map()?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/228)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @eleksi, @pallost, @Skip, and @tk-vincit)

___
*[`src/graphql/route.ts` line 120 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gJOwU4GkU0j8pqWSH:-N4x8Dk80LUHdcAGSSXs:b-su4p7x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L120)):*

naming suggestion: SearchLinesAndRoutes
</blockquote></details>

True! The "all" is redundant here.
___
*[`src/graphql/route.ts` line 127 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gMMvH2t60gnjsX4Yv:-N4x9cCW6MCDj1k7sMyb:b-tld5ew) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L127)):*

good idea to combine these into a single query! On the other hand I can't find the place where do you only search for lines or only for routes if the user has decided so? Currently the same filter is applied to both routes and lines, so we always execute both searches even though we might not even need one another... 

You might need to add the @skip directive here https://hasura.io/docs/latest/graphql/core/databases/postgres/queries/variables-aliases-fragments-directives/#skipif-boolean which allows you not to search 
</blockquote></details>

The search result page has this toggle to show `Routes` or `Lines` in the result list. So when we execute search, we execute both searches and then we have all the needed results if the user wants to change between routes and lines.
___
*[`src/graphql/route.ts` line 137 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gJW5E38iF0NBP7q5U:-N4xAA3c1Kg0p1wiyFlb:bmtzf36) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L137)):*

before the response arrives, the data object might be undefined. You could however default to empty array here:
```
{
lines: (result.data?.route_line || []) as RouteLine[],
routes: (result.data?.route_route || []) as RouteRoute[],
}
```

</blockquote></details>

Ahh, that is actually very nice solution, this could be applied to some other mappers as well! Thanks
___
*[`src/utils/search.ts` line 76 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gM20R51V8vBOuEw4r:-N4xDdms70ibA6GGRGh7:bjgcl2r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/utils/search.ts#L76)):*

I can't follow 100% what's happening here, are you trying to flatten this previous array? Wouldn't it work to just call .flatMap((entry) => {...}) instead of the current .map()?
</blockquote></details>

So what this does is it will turn the array in to an object. This will make the result to be structured as:
```
{
  {...},
  {...}
}
```
instead of 
```
[{...}, {...}]
```
But actually I think I can achieve the same with this: 
`  const filter = Object.assign({}, ...result);`



_Code quote:_
```XML
result.reduce((next, curr) => ({ ...curr, ...next }));
```

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/228)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @eleksi, @HenrikHartiala, @Skip, and @tk-vincit)

___
*[`src/graphql/route.ts` line 127 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gMMvH2t60gnjsX4Yv:-N4zfK9yBrMAbtWaA5Q7:b-2fkx7v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L127)):*

The search result page has this toggle to show `Routes` or `Lines` in the result list. So when we execute search, we execute both searches and then we have all the needed results if the user wants to change between routes and lines.
</blockquote></details>

yes, but the same query gets executed also on the frontend, so it will be still very responsive, even if the actual query goes to the backend as well. Not an essential thing to do in this last day of crunch, but would be nice if you have the time/energy
___
*[`src/utils/search.ts` line 76 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gM20R51V8vBOuEw4r:-N4zfix11WkNMxeKYB94:btliive) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/utils/search.ts#L76)):*

So what this does is it will turn the array in to an object. This will make the result to be structured as:
```
{
  {...},
  {...}
}
```
instead of 
```
[{...}, {...}]
```
But actually I think I can achieve the same with this: 
`  const filter = Object.assign({}, ...result);`
</blockquote></details>

`{{...}, {...}}`? that's not possible ;)
`Object.assign({}, ['a', 2], [3, 'b']) == {0: 3, 1: 'b'}`

were you looking for this...? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/228)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @eleksi, @pallost, @Skip, and @tk-vincit)

___
*[`src/graphql/route.ts` line 127 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gMMvH2t60gnjsX4Yv:-N4zke3wAOee6UJMvT52:bwwh2xg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/graphql/route.ts#L127)):*

yes, but the same query gets executed also on the frontend, so it will be still very responsive, even if the actual query goes to the backend as well. Not an essential thing to do in this last day of crunch, but would be nice if you have the time/energy
</blockquote></details>

Right yeah, I can still look in to this.
___
*[`src/utils/search.ts` line 76 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gM20R51V8vBOuEw4r:-N4zkKLC0gTaTJ4w2Htr:bx0kmd1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/utils/search.ts#L76)):*

`{{...}, {...}}`? that's not possible ;)
`Object.assign({}, ['a', 2], [3, 'b']) == {0: 3, 1: 'b'}`

were you looking for this...? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries
</blockquote></details>

Note the `...result` instead of `result`. That does the trick.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/228)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @eleksi, @pallost, @Skip, and @tk-vincit)

___
*[`src/utils/search.ts` line 76 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/228#-N4gM20R51V8vBOuEw4r:-N5-cfH97bCHfM9CGtHh:b-4mwcf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0ce405f2ac917ea4c4c1e9ff4bfc632b9f46e9a/src/utils/search.ts#L76)):*

aah, true! Could you add a comment here about what you are doing? E.g.
`// combining [{key: value}, ...] into a single object`
</blockquote></details>

:+1: I'll add a comment!

PR #229, opened 2022-06-16
Fix journey pattern losing data when manipulating route's stops

eleksi:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/229)


eleksi:
This PR was approved by @pallost in a call.

PR #237, opened 2022-06-22
Remove observation nullability

HenrikHartiala:
This PR
* **Removes** the possibility to set observationDate to null.
* **Fixes a bug** where line details page fetched information with null parameters, which caused the result to be one random line information from the database. Now we check that the parameter exists before fetching the information.
* **Fixes a bug** where we did not support the possibility that `validity_start` could be null. Now the support is there and if the `validity_start` is null for a line, it will be considered as active (and if the validity_end is in the future or null)

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/237)

### DIFF
--- a/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
+++ b/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
@@ -32,10 +32,6 @@ export const ActionsRow = ({
     const updatedUrlQuery = produce(queryParams, (draft) => {
       if (date.isValid) {
         draft.selectedDate = date.toISODate();
-        delete draft.showAll;
-      } else {
-        draft.showAll = true.toString();
-        delete draft.selectedDate;
       }
     });
 
@@ -54,7 +50,8 @@ export const ActionsRow = ({
         <Column className="w-1/4">
           <input
             type="date"
-            value={observationDate?.toISODate() || ''}
+            required
+            value={observationDate?.toISODate()}
             onChange={(e) => onDateChange(DateTime.fromISO(e.target.value))}
             className="flex-1"
           />
--- a/ui/src/hooks/line-details/useGetLineDetails.ts
+++ b/ui/src/hooks/line-details/useGetLineDetails.ts
@@ -66,7 +66,7 @@ const filterLineDetailsByDate = (line: RouteLine) => {
 
 /** Returns the initial observation date depending on the parameters. */
 const getInitialDate = (
-  selectedISODate: string,
+  selectedISODate: string | undefined,
   validityStart?: DateTime | null,
   validityEnd?: DateTime | null,
 ) => {
@@ -80,8 +80,7 @@ const getInitialDate = (
     (!validityEnd || validityEnd >= DateTime.now());
 
   if (isActiveToday) {
-    // DateTime.now() triggers infinite re-renders, so need to doublecast it
-    return DateTime.fromISO(DateTime.now().toISODate());
+    return DateTime.now().startOf('day');
   }
 
   return validityStart;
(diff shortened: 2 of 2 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/237)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`src/hooks/line-details/useGetLineDetails.ts` line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5AEKY6B3J_DO2ca-xi:-N5AEKY6B3J_DO2ca-xj:boly7z) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L77)):*
> ```XML
>   }
> 
>   const isActiveToday =
> ```

I think this logic is actually incorrect in case validity start is null. Would it actually be
```
  const isActiveToday =
    (!validityStart || validityStart <= DateTime.now()) &&
    (!validityEnd || validityEnd >= DateTime.now());
```
?
___
*[`src/hooks/line-details/useGetLineDetails.ts` line 84 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5ACnWI1TwLmK8MdEQR:-N5ACnWI1TwLmK8MdEQS:b-a59lok) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L84)):*
> ```XML
>   if (isActiveToday) {
>     // DateTime.now() triggers infinite re-renders, so need to doublecast it
>     return DateTime.fromISO(DateTime.now().toISODate());
> ```

this would be clearer like this, there's no need for double casting: we just need it to have a static value
```
DateTime.now().startOf('day')
```
___
*[`src/hooks/line-details/useGetLineDetails.ts` line 127 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5ABnjUAltSb6wAbkW9:-N5ABnjV8ow5Eae5Pzd5:b-pp66fw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L127)):*
> ```XML
>   const [observationDate, setObservationDate] = useState<DateTime>();
> 
>   const selectedDate = queryParams.selectedDate as string;
> ```

`as string | undefined`, because it might not be present in the querystring
___
*[`src/hooks/line-details/useGetLineDetails.ts` line 129 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5AElSX4b9P7UBIP9xK:-N5AElSX4b9P7UBIP9xL:bhj92ge) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L129)):*
> ```XML
>   const selectedDate = queryParams.selectedDate as string;
> 
>   const fetchLineDetails = async () => {
> ```

this logic is still really weird for me... what should happen here is:

```
const selectedDate = queryParams.selectedDate;

const initializeSelectedDate = async() {
  const results = await getMinimumInfoAboutLine(line_id);
  const fixedDate = (DateTime.now between validity start and end) ? DateTime.now : validity start;
  return fixedDate;
}

useEffect(() {
  // if wanted, you can run this with every render, regardless if selectedDate exists or not: then it will also fix the selectedDate if the user chose a date that's outside of the allowed boundaries
  if (!selectedDate) {
    const fixedDate = await initializeSelectedDate()
    write fixedDate to queryParams.selectedDate
  }
})

    const lineByDateResult = await getHighestPriorityLineDetails({
      ...constructLineDetailsGqlFilters(lineDetails, selectedDate),
     skip: !selectedDate
    });

```

so basically if the selectedDate does not exist in the query, retrieve it and put it to the query. With the next render loop, it can be fetched from query and will be available.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/237)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`src/hooks/line-details/useGetLineDetails.ts` line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5AEKY6B3J_DO2ca-xi:-N5E8O1n-pF_CIN1zqA9:b-nhlhvs) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L77)):*

I think this logic is actually incorrect in case validity start is null. Would it actually be
```
  const isActiveToday =
    (!validityStart || validityStart <= DateTime.now()) &&
    (!validityEnd || validityEnd >= DateTime.now());
```
?
</blockquote></details>

Hmm, what am I missing? I think I have it just like that.
___
*[`src/hooks/line-details/useGetLineDetails.ts` line 84 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5ACnWI1TwLmK8MdEQR:-N5EAXbw5W5dNu_OkRnW:b-3cp8y2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L84)):*

this would be clearer like this, there's no need for double casting: we just need it to have a static value
```
DateTime.now().startOf('day')
```
</blockquote></details>

Thanks, didn't know about this. 
___
*[`src/hooks/line-details/useGetLineDetails.ts` line 127 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5ABnjUAltSb6wAbkW9:-N5EBS1nByiNYizWlvdZ:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L127)):*

`as string | undefined`, because it might not be present in the querystring
</blockquote></details>

:+1: 
___
*[`src/hooks/line-details/useGetLineDetails.ts` line 129 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/237#-N5AElSX4b9P7UBIP9xK:-N5EI-0W-dWfCfZdxWYS:b-641vrd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/be2b0a205f7813851abc8f5a445e5beb78b2bb54/src/hooks/line-details/useGetLineDetails.ts#L129)):*

this logic is still really weird for me... what should happen here is:

```
const selectedDate = queryParams.selectedDate;

const initializeSelectedDate = async() {
  const results = await getMinimumInfoAboutLine(line_id);
  const fixedDate = (DateTime.now between validity start and end) ? DateTime.now : validity start;
  return fixedDate;
}

useEffect(() {
  // if wanted, you can run this with every render, regardless if selectedDate exists or not: then it will also fix the selectedDate if the user chose a date that's outside of the allowed boundaries
  if (!selectedDate) {
    const fixedDate = await initializeSelectedDate()
    write fixedDate to queryParams.selectedDate
  }
})

    const lineByDateResult = await getHighestPriorityLineDetails({
      ...constructLineDetailsGqlFilters(lineDetails, selectedDate),
     skip: !selectedDate
    });

```

so basically if the selectedDate does not exist in the query, retrieve it and put it to the query. With the next render loop, it can be fetched from query and will be available.
</blockquote></details>

Can I do a separate PR about this refactor?

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/237)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @pallost and @tk-vincit)

PR #244, opened 2022-06-27
Refactor useGetLineDetails hook

HenrikHartiala:
Refactor hook to always update `observationDate` to queryString before fetching by the observationDate:
* If observationDate is not present in queryString, we check if the line is active today:
  * If line **is active today**, we set *today's date* to queryString
  * If line **is not active today**, we set line's *validity_start* to queryString

Note: If line priority is removed (https://github.com/HSLdevcom/jore4/issues/870), then this hook can be simplified a lot.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/244)

### DIFF
--- a/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
+++ b/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
@@ -1,7 +1,6 @@
 import produce from 'immer';
 import { DateTime } from 'luxon';
 import qs from 'qs';
-import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router';
 import { RouteLine } from '../../../generated/graphql';
@@ -31,7 +30,7 @@ export const ActionsRow = ({
   const onDateChange = (date: DateTime) => {
     const updatedUrlQuery = produce(queryParams, (draft) => {
       if (date.isValid) {
-        draft.selectedDate = date.toISODate();
+        draft.observationDate = date.toISODate();
       }
     });
 
@@ -51,7 +50,7 @@ export const ActionsRow = ({
           <input
             type="date"
             required
-            value={observationDate?.toISODate()}
+            value={observationDate?.toISODate() || ''}
             onChange={(e) => onDateChange(DateTime.fromISO(e.target.value))}
             className="flex-1"
           />
--- a/ui/src/generated/graphql.tsx
+++ b/ui/src/generated/graphql.tsx
@@ -6999,6 +6999,13 @@ export type GetLineDetailsByIdQueryVariables = Exact<{
 
 export type GetLineDetailsByIdQuery = { __typename?: 'query_root', route_line_by_pk?: { __typename?: 'route_line', line_id: UUID, name_i18n: LocalizedString, short_name_i18n: LocalizedString, primary_vehicle_mode: ReusableComponentsVehicleModeEnum, type_of_line: RouteTypeOfLineEnum, transport_target: HslRouteTransportTargetEnum, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined, priority: number, label: string } | null | undefined };
 
+export type GetLineValidityPeriodByIdQueryVariables = Exact<{
+  line_id: Scalars['uuid'];
+}>;
+
+
+export type GetLineValidityPeriodByIdQuery = { __typename?: 'query_root', route_line_by_pk?: { __typename?: 'route_line', line_id: UUID, validity_start?: luxon.DateTime | null | undefined, validity_end?: luxon.DateTime | null | undefined } | null | undefined };
+
 export type GetLinesByValidityQueryVariables = Exact<{
   filter?: Maybe<RouteLineBoolExp>;
 }>;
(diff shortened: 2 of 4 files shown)


HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/244)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 110 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5ZqLnoCL7fS1Hcm7dy:-N5_1var96ljGJscobnD:bvua71v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/db6f7783123ae1f82fb32e053e17a9a0e8fdc7ce/ui/src/hooks/line-details/useGetLineDetails.ts#L110)):*

You commented to PR itself:
>Note: If line priority is removed (HSLdevcom/jore4#870), then this hook can be simplified a lot.

If this was something that we are going to do, then maybe you could add here // TODO: or // NOTE: comment and describe shortly what kind of changes we should do in that case?
</blockquote></details>

I can add some comments yeah, but it is still under discussion whether we remove the priority from lines.
___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 123 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5Zqes4BGVJ4gG4Q7RS:-N5_SHbp1c0vDYkTJk6A:b-c0xkl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/db6f7783123ae1f82fb32e053e17a9a0e8fdc7ce/ui/src/hooks/line-details/useGetLineDetails.ts#L123)):*

I'm a bit confused when reading this hook - why do we need both `dateQs` and `observationDate` variables?  

`dateQs` is a bad variable name - Qs doesn't tell anything, and even if one figures out that it means QueryString, dateQueryString still doesnt tell that that variable is about.
My suggestions:
1) if the variable is not actually needed, initialize observationDate like this: `const observationDate = parseDate(queryParams.observationDate);`
2) if we really need it, then give it more descriptive name
</blockquote></details>

The problem here is that if we call parseDate(...) when initialising observationDate and the queryParameter is invalid, then the page gets stuck (infinite errors). 
___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 137 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5ZrsIm-g03vG1W-Rgj:-N5_6tDWAiaeP9lnyQAS:b6z59hq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/db6f7783123ae1f82fb32e053e17a9a0e8fdc7ce/ui/src/hooks/line-details/useGetLineDetails.ts#L137)):*

Signature of getInitialDate seems to be:
```
const getInitialDate = (
  selectedISODate: string | undefined,
  validityStart?: DateTime | null,
  validityEnd?: DateTime | null,
) 
```
So, if you just drop first parameter does it anymore do what you want it to do? Seems like you either want to call it like `getInitialDate(undefined, validity_start, validity_end)` or then at least explain WTF is going on here. And then we could maybe refactor this to be less confusing...
</blockquote></details>

Hmm, what do you mean "signature"? The selectedISODate is not there anymore.
Maybe some reviewable mistake etc.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/244)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @tk-vincit)

___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 123 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5Zqes4BGVJ4gG4Q7RS:-N5d1Kev7TvIupcMrOVM:b-t7221t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/db6f7783123ae1f82fb32e053e17a9a0e8fdc7ce/ui/src/hooks/line-details/useGetLineDetails.ts#L123)):*

The problem here is that if we call parseDate(...) when initialising observationDate and the queryParameter is invalid, then the page gets stuck (infinite errors). 
</blockquote></details>

Sounds weird. Anyway, this looks better now. :+1: 
___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 137 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5ZrsIm-g03vG1W-Rgj:-N5d00cDBzHJj9PRup7g:b90uqn1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/db6f7783123ae1f82fb32e053e17a9a0e8fdc7ce/ui/src/hooks/line-details/useGetLineDetails.ts#L137)):*

Hmm, what do you mean "signature"? The selectedISODate is not there anymore.
Maybe some reviewable mistake etc.
</blockquote></details>

True, seems like I missed it on first review. Sorry!

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/244)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 124 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5dfcwm-9IJYNrVl_7C:-N5dz5fjAdR_7HhfgkF6:b-abhid7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ab04056159b5ff85f455129f6a739d3bb8706a8c/ui/src/hooks/line-details/useGetLineDetails.ts#L124)):*

Do you have to do unsafe type casting? Seems like queryParams already covers both string and undefined types.
</blockquote></details>

It is needed for the parseDate, because queryParam can also be these types: `qs.ParsedQs | qs.ParsedQs[]`. And the parameter for parseDate can't be neither of these two.
___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 149 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5dgJqJ4-l0Rm4beLcH:-N5dwPz54toQwxQrDbbj:bopzvk1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ab04056159b5ff85f455129f6a739d3bb8706a8c/ui/src/hooks/line-details/useGetLineDetails.ts#L149)):*

Sometimes similar thing has been done like this:
```
history.push({
      search: qs.stringify(updatedUrlQuery),
    });
``` 
(see `ui/src/hooks/usePagination.ts`). Does it matter if we want to push or replace history?

Also started to thing if we would want to have generalized solution for updating url query when needed, but maybe its not yet important thing to do.
</blockquote></details>

Hmm, there was supposed to be a bit different behaviour when using back button, but now that you mentioned it and I tested it, I can't see any difference in the behaviour. 

```
push(path, [state]) - (function) Pushes a new entry onto the history stack
replace(path, [state]) - (function) Replaces the current entry on the history stack
```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/244)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/hooks/line-details/useGetLineDetails.ts` line 149 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/244#-N5dgJqJ4-l0Rm4beLcH:-N5dzlaxCae4NwQzcOw7:bkw2p22) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/ab04056159b5ff85f455129f6a739d3bb8706a8c/ui/src/hooks/line-details/useGetLineDetails.ts#L149)):*

Hmm, there was supposed to be a bit different behaviour when using back button, but now that you mentioned it and I tested it, I can't see any difference in the behaviour. 

```
push(path, [state]) - (function) Pushes a new entry onto the history stack
replace(path, [state]) - (function) Replaces the current entry on the history stack
```
</blockquote></details>

And yeah, generalized solution might come in handy in the future, so I'll keep that in mind as well.

PR #246, opened 2022-06-29
Fix Routestops to filter and sort eligible stops along route

HenrikHartiala:
RouteStopsTable showed stops that were not eligible for route. This PR fixes that issue. Also there was a bug that stops on the same infrastructure link were not sorted to the correct order for the route, which caused the list to be in wrong order and also prevented the stop from being added to the route. This is also fixed in this PR.

There were similarities in the logic that the map stoplist uses and these should be combined in the future if possible. This is already step for the right way and the combination could be done in another iteration.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/246)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/246)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`ui/src/graphql/servicePattern.ts` line 167 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/246#-N5jHmHJ-h9d3byVoNKs:-N5jHmHJ-h9d3byVoNKt:b-2gdnyw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d905c4b31bc3dc23a665607d68ca58addbebdcd0/ui/src/graphql/servicePattern.ts#L167)):*
> ```XML
> 
> // TODO: This should also check that vehicleMode of route matches stop
> // TODO: This should be combined with useExctractRouteFromFeature.ts business logics
> ```

Why not combined already? 
___
*[`ui/src/graphql/servicePattern.ts` line 181 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/246#-N5jHfuA89xkzGz8Kq7m:-N5jHfuA89xkzGz8Kq7n:b-j0v2sv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d905c4b31bc3dc23a665607d68ca58addbebdcd0/ui/src/graphql/servicePattern.ts#L181)):*
> ```XML
>   );
> 
> // TODO: This should be combined with useExctractRouteFromFeature business logics
> ```

Why not combined already? Seems like this is direct copypasta. Thus it would be easy to extract this (=at least sorting function, `(stop1, stop2) => ...` itself) to util function.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/246)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`ui/src/graphql/servicePattern.ts` line 167 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/246#-N5jHmHJ-h9d3byVoNKs:-N5nqatFEPREp0inxX_G:b-kgf85) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d905c4b31bc3dc23a665607d68ca58addbebdcd0/ui/src/graphql/servicePattern.ts#L167)):*

Because the logics don't match 1:1, and this might need a bit bigger refactor so I don't want to expand this PR too much. I also do not fully understand the useExtractRouteFromFeature's logic, so need a bit more research for it. 
</blockquote></details>

Well, OK. Its always a bit fishy to add duplicated logics to repository, but if its not so straighforward to combine the logics then maybe this is good enough for now.
___
*[`ui/src/utils/stops.ts` line 8 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/246#-N5npc-G1ZoD0QcTLG2I:-N5npc-G1ZoD0QcTLG2J:b-yetui5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8212219e475c3d17540a1f1631469375d0517b1e/ui/src/utils/stops.ts#L8)):*
> ```XML
>   isTraversalForwards: boolean,
> ) =>
>   stopPoints.sort((stop1, stop2) =>
> ```

As I tried to explain earlier: it would have been better to extract the whole sort function, so that it would be easier to test and be used separately.
And isTravelsalForwards could also have been extracted from sorting function.

Something like this:

```
const sortStopsByTravelsalForwards = (stop1, stop2) => ...
const sortStopsByTravelsalBackwards = (stop1, stop2) => ...

stopPoints.sort(isTravelsalForwards ? sortStopsByTravelsalForwards : sortStopsByTravelsalBackwards)
```

But OK, now we are not anymore adding copypasta so this is non-blocking thing.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/246)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/graphql/servicePattern.ts` line 167 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/246#-N5jHmHJ-h9d3byVoNKs:-N5o4ArY8v8lYKrM4MXl:by28udt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d905c4b31bc3dc23a665607d68ca58addbebdcd0/ui/src/graphql/servicePattern.ts#L167)):*

Well, OK. Its always a bit fishy to add duplicated logics to repository, but if its not so straighforward to combine the logics then maybe this is good enough for now.
</blockquote></details>

I'll create a ticket about it :+1: so it won't be forgotten.
___
*[`ui/src/utils/stops.ts` line 8 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/246#-N5npc-G1ZoD0QcTLG2I:-N5o2ET769bfiwY4pvln:bhcdo5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8212219e475c3d17540a1f1631469375d0517b1e/ui/src/utils/stops.ts#L8)):*

As I tried to explain earlier: it would have been better to extract the whole sort function, so that it would be easier to test and be used separately.
And isTravelsalForwards could also have been extracted from sorting function.

Something like this:

```
const sortStopsByTravelsalForwards = (stop1, stop2) => ...
const sortStopsByTravelsalBackwards = (stop1, stop2) => ...

stopPoints.sort(isTravelsalForwards ? sortStopsByTravelsalForwards : sortStopsByTravelsalBackwards)
```

But OK, now we are not anymore adding copypasta so this is non-blocking thing.
</blockquote></details>

Ahh, right yeah I missed it the first time sorry, but got it now. Good point! I'll do this fix

PR #262, opened 2022-07-07
Fix multiple stop versions on map

HenrikHartiala:
Resolves HSLdevcom/jore4#860
This also refactors useMapStops logic to make less graphQL fetches and getting the same information over and over again.

Additional note: Not sure if these two functionalities (getStopHighlighted, getStopVehicleMode) benefit from being in the same hook, or should they be split in to their own hooks.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/262)

### DIFF
--- a/ui/src/components/map/RouteStopsOverlay.tsx
+++ b/ui/src/components/map/RouteStopsOverlay.tsx
@@ -129,9 +129,12 @@ export const RouteStopsOverlay = ({ className = '' }: Props) => {
           </div>
         </div>
       </div>
-      {routeStops?.map((routeStop) => (
+      {routeStops?.map((routeStop, index) => (
         <StopRow
-          key={routeStop.label}
+          // This list is recreated every time when changes happen, so we can
+          // use index as key here
+          // eslint-disable-next-line react/no-array-index-key
+          key={`${routeStop.label}_${index}`}
           label={routeStop.label}
           onRoute={routeStop.belongsToJourneyPattern}
           isReadOnly={!routeEditingInProgress}
--- a/ui/src/components/map/routes/DrawRouteLayer.tsx
+++ b/ui/src/components/map/routes/DrawRouteLayer.tsx
@@ -43,7 +43,7 @@ import {
   setDraftRouteGeometryAction,
   stopRouteEditingAction,
 } from '../../../redux';
-import { showToast } from '../../../utils';
+import { filterDistinctConsecutiveRouteStops, showToast } from '../../../utils';
 import { addRoute, removeRoute } from '../mapUtils';
 import { featureStyle, handleStyle } from './editorStyles';
 
@@ -166,7 +166,12 @@ const DrawRouteLayerComponent = (
         editedRouteData?.id,
       );
 
-      dispatch(setDraftRouteGeometryAction({ stops: routeStops, infraLinks }));
+      dispatch(
+        setDraftRouteGeometryAction({
+          stops: filterDistinctConsecutiveRouteStops(routeStops),
+          infraLinks,
+        }),
+      );
 
       if (geometry) {
         addRoute(map, SNAPPING_LINE_LAYER_ID, geometry);
@@ -179,7 +184,7 @@ const DrawRouteLayerComponent = (
     },
     [
       baseRoute,
-      editedRouteData.id,
+      editedRouteData?.id,
       editedRouteData.stops,
       editedRouteData.infraLinks,
       creatingNewRoute,
--- a/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx
+++ b/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx
@@ -111,9 +111,12 @@ export const RouteStopsSection = ({
         onToggle={onToggle}
       />
       {isOpen &&
-        displayedStops.map((item) => (
+        displayedStops.map((item, index) => (
           <RouteStopsRow
-            key={item.scheduled_stop_point_id}
+            // This list is recreated every time when changes happen, so we can
+            // use index as key here
+            // eslint-disable-next-line react/no-array-index-key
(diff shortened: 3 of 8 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/262)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala and @pallost)

___
*[`ui/src/components/map/RouteStopsOverlay.tsx` line 137 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/262#-N6g0kH30pbgS1R-ra2K:-N6g0kH30pbgS1R-ra2L:b3c2yr1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/01bb4466ea828c994b8969a55f4c6725a8517779/ui/src/components/map/RouteStopsOverlay.tsx#L137)):*
> ```XML
>           // use index as key here
>           // eslint-disable-next-line react/no-array-index-key
>           key={routeStop.label + index}
> ```

good fix! I didn't realize we could actually use the index here...

but let's rather use template strings here for clarity
```
key={`${routeStop.label}_${index}`}
```
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx` line 119 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/262#-N6g1DT7-YfVShU4ygzh:-N6g1DT7-YfVShU4ygzi:bq84ick) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/01bb4466ea828c994b8969a55f4c6725a8517779/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx#L119)):*
> ```XML
>             // use index as key here
>             // eslint-disable-next-line react/no-array-index-key
>             key={item.scheduled_stop_point_id + index}
> ```

let's use template string here too
___
*[`ui/src/hooks/stops/useMapStops.ts` line 64 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/262#-N6g96TnDRbhOUZy7wY7:-N6g96TnDRbhOUZy7wY8:bo7rqvr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/01bb4466ea828c994b8969a55f4c6725a8517779/ui/src/hooks/stops/useMapStops.ts#L64)):*
> ```XML
> 
>   const selectedRouteStops = selectedRoute
>     ? extractHighestPriorityStopsFromRoute(selectedRoute, observationDate)
> ```

why do we only select the highest priority stops for the selected route? That would be misleading for the user as e.g. the route can contain both the standard and the temporary version of the same stop and it's all fine and intended.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/262)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @pallost)

___
*[`ui/src/components/map/RouteStopsOverlay.tsx` line 137 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/262#-N6g0kH30pbgS1R-ra2K:-N6gAdjq57QOaSohuDmT:b-4t9mow) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/01bb4466ea828c994b8969a55f4c6725a8517779/ui/src/components/map/RouteStopsOverlay.tsx#L137)):*

good fix! I didn't realize we could actually use the index here...

but let's rather use template strings here for clarity
```
key={`${routeStop.label}_${index}`}
```
</blockquote></details>

Yeah, using index as key is not optimal in all cases, but in this case it should not cause any additional renders. And good one, I'll change it to template string.
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx` line 119 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/262#-N6g1DT7-YfVShU4ygzh:-N6gB1Md0lkwALo9hEK0:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/01bb4466ea828c994b8969a55f4c6725a8517779/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx#L119)):*

let's use template string here too
</blockquote></details>

Done.
___
*[`ui/src/hooks/stops/useMapStops.ts` line 64 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/262#-N6g96TnDRbhOUZy7wY7:-N6gPAGb7kIW2oQkoLTl:by2l6wf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/01bb4466ea828c994b8969a55f4c6725a8517779/ui/src/hooks/stops/useMapStops.ts#L64)):*

why do we only select the highest priority stops for the selected route? That would be misleading for the user as e.g. the route can contain both the standard and the temporary version of the same stop and it's all fine and intended.
</blockquote></details>

We select only the highest priority on observation date to be hilighted when the route is clicked. It will still show the two other versions as blue, but only the active stop is hilighted as red.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/262)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala and @pallost)

___
*[`ui/src/hooks/stops/useMapStops.ts` line 64 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/262#-N6g96TnDRbhOUZy7wY7:-N6gfC7C4ft74pY_znFw:b1easkr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/01bb4466ea828c994b8969a55f4c6725a8517779/ui/src/hooks/stops/useMapStops.ts#L64)):*

We select only the highest priority on observation date to be hilighted when the route is clicked. It will still show the two other versions as blue, but only the active stop is hilighted as red.
</blockquote></details>

I understand. Let's find a better name then. Should it be "activeRouteStops" instead?

PR #267, opened 2022-07-13
Rework line drafts page

HenrikHartiala:
LineDraftsPage will now show all the draft routes which are under the given line label. This is mostly for MVP and this page will be modified in the future to have much more details about route versions etc.

There is also two commits to:
* Decouple useGetLineDetails from RouteStopsSection, so that RouteStopsSection is reusable.
* Move setObservationDate (to query parameters) to useUrlQuery hook, so that it is also reusable.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)

### DIFF
--- a/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
+++ b/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
@@ -1,10 +1,10 @@
-import produce from 'immer';
-import { DateTime } from 'luxon';
 import qs from 'qs';
 import { useTranslation } from 'react-i18next';
-import { useHistory } from 'react-router';
 import { RouteLine } from '../../../generated/graphql';
-import { useGetLineDetails, useUrlQuery } from '../../../hooks';
+import {
+  useGetLineDetails,
+  useObservationDateQueryParam,
+} from '../../../hooks';
 import { Column, Container, Row } from '../../../layoutComponents';
 import { Path, routeDetails } from '../../../router/routeDetails';
 import { SimpleButton } from '../../../uiComponents';
@@ -22,22 +22,14 @@ export const ActionsRow = ({
   className?: string;
 }): JSX.Element => {
   const { t } = useTranslation();
-  const history = useHistory();
-  const queryParams = useUrlQuery();
 
-  const { line, observationDate } = useGetLineDetails();
+  const { line } = useGetLineDetails();
 
-  const onDateChange = (date: DateTime) => {
-    const updatedUrlQuery = produce(queryParams, (draft) => {
-      if (date.isValid) {
-        draft.observationDate = date.toISODate();
-      }
-    });
+  const { observationDate, setObservationDateToUrl } =
+    useObservationDateQueryParam();
 
-    const queryString = qs.stringify(updatedUrlQuery);
-    history.push({
-      search: `?${queryString}`,
-    });
+  const onDateChange = (value: string) => {
+    setObservationDateToUrl(value);
   };
 
   return (
@@ -51,7 +43,7 @@ export const ActionsRow = ({
             type="date"
             required
             value={observationDate?.toISODate() || ''}
-            onChange={(e) => onDateChange(DateTime.fromISO(e.target.value))}
+            onChange={(e) => onDateChange(e.target.value)}
             className="flex-1"
           />
         </Column>
--- a/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 import {
@@ -31,7 +31,7 @@ export const LineDetailsPage = (): JSX.Element => {
   const dispatch = useAppDispatch();
   const { addMapOpenQueryParameter } = useMapUrlQuery();
 
-  const { line } = useGetLineDetails();
+  const { line, observationDate } = useGetLineDetails();
 
   const onCreateRoute = () => {
     dispatch(resetMapEditorStateAction());
@@ -74,7 +74,10 @@ export const LineDetailsPage = (): JSX.Element => {
                   {t('lines.routes')}
                 </h1>
                 {line.line_routes?.length > 0 ? (
-                  <RouteStopsTable routes={line.line_routes} />
+                  <RouteStopsTable
+                    routes={line.line_routes}
+                    observationDate={observationDate ?? DateTime.now()}
+                  />
                 ) : (
                   <CreateRouteBox onCreateRoute={onCreateRoute} />
                 )}
--- a/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
@@ -1,4 +1,3 @@
-import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 import {
@@ -6,6 +5,7 @@ import {
   useAppSelector,
   useGetLineDetails,
   useMapUrlQuery,
+  useObservationDateQueryParam,
 } from '../../../hooks';
 import { Column, Container, Row, Visible } from '../../../layoutComponents';
 import {
@@ -31,7 +31,8 @@ export const LineDetailsPage = (): JSX.Element => {
   const dispatch = useAppDispatch();
   const { addMapOpenQueryParameter } = useMapUrlQuery();
 
-  const { line, observationDate } = useGetLineDetails();
+  const { line } = useGetLineDetails();
+  const { observationDate } = useObservationDateQueryParam();
 
   const onCreateRoute = () => {
     dispatch(resetMapEditorStateAction());
@@ -73,10 +74,10 @@ export const LineDetailsPage = (): JSX.Element => {
                 <h1 className="mt-8 text-3xl font-semibold">
                   {t('lines.routes')}
                 </h1>
-                {line.line_routes?.length > 0 ? (
+                {line.line_routes?.length > 0 && observationDate ? (
                   <RouteStopsTable
                     routes={line.line_routes}
-                    observationDate={observationDate ?? DateTime.now()}
+                    observationDate={observationDate}
                   />
                 ) : (
                   <CreateRouteBox onCreateRoute={onCreateRoute} />
--- a/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx
+++ b/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx
@@ -9,7 +9,6 @@ import {
 import {
   filterHighestPriorityCurrentStops,
   useEditRouteJourneyPattern,
-  useGetLineDetails,
 } from '../../../hooks';
 import { Priority } from '../../../types/Priority';
 import { showDangerToast, showSuccessToast } from '../../../utils';
@@ -19,12 +18,14 @@ import { RouteStopsRow } from './RouteStopsRow';
 interface Props {
   className?: string;
   route: RouteRoute;
+  observationDate: DateTime;
   showUnusedStops: boolean;
 }
 
 export const RouteStopsSection = ({
   className = '',
   route,
+  observationDate,
   showUnusedStops,
 }: Props) => {
   const [isOpen, setOpen] = useState(false);
@@ -40,8 +41,6 @@ export const RouteStopsSection = ({
   const onToggle = () => {
     setOpen(!isOpen);
   };
-
-  const { observationDate } = useGetLineDetails();
 
   const stopsAlongRoute = getEligibleStopsAlongRouteGeometry(route);
 
--- a/ui/src/components/routes-and-lines/line-details/RouteStopsTable.spec.tsx
+++ b/ui/src/components/routes-and-lines/line-details/RouteStopsTable.spec.tsx
@@ -198,7 +198,11 @@ describe(`<${RouteStopsTable.name} />`, () => {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const line = mapLineDetailsWithRoutesResult(mockResponse)!;
     const { container, asFragment } = render(
-      <RouteStopsTable testId={testId} routes={line.line_routes} />,
+      <RouteStopsTable
+        observationDate={DateTime.fromISO('2022-01-01')}
+        testId={testId}
+        routes={line.line_routes}
+      />,
     );
 
     // the stops don't show as the accordion is not open
--- a/ui/src/components/routes-and-lines/line-details/RouteStopsTable.tsx
+++ b/ui/src/components/routes-and-lines/line-details/RouteStopsTable.tsx
@@ -1,5 +1,6 @@
 import { Switch as HuiSwitch } from '@headlessui/react';
 import orderBy from 'lodash/orderBy';
+import { DateTime } from 'luxon/src/datetime';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { RouteRoute } from '../../../generated/graphql';
@@ -9,10 +10,16 @@ import { RouteStopsSection } from './RouteStopsSection';
 interface Props {
   className?: string;
   routes: RouteRoute[];
+  observationDate: DateTime;
   testId?: string;
 }
 
-export const RouteStopsTable = ({ className = '', routes, testId }: Props) => {
+export const RouteStopsTable = ({
+  className = '',
+  routes,
+  observationDate,
+  testId,
+}: Props) => {
   const { t } = useTranslation();
   const [showUnusedStops, setShowUnusedStops] = useState(false);
   const sortedRoutes = orderBy(routes, ['label', 'direction'], ['asc', 'desc']);
@@ -38,6 +45,7 @@ export const RouteStopsTable = ({ className = '', routes, testId }: Props) => {
             <RouteStopsSection
               key={item.route_id}
               route={item}
+              observationDate={observationDate}
               showUnusedStops={showUnusedStops}
             />
           );
--- a/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
@@ -1,21 +1,20 @@
 import { useTranslation } from 'react-i18next';
 import { useHistory, useParams } from 'react-router';
-import { RouteLine } from '../../../generated/graphql';
-import { useUrlQuery } from '../../../hooks';
+import { useObservationDateQueryParam, useUrlQuery } from '../../../hooks';
 import { useGetLineDraftDetails } from '../../../hooks/line-drafts/useGetLineDraftDetails';
-import { Container, Row } from '../../../layoutComponents';
+import { Column, Container, Row } from '../../../layoutComponents';
 import { Path, routeDetails } from '../../../router/routeDetails';
 import { CloseIconButton } from '../../../uiComponents';
-import { RoutesTable } from '../main/RoutesTable';
-import { LineDraftTableHeader } from './LineDraftsTableHeader';
-import { LineDraftTableRow } from './LineDraftTableRow';
+import { RouteStopsTable } from '../line-details/RouteStopsTable';
 
 export const LineDraftsPage = (): JSX.Element => {
   const { t } = useTranslation();
   const history = useHistory();
   const { label } = useParams<{ label: string }>();
-  const queryParams = useUrlQuery();
-  const { lines } = useGetLineDraftDetails();
+  const { queryParams } = useUrlQuery();
+  const { observationDate, setObservationDateToUrl } =
+    useObservationDateQueryParam();
+  const { routes } = useGetLineDraftDetails();
 
   const onClose = () => {
     // If there is no returnTo set, we return to 'routes and lines' page
@@ -26,6 +25,10 @@ export const LineDraftsPage = (): JSX.Element => {
     history.push({
       pathname,
     });
+  };
+
+  const onDateChange = (value: string) => {
+    setObservationDateToUrl(value);
   };
 
   return (
@@ -40,13 +43,23 @@ export const LineDraftsPage = (): JSX.Element => {
           onClick={onClose}
         />
       </Row>
-      {lines?.length ? (
-        <RoutesTable>
-          <LineDraftTableHeader />
-          {lines?.map((item: RouteLine) => (
-            <LineDraftTableRow key={item.line_id} line={item} />
-          ))}
-        </RoutesTable>
+      <Row>
+        <h2 className="text-sm font-bold">{t('filters.observationDate')}</h2>
+      </Row>
+      <Row>
+        <Column className="w-1/4">
+          <input
+            type="date"
+            required
+            value={observationDate?.toISODate() || ''}
+            onChange={(e) => onDateChange(e.target.value)}
+            className="flex-1"
+          />
+        </Column>
+      </Row>
+
+      {routes?.length && observationDate ? (
+        <RouteStopsTable routes={routes} observationDate={observationDate} />
       ) : (
         <Row className="py-20">
           <p className="mx-auto flex text-2xl font-bold">
--- a/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
@@ -1,4 +1,3 @@
-import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory, useParams } from 'react-router';
 import { RouteLine } from '../../../generated/graphql';
@@ -15,13 +14,13 @@ export const LineDraftsPage = (): JSX.Element => {
   const { t } = useTranslation();
   const history = useHistory();
   const { label } = useParams<{ label: string }>();
-  const { returnTo } = useUrlQuery();
+  const queryParams = useUrlQuery();
   const { lines } = useGetLineDraftDetails();
 
   const onClose = () => {
     // If there is no returnTo set, we return to 'routes and lines' page
-    const pathname = returnTo
-      ? routeDetails[Path.lineDetails].getLink(returnTo as string)
+    const pathname = queryParams.returnTo
+      ? routeDetails[Path.lineDetails].getLink(queryParams.returnTo as string)
       : routeDetails[Path.routes].getLink();
 
     history.push({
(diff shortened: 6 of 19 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx` line 30 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SpzsYEQHgXxwW9qJ6:-N8SpzsYEQHgXxwW9qJ7:bhw76ky) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx#L30)):*
> ```XML
>   };
> 
>   /** Determines and sets date to query parameters if it's not there */
> ```

Why do we have to set "initial" observation date to query? IMO it is valid case that it is not there (and then we would fallback to e.g. today in the UI)
___
*[`ui/src/hooks/useUrlQuery.ts` line 6 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SqxSZ8JPV6Thv_GSy:-N8SqxSZ8JPV6Thv_GSz:b-aby0kk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/useUrlQuery.ts#L6)):*
> ```XML
> import { useHistory, useLocation } from 'react-router-dom';
> 
> export const useUrlQuery = () => {
> ```

Do we want to expand responsibilites of this hook or rather create new hook to do new things?

Seems like we could split this easily to 2 different hooks that would do just one thing each:
- `useUrlQuery()` (or readUrlQuery or parseUrlQuery or something) to do what this originally did
- `setToUrlQuery(newParam: Object)`: calls useUrlQuery and sets new query param based to it

Then we could have convenience hooks for setting different parts of url query (e.g. observationDate) to avoid having define same (string) names for parts of url query in different places. And then you could expose all the previous hooks from one general hook, so that UI components could access all of those from one import and from one useSomeHok call.
___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 56 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SsjKbBMDChXhi0Vha:-N8SsjKbBMDChXhi0Vhb:b-o65tbq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L56)):*
> ```XML
>   const routes = mapRouteResultToRoutes(result);
> 
>   // Filter routes by observationDate in front end (and not in gql query) to avoid
> ```

"in ui" or "in frontend"

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 15 files reviewed, 8 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/ActionsRow.tsx` line 40 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8StRsA5DqP0s38p7VV:-N8StRsA5DqP0s38p7VW:b-s0bulc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx#L40)):*
> ```XML
>             value={observationDate?.toISODate() || ''}
>             onChange={(e) =>
>               setObservationDate(DateTime.fromISO(e.target.value))
> ```

let's not directly write to the url query from an inline function, it will be hard to follow... Let's rather wrap this into an `onDateChange` function as it was before
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx` line 21 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SlMpN0ZmproUJr-lk:-N8SlMpN0ZmproUJr-ll:bg7op39) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx#L21)):*
> ```XML
>   className?: string;
>   route: RouteRoute;
>   observationDate?: DateTime;
> ```

let's not allow undefined here. The useGetLineDetails should anyway always default to the current date for observationDate, so this is always set
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsTable.tsx` line 13 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SlbonEZG4mTCbwPYJ:-N8SlbonEZG4mTCbwPYK:bg7op39) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-details/RouteStopsTable.tsx#L13)):*
> ```XML
>   className?: string;
>   routes: RouteRoute[];
>   observationDate?: DateTime;
> ```

let's not allow undefined here. The useGetLineDetails should anyway always default to the current date for observationDate, so this is always set
___
*[`ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx` line 30 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SpzsYEQHgXxwW9qJ6:-N8Su-4f2wPYIwwYCO-D:b-h9vjrs) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx#L30)):*

Why do we have to set "initial" observation date to query? IMO it is valid case that it is not there (and then we would fallback to e.g. today in the UI)
</blockquote></details>

multiple hooks and components are communicating the observationDate state through the query. In case of e.g. useGetLineDetails hook, this is a must as we cannot necessarily always fall back to "today" as default, as it might not be within the line's validity period. IMO, this is perfectly fine solution. (I suggested it ;) ) Though maybe I'd rather move this initialization logic directly to useGetLineDraftDetails hook, instead of having it here in the component
___
*[`ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx` line 63 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Sv9ea5T_Q3LZ6-5k0:-N8Sv9ea5T_Q3LZ6-5k1:b6xpsow) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx#L63)):*
> ```XML
>             value={observationDate?.toISODate() || ''}
>             onChange={(e) =>
>               setObservationDate(DateTime.fromISO(e.target.value))
> ```

instead of modifying the query directly from an inline function, let's wrap it to an event handler function for clarity and trackability
___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8St-CK92sBPbY033t9:-N8St-CK92sBPbY033tA:bkfj344) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L42)):*
> ```XML
>     () =>
>       parseDate(queryParams.observationDate as string | undefined) ??
>       DateTime.now(),
> ```

could you put a comment here that useMemo was used because DateTime.now is constantly changing?

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 15 files reviewed, 8 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/hooks/useUrlQuery.ts` line 6 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SqxSZ8JPV6Thv_GSy:-N8SwYb04L2S_QVTbiDW:b-8jwylb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/useUrlQuery.ts#L6)):*

Do we want to expand responsibilites of this hook or rather create new hook to do new things?

Seems like we could split this easily to 2 different hooks that would do just one thing each:
- `useUrlQuery()` (or readUrlQuery or parseUrlQuery or something) to do what this originally did
- `setToUrlQuery(newParam: Object)`: calls useUrlQuery and sets new query param based to it

Then we could have convenience hooks for setting different parts of url query (e.g. observationDate) to avoid having define same (string) names for parts of url query in different places. And then you could expose all the previous hooks from one general hook, so that UI components could access all of those from one import and from one useSomeHok call.
</blockquote></details>

Similar thoughts... I could imagine extending this hook with some generic functions to serialize and deserialize complex objects, like DateTimes, but the actual observationDate handling should be in its own hook. Maybe just baked into the useGetLineDetails and useGetLineDraftDetails hooks...?

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 15 files reviewed, 10 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 58 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SzLHaDrsq4csGIUHQ:-N8SzLHaDrsq4csGIUHR:bwknbd7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L58)):*
> ```XML
>   // Filter routes by observationDate in front end (and not in gql query) to avoid
>   // unnecessary graphql queries which would cause the list to reload on every date change
>   const filteredRoutes = filterRoutesByObservationDate(routes, observationDate);
> ```

a bit of functional programming thinking: instead of writing a wrapper function like `filterRoutesByObservationDate`, it would be more reusable if you wrote the checker function separately, like this:

```
const isRouteActiveOnObservationDate = (route: RouteRoute, observationDate: DateTime) =>
  isDateInRange(observationDate, route.validity_start, route.validity_end);
```

and then write the filter like this:

```
const filteredRoutes = routes.filter((route) => isRouteActiveOnObservationDate(route, observationDate));
```

or even better (if you learn to read these):
```
const filteredRoutes = routes.filter(partialRight(isRouteActiveOnObservationDate, observationDate));
```
___
*[`ui/src/utils/gql.ts` line 52 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8T-Wmx1o_vAwh18W-e:-N8T-Wmy1myrqPngrfhO:b8i3jw1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/utils/gql.ts#L52)):*
> ```XML
>     label: { _eq: label },
>   },
> });
> ```

why not reuse it? ;)

```
export const constructRouteLineLabelGqlFilter = (label: string) => ({
  route_line: constructLabelGqlFilter(label)
});

```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 15 files reviewed, 10 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/ActionsRow.tsx` line 40 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8StRsA5DqP0s38p7VV:-N8TACH28OwyQRgLp4fe:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx#L40)):*

let's not directly write to the url query from an inline function, it will be hard to follow... Let's rather wrap this into an `onDateChange` function as it was before
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx` line 21 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SlMpN0ZmproUJr-lk:-N8TAedl57Lk2ZbXFWMx:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx#L21)):*

let's not allow undefined here. The useGetLineDetails should anyway always default to the current date for observationDate, so this is always set
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsTable.tsx` line 13 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SlbonEZG4mTCbwPYJ:-N8TAfdC7TQVaDGILUNB:bs7wumx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-details/RouteStopsTable.tsx#L13)):*

let's not allow undefined here. The useGetLineDetails should anyway always default to the current date for observationDate, so this is always set
</blockquote></details>

Done. 
___
*[`ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx` line 30 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SpzsYEQHgXxwW9qJ6:-N8TamHf8WD24ZziALKj:b-spwz4d) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx#L30)):*

multiple hooks and components are communicating the observationDate state through the query. In case of e.g. useGetLineDetails hook, this is a must as we cannot necessarily always fall back to "today" as default, as it might not be within the line's validity period. IMO, this is perfectly fine solution. (I suggested it ;) ) Though maybe I'd rather move this initialization logic directly to useGetLineDraftDetails hook, instead of having it here in the component
</blockquote></details>

Good point, no business being in the component.
___
*[`ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx` line 63 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Sv9ea5T_Q3LZ6-5k0:-N8TIJC88a0hTFiyjX6g:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx#L63)):*

instead of modifying the query directly from an inline function, let's wrap it to an event handler function for clarity and trackability
</blockquote></details>

Done.
___
*[`ui/src/hooks/useUrlQuery.ts` line 6 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SqxSZ8JPV6Thv_GSy:-N8U2SvB5yTkB95uleRt:b-1q51de) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/useUrlQuery.ts#L6)):*

Similar thoughts... I could imagine extending this hook with some generic functions to serialize and deserialize complex objects, like DateTimes, but the actual observationDate handling should be in its own hook. Maybe just baked into the useGetLineDetails and useGetLineDraftDetails hooks...?
</blockquote></details>

This was a great idea. I did retrace my steps and created this useSetToUrlQuery -hook and useSetObservationToUrl -hook. This took a while, but at least I now learned why I should've not extend the useUrlQuery and rather make a new one. We'll see if my solution is fine now.
___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8St-CK92sBPbY033t9:-N8TMOat5_W4clv7uKUG:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L42)):*

could you put a comment here that useMemo was used because DateTime.now is constantly changing?
</blockquote></details>

Done.
___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 56 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SsjKbBMDChXhi0Vha:-N8TMSH63sDtq_TOwR3f:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L56)):*

"in ui" or "in frontend"
</blockquote></details>

Done.
___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 58 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SzLHaDrsq4csGIUHQ:-N8TRgFu8aHR9hnhBscU:b-vides7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L58)):*

a bit of functional programming thinking: instead of writing a wrapper function like `filterRoutesByObservationDate`, it would be more reusable if you wrote the checker function separately, like this:

```
const isRouteActiveOnObservationDate = (route: RouteRoute, observationDate: DateTime) =>
  isDateInRange(observationDate, route.validity_start, route.validity_end);
```

and then write the filter like this:

```
const filteredRoutes = routes.filter((route) => isRouteActiveOnObservationDate(route, observationDate));
```

or even better (if you learn to read these):
```
const filteredRoutes = routes.filter(partialRight(isRouteActiveOnObservationDate, observationDate));
```
</blockquote></details>

Ah good one! I'll go with the first solution, since I think its more readable.
Also checked this `isDateInRange` function in `time.ts` and modified it to handle null startDate's as well. Will put this in a separate commit.
___
*[`ui/src/utils/gql.ts` line 52 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8T-Wmx1o_vAwh18W-e:-N8TSdCd70B5Fmqdj3eq:b3ax0it) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/utils/gql.ts#L52)):*

why not reuse it? ;)

```
export const constructRouteLineLabelGqlFilter = (label: string) => ({
  route_line: constructLabelGqlFilter(label)
});

```
</blockquote></details>

Well, I think that it is a bit more readable in this simple case to write the whole structure on this function without using the function above. If it's 
```
export const constructRouteLineLabelGqlFilter = (label: string) => ({
  route_line: constructLabelGqlFilter(label),
});
```
Then you would need to check what does this `constructLabelGqlFilter(label)` do, and so on. On the other hand, it would be more clear what is the difference between these two functions, so actually I'll do the change after all.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 20 files reviewed, 9 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/hooks/useUrlQuery.ts` line 6 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8SqxSZ8JPV6Thv_GSy:-N8XJoNSDy_GSwsbrpun:b-7smowg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/hooks/useUrlQuery.ts#L6)):*

This was a great idea. I did retrace my steps and created this useSetToUrlQuery -hook and useSetObservationToUrl -hook. This took a while, but at least I now learned why I should've not extend the useUrlQuery and rather make a new one. We'll see if my solution is fine now.
</blockquote></details>

IMO this is fine now.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 20 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 79 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Y8tKdCKVOtGnSoZeC:-N8Y8tKdCKVOtGnSoZeD:b-31h3fb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L79)):*
> ```XML
>                   <RouteStopsTable
>                     routes={line.line_routes}
>                     observationDate={observationDate ?? DateTime.now()}
> ```

the useGetLineDetails should always return a valid value for observationDate. If it's undefined in the beginning, you could extend the render condition to:

```
                {line.line_routes?.length > 0 && observationDate ? (
                  <RouteStopsTable
                    routes={line.line_routes}
                    observationDate={observationDate}
                ...
```
___
*[`ui/src/hooks/useSetObservationDateToUrl.ts` line 3 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8YA_e3ExDPro_K5Kq6:-N8YA_e3ExDPro_K5Kq7:bhmykcr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/hooks/useSetObservationDateToUrl.ts#L3)):*
> ```XML
> import { useSetToUrlQuery } from './useSetToUrlQuery';
> 
> export const useSetObservationDateToUrl = () => {
> ```

umm, I was probably unclear about the responsibilities of the hooks.
- `useUrlQuery` should be responsible for generally reading and writing to the query, without actually knowing what parameters there are and what they mean
- `useObservationDateQueryParam` (so this file) should be responsible for reading and writing the observationDate parameter to the url. To give a meaning to the parameter and additionally do some type conversion. It should look roughly like this:

```
export const useObservationDateQueryParam = () => {
  const queryParams = useUrlQuery();
  const observationDate = parseDate(queryParams['observationDate']);

  const setObservationDate = (date: string, replace = false) => {
    setToUrlQuery({ paramName: 'observationDate', value: date, replace });
  };

  return {observationDate, setObservationDate}
}
```
___
*[`ui/src/hooks/useSetToUrlQuery.ts` line 10 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Y9sqfC3BULx33OV2S:-N8Y9sqfC3BULx33OV2T:bbdittl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/hooks/useSetToUrlQuery.ts#L10)):*
> ```XML
> 
>   /** Sets parameter to URL query
> ```

thanks for the comment!

I think function should actually go rather to useUrlQuery, no need to create a separate hooks for reading from and writing to the query, they belong to the same utility
___
*[`ui/src/utils/gql.ts` line 52 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8T-Wmx1o_vAwh18W-e:-N8YAU8h2g7K9R7TaBf3:blsfwj9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8696c700246efda7ff3ac50545a66a82830d62df/ui/src/utils/gql.ts#L52)):*

Well, I think that it is a bit more readable in this simple case to write the whole structure on this function without using the function above. If it's 
```
export const constructRouteLineLabelGqlFilter = (label: string) => ({
  route_line: constructLabelGqlFilter(label),
});
```
Then you would need to check what does this `constructLabelGqlFilter(label)` do, and so on. On the other hand, it would be more clear what is the difference between these two functions, so actually I'll do the change after all.
</blockquote></details>

Up to you :)

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 21 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 79 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Y8tKdCKVOtGnSoZeC:-N8YDaqJ8_AyxfUdxQ0X:b-80bnvp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L79)):*

the useGetLineDetails should always return a valid value for observationDate. If it's undefined in the beginning, you could extend the render condition to:

```
                {line.line_routes?.length > 0 && observationDate ? (
                  <RouteStopsTable
                    routes={line.line_routes}
                    observationDate={observationDate}
                ...
```

</blockquote></details>

True.
___
*[`ui/src/hooks/useSetToUrlQuery.ts` line 10 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Y9sqfC3BULx33OV2S:-N8YFjDU3pWZEDrIBAir:bf5og1o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/hooks/useSetToUrlQuery.ts#L10)):*

thanks for the comment!

I think function should actually go rather to useUrlQuery, no need to create a separate hooks for reading from and writing to the query, they belong to the same utility
</blockquote></details>

It was like that before, but as suggested by @tk-vincit I split them up in to two, which both now have a single responsibility. If we still decide that we want to combine these yet again, we do it in some other separate PR. 
___
*[`ui/src/hooks/useSetObservationDateToUrl.ts` line 3 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8YA_e3ExDPro_K5Kq6:-N8YFbusEcovQBspbQcx:bxd8bsx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/hooks/useSetObservationDateToUrl.ts#L3)):*

umm, I was probably unclear about the responsibilities of the hooks.
- `useUrlQuery` should be responsible for generally reading and writing to the query, without actually knowing what parameters there are and what they mean
- `useObservationDateQueryParam` (so this file) should be responsible for reading and writing the observationDate parameter to the url. To give a meaning to the parameter and additionally do some type conversion. It should look roughly like this:

```
export const useObservationDateQueryParam = () => {
  const queryParams = useUrlQuery();
  const observationDate = parseDate(queryParams['observationDate']);

  const setObservationDate = (date: string, replace = false) => {
    setToUrlQuery({ paramName: 'observationDate', value: date, replace });
  };

  return {observationDate, setObservationDate}
}
```
</blockquote></details>

Ah yeah, `useObservationDateQueryParm` with the parsing makes sense.  Thanks

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/hooks/useSetToUrlQuery.ts` line 10 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Y9sqfC3BULx33OV2S:-N8YWfwu-Sps4I032N7O:b-eblnqi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/hooks/useSetToUrlQuery.ts#L10)):*

It was like that before, but as suggested by @tk-vincit I split them up in to two, which both now have a single responsibility. If we still decide that we want to combine these yet again, we do it in some other separate PR. 
</blockquote></details>

we just discussed it with @tk-vincit in the office and probably still having issues communicating the idea:
useUrlQuery should be responsible for generally reading and writing query parameters
useObservationDateQueryParam should give a meaning to a set of query parameters (in this case, 'observationDate') and do some additional conversion/parsing to it.

there's no need to split useUrlQuery and useSetToUrlQuery, they are both responsible for providing a generic interface to getting and setting parameters to the query

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/hooks/useSetToUrlQuery.ts` line 10 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8Y9sqfC3BULx33OV2S:-N8Ym3bR9qUsW5ntkL3s:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8dceb6d9ba1c97ba7075e01665c8f969f8f6596a/ui/src/hooks/useSetToUrlQuery.ts#L10)):*

we just discussed it with @tk-vincit in the office and probably still having issues communicating the idea:
useUrlQuery should be responsible for generally reading and writing query parameters
useObservationDateQueryParam should give a meaning to a set of query parameters (in this case, 'observationDate') and do some additional conversion/parsing to it.

there's no need to split useUrlQuery and useSetToUrlQuery, they are both responsible for providing a generic interface to getting and setting parameters to the query
</blockquote></details>

Done.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala, @pallost, and @tk-vincit)

___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 30 at r13](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8bLAcVAOvqCwucjOkC:-N8bLAcVAOvqCwucjOkD:b-h4999x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/15883997dde84fde8482e807208edd1544e38c50/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L30)):*
> ```XML
>     useObservationDateQueryParam();
> 
>   // useMemo hook is used because DateTime.now() is constantly changing
> ```

this comment was forgotten here

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/267)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`ui/src/hooks/line-drafts/useGetLineDraftDetails.ts` line 30 at r13](https://reviewable.io/reviews/hsldevcom/jore4-ui/267#-N8bLAcVAOvqCwucjOkC:-N8bdggj-9fr5id6zvlR:b-ucfknh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/15883997dde84fde8482e807208edd1544e38c50/ui/src/hooks/line-drafts/useGetLineDraftDetails.ts#L30)):*

this comment was forgotten here
</blockquote></details>

Thanks!

PR #273, opened 2022-07-18
Add map position to url

HenrikHartiala:
This PR 
* Adds Map position (longitude, latitude & zoom) to query parameters.
* Adds testIds to stops
* Modifies and extends createRoutes e2e test

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)


culka (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-:-N83nZy57vn1GpHqIice:bnfp4nl)* status: 0 of 11 files reviewed, all discussions resolved

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)* status: 0 of 11 files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 24 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Stz4f9a-0Xu6kSupB:-N8Stz4f9a-0Xu6kSupC:bkuvbko) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/ConfirmSaveForm.ts#L24)):*
> ```XML
>     return cy
>       .getByTestId('confirmSaveForm:endDateInput')
>       .invoke('removeAttr', 'type')
> ```

Just curious: why we have to do this?
___
*[`cypress/pageObjects/MapEditor.ts` line 26 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SuFg9CAcXjkKSWcHd:-N8SuFg9CAcXjkKSWcHe:bigw94n) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/MapEditor.ts#L26)):*
> ```XML
>   ) {
>     cy.getByTestId(testId)
>       .parent()
> ```

Why we are accessing parent element in test (instead of targeting it directly)? This makes coupling between test and parent of element with test id, which sounds tricky to maintain.
___
*[`cypress/pageObjects/MapEditor.ts` line 30 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Suh9x6Z0aQSoHRE9h:-N8Suh9x6Z0aQSoHRE9i:b-4bckkj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/MapEditor.ts#L30)):*
> ```XML
>         const x = mark.position().left;
> 
>         // Map header heigth needs to be subtracted from the y coordinate
> ```

Not sure what is going on here, but just an idea: could we target map element itself to get rid of this problem?
___
*[`cypress/pageObjects/MapFooter.ts` line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Sv1JR0Z8HoGDDPgxO:-N8Sv1JR0Z8HoGDDPgxP:b-duhf4y) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/MapFooter.ts#L22)):*
> ```XML
>   }
> 
>   checkSubmitSuccess() {
> ```

`checkRouteSubmitSuccess` or something like that. Now this is confusing e.g. if user saves stop instance and tries to check submit success with this.

Also not sure if this belongs to MapFooter as its not responsibility of corrsponding UI component to show any kind of toast messages.
___
*[`ui/src/components/map/Maplibre.tsx` line 41 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SvZAcDAXdmLnBPM_L:-N8SvZAd5iPqYeg_gaR2:b-l3ue8i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/components/map/Maplibre.tsx#L41)):*
> ```XML
> 
>   const { mapPosition, setMapPositionQueryParametersDebounced } =
>     useMapUrlQuery();
> ```

In previous PR we had `useUrlQuery` hook that made changes to observationDate in url query, but here we have different hook for setting mapPosition in url query... Wheres the consistency?  Please refactor this according to the decisions we make in the previous PR. (https://reviewable.io/reviews/hsldevcom/jore4-ui/267)
___
*[`ui/src/components/map/markers/Circle.tsx` line 25 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SwPJX1snkkWf4Wl84:-N8SwPJX1snkkWf4Wl85:b-aptgf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/components/map/markers/Circle.tsx#L25)):*
> ```XML
>     <svg height={size} width={size} onClick={onClick}>
>       <circle
>         data-testid={testId}
> ```

Does this cause`data-testid="undefined"` to end up into DOM? If yes, then please at least give default value "" to testId property.

Same for other data-testid props in this PR.
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 10 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8StboW1qH1U7_OMjdz:-N8StboXEUzrU32MixF6:b-eoujsa) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L10)):*
> ```XML
>   const queryParameterNameMapOpen = 'mapOpen' as const;
>   const queryParameterNameLongitude = 'lng' as const;
>   const queryParameterNameLatitude = 'ltd' as const;
> ```

I'm pretty sure we want rather call this `'lat'`
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SxbcXB0SwgVMPwvw7:-N8SxbcXB0SwgVMPwvw8:b-gwuimk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L15)):*
> ```XML
> 
>   const queryParamObject = useMemo(
>     () => new URLSearchParams(history.location.search),
> ```

Are URLSearchParams same as url query params? Please use existing `useUrlQuery` hook for accessing those, and refactor its implementation if to use `URLSearchParams` if needed. We don't want to have multiple ways to do same things.
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 19 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SxzqM0dxj0v4HQPmX:-N8SxzqM0dxj0v4HQPmY:b-rbd19k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L19)):*
> ```XML
>   );
> 
>   const parseFloatFromQueryParam = (value: string) =>
> ```

`value` -> `paramName` or`fieldName`?
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 44 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SyFRlBpdwpGGg-8ir:-N8SyFRlBpdwpGGg-8is:b-ha21bv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L44)):*
> ```XML
>   const setMapPositionQueryParametersDebounced = useMemo(
>     () =>
>       debounce((latitude, longitude, zoom) => {
> ```

Please comment why this has to be debounced
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 81 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Sw9ar3RpUeuO2jmwj:-N8Sw9ar3RpUeuO2jmwk:b2vxz7i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L81)):*
> ```XML
>     addMapOpenQueryParameter,
>     mapPosition,
>     setMapPositionQueryParametersDebounced,
> ```

Does the user of this api care if this call is debounced or not? Why this has to be debounced and corresponding delete call does not?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)* status: 0 of 11 files reviewed, 11 unresolved discussions (waiting on @tk-vincit)

___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 24 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Stz4f9a-0Xu6kSupB:-N8XIuY7C6CWb7zKQrHa:bxfxe5i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/ConfirmSaveForm.ts#L24)):*

Just curious: why we have to do this?
</blockquote></details>

this invoke? Cypress types in the ISODate which has `-` character between year, month and day. These will cause the map to zoom out in the background. This might be something that we want to fix on the map as well (`+` and `-` shortcuts apply on map only if focus is there etc), but did not spend time with that on this PR, since the user really uses the datepicker to get the date. Or even if the date is typed, the dashes are not needed there.
If some better solution comes in to mind with this, Im open for suggestions.
___
*[`cypress/pageObjects/MapEditor.ts` line 26 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SuFg9CAcXjkKSWcHd:-N8XPXFP7-2ccM9cpbev:b-89beob) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/MapEditor.ts#L26)):*

Why we are accessing parent element in test (instead of targeting it directly)? This makes coupling between test and parent of element with test id, which sounds tricky to maintain.
</blockquote></details>

Actually, this does not need the parent() to work, but for some reason the types get messed up. without the `parent()`, the "mark" is marked as being `Element` and with the `parent()` it is JQuery<HtmlElement> which we want. But I need to figure out how can i get this same type without the `parent()`.

Ahhhhh.. I think I got it, I think we had the typing wrong in `cypress/support/index.d.ts`. I changed it to `getByTestId(value: string): Chainable<JQuery<HTMLElement>>;` Now everything is working correctly and no need for the parent.
**Update **: The typing error is now fixed in main.
___
*[`cypress/pageObjects/MapEditor.ts` line 30 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Suh9x6Z0aQSoHRE9h:-N8XUjzL769OHm28ftPr:bv8ya6m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/MapEditor.ts#L30)):*

Not sure what is going on here, but just an idea: could we target map element itself to get rid of this problem?
</blockquote></details>

We are targeting the map element, but the coordinate we get from the `mark.position().y` is the coordinate from top of the screen. 
I would like to use these coordinates to click on the screen, but the map editor does not register those clicks for some reason. When observing the cypress test, it can be seen that the right positions are clicked. Here is one alternative code that does not work even though the correct coordinates are clicked:
```
    cy.getByTestId(testId).then((mark) => {
      const x = mark.position().left;
      const y = mark.position().top;
      cy.root().click(x + xpos, y+ypos, {force: true} )
    });
```

I think I now got a better solution now. And also using the `getBoundingClientRect()` instead of `top` and `left. So now we calculate the position of the map and subtract those coordinates from the marker to click the right position.
___
*[`cypress/pageObjects/MapFooter.ts` line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Sv1JR0Z8HoGDDPgxO:-N8XgYFR7wG5yvw2SOTV:by57tlp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/MapFooter.ts#L22)):*

`checkRouteSubmitSuccess` or something like that. Now this is confusing e.g. if user saves stop instance and tries to check submit success with this.

Also not sure if this belongs to MapFooter as its not responsibility of corrsponding UI component to show any kind of toast messages.
</blockquote></details>

These were actually already changed in the later PR's exactly like this. The name is checkRouteSubmitSuccess and they are in `Toast` page object. My bad that it wasn't already in this PR. I'll add them here already.
___
*[`ui/src/components/map/Maplibre.tsx` line 41 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SvZAcDAXdmLnBPM_L:-N8wjAjG74UBHKtIsR4Z:bsg408e) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/components/map/Maplibre.tsx#L41)):*

In previous PR we had `useUrlQuery` hook that made changes to observationDate in url query, but here we have different hook for setting mapPosition in url query... Wheres the consistency?  Please refactor this according to the decisions we make in the previous PR. (https://reviewable.io/reviews/hsldevcom/jore4-ui/267)
</blockquote></details>

This is now reworked in main as well and used here correspondingly.
___
*[`ui/src/components/map/markers/Circle.tsx` line 25 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SwPJX1snkkWf4Wl84:-N8XrxBLBmxXcWN-qD1T:b1czmbu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/components/map/markers/Circle.tsx#L25)):*

Does this cause`data-testid="undefined"` to end up into DOM? If yes, then please at least give default value "" to testId property.

Same for other data-testid props in this PR.
</blockquote></details>

No, the attribute is removed from the element completely if there is no value (which is nice).
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 10 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8StboW1qH1U7_OMjdz:-N8XsELQ6AFW6Dp4TVYx:b-s8t86a) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L10)):*

I'm pretty sure we want rather call this `'lat'`
</blockquote></details>

check!
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SxbcXB0SwgVMPwvw7:-N8wjGDSCMC8IiZRvr7P:b488osc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L15)):*

Are URLSearchParams same as url query params? Please use existing `useUrlQuery` hook for accessing those, and refactor its implementation if to use `URLSearchParams` if needed. We don't want to have multiple ways to do same things.
</blockquote></details>

This is reworked now.
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 19 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SxzqM0dxj0v4HQPmX:-N8wjMtr7mW5Lh0clf56:bndhkj4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L19)):*

`value` -> `paramName` or`fieldName`?
</blockquote></details>

Ah, that was a mistake. Nice spot! But actually I will also move this function to useUrlQuery with a different name to be consistent with (now implemented) `get` methods in useUrlQuery.
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 44 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SyFRlBpdwpGGg-8ir:-N8wpbKLEuSotiMwIPdB:b-i3ns0u) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L44)):*

Please comment why this has to be debounced
</blockquote></details>

This debouncing is now moved to Maplibre.tsx. Added a comment there though, thanks
___
*[`ui/src/hooks/useMapUrlQuery.ts` line 81 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Sw9ar3RpUeuO2jmwj:-N8wq7NY48QCdacV2L_Y:b488osc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/hooks/useMapUrlQuery.ts#L81)):*

Does the user of this api care if this call is debounced or not? Why this has to be debounced and corresponding delete call does not?
</blockquote></details>

This is reworked now.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)* status: 0 of 25 files reviewed, 8 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`cypress/e2e/createRoute.cy.ts` line 24 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91-I9w7vR3b7PikhEy:-N91-I9x-OHM6v9shIP2:bhxwq08) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4247ac106b3e3732dcce2edf77aa7982582fc294/cypress/e2e/createRoute.cy.ts#L24)):*
> ```XML
>       cy.mockLogin();
>       cy.visit(
>         '/routes?mapOpen=true&lng=24.93021804533524&lat=60.164074274478054&z=15',
> ```

Is it really relevant for the test to add these coordinates here? IMO user should be able to access `/routes?mapOpen=true` route and  it is responsibility to the app itself to populate these parameters to url if needed.
___
*[`cypress/e2e/createRoute.cy.ts` line 87 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N910HN41IW_xBQaHZpB:-N910HN41IW_xBQaHZpC:bfblco3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/cypress/e2e/createRoute.cy.ts#L87)):*
> ```XML
>           -10,
>           25,
>           'map::stopMarker::e3528755-711f-4e4f-9461-7931a2c4bc6d',
> ```

Where does this magical uuid come from? This kind of selector is impossible to maintain. Please import the uuid from the file where it is originally defined, or at least define it in a variable with descriptive name and some documentation, e.g. somehow like this:
```
// this stop is created when seeding test db
const prepopulatedStopUuid = ....;
```
___
*[`cypress/e2e/createRoute.cy.ts` line 92 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N910p1Q1BSF2S0VN4ly:-N910p1Q1BSF2S0VN4lz:bcvbenc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/cypress/e2e/createRoute.cy.ts#L92)):*
> ```XML
>           25,
>           5,
>           'map::stopMarker::f8eace87-7901-4438-bfee-bb6f24f1c4c4',
> ```

Magical uuid
___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 24 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Stz4f9a-0Xu6kSupB:-N911BcA-2OuUo_1B8X0:bx79wca) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/ConfirmSaveForm.ts#L24)):*

this invoke? Cypress types in the ISODate which has `-` character between year, month and day. These will cause the map to zoom out in the background. This might be something that we want to fix on the map as well (`+` and `-` shortcuts apply on map only if focus is there etc), but did not spend time with that on this PR, since the user really uses the datepicker to get the date. Or even if the date is typed, the dashes are not needed there.
If some better solution comes in to mind with this, Im open for suggestions.
</blockquote></details>

So if I understood correcly, when user types '-' or '+' char those actually zoom map in/out and this is some kind of hack to prevent this from happening? Then I suggest we at least add comment about this here, this is not self-evident thing and makes this function confusing.
___
*[`ui/src/components/map/markers/Circle.tsx` line 25 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8SwPJX1snkkWf4Wl84:-N914Ba5CfbrCSO1UwpY:b32w0x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/ui/src/components/map/markers/Circle.tsx#L25)):*

No, the attribute is removed from the element completely if there is no value (which is nice).
</blockquote></details>

OK
___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 2 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91CSz94oXkh9RGP_Xj:-N91CSz94oXkh9RGP_Xk:b-drc3g8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/ui/src/hooks/urlQuery/useUrlQuery.ts#L2)):*
> ```XML
> import produce from 'immer';
> import { isBoolean, isNumber } from 'lodash';
> ```

This imports the whole lodash library. 
Please import these like this:
`import isBoolean from 'lodash/isBoolean'`, and same for isNumber
___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 58 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91C3_H70cKn6u91D1i:-N91C3_H70cKn6u91D1j:b-2f3q0r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/ui/src/hooks/urlQuery/useUrlQuery.ts#L58)):*
> ```XML
>   };
> 
>   /** Sets boolean parameter to URL query
> ```

Nice documentation!

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)* status: 0 of 25 files reviewed, 8 unresolved discussions (waiting on @tk-vincit)

___
*[`cypress/e2e/createRoute.cy.ts` line 24 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91-I9w7vR3b7PikhEy:-N91Egwu9Ntp13C7gLsb:b-9kk66c) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4247ac106b3e3732dcce2edf77aa7982582fc294/cypress/e2e/createRoute.cy.ts#L24)):*

Is it really relevant for the test to add these coordinates here? IMO user should be able to access `/routes?mapOpen=true` route and  it is responsibility to the app itself to populate these parameters to url if needed.
</blockquote></details>

This is a shortcut to a precise location to ease up the test. The App does populate the coordinates when you open the map, but that is the default Helsinki city center coordinates, which I do not want to use in these tests.

So the alternative would be to open this with just `/routes?mapOpen=true` and then use the UI to get a nice zoom and correct position for the tests.
___
*[`cypress/e2e/createRoute.cy.ts` line 87 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N910HN41IW_xBQaHZpB:-N91FWjb1PVSYxqtHPp_:b-dkoztv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/cypress/e2e/createRoute.cy.ts#L87)):*

Where does this magical uuid come from? This kind of selector is impossible to maintain. Please import the uuid from the file where it is originally defined, or at least define it in a variable with descriptive name and some documentation, e.g. somehow like this:
```
// this stop is created when seeding test db
const prepopulatedStopUuid = ....;
```
</blockquote></details>

Hmm, right, this this actually a result of merge conflict. My initial test id for this was <label>_<priority>, but it was changed to this UUID version. I forgot to change this back, because I also came in to the conclusion that we can't use this UUID test-id for any stop that we create via the UI.
___
*[`cypress/e2e/createRoute.cy.ts` line 92 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N910p1Q1BSF2S0VN4ly:-N91VskCA0oyxeor6toT:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/cypress/e2e/createRoute.cy.ts#L92)):*

Magical uuid
</blockquote></details>

Done.
___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 24 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Stz4f9a-0Xu6kSupB:-N91VM2h1pFYKiRPzvMs:b6y857f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/ConfirmSaveForm.ts#L24)):*

So if I understood correcly, when user types '-' or '+' char those actually zoom map in/out and this is some kind of hack to prevent this from happening? Then I suggest we at least add comment about this here, this is not self-evident thing and makes this function confusing.
</blockquote></details>

Added comment. This invoke was already in the existing code and did not pay that much attention to it that time. But it is true that this is confusing with the `'removeAttr', 'type'` as well and maybe the map zoom should be fixed soon to also get rid of these invokes as well.
___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 2 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91CSz94oXkh9RGP_Xj:-N91S5Jb1VavJTVsbwr-:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/ui/src/hooks/urlQuery/useUrlQuery.ts#L2)):*

This imports the whole lodash library. 
Please import these like this:
`import isBoolean from 'lodash/isBoolean'`, and same for isNumber
</blockquote></details>

Done.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)* status: 0 of 25 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`cypress/e2e/createRoute.cy.ts` line 24 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91-I9w7vR3b7PikhEy:-N91dT5e-mAmGpM3-Kn_:b-5bqigb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/4247ac106b3e3732dcce2edf77aa7982582fc294/cypress/e2e/createRoute.cy.ts#L24)):*

This is a shortcut to a precise location to ease up the test. The App does populate the coordinates when you open the map, but that is the default Helsinki city center coordinates, which I do not want to use in these tests.

So the alternative would be to open this with just `/routes?mapOpen=true` and then use the UI to get a nice zoom and correct position for the tests.
</blockquote></details>

OK, if these are relevant for the test then no problem here :ok_hand: 
___
*[`cypress/e2e/createRoute.cy.ts` line 87 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N910HN41IW_xBQaHZpB:-N91deVQ5DfeMRVwf8Ku:b-w0yibo) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/7e450acab090c888ad2af7fec699aafccb3684c9/cypress/e2e/createRoute.cy.ts#L87)):*

Hmm, right, this this actually a result of merge conflict. My initial test id for this was <label>_<priority>, but it was changed to this UUID version. I forgot to change this back, because I also came in to the conclusion that we can't use this UUID test-id for any stop that we create via the UI.
</blockquote></details>

Much more readable and maintainable now. Thanks.
___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 24 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N8Stz4f9a-0Xu6kSupB:-N91fEmG72k1IXg9M9V_:b-hay274) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e0806c91b0e9488aa8521c5b74c1bcc7c039b568/cypress/pageObjects/ConfirmSaveForm.ts#L24)):*

Added comment. This invoke was already in the existing code and did not pay that much attention to it that time. But it is true that this is confusing with the `'removeAttr', 'type'` as well and maybe the map zoom should be fixed soon to also get rid of these invokes as well.
</blockquote></details>

Thanks. 
___
*[`cypress/pageObjects/MapEditor.ts` line 24 at r6](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91e5XgBttovk6vB-zF:-N91e5XgBttovk6vB-zG:b9rxp2m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8f023e1978dd14291daee0f14a618f9534793d72/cypress/pageObjects/MapEditor.ts#L24)):*
> ```XML
>     cy.get('#editor').then((editor) => {
>       cy.getByTestId(testId).then((mark) => {
>         // Because we are clicking the editor, we need to subtract the editors coordinates
> ```

Nice that we have some kind of documentation! Anyway, not sure if I understand this comment (at all). 

- "Were are clicking the editor" (Map Editor, what is map editor, we don't have that kind of concept in UI side, so now I realised that is is a bit confusing that we have "Map Editor" page object. What is the page that it handles? Map Itself?)

- "we need to subtract the editors coordinates from the markers coodrdinates" what? What coordinates? if the editor is the map itself, what are its coordinates? top right corner? bottom left corner? what?

- "to get the correct coordinate" - not sure what it might be and what we are aiming to do here. Seems like we want to hit some kind of map markers, but why do we need the coordinates at all if we already have testId to target the marker?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)* status: 0 of 26 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[`cypress/pageObjects/MapEditor.ts` line 24 at r6](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N91e5XgBttovk6vB-zF:-N95detI92Zjq22Ihszl:bfm0ple) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8f023e1978dd14291daee0f14a618f9534793d72/cypress/pageObjects/MapEditor.ts#L24)):*

Nice that we have some kind of documentation! Anyway, not sure if I understand this comment (at all). 

- "Were are clicking the editor" (Map Editor, what is map editor, we don't have that kind of concept in UI side, so now I realised that is is a bit confusing that we have "Map Editor" page object. What is the page that it handles? Map Itself?)

- "we need to subtract the editors coordinates from the markers coodrdinates" what? What coordinates? if the editor is the map itself, what are its coordinates? top right corner? bottom left corner? what?

- "to get the correct coordinate" - not sure what it might be and what we are aiming to do here. Seems like we want to hit some kind of map markers, but why do we need the coordinates at all if we already have testId to target the marker?
</blockquote></details>

Thanks for pushing this, I checked this again and finally found a proper solution. It was hard to think of nice parameter names and the comment, but please say if they should still be changed.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/273)* status: 0 of 26 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`cypress/pageObjects/MapEditor.ts` line 19 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/273#-N96nI9m83iNcLnV3WoM:-N96nI9m83iNcLnV3WoN:b-7vysuf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2bd306ee446c71b03a33cd6a1bf13dde7795b6a2/cypress/pageObjects/MapEditor.ts#L19)):*
> ```XML
>   /** Clicks position relative to stopMarker. So this uses given stopMarker
>    * as origin and then adds 'right' to x coordinate and 'down' to y coordinates
>    * and clicks that position. You can go left and up by giving negative numbers. */
> ```

Still don't understand what is the use case to click some other position other than stop marker itself, relative to the stop marker that is selected. (And usually x axel points to the right and y axel UP, not down, so comment is slightly confusing in that regard, but I think it is clear enough anyway.)

But yeah, this is much cleaner now.

PR #278, opened 2022-08-02
Fix e2e tests

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/278)

### DIFF
--- a/cypress/e2e/changeLanguage.cy.ts
+++ b/cypress/e2e/changeLanguage.cy.ts
@@ -1,6 +1,6 @@
 import { Navbar } from '../pageObjects';
 
-describe('Verify that language chancing works', () => {
+describe('Verify that language changing works', () => {
   let navbar: Navbar;
   beforeEach(() => {
     navbar = new Navbar();
--- a/cypress/e2e/createRoute.cy.ts
+++ b/cypress/e2e/createRoute.cy.ts
@@ -60,8 +60,8 @@ if (!Cypress.env('SKIP_MAP_TESTS')) {
         routePropertiesForm.save();
 
         mapEditor.clickAtPositionFromNthMapMarker(-10, 25, 1);
-        mapEditor.clickAtPositionFromNthMapMarker(25, 5, 1);
-        mapEditor.clickNthCreatedRectangle(1);
+        mapEditor.clickAtPositionFromNthMapMarker(25, 5, 3);
+        mapEditor.clickNthSnappingPointHandle(1);
 
         cy.getByTestId('RouteStopsOverlay:mapOverlayHeader')
           .get('div')
--- a/cypress/package.json
+++ b/cypress/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@hsl/jore4-test-db-manager": "1.0.0",
-    "cypress": "^10.1.0"
+    "cypress": "^10.3.1"
   },
   "devDependencies": {
     "eslint-plugin-cypress": "^2.12.1"
--- a/cypress/pageObjects/MapEditor.ts
+++ b/cypress/pageObjects/MapEditor.ts
@@ -1,14 +1,3 @@
-function positionStringToCoordinates(value: string | null) {
-  const matcher = /translate\((\d+)px, (\d+)px\)/;
-  if (value) {
-    const position = value.match(matcher)?.slice(1);
-    if (position != null && position.length > 1) {
-      return { x: +position[0], y: +position[1] };
-    }
-  }
-  return { x: 0, y: 0 };
-}
-
 export class MapEditor {
   zoomIn() {
     cy.get('.mapboxgl-ctrl-zoom-in').click();
@@ -31,17 +20,15 @@ export class MapEditor {
     ypos: number,
     markerNumber: number,
   ) {
-    this.getNthMarker(markerNumber).then((mark) => {
-      const position = positionStringToCoordinates(
-        mark[0].getAttribute('style'),
-      );
-      cy.get('#editor').click(position.x + xpos, position.y + ypos, {
+    this.getNthMarker(markerNumber).then((marker) => {
+      const position = marker.position();
+      cy.get('#editor').click(position.left + xpos, position.top + ypos, {
         force: true,
       });
     });
   }
 
-  clickNthCreatedRectangle(nth: number) {
+  clickNthSnappingPointHandle(nth: number) {
     cy.get(`rect[data-index="${nth}"]`).first().click({ force: true });
   }
 }
--- a/ui/src/components/map/stops/EditStopLayer.tsx
+++ b/ui/src/components/map/stops/EditStopLayer.tsx
@@ -262,6 +262,7 @@ export const EditStopLayer: React.FC<Props> = ({
     <>
       {/* Creates new <Stop> instance for both existing and draft stops */}
       <Stop
+        testId="map:stopMarker:editedStop"
         selected
         longitude={editedStopLocation.longitude}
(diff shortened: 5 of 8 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/278)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`cypress/pageObjects/MapEditor.ts` line 1 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XB_gL7jg_oqWdAsKV:-N8XB_gL7jg_oqWdAsKW:b-vtqnr4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/cypress/pageObjects/MapEditor.ts#L1)):*
> ```XML
> function positionStringToCoordinates(value: string | null) {
> ```

Maybe we should have documentation (comment) about what kind of "position strings" this does parse. And having small test suite for this wouldn't hurt either, reading (and possibly editing in the future) regexes is a bit tricky.
___
*[`ui/src/components/map/stops/Stop.tsx` line 21 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XB0Hl0zsK5GpQoK9-:-N8XB0Hl0zsK5GpQoK90:b-ucvwpy) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/ui/src/components/map/stops/Stop.tsx#L21)):*
> ```XML
> 
> export const Stop = ({
>   testId,
> ```

I think we should have default value ("") for this to avoid "undefined" being ending up to DOM if this is not given. OK, now this is not optional but maybe it should be..?
___
*[`ui/src/components/map/stops/Stop.tsx` line 54 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XBQxg6qGSXmCFfA64:-N8XBQxg6qGSXmCFfA65:bnzcksk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/ui/src/components/map/stops/Stop.tsx#L54)):*
> ```XML
>       draggable={draggable}
>       onDragEnd={onDragEnd}
>       className={`${testId} map:stopMarker rounded-full`} // the Marker component does not have testid property, neither does data-testid work, so have to use className
> ```

Why "map:stopMarker" class?

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/278)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`ui/src/components/map/stops/Stop.tsx` line 54 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XBQxg6qGSXmCFfA64:-N8XD5jt86D0IxjJ--Xl:b-vx10cb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/ui/src/components/map/stops/Stop.tsx#L54)):*

Why "map:stopMarker" class?
</blockquote></details>

If it is supposed to be testId, then I would rather at least define local variable `const testIds = `map:stopMarker ${testId}`` to make it self-evident that it is supposed to be test id and not css class name.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/278)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[`cypress/pageObjects/MapEditor.ts` line 1 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XB_gL7jg_oqWdAsKV:-N8XKwDW7aZlZ3U1UGjT:bnhvl77) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/cypress/pageObjects/MapEditor.ts#L1)):*

Maybe we should have documentation (comment) about what kind of "position strings" this does parse. And having small test suite for this wouldn't hurt either, reading (and possibly editing in the future) regexes is a bit tricky.
</blockquote></details>

actually, I found that cypress's jquery can get the position of this element just fine, so no need to mess around with this unreliable method
___
*[`ui/src/components/map/stops/Stop.tsx` line 21 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XB0Hl0zsK5GpQoK9-:-N8XGLt56ejSiT5x-iO7:bb60mde) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/ui/src/components/map/stops/Stop.tsx#L21)):*

I think we should have default value ("") for this to avoid "undefined" being ending up to DOM if this is not given. OK, now this is not optional but maybe it should be..?
</blockquote></details>

I don't think we should allow undefined values or empty strings as test ids. Not for stop markers neither for any other interactive elements... It's always a bigger hassle to track and fill the missing test ids down when creating e2e tests than already filling them in at the first place.
___
*[`ui/src/components/map/stops/Stop.tsx` line 54 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XBQxg6qGSXmCFfA64:-N8XGtgD32sRhrpTVXJg:b-5ill21) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/ui/src/components/map/stops/Stop.tsx#L54)):*

If it is supposed to be testId, then I would rather at least define local variable `const testIds = `map:stopMarker ${testId}`` to make it self-evident that it is supposed to be test id and not css class name.
</blockquote></details>

good point! I don't actually think I need the `map:stopMarker` here, so I'll just remove it. Then the testId parameter will be obvious in the template string

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/278)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @tk-vincit)

___
*[`cypress/pageObjects/MapEditor.ts` line 1 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XB_gL7jg_oqWdAsKV:-N8XO8Wu64OoXH6ZnvsK:b-83sjoh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/cypress/pageObjects/MapEditor.ts#L1)):*

actually, I found that cypress's jquery can get the position of this element just fine, so no need to mess around with this unreliable method
</blockquote></details>

Nice!
___
*[`ui/src/components/map/stops/Stop.tsx` line 21 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/278#-N8XB0Hl0zsK5GpQoK9-:-N8XNql6CPvyGn5WG7sY:b32w0x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b9a9931adcd5c141f638f747f02d0531fd3c5b4e/ui/src/components/map/stops/Stop.tsx#L21)):*

I don't think we should allow undefined values or empty strings as test ids. Not for stop markers neither for any other interactive elements... It's always a bigger hassle to track and fill the missing test ids down when creating e2e tests than already filling them in at the first place.
</blockquote></details>

OK

PR #281, opened 2022-08-04
Validate all form fields

eleksi:
- Validate route form fields
- Validate stop form fields
- Fix visual bug in loading overlay

Resolves https://github.com/HSLdevcom/jore4/issues/899

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`ui/src/components/forms/common/ValidationErrorList.tsx` line 45 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bwiJoBI1yrgDV8VSs:-N8bwiJoBI1yrgDV8VSt:b-jpnziz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/common/ValidationErrorList.tsx#L45)):*
> ```XML
>     // but mapping by type works neither as same error type has multiple different messages to show.
>     // TODO: Figure out a better way to do this.
>     switch (message) {
> ```

well, to do this properly, we'll need to use https://github.com/colinhacks/zod#superrefine , so that we can create custom zod types. Something like...
```
const requiredString = z.string().required().superRefine((val, ctx) => {
  if (val.length === 0) {
    ctx.addIssue({
      code: z.ZodIssueCode.required,
       ...    
    });
  }

const requiredUuid = ...
const requiredNativeEnum = ...
```
(could put this into a file called components/forms/common/customZodSchemas.ts - also then merge it with localizedStringSchema.ts)


This would allow us to do some additional check for the value (e.g. string length > 0) and still mark it as a "required" type. Then this type should be used in zod schemas, like:

```
export const routeFormSchema = z
  .object({
    label: requiredString,
    finnishName: requiredString,
    onLineId: requiredUuid,
    direction: z.nativeEnum(RouteDirection),
    origin: namesSchema.required(),
    destination: namesSchema.required(),
  })
  .merge(confirmSaveFormSchema);
```

___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 105 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bvbU7Ddcq2Y5ofAwW:-N8bvbU7Ddcq2Y5ofAwX:b-fixqal) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/route/RoutePropertiesForm.tsx#L105)):*
> ```XML
>             />
>           </Column>
>           <Column className="w-80 flex-auto">
> ```

please also start using the `<FormColumn>` and `<FormRow>` components for making the layout responsive
___
*[`ui/src/components/forms/stop/StopForm.tsx` line 162 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bvp0VAvxnnXGcpzGT:-N8bvp0VAvxnnXGcpzGU:b-fixqal) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/stop/StopForm.tsx#L162)):*
> ```XML
>             <Column className="space-y-2">
>               <h3 className="text-lg font-bold">{t('stops.nameAddress')}</h3>
>               <Column>
> ```

please also start using the `<FormColumn>` and `<FormRow>` components for making the layout responsive

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @pallost)

___
*[`ui/src/components/forms/common/ValidationErrorList.tsx` line 45 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bwiJoBI1yrgDV8VSs:-N8cvuVDDnfJu8dFGish:b-1z2k85) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/common/ValidationErrorList.tsx#L45)):*

well, to do this properly, we'll need to use https://github.com/colinhacks/zod#superrefine , so that we can create custom zod types. Something like...
```
const requiredString = z.string().required().superRefine((val, ctx) => {
  if (val.length === 0) {
    ctx.addIssue({
      code: z.ZodIssueCode.required,
       ...    
    });
  }

const requiredUuid = ...
const requiredNativeEnum = ...
```
(could put this into a file called components/forms/common/customZodSchemas.ts - also then merge it with localizedStringSchema.ts)


This would allow us to do some additional check for the value (e.g. string length > 0) and still mark it as a "required" type. Then this type should be used in zod schemas, like:

```
export const routeFormSchema = z
  .object({
    label: requiredString,
    finnishName: requiredString,
    onLineId: requiredUuid,
    direction: z.nativeEnum(RouteDirection),
    origin: namesSchema.required(),
    destination: namesSchema.required(),
  })
  .merge(confirmSaveFormSchema);
```

</blockquote></details>

I tried to do it this way, but unfortunately those custom zod types are not extendable.

https://github.com/colinhacks/zod/issues/454

This means for example if we create `requiredNumber` type, we cannot extend it by adding min and max. I'm working on a solution that is better than the original one but not as good as the one you proposed.

We could solve this problem by not extending our custom types at all, but instead creating new types for each field type that has unique validation, e.g. `requiredCoordinate`. I'm not sure if we want to prioritize reusability or readability.


In addition, when validating a number, the error is always type error, because empty field is a NaN. We cannot make a field that has type number and get some other type of error, because the type is always checked first.

Combining the points above, we would have an ugly custom type for `requiredCoordinate` like this:



_Code snippet:_
```XML
export const REQUIRED_ISSUE = {
  code: z.ZodIssueCode.custom, // There is no "required" code
  message: 'required',
  fatal: true,
};

// Instead of this, I would like to do requiredCoordinate = requiredNumber.min(-180).max(180)

export const requiredCoordinate = z.unknown().superRefine((val, ctx) => { // If z.number() used, empty field always returns type error
  if (Number.isNaN(val)) {
    ctx.addIssue(REQUIRED_ISSUE);
  }

  const numberVal = Number(val);
  
  // Min and max have to be checked here because we cannot extend requiredCoordinate, as it is ZodEffects instead of ZodNumber
  if (numberVal < -180) {
    ctx.addIssue({
      code: z.ZodIssueCode.too_small,
      minimum: 180,
      inclusive: false,
      type: 'number',
    });
  } else if (numberVal > 180) {
    ctx.addIssue({
      code: z.ZodIssueCode.too_big,
      maximum: 180,
      inclusive: false,
      type: 'number',
    });
  }
});
```
___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 105 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bvbU7Ddcq2Y5ofAwW:-N8cwzErAUZ4Xt1bg56m:b-deessm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/route/RoutePropertiesForm.tsx#L105)):*

please also start using the `<FormColumn>` and `<FormRow>` components for making the layout responsive
</blockquote></details>

This form already has quite responsive layout that might not be easily kept if we start using those.
___
*[`ui/src/components/forms/stop/StopForm.tsx` line 162 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bvp0VAvxnnXGcpzGT:-N8cxK3_4lNbuzXPenRe:btfrn50) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/stop/StopForm.tsx#L162)):*

please also start using the `<FormColumn>` and `<FormRow>` components for making the layout responsive
</blockquote></details>

I will do that.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`ui/src/components/forms/common/ValidationErrorList.tsx` line 45 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bwiJoBI1yrgDV8VSs:-N8wNPGJ9B6_rMYJsyt_:b-qh0gs5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/common/ValidationErrorList.tsx#L45)):*

I tried to do it this way, but unfortunately those custom zod types are not extendable.

https://github.com/colinhacks/zod/issues/454

This means for example if we create `requiredNumber` type, we cannot extend it by adding min and max. I'm working on a solution that is better than the original one but not as good as the one you proposed.

We could solve this problem by not extending our custom types at all, but instead creating new types for each field type that has unique validation, e.g. `requiredCoordinate`. I'm not sure if we want to prioritize reusability or readability.


In addition, when validating a number, the error is always type error, because empty field is a NaN. We cannot make a field that has type number and get some other type of error, because the type is always checked first.

Combining the points above, we would have an ugly custom type for `requiredCoordinate` like this:

</blockquote></details>

damn... that sucks... Well, could try this kind of approach then:

```
const requiredString = () => z.string().min(1, {message: 'zod_required'})
const requiredEnum = (enumType: Object) => z.nativeEnum(enumType, {required_error: 'zod_required', invalid_type_error: 'zod_required'})
const requiredUuid = () => requiredString().uuid()
```

at least you'd have a more generic message to use as type...
___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 105 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bvbU7Ddcq2Y5ofAwW:-N8wN41V-EZJWCGpNU0_:b-n0s73) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/route/RoutePropertiesForm.tsx#L105)):*

This form already has quite responsive layout that might not be easily kept if we start using those.
</blockquote></details>

well, w-80 is still just hardcoding the width for the input field, so it won't scale for bigger and/or smaller screens. If it's very impossible, leave it, but I think the new FormColumn and FormRow are pretty simple to use and would unify the design all over the UI

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)* status: 0 of 25 files reviewed, 3 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 105 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bvbU7Ddcq2Y5ofAwW:-N8xMClo4I5OBzR8X9tF:b6uzshz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/route/RoutePropertiesForm.tsx#L105)):*

well, w-80 is still just hardcoding the width for the input field, so it won't scale for bigger and/or smaller screens. If it's very impossible, leave it, but I think the new FormColumn and FormRow are pretty simple to use and would unify the design all over the UI
</blockquote></details>

In this case paired with the proper flex styles it does scale up and down, w-* styles are just for fields' relative widths.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)* status: 0 of 25 files reviewed, 8 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`ui/src/components/forms/common/ConfirmSaveForm.tsx` line 16 at r12](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N90qW0z9JFGKavrutQN:-N90qW0z9JFGKavrutQO:b-45l3nz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/921aa22ea9b611bb15c81341c004ab6146aa19a2/ui/src/components/forms/common/ConfirmSaveForm.tsx#L16)):*
> ```XML
>   validityStart: z
>     .string()
>     .min(1, REQUIRED_FIELD_ERROR_MESSAGE)
> ```

let's use that `requiredString` directly instead of copy-pasting the message here
___
*[`ui/src/components/forms/common/customZodSchemas.ts` line 13 at r12](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N90qmSa09LflNB1dUM6:-N90qmSa09LflNB1dUM7:b-3rag92) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/921aa22ea9b611bb15c81341c004ab6146aa19a2/ui/src/components/forms/common/customZodSchemas.ts#L13)):*
> ```XML
>  * automatically converts value to number resulting empty
>  * field having value NaN. The flag is required for zod to
>  * accept the value as a number.
> ```

thanks for the comment!
___
*[`ui/src/components/forms/common/customZodSchemas.ts` line 26 at r12](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N90tAr-EE_4O8M8Zs7I:-N90tAr-EE_4O8M8Zs7J:bu7ip4y) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/921aa22ea9b611bb15c81341c004ab6146aa19a2/ui/src/components/forms/common/customZodSchemas.ts#L26)):*
> ```XML
> });
> 
> export const localizedStringOptional = z.object({
> ```

there's no need to do this now, but just a wink: it's going to be more future-proof if these schemas are actually functions, like this:

```
export const requiredString = () => z.string().min(1, REQUIRED_FIELD_ERROR_MESSAGE);
```

it will allow us to later give some optional parameters if zod cannot be bent to our will in some case but need to go for `refine` or `transform`
___
*[`ui/src/components/forms/common/ValidationErrorList.tsx` line 46 at r12](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N90r4P4CK474WYuZgJo:-N90r4P4CK474WYuZgJp:b5lwnro) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/921aa22ea9b611bb15c81341c004ab6146aa19a2/ui/src/components/forms/common/ValidationErrorList.tsx#L46)):*
> ```XML
>     // invalid type error messages to 'required' and then map them
>     // to error messages here
>     if (message?.toLowerCase() === REQUIRED_FIELD_ERROR_MESSAGE) {
> ```

why do you need toLowerCase here?
___
*[`ui/src/components/forms/common/ValidationErrorList.tsx` line 56 at r12](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N90rBYjDELKVQUYUFwg:-N90rBYjDELKVQUYUFwh:b-9fq79r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/921aa22ea9b611bb15c81341c004ab6146aa19a2/ui/src/components/forms/common/ValidationErrorList.tsx#L56)):*
> ```XML
>         return t('formValidation.tooBig');
>       case 'custom':
>         return t(`formValidation.${message}`);
> ```

ouu, very nice idea! Mapping the validation error message like this...
___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 105 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N8bvbU7Ddcq2Y5ofAwW:-N90spS27nHobRiFWUux:b-apw4j7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fd39acf5f6f71b0abadaea7f57aee62fdf70c66b/ui/src/components/forms/route/RoutePropertiesForm.tsx#L105)):*

In this case paired with the proper flex styles it does scale up and down, w-* styles are just for fields' relative widths.
</blockquote></details>

but is there a special reason why this form should not follow the standard-ish layout components but use its own styling...?
___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 75 at r12](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N90rPwf4ahnq_YdSv6p:-N90rPwf4ahnq_YdSv6q:bjaajyw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/921aa22ea9b611bb15c81341c004ab6146aa19a2/ui/src/components/forms/route/RoutePropertiesForm.tsx#L75)):*
> ```XML
>           </Row>
>         )}
>         <Row className="mb-5 grid-cols-1 flex-wrap gap-y-5 md:gap-x-8">
> ```

Instead of copy-pasting the FormRow classes here, why didn't you use the FormRow component directly? Also, the grid-cols-1 class won't work without the base 'grid' class. It's also not compatible with the flex-wrap...
___
*[`ui/src/components/forms/stop/StopForm.tsx` line 31 at r12](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N90tbdQ4Tewd9qB8pm-:-N90tbdQ4Tewd9qB8pm0:b-bj1nm3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/921aa22ea9b611bb15c81341c004ab6146aa19a2/ui/src/components/forms/stop/StopForm.tsx#L31)):*
> ```XML
>   schema as confirmSaveFormSchema,
> } from '../common/ConfirmSaveForm';
> import { requiredNumber, requiredString } from '../common/customZodSchemas';
> ```

let's group the imports as `from '../common';`

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)* status: 0 of 25 files reviewed, 2 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 76 at r20](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N91E87-B1a4JrzzDAru:-N91E87-B1a4JrzzDArv:b-srpunw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9d7a4f7a0370da5b4687b13e99fe00052234c03d/ui/src/components/forms/route/RoutePropertiesForm.tsx#L76)):*
> ```XML
>         )}
>         <FormColumn>
>           <FormRow columns={3} className="sm:grid-cols-2">
> ```

I see you defined a custom styling for the `sm` case to put "label" and "direction" next to each other. It looks nice, but the gaps between these input fields are missing (`sm:gap-x-8`). Also you needed to define custom class names everywhere, might not be very future-proof. Wouldn't it be easier to implement this directly within the FormRow component? Something like this?

```
export const FormRow: React.FC<Props> = ({
  className = '',
  mdColumns = 1,
  smColumns = mdColumns,
  children,
}) => {
  const baseClassName = 'grid grid-cols-1 gap-y-5';
  const mdClassName = `md:grid-cols-${mdColumns} md:gap-x-8`;
  const smClassName = `sm:grid-cols-${smColumns} sm:gap-x-8`;
  return (
    <div className={`${className} ${baseClassName} ${mdClassName} ${smClassName}`}>
      {children}
    </div>
  );
};

```

Would this work?

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/281)* status: 0 of 25 files reviewed, 2 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 76 at r20](https://reviewable.io/reviews/hsldevcom/jore4-ui/281#-N91E87-B1a4JrzzDAru:-N91TVom48nskvulx1Kx:b-qe63cn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9d7a4f7a0370da5b4687b13e99fe00052234c03d/ui/src/components/forms/route/RoutePropertiesForm.tsx#L76)):*

I see you defined a custom styling for the `sm` case to put "label" and "direction" next to each other. It looks nice, but the gaps between these input fields are missing (`sm:gap-x-8`). Also you needed to define custom class names everywhere, might not be very future-proof. Wouldn't it be easier to implement this directly within the FormRow component? Something like this?

```
export const FormRow: React.FC<Props> = ({
  className = '',
  mdColumns = 1,
  smColumns = mdColumns,
  children,
}) => {
  const baseClassName = 'grid grid-cols-1 gap-y-5';
  const mdClassName = `md:grid-cols-${mdColumns} md:gap-x-8`;
  const smClassName = `sm:grid-cols-${smColumns} sm:gap-x-8`;
  return (
    <div className={`${className} ${baseClassName} ${mdClassName} ${smClassName}`}>
      {children}
    </div>
  );
};

```

Would this work?
</blockquote></details>

It would work, I will do that.

PR #289, opened 2022-08-08
Add DateTime get and set methods to useUrlQuery

HenrikHartiala:
Add and use DateTime getter and setter for urlQuery. Also add useMemo to queryParameters.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/289)

### DIFF
--- a/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
+++ b/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
@@ -1,3 +1,4 @@
+import { DateTime } from 'luxon';
 import qs from 'qs';
 import { useTranslation } from 'react-i18next';
 import { RouteLine } from '../../../generated/graphql';
@@ -29,7 +30,7 @@ export const ActionsRow = ({
     useObservationDateQueryParam();
 
   const onDateChange = (value: string) => {
-    setObservationDateToUrl(value);
+    setObservationDateToUrl(DateTime.fromISO(value));
   };
 
   return (
--- a/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
@@ -1,3 +1,4 @@
+import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
 import { useHistory, useParams } from 'react-router';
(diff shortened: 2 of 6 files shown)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/289)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala and @pallost)

___
*[`ui/src/hooks/urlQuery/useObservationDateQueryParam.ts` line 10 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/289#-N90PEwXDyrYOcX4wpiP:-N90PEwXDyrYOcX4wpiQ:b8f60wh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f84a40631699cb4896b9b3967dfeabd2ad521a5a/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts#L10)):*
> ```XML
>   // Memoize the actual value to prevent unnecessary updates
>   const observationDate = useMemo(
>     () => getDateTimeFromUrlQuery('observationDate'),
> ```

the getDateTimeFromUrlQuery function is already wrapped into useCallback. I don't think you need to use useMemo here, but I'm no expert...

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/289)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[`ui/src/hooks/urlQuery/useObservationDateQueryParam.ts` line 10 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/289#-N90PEwXDyrYOcX4wpiP:-N90S6XC4GD-9aXizPtF:b-2b3egh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f84a40631699cb4896b9b3967dfeabd2ad521a5a/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts#L10)):*

the getDateTimeFromUrlQuery function is already wrapped into useCallback. I don't think you need to use useMemo here, but I'm no expert...
</blockquote></details>

We do need the useMemo here as well. The useCallback is to prevent the the re-creation of the function between re-renders and useMemo is to prevent the call of the function between re-renders.

If you remove the useMemo here, it will cause an infinite loop.

This might not be the only solution to the problem, but at least for this iteration I think it is fine and it is commented as well.

PR #292, opened 2022-08-10
Fix stop list menu creating horizontal scrollbars to the list

eleksi:
Ideally this would not be fixed by changing the alignment, but playing with overlays proved to be quite difficult; Setting overflow-y to auto would not allow setting overflow-x to visible (https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue)

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/292)

### DIFF
--- a/ui/src/components/map/RouteStopsOverlay.tsx
+++ b/ui/src/components/map/RouteStopsOverlay.tsx
@@ -10,7 +10,11 @@ import {
   setRouteMetadataFormOpenAction,
   setStopOnRouteAction,
 } from '../../redux';
-import { EditButton, SimpleDropdownMenu } from '../../uiComponents';
+import {
+  AlignDirection,
+  EditButton,
+  SimpleDropdownMenu,
+} from '../../uiComponents';
 import { mapToVariables } from '../../utils';
 import { MapOverlay, MapOverlayHeader } from './MapOverlay';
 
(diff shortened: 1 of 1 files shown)


tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/292)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @eleksi and @tk-vincit)

___
*[`ui/src/components/map/RouteStopsOverlay.tsx` line 14 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/292#-N9AVJmM8vw7eqEemtbm:-N9AVJmM8vw7eqEemtbn:b4accdf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/54a3ae9932f9a07bf23078cad211db1197e5242b/ui/src/components/map/RouteStopsOverlay.tsx#L14)):*
> ```XML
> } from '../../redux';
> import {
>   AlignDirection,
> ```

Not part of this PR, but now I noticed that we have "AlignDirection" exposed from `uiComponents`, which makes no sense as it is not generic align direction, but should rather be something like "(Simple)Dropdown(Menu)AlignDirection".

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/292)* status: all files reviewed, 1 unresolved discussion (waiting on @eleksi)

PR #297, opened 2022-08-15
Journey pattern stops refactor

pallost:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/297)

Closes https://github.com/HSLdevcom/jore4/issues/876

### DIFF
--- a/codegen/schemaLoader.ts
+++ b/codegen/schemaLoader.ts
@@ -116,6 +116,11 @@ const hasuraOverrideTypeDefs = gql`
   type infrastructure_network_infrastructure_link {
     shape: geography_linestring! # fix the geography type
   }
+
+  interface validity_period {
+    validity_start: timestamptz
+    validity_end: timestamptz
+  }
 `;
 
 // Note: if want to make custom directives for the schema, here are some examples on how to do so:
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test": "yarn ws:ui test",
     "test:e2e": "yarn ws:db build && yarn ws:e2e cypress run --browser chrome",
     "test:e2e:open": "yarn ws:db build && yarn ws:e2e cypress open",
-    "qa": "yarn --frozen-lockfile && yarn ts:check && yarn prettier:check && yarn lint && yarn test && yarn test:e2e"
+    "qa": "yarn --frozen-lockfile && yarn ws:db build && yarn ts:check && yarn prettier:check && yarn lint && yarn test && yarn test:e2e"
   },
   "dependencies": {},
   "devDependencies": {
--- a/test-db-manager/src/generated/graphql.ts
+++ b/test-db-manager/src/generated/graphql.ts
@@ -6876,6 +6876,11 @@ export type UuidComparisonExp = {
   _nin?: Maybe<Array<Scalars['uuid']>>;
 };
 
+export type ValidityPeriod = {
+  validity_end?: Maybe<Scalars['timestamptz']>;
+  validity_start?: Maybe<Scalars['timestamptz']>;
+};
+
 export type InsertLinesMutationVariables = Exact<{
   objects: Array<RouteLineInsertInput> | RouteLineInsertInput;
 }>;
--- a/ui/graphql.schema.json
+++ b/ui/graphql.schema.json
@@ -41818,6 +41818,41 @@
         "interfaces": null,
         "enumValues": null,
         "possibleTypes": null
+      },
+      {
+        "kind": "INTERFACE",
+        "name": "validity_period",
+        "description": null,
+        "fields": [
+          {
+            "name": "validity_end",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "SCALAR",
+              "name": "timestamptz",
+              "ofType": null
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
+            "name": "validity_start",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "SCALAR",
+              "name": "timestamptz",
+              "ofType": null
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          }
+        ],
+        "inputFields": null,
+        "interfaces": [],
+        "enumValues": null,
+        "possibleTypes": []
       }
     ],
     "directives": [
--- a/ui/src/components/map/routes/DrawRouteLayer.tsx
+++ b/ui/src/components/map/routes/DrawRouteLayer.tsx
@@ -26,11 +26,16 @@ import {
   mapRouteToInfraLinksAlongRoute,
 } from '../../../graphql';
 import {
+  extractJourneyPatternCandidateStops,
+  filterDistinctConsecutiveRouteStops,
+  getOldRouteGeometryVariables,
   getRouteStops,
   LineStringFeature,
+  mapInfraLinksToFeature,
   useAppDispatch,
   useAppSelector,
   useExtractRouteFromFeature,
+  useFilterStops,
   useLoader,
 } from '../../../hooks';
 import {
@@ -42,7 +47,8 @@ import {
   setDraftRouteGeometryAction,
   stopRouteEditingAction,
 } from '../../../redux';
-import { showToast } from '../../../utils';
+import { parseDate } from '../../../time';
+import { log, showToast } from '../../../utils';
 import { addRoute, removeRoute } from '../../../utils/map';
 import { featureStyle, handleStyle } from './editorStyles';
 
@@ -83,14 +89,9 @@ const DrawRouteLayerComponent = (
 
   const [selectedSnapPoints, setSelectedSnapPoints] = useState<number[]>([]);
 
-  const {
-    extractScheduledStopPoints,
-    getInfraLinksWithStopsForGeometry,
-    mapInfraLinksToFeature,
-    getRemovedStopLabels,
-    getOldRouteGeometryVariables,
-    filterDistinctConsecutiveRouteStops,
-  } = useExtractRouteFromFeature();
+  const { getInfraLinksWithStopsForGeometry, getRemovedStopLabels } =
+    useExtractRouteFromFeature();
+  const { filter } = useFilterStops();
 
   const { setIsLoading } = useLoader(Operation.MatchRoute);
 
@@ -133,18 +134,22 @@ const DrawRouteLayerComponent = (
       // we are editing an existing or a template route, but haven't (yet) received the graphql
       // response with its data -> return early
       if (!baseRoute && !creatingNewRoute) {
-        // eslint-disable-next-line no-console
-        console.debug(
+        log.error(
           'Trying to edit an existing route but could not find a base route (yet)',
         );
         return;
       }
 
-      // check if we already have the line info available or is it yet to come
-      if (!editedRouteData?.lineInfo) {
-        // eslint-disable-next-line no-console
-        console.debug(
-          'Trying to update route geometry but line info is not (yet) available',
+      // we can only find the stops belonging to the route if its metadata is available
+      // (when editing, fetch it from graphql; when creating, filled in through form)
+      if (
+        !editedRouteData.metaData ||
+        !editedRouteData.metaData.validityStart ||
+        !editedRouteData.metaData.validityEnd ||
+        !editedRouteData.metaData.priority
+      ) {
+        log.error(
+          'Trying to update route geometry but route metadata is not (yet) available',
         );
         return;
       }
@@ -172,14 +177,23 @@ const DrawRouteLayerComponent = (
         oldStopLabels,
       );
 
-      // Extract list of the stops to be included in the route
-      const stops = extractScheduledStopPoints(
-        infraLinksWithStops,
-        editedRouteData.lineInfo.primary_vehicle_mode,
+      // Extract list of the stops that are eligible to be included in route's journey pattern
+      const routeMetadata = {
+        validity_start: parseDate(editedRouteData.metaData.validityStart),
+        validity_end: parseDate(editedRouteData.metaData.validityEnd),
+        priority: editedRouteData.metaData.priority,
+      };
+      const stopsEligibleForJourneyPattern =
+        extractJourneyPatternCandidateStops(infraLinksWithStops, routeMetadata);
+
+      // FIXME. We shouldn't filter stops by being visible on the map as it might rule
+      // out some of the possibly eligible stops from the journey pattern.
+      const visibleStopsEligibleForJourneyPattern = filter(
+        stopsEligibleForJourneyPattern,
       );
 
       const routeStops = getRouteStops(
-        stops,
+        visibleStopsEligibleForJourneyPattern,
         removedStopLabels || [],
         editedRouteData?.id,
       );
@@ -187,6 +201,7 @@ const DrawRouteLayerComponent = (
       dispatch(
         setDraftRouteGeometryAction({
           stops: filterDistinctConsecutiveRouteStops(routeStops),
+          stopsEligibleForJourneyPattern,
           infraLinks: infraLinksWithStops,
         }),
       );
@@ -206,12 +221,10 @@ const DrawRouteLayerComponent = (
       dispatch,
       editedRouteData?.id,
       editedRouteData.infraLinks,
-      editedRouteData.lineInfo,
+      editedRouteData.metaData,
       editedRouteData.stops,
-      extractScheduledStopPoints,
-      filterDistinctConsecutiveRouteStops,
+      filter,
       getInfraLinksWithStopsForGeometry,
-      getOldRouteGeometryVariables,
       getRemovedStopLabels,
       map,
       onDelete,
@@ -239,7 +252,6 @@ const DrawRouteLayerComponent = (
       setSnappingLine(undefined);
     }
   }, [
-    mapInfraLinksToFeature,
     onUpdateRouteGeometry,
     mode,
     creatingNewRoute,
--- a/ui/src/generated/graphql.tsx
+++ b/ui/src/generated/graphql.tsx
@@ -6880,6 +6880,11 @@ export type UuidComparisonExp = {
   _nin?: Maybe<Array<Scalars['uuid']>>;
 };
 
+export type ValidityPeriod = {
+  validity_end?: Maybe<Scalars['timestamptz']>;
+  validity_start?: Maybe<Scalars['timestamptz']>;
+};
+
 export type StopPopupInfoFragment = {
   __typename?: 'service_pattern_scheduled_stop_point';
   label: string;
@@ -7244,6 +7249,13 @@ export type LineAllFieldsFragment = {
   validity_end?: luxon.DateTime | null | undefined;
   priority: number;
   label: string;
+};
+
+export type RouteValidityFragment = {
+  __typename?: 'route_route';
+  validity_start?: luxon.DateTime | null | undefined;
+  validity_end?: luxon.DateTime | null | undefined;
+  priority: number;
 };
 
 export type RouteAllFieldsFragment = {
@@ -9110,6 +9122,13 @@ export const LineDefaultFieldsFragmentDoc = gql`
     short_name_i18n
     validity_start
     validity_end
+  }
+`;
+export const RouteValidityFragmentDoc = gql`
+  fragment route_validity on route_route {
+    validity_start
+    validity_end
+    priority
   }
 `;
 export const RouteDefaultFieldsFragmentDoc = gql`
--- a/ui/src/graphql/infrastructureNetwork.ts
+++ b/ui/src/graphql/infrastructureNetwork.ts
@@ -5,8 +5,10 @@ import {
   InfraLinkMatchingFieldsFragment,
   InfrastructureNetworkDirectionEnum,
   InfrastructureNetworkInfrastructureLink,
+  InfrastructureNetworkInfrastructureLinkInsertInput,
   QueryClosestLinkQuery,
   QueryPointDirectionOnLinkQuery,
+  ReusableComponentsVehicleSubmodeEnum,
   RouteInfraLinkFieldsFragment,
   RouteRoute,
 } from '../generated/graphql';
@@ -60,6 +62,16 @@ export const orderInfraLinksByExternalLinkId = <
 
     return infraLinkWithStop;
   });
+
+const input: InfrastructureNetworkInfrastructureLinkInsertInput = {
+  vehicle_submode_on_infrastructure_link: {
+    data: [
+      {
+        vehicle_submode: ReusableComponentsVehicleSubmodeEnum.GenericBus,
+      },
+    ],
+  },
+};
 
 const INFRASTRUCTURE_LINK_ALL_FIELDS = gql`
   fragment infrastructure_link_all_fields on infrastructure_network_infrastructure_link {
--- a/ui/src/graphql/route.ts
+++ b/ui/src/graphql/route.ts
@@ -41,6 +41,14 @@ const LINE_ALL_FIELDS = gql`
   }
 `;
 
+const ROUTE_VALIDITY = gql`
+  fragment route_validity on route_route {
+    validity_start
+    validity_end
+    priority
+  }
+`;
+
 const ROUTE_ALL_FIELDS = gql`
   fragment route_all_fields on route_route {
     route_id
--- a/ui/src/hooks/lines/useValidateLine.ts
+++ b/ui/src/hooks/lines/useValidateLine.ts
@@ -4,7 +4,7 @@ import {
   useGetLineDetailsWithRoutesByIdAsyncQuery,
 } from '../../generated/graphql';
 import { mapLineDetailsWithRoutesResult } from '../../graphql';
-import { useValidateRoute } from '../routes';
+import { useValidateRoute } from '../routes/useValidateRoute';
 
 interface ValidateParams {
   lineId: UUID;
--- a/ui/src/hooks/routes/useEditRouteJourneyPattern.ts
+++ b/ui/src/hooks/routes/useEditRouteJourneyPattern.ts
@@ -11,7 +11,7 @@ import {
   stopBelongsToJourneyPattern,
 } from '../../graphql';
 import { removeFromApolloCache } from '../../utils';
-import { useExtractRouteFromFeature } from './useExtractRouteFromFeature';
+import { filterDistinctConsecutiveRouteStops } from './useExtractRouteFromFeature';
 import { useValidateRoute } from './useValidateRoute';
 
 interface DeleteStopParams {
@@ -33,7 +33,6 @@ export const useEditRouteJourneyPattern = () => {
   const [updateRouteJourneyPatternMutation] =
     useUpdateRouteJourneyPatternMutation();
   const { validateStopCount } = useValidateRoute();
-  const { filterDistinctConsecutiveRouteStops } = useExtractRouteFromFeature();
 
   const prepareUpdateJourneyPattern = (
     params: AddStopParams | DeleteStopParams,
--- a/ui/src/hooks/routes/useExtractRouteFromFeature.ts
+++ b/ui/src/hooks/routes/useExtractRouteFromFeature.ts
@@ -3,8 +3,10 @@ import { useCallback } from 'react';
 import { getBusRoute } from '../../api/routing';
 import {
   InfrastructureNetworkDirectionEnum,
-  ReusableComponentsVehicleModeEnum,
   RouteRoute,
+  RouteStopFieldsFragment,
+  RouteValidityFragment,
+  ScheduledStopPointDefaultFieldsFragment,
   ServicePatternScheduledStopPoint,
   StopWithJourneyPatternFieldsFragment,
   useGetLinksWithStopsByExternalLinkIdsAsyncQuery,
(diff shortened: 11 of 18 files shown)


HenrikHartiala (changes requested):
> Reviewed 6 of 8 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/297)* status: all files reviewed, 12 unresolved discussions (waiting on @pallost)

___
*[`ui/src/time.ts` line 90 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fDEbxELXCSSY6zpOd:-N9fDEbxELXCSSY6zpOe:b-qbhg55) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/time.ts#L90)):*
> ```XML
>   );
> 
> export const doValiditiesOverlap = (
> ```

Hmm, this `do` prefix might be a bit confusing because we have functions in the code named `doEditStop(..)` which implies that some action is taken. But here it is a check returning true or false. Not sure which would be better naming. Maybe `areValiditiesOverlapping` or something.
Maybe this is no issue, and maybe actually the `doEditStop()` - kind of function names should rather be changed. 
What do you think?
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 153 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f-Fbf-BSspWTONCvH:-N9f-Fbf-BSspWTONCvI:b-yi2blm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L153)):*
> ```XML
>       }
> 
>       // we can only find the stops belonging to the route its metadata is available
> ```

"if* its metadata is available"
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 161 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f-ReI5jeI1YHiPWrv:-N9f-ReI5jeI1YHiPWrw:b-5c1992) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L161)):*
> ```XML
>         !editedRouteData.metaData.priority
>       ) {
>         // eslint-disable-next-line no-console
> ```

We already disable a lot of es-lint warnings. If we are to use this kind of debugging, could we at least create some debug function and then use that for debug logging and only have this es-lint disable there?
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 191 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f2jJF8eMo9V27ZQlY:-N9f2jJF8eMo9V27ZQlZ:b-uvpi9p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L191)):*
> ```XML
>       );
> 
>       // Extract list of the stops that are eligible to be included in route's journey pattern
> ```

I was about to say that this comment should be on top of the function call, rather than this initialization. And then was about to say that this should actually be as JSDoc comment on top of the extractJourneyPatternCandidateStops inside `useExtractRouteFromFeature.ts` but it is already there with nice documentation.

So this comment can be removed completely.
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 203 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f5lHGDS5EzYbpAVkw:-N9f5lHGDS5EzYbpAVkx:bgx15n1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L203)):*
> ```XML
>       );
> 
>       // FIXME. We shouldn't filter stops by being visible on the map as it might rule
> ```

First: does this `FIXME` have some other purpose than `TODO` which we have used a bit more? If not then we should only use one of those.

Second: So why is the stops filtered by visibility on the map? I think the route stop list should always be the same regardless of what we see on map. 

Now if you would like to include all the stops along the route, even the ones that's validityStart is in the future, you would need to change the filters to include all the stops. But I'm also not completely sure how the end users would use this.

I think we had a discussion about this earlier this week, but not sure what we concluded anything, maybe this is something that still needs a bit of thought. 
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 205 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f9xLqCMffe3mODonP:-N9f9xLqCMffe3mODonQ:b-gwmcsj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L205)):*
> ```XML
>       // FIXME. We shouldn't filter stops by being visible on the map as it might rule
>       // out some of the possibly eligible stops from the journey pattern.
>       const visibleJourneyPatternCandidateStops = filter(
> ```

This `filter(..)` function could use a more descriptive name (`filterVisibleStopsByMapFilters` or `filterVisibleStopsByMapStopFilters` etc.) and maybe a JSDoc explanation. 
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 218 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fBwf11kVBTFDgT4Rb:-N9fBwf11kVBTFDgT4Rc:bpkfvoe) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L218)):*
> ```XML
>         setDraftRouteGeometryAction({
>           stops: filterDistinctConsecutiveRouteStops(routeStops),
>           stopsEligibleForJourneyPattern: journeyPatternCandidateStops,
> ```

Are these the same thing? If so, maybe we should use only one. I think either is fine though.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 72 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fJYkXAkMBWDKWmNB5:-N9fJYkXAkMBWDKWmNB6:b-83sjoh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L72)):*
> ```XML
>     .map((item) => item.label);
> 
> /**
> ```

Nice!
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 175 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fMCv2-Z-JzmL3GOue:-N9fMCv2-Z-JzmL3GOuf:b-2f3q0r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L175)):*
> ```XML
> ) => stops.filter((stop, index) => stops[index - 1]?.label !== stop.label);
> 
> /**
> ```

Nice documentation!
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 197 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fNjEH0-_kYvwyu1xe:-N9fNjEH0-_kYvwyu1xf:boe87ed) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L197)):*
> ```XML
>   infraLinksWithStops: RouteInfraLink[],
>   routeMetadata: RouteValidityFragment,
>   lineMetadata: Pick<RouteLine, 'primary_vehicle_mode'>,
> ```

Is there a benefit of calling this "lineMetadata" and then picking just one type, rather than just calling this `vehicleMode` with the type of `ReusableComponentsVehicleMode`?
___
*[`ui/src/hooks/stops/useFilterStops.ts` line 105 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fgjk_6PttYXzM6_Sz:-N9fgjk_6PttYXzM6_T-:b-u83g28) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/stops/useFilterStops.ts#L105)):*
> ```XML
>   // Pick stop instance with the highest priority for each stop label
>   return Object.values(groupedCurrentStops).flatMap(
>     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
> ```

Could this be handled with a null check or some other way than non-null-assertion? Or at least add a comment why this assertion can be done.
___
*[`ui/src/redux/slices/mapEditor.ts` line 58 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fhcUz2q7x8UqdgzEb:-N9fhcUz2q7x8UqdgzEc:brtqy4h) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/redux/slices/mapEditor.ts#L58)):*
> ```XML
>      * Array of stops that are eligible to be added to the journey pattern
>      *
>      * TODO: should start using this instead of the "stops" variable above
> ```

Good explanation :+1:

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/297)* status: 8 of 10 files reviewed, 12 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/time.ts` line 90 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fDEbxELXCSSY6zpOd:-NADhYoJ4ueRyFxk09Hj:byy5pdx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/time.ts#L90)):*

Hmm, this `do` prefix might be a bit confusing because we have functions in the code named `doEditStop(..)` which implies that some action is taken. But here it is a check returning true or false. Not sure which would be better naming. Maybe `areValiditiesOverlapping` or something.
Maybe this is no issue, and maybe actually the `doEditStop()` - kind of function names should rather be changed. 
What do you think?
</blockquote></details>

good point. Renamed it to `areValidityPeriodsOverlapping`
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 153 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f-Fbf-BSspWTONCvH:-NADkmQ_9HZAghwVn15r:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L153)):*

"if* its metadata is available"
</blockquote></details>

Done.



_Code quote:_
```XML
 we can only find the stops belonging to the route its metadata is available
```
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 161 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f-ReI5jeI1YHiPWrv:-NADmBz8-3IabNNjIV6e:b-9ncllj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L161)):*

We already disable a lot of es-lint warnings. If we are to use this kind of debugging, could we at least create some debug function and then use that for debug logging and only have this es-lint disable there?
</blockquote></details>

well, this is a bit of a weird place here... this function needs horribly lots of dependencies and I couldn't find a reliable way for tracking all of them down. Many of the required variables are only bound later on, which is fine, that's not an error. But in some weird cases they are never bound, which is not fine and I'd like to see that from some log message at least. The optimal solution would be to have an actual logger set up, but it still hasn't been scheduled for a sprint... 
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 191 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f2jJF8eMo9V27ZQlY:-NADzdVy7w3MyQgmTzss:b-n82nma) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L191)):*

I was about to say that this comment should be on top of the function call, rather than this initialization. And then was about to say that this should actually be as JSDoc comment on top of the extractJourneyPatternCandidateStops inside `useExtractRouteFromFeature.ts` but it is already there with nice documentation.

So this comment can be removed completely.
</blockquote></details>

I disagree. The comment is sitting on top a code block, explaining what happens there, so that it's easier to follow the flow of what's happening in this long and messy function. The routeMetadata variable is scoped only for this block, so that also belongs under the comment
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 203 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f5lHGDS5EzYbpAVkw:-NADztg75EyS7bSU8THe:b-stnjyi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L203)):*

First: does this `FIXME` have some other purpose than `TODO` which we have used a bit more? If not then we should only use one of those.

Second: So why is the stops filtered by visibility on the map? I think the route stop list should always be the same regardless of what we see on map. 

Now if you would like to include all the stops along the route, even the ones that's validityStart is in the future, you would need to change the filters to include all the stops. But I'm also not completely sure how the end users would use this.

I think we had a discussion about this earlier this week, but not sure what we concluded anything, maybe this is something that still needs a bit of thought. 
</blockquote></details>

FIXME means that it's a known bug and should be fixed soon. TODO only means that something should be done in the future, does not indicate that it's a bug. This is a bug and should be fixed soon, that's why I used FIXME. I don't think we should limit the use to only one of them.

The stops have been filtered by visibility on the map, because we used to think that that is how it should work. However, this filtering should be done on the map and on the overlays, not in the datamodel as it was done here previously. However this needs more thinking and more refactoring that is not intended to be in the scope of this PR. That's why I created the stopsEligibleForJourneyPattern variable in the redux store that we would start taking into use and do post-filtering on the UI.
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 205 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f9xLqCMffe3mODonP:-NADrkETDHEnRBH1CiXZ:b-r48vre) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L205)):*

This `filter(..)` function could use a more descriptive name (`filterVisibleStopsByMapFilters` or `filterVisibleStopsByMapStopFilters` etc.) and maybe a JSDoc explanation. 
</blockquote></details>

I am anyway against any functions that start with the word `filter`. It should look rather like this to be more reusable:
`const visibleJourneyPatternCandidateStops = journeyPatternCandidateStops.filter(isStopVisibleOnMap)`
But refactoring this is not in the scope of this PR.
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 218 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fBwf11kVBTFDgT4Rb:-NADsHdPEO67oq1eKbgQ:b-am7kdp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L218)):*

Are these the same thing? If so, maybe we should use only one. I think either is fine though.
</blockquote></details>

they are not exactly the same thing. The `stops` is the compatibility version which is filtered by what is visible on the UI. The `stopsEligibleForJourneyPattern` is the actual thing that should be used in the redux data model and then post-filtered for the different UI components based on their own needs.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 72 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fJYkXAkMBWDKWmNB5:-NADuhENElllxoRNhykQ:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L72)):*

Nice!
</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 175 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fMCv2-Z-JzmL3GOue:-NADuhrP8cxa5sW2GYhR:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L175)):*

Nice documentation!
</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 197 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fNjEH0-_kYvwyu1xe:-NADune54speMtWXoN_m:b66rpst) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L197)):*

Is there a benefit of calling this "lineMetadata" and then picking just one type, rather than just calling this `vehicleMode` with the type of `ReusableComponentsVehicleMode`?
</blockquote></details>

yes, we might also need other metadata fields from the line in the future. I don't know yet how e.g. the line's priority and/or validity period may affect the stops that can belong to a journey pattern
___
*[`ui/src/hooks/stops/useFilterStops.ts` line 105 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fgjk_6PttYXzM6_Sz:-NADzEzcBX31xc08NPeG:bwntp43) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/stops/useFilterStops.ts#L105)):*

Could this be handled with a null check or some other way than non-null-assertion? Or at least add a comment why this assertion can be done.
</blockquote></details>

refactored this function to use a functional programming flow. Hopefully this is clearer now
___
*[`ui/src/redux/slices/mapEditor.ts` line 58 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fhcUz2q7x8UqdgzEb:-NADzTFWEiH4tpe0Ix4C:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/redux/slices/mapEditor.ts#L58)):*

Good explanation :+1: 
</blockquote></details>

Done.

HenrikHartiala (changes requested):
> Reviewed 5 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/297)* status: 9 of 15 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala and @pallost)

___
*[`codegen/schemaLoader.ts` line 118 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-NAJa1I6EoPdYShcrAB9:-NAJa1I6EoPdYShcrABA:blll6lx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d7b2a385e86d4c5b6a70e8edf99d7cded1ed56dc/codegen/schemaLoader.ts#L118)):*
> ```XML
>   }
> 
>   interface validity_period {
> ```

All the other definitions here uses `type` instead of `interface`. Is this intentional?
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 161 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f-ReI5jeI1YHiPWrv:-NAJakCh-o73Mf1ZBacc:blqxxn6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L161)):*

well, this is a bit of a weird place here... this function needs horribly lots of dependencies and I couldn't find a reliable way for tracking all of them down. Many of the required variables are only bound later on, which is fine, that's not an error. But in some weird cases they are never bound, which is not fine and I'd like to see that from some log message at least. The optimal solution would be to have an actual logger set up, but it still hasn't been scheduled for a sprint... 
</blockquote></details>

Yeah, that is true and would also like some actual logger. The logger should also be easier to set up if we already use even this kind of own logging for the time being. Thanks!
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 191 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f2jJF8eMo9V27ZQlY:-NAJbTHKA3HalpcmL1wE:be0b4qw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L191)):*

I disagree. The comment is sitting on top a code block, explaining what happens there, so that it's easier to follow the flow of what's happening in this long and messy function. The routeMetadata variable is scoped only for this block, so that also belongs under the comment
</blockquote></details>

Okay, I see. If it helps the readability then it is fine :+1: 
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 203 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f5lHGDS5EzYbpAVkw:-NAJbhpq9WzwGtoRywuZ:bygpasw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L203)):*

FIXME means that it's a known bug and should be fixed soon. TODO only means that something should be done in the future, does not indicate that it's a bug. This is a bug and should be fixed soon, that's why I used FIXME. I don't think we should limit the use to only one of them.

The stops have been filtered by visibility on the map, because we used to think that that is how it should work. However, this filtering should be done on the map and on the overlays, not in the datamodel as it was done here previously. However this needs more thinking and more refactoring that is not intended to be in the scope of this PR. That's why I created the stopsEligibleForJourneyPattern variable in the redux store that we would start taking into use and do post-filtering on the UI.

</blockquote></details>

Thanks for explanation.
True, not in the scope of this PR so this is OK :+1: 
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 205 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f9xLqCMffe3mODonP:-NAJc8L2CFzSpqCVQ5KB:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L205)):*

I am anyway against any functions that start with the word `filter`. It should look rather like this to be more reusable:
`const visibleJourneyPatternCandidateStops = journeyPatternCandidateStops.filter(isStopVisibleOnMap)`
But refactoring this is not in the scope of this PR.
</blockquote></details>

:+1: 
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 218 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fBwf11kVBTFDgT4Rb:-NAJcRM62iJU9SmNyaxm:b-toai30) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L218)):*

they are not exactly the same thing. The `stops` is the compatibility version which is filtered by what is visible on the UI. The `stopsEligibleForJourneyPattern` is the actual thing that should be used in the redux data model and then post-filtered for the different UI components based on their own needs.
</blockquote></details>

Ah, sorry, I meant that are `stopsEligibleForJourneyPattern` and `journeyPatternCandidateStops` the same thing? Or do we need to rename this for some other particular reason?
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 197 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fNjEH0-_kYvwyu1xe:-NAJd0l-4o3EHoQEp9p6:b-rcqfvl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L197)):*

yes, we might also need other metadata fields from the line in the future. I don't know yet how e.g. the line's priority and/or validity period may affect the stops that can belong to a journey pattern
</blockquote></details>

Okay :+1: 
___
*[`ui/src/hooks/stops/useFilterStops.ts` line 105 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fgjk_6PttYXzM6_Sz:-NAJdIMa8oKFnYZrGGEe:b-ucfknh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/stops/useFilterStops.ts#L105)):*

refactored this function to use a functional programming flow. Hopefully this is clearer now
</blockquote></details>

Thanks!

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/297)* status: 0 of 16 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`codegen/schemaLoader.ts` line 118 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-NAJa1I6EoPdYShcrAB9:-NAntRIwDssnyFe1qr6a:b-zf322i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d7b2a385e86d4c5b6a70e8edf99d7cded1ed56dc/codegen/schemaLoader.ts#L118)):*

All the other definitions here uses `type` instead of `interface`. Is this intentional?
</blockquote></details>

yes :) interfaces can be extended, "used" by types. Here the idea is that lines, routes, stops all have validity_start and validity_end fields, thus this is not a type by itself but an interface
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 161 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9f-ReI5jeI1YHiPWrv:-NAnsyEU3IhVyeeKcZoC:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L161)):*

Yeah, that is true and would also like some actual logger. The logger should also be easier to set up if we already use even this kind of own logging for the time being. Thanks!
</blockquote></details>

Done.
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 218 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fBwf11kVBTFDgT4Rb:-NAntIPa1qTR3kIOLDuq:bh80gkx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L218)):*
> journeyPatternCandidateStops
good point, combined them
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 197 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fNjEH0-_kYvwyu1xe:-NAntLB1-NSE2CjdUpiE:b7fogo4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L197)):*

Okay :+1: 
</blockquote></details>

actually, removed the whole check. Added a few lines about the reason

HenrikHartiala (approved):
> Reviewed 16 of 16 files at r4, 15 of 15 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/297)* status: all files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[`codegen/schemaLoader.ts` line 118 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-NAJa1I6EoPdYShcrAB9:-NBHuIP-61wRVGTCbltS:b-1qdx8w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/d7b2a385e86d4c5b6a70e8edf99d7cded1ed56dc/codegen/schemaLoader.ts#L118)):*

yes :) interfaces can be extended, "used" by types. Here the idea is that lines, routes, stops all have validity_start and validity_end fields, thus this is not a type by itself but an interface
</blockquote></details>

Okay, nice :+1: 
___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 218 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/297#-N9fBwf11kVBTFDgT4Rb:-NBHuWXKDT35_cXNDhw1:b-ucfknh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/18c73e02542039b3822baf2306fc12d2e1b3cac2/ui/src/components/map/routes/DrawRouteLayer.tsx#L218)):*

> journeyPatternCandidateStops
good point, combined them
</blockquote></details>

Thanks!

PR #303, opened 2022-08-17
Fix line search results marker button

eleksi:
Resolves HSLdevcom/jore4#915

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/303)

### DIFF
--- a/ui/src/components/routes-and-lines/main/LineTableRow.tsx
+++ b/ui/src/components/routes-and-lines/main/LineTableRow.tsx
@@ -1,6 +1,7 @@
+import { gql } from '@apollo/client';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
-import { RouteLine } from '../../../generated/graphql';
+import { LineTableRowFragment } from '../../../generated/graphql';
 import { useAlertsAndHighLights, useShowRoutesOnModal } from '../../../hooks';
 import { Column, Row } from '../../../layoutComponents';
 import { Path, routeDetails } from '../../../router/routeDetails';
@@ -9,8 +10,17 @@ import { LocatorButton } from '../../../uiComponents';
 
 interface Props {
   className?: string;
-  line: RouteLine;
+  line: LineTableRowFragment;
 }
+
+const GQL_LINE_TABLE_ROW = gql`
+  fragment line_table_row on route_line {
+    ...line_all_fields
+    line_routes {
+      route_id
+    }
+  }
+`;
 
 export const LineTableRow = ({ className = '', line }: Props): JSX.Element => {
   const { t } = useTranslation();
--- a/ui/src/components/routes-and-lines/main/LinesList.tsx
+++ b/ui/src/components/routes-and-lines/main/LinesList.tsx
@@ -1,15 +1,14 @@
-import React from 'react';
-import { RouteLine } from '../../../generated/graphql';
+import { LineTableRowFragment } from '../../../generated/graphql';
 import { LineTableRow } from './LineTableRow';
 import { RoutesTable } from './RoutesTable';
 
 type Props = {
-  lines?: RouteLine[];
+  lines?: LineTableRowFragment[];
 };
 
 export const LinesList = ({ lines }: Props): JSX.Element => (
   <RoutesTable>
-    {lines?.map((item: RouteLine) => (
+    {lines?.map((item: LineTableRowFragment) => (
       <LineTableRow key={item.line_id} line={item} />
     ))}
   </RoutesTable>
--- a/ui/src/components/routes-and-lines/main/RoutesList.tsx
+++ b/ui/src/components/routes-and-lines/main/RoutesList.tsx
@@ -1,15 +1,14 @@
-import React from 'react';
-import { RouteRoute } from '../../../generated/graphql';
+import { RouteAllFieldsFragment } from '../../../generated/graphql';
 import { RoutesTable } from './RoutesTable';
 import { RoutesTableRow } from './RoutesTableRow';
 
 type Props = {
-  routes?: RouteRoute[];
+  routes?: RouteAllFieldsFragment[];
 };
 
 export const RoutesList = ({ routes }: Props): JSX.Element => (
   <RoutesTable>
-    {routes?.map((item: RouteRoute) => (
+    {routes?.map((item: RouteAllFieldsFragment) => (
       <RoutesTableRow key={item.route_id} route={item} />
     ))}
   </RoutesTable>
--- a/ui/src/components/routes-and-lines/main/RoutesTableRow.tsx
+++ b/ui/src/components/routes-and-lines/main/RoutesTableRow.tsx
@@ -1,6 +1,6 @@
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
-import { RouteRoute } from '../../../generated/graphql';
+import { RouteAllFieldsFragment } from '../../../generated/graphql';
 import { useAlertsAndHighLights, useShowRoutesOnModal } from '../../../hooks';
 import { Column, Row } from '../../../layoutComponents';
 import { Path, routeDetails } from '../../../router/routeDetails';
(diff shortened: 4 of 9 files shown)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/303)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @eleksi and @pallost)

___
*[`ui/src/hooks/search/useSearchResults.ts` line 32 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jirr0DSKkAoD76Rco:-N9jirr0DSKkAoD76Rcp:b2abkf2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/hooks/search/useSearchResults.ts#L32)):*
> ```XML
>   result: ReturnType<typeof useSearchLinesAndRoutesQuery>,
> ) => ({
>   lines: (result.data?.route_line || []) as LineTableRowFragment[],
> ```

you should not need this mapper anymore, just use the result.data directly in the consumers

HenrikHartiala (changes requested):
> Reviewed 7 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/303)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @eleksi, @HenrikHartiala, and @pallost)

___
*[`ui/src/components/routes-and-lines/main/LineTableRow.tsx` line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jgLxcEyBwWRy0C-CD:-N9jgLxcEyBwWRy0C-CE:bd3zj5o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/components/routes-and-lines/main/LineTableRow.tsx#L17)):*
> ```XML
> 
> const GQL_LINE_TABLE_ROW = gql`
>   fragment line_table_row on route_line {
> ```

I think this should be named `fragment line_table_row_fields` to match the other fragment names
___
*[`ui/src/hooks/search/useSearchResults.ts` line 38 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jhX0M2g-V427LqZ-_:-N9jhX0M2g-V427LqZ-a:b-4xr1vw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/hooks/search/useSearchResults.ts#L38)):*
> ```XML
> export const useSearchResults = (): {
>   lines: LineTableRowFragment[];
>   routes: RouteAllFieldsFragment[];
> ```

Should we have fragment for RouteTableRow as well? (btw, just realised that the component should probably be called `RouteTableRow` instead of `RoutesTableRow` (plural)).

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/303)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala and @pallost)

___
*[`ui/src/components/routes-and-lines/main/LineTableRow.tsx` line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jgLxcEyBwWRy0C-CD:-N9jqL4gBThBYUOshMZ0:biuvijs) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/components/routes-and-lines/main/LineTableRow.tsx#L17)):*

I think this should be named `fragment line_table_row_fields` to match the other fragment names
</blockquote></details>

Not all the fragments are named like that e.g. (`route_with_journey_pattern_stops`). I think the fields postfix is there to avoid having name like `route_all` when we literally mean route with all fields.

In my opinion I don't think we have to add "fields" to all fragments, as fragment-postfix is already added to typescript types and when using these inside graphql it is clear that those are fragments that contain fields.
___
*[`ui/src/hooks/search/useSearchResults.ts` line 32 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jirr0DSKkAoD76Rco:-N9jr9fh0Ipjh5AChGyG:b-uy4i1k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/hooks/search/useSearchResults.ts#L32)):*

you should not need this mapper anymore, just use the result.data directly in the consumers
</blockquote></details>

Yep maybe it is cleaner. Fixed.
___
*[`ui/src/hooks/search/useSearchResults.ts` line 38 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jhX0M2g-V427LqZ-_:-N9jrG3y4xRo8BCBPyHN:b-nown5r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/hooks/search/useSearchResults.ts#L38)):*

Should we have fragment for RouteTableRow as well? (btw, just realised that the component should probably be called `RouteTableRow` instead of `RoutesTableRow` (plural)).
</blockquote></details>

I was thinking of this as well, but now that it is same as `RouteAllFieldsFragment` I'm wondering if we really want to make a copy of that. 

`LineTableRowFragment` also has a dependency to `LineAllFieldsFragment`, it could be a bad thing but do we really want to start copying every single line and route field these types... Where is the optimal point..?

HenrikHartiala (approved):
Nice!

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-:-N9k0yyl8miKyoVrw8Bs:b-83sjoh)* status: 7 of 9 files reviewed, 2 unresolved discussions (waiting on @eleksi, @HenrikHartiala, and @pallost)

___
*[`ui/src/components/routes-and-lines/main/LineTableRow.tsx` line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jgLxcEyBwWRy0C-CD:-N9jxth_3GRepiwApdlP:b5saakz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/components/routes-and-lines/main/LineTableRow.tsx#L17)):*

Not all the fragments are named like that e.g. (`route_with_journey_pattern_stops`). I think the fields postfix is there to avoid having name like `route_all` when we literally mean route with all fields.

In my opinion I don't think we have to add "fields" to all fragments, as fragment-postfix is already added to typescript types and when using these inside graphql it is clear that those are fragments that contain fields.

</blockquote></details>

Ah, my skim search did not spot that there were also fragments without the `fields` postfix. Also good point about the fragment-postfix. This is good then :+1: 
___
*[`ui/src/hooks/search/useSearchResults.ts` line 38 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/303#-N9jhX0M2g-V427LqZ-_:-N9jyZQZ9iEYJpNDWrd9:b-f1ys4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/896cdc9575d3b7c4bd7d812aad6ad39f4afc53ef/ui/src/hooks/search/useSearchResults.ts#L38)):*

I was thinking of this as well, but now that it is same as `RouteAllFieldsFragment` I'm wondering if we really want to make a copy of that. 

`LineTableRowFragment` also has a dependency to `LineAllFieldsFragment`, it could be a bad thing but do we really want to start copying every single line and route field these types... Where is the optimal point..?
</blockquote></details>

But do we actually need *all* the fields? Hmm, as I checked: maybe we do :D. But yeah I only wondered if at some point we add fields to route that are not needed in this particular component, it wouldn't affect this if it had it's own fields. I think we also discussed that it is OK to have the same fields in different fragments. I think this is okay for now and up to you if you still want to change it.

What comes to the `LineAllFieldsFragment` dependency, it is still much easier to change to the desired fields when we already have this separated `LineTableRowFragment`. So I think it is okay to use the AllFields fragment as dependency as long as it uses all the fields :+1:

PR #306, opened 2022-08-18
Fix modal covering loading overlay bug

eleksi:
Edit confirmation dialogs would be rendered on top of loading overlay. Changing the order with z-index does not work if the components are in different level in the DOM hierarchy. Therefore loading overlay was moved to upper level in the component hierarchy.

Separate MapLoader component was created, because Router component should have nothing to do with deciding whether or not a loading overlay should be displayed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/306)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/306)* status: 0 of 4 files reviewed, all discussions resolved

PR #308, opened 2022-08-22
Keep observation date in query param

culka:
Keep observation date in the query parameter when viewing route details, so it is possible to return to the same date in line details view.


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/308)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @culka and @tk-vincit)

___
*[`ui/src/components/common/RedirectWithQuery.tsx` line 13 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NAdHR0B-GwFnReHIONc:-NAdHR0B-GwFnReHIONd:b-2apny) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c07e786b1cdb5565df3160abfe85e577f6f4712b/ui/src/components/common/RedirectWithQuery.tsx#L13)):*
> ```XML
>       {...props}
>       to={{
>         // eslint-disable-next-line react/destructuring-assignment
> ```

I don't see reason to break this rule.

You could do something like this at the beginning of the component definition:
`const { to, ...propsWithoutTo } = props;` 

Then you could spread `{ ...propsWithoutTo }` instead of `{ ...props }` and access `to` on the next block. That would remove need for overriding eslint.
___
*[`ui/src/components/common/RedirectWithQuery.tsx` line 14 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NAdIERq2aciM1VBW5Kw:-NAdIERq2aciM1VBW5Kx:b3x4rk7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c07e786b1cdb5565df3160abfe85e577f6f4712b/ui/src/components/common/RedirectWithQuery.tsx#L14)):*
> ```XML
>       to={{
>         // eslint-disable-next-line react/destructuring-assignment
>         ...(typeof props.to === 'object' ? props.to : {}),
> ```

I looked at the typings and seems like `to` prop can be either string (`Path`) or `LocationDescriptorObject`. If it happens to be string, then I don't know what we should do, but now we are just ignoring it then, and that sounds like error. Maybe fallback to `{ pathname: to }` instead of `{}`?

If it happens to be "LocationDescriptorObject", then this makes more sense and is probably OK.
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsHeaderRow.tsx` line 60 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NAdCUlJ3gnhwgojk1MU:-NAdCUlJ3gnhwgojk1MV:b9pbche) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c07e786b1cdb5565df3160abfe85e577f6f4712b/ui/src/components/routes-and-lines/line-details/RouteStopsHeaderRow.tsx#L60)):*
> ```XML
>             href={`${routeDetails[Path.editRoute].getLink(
>               route.route_id,
>             )}?observationDate=${observationDate.toISODate()}`}
> ```

Please avoid buildin query strings manually as that is error-prone especially if there are multiple url params.
We have `qs` library for parsing query strings and building those from JS objects. There might be better ways to do this, but IMO minimum solution would be something like:
```
            href={`${routeDetails[Path.editRoute].getLink(
              route.route_id,
            )}?qs.stringify({ observationDate: observationDate.toISODate() })`}
```

Please note that even that might not be optimal as it swallows possibly previously existing query.

We also seem to have `ui/src/hooks/urlQuery/useUrlQuery.ts` for doing various url query operations, but not sure if it works in this case as now we are entering to whole new route.

If `observationDate` is something that is always needed when we are at "Path.editRoute" route, then it might even make sense to include this logic in `ui/src/router/routeDetails.ts` file's corresponding `getLink` function definition.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/308)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @culka, @pallost, and @tk-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsHeaderRow.tsx` line 60 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NAdCUlJ3gnhwgojk1MU:-NAdXvfH8IWf7IkOZ_Z1:bte31hd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c07e786b1cdb5565df3160abfe85e577f6f4712b/ui/src/components/routes-and-lines/line-details/RouteStopsHeaderRow.tsx#L60)):*

Please avoid buildin query strings manually as that is error-prone especially if there are multiple url params.
We have `qs` library for parsing query strings and building those from JS objects. There might be better ways to do this, but IMO minimum solution would be something like:
```
            href={`${routeDetails[Path.editRoute].getLink(
              route.route_id,
            )}?qs.stringify({ observationDate: observationDate.toISODate() })`}
```

Please note that even that might not be optimal as it swallows possibly previously existing query.

We also seem to have `ui/src/hooks/urlQuery/useUrlQuery.ts` for doing various url query operations, but not sure if it works in this case as now we are entering to whole new route.

If `observationDate` is something that is always needed when we are at "Path.editRoute" route, then it might even make sense to include this logic in `ui/src/router/routeDetails.ts` file's corresponding `getLink` function definition.
</blockquote></details>

Yeah, I'd also rather bake the observationDate parameter into the getLink function

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/308)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`ui/src/components/common/RedirectWithQuery.tsx` line 13 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NAdHR0B-GwFnReHIONc:-NAobVsC7xI65FPpg9p9:bfoxbwn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c07e786b1cdb5565df3160abfe85e577f6f4712b/ui/src/components/common/RedirectWithQuery.tsx#L13)):*

I don't see reason to break this rule.

You could do something like this at the beginning of the component definition:
`const { to, ...propsWithoutTo } = props;` 

Then you could spread `{ ...propsWithoutTo }` instead of `{ ...props }` and access `to` on the next block. That would remove need for overriding eslint.
</blockquote></details>

Done. I couldn't figure out how to replace one of the properties, so this is much better.
___
*[`ui/src/components/common/RedirectWithQuery.tsx` line 14 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NAdIERq2aciM1VBW5Kw:-NAoc2aIE_87-bTieWSl:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c07e786b1cdb5565df3160abfe85e577f6f4712b/ui/src/components/common/RedirectWithQuery.tsx#L14)):*

I looked at the typings and seems like `to` prop can be either string (`Path`) or `LocationDescriptorObject`. If it happens to be string, then I don't know what we should do, but now we are just ignoring it then, and that sounds like error. Maybe fallback to `{ pathname: to }` instead of `{}`?

If it happens to be "LocationDescriptorObject", then this makes more sense and is probably OK.
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopsHeaderRow.tsx` line 60 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NAdCUlJ3gnhwgojk1MU:-NAp0IQ5CZDugvf8uuLM:b-z1w6cz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c07e786b1cdb5565df3160abfe85e577f6f4712b/ui/src/components/routes-and-lines/line-details/RouteStopsHeaderRow.tsx#L60)):*

Yeah, I'd also rather bake the observationDate parameter into the getLink function
</blockquote></details>

ObservationDate is not used in every place, so added it as an optional parameter to the getLink function.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/308)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @culka, @pallost, and @tk-vincit)

___
*[`ui/src/router/routeDetails.ts` line 39 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NBMJ--r-d6qUKK4oblz:-NBMJ--r-d6qUKK4obm-:b-rmu32f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/cd1596f6c395b836c7cb44bbe39e7cbc8b7e8b6d/ui/src/router/routeDetails.ts#L39)):*
> ```XML
>   },
>   [Path.editRoute]: {
>     getLink: (id: string, observationDate: string) =>
> ```

I'd rather call this function parameter `observationIsoDate` (assuming it is ISO date).

That hugely improves code readability and maintanability in the long run. If date string is called just "date", no one knows what format it is supposed to be and has to figure it out somehow.

Please note url param can still be called observationDate:
`... qs.stringify ({ observationDate: observationIsoDate })`

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/308)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @pallost and @tk-vincit)

___
*[`ui/src/router/routeDetails.ts` line 39 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/308#-NBMJ--r-d6qUKK4oblz:-NBMQGT-4MKbD5MJRLNK:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/cd1596f6c395b836c7cb44bbe39e7cbc8b7e8b6d/ui/src/router/routeDetails.ts#L39)):*

I'd rather call this function parameter `observationIsoDate` (assuming it is ISO date).

That hugely improves code readability and maintanability in the long run. If date string is called just "date", no one knows what format it is supposed to be and has to figure it out somehow.

Please note url param can still be called observationDate:
`... qs.stringify ({ observationDate: observationIsoDate })`
</blockquote></details>

Done.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/308)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @pallost and @tk-vincit)

PR #314, opened 2022-08-26
Add apollo error message handler

HenrikHartiala:
#Resolves HSLdevcom/jore4#739

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)

### DIFF
--- a/ui/src/utils/apolloErrors.ts
+++ b/ui/src/utils/apolloErrors.ts
@@ -1,0 +1,33 @@
+import { ApolloError } from '@apollo/client';
+
+/** This will get the actual error message from ApolloError.
+ * There are three different types of errors: network error,
+ * graphQLError and client error. Returned message corresponds
+ * to the type of the message. If there are multiple error messages,
+ * they will be all included.
+ */
+export const getApolloErrorMessage = (err: ApolloError) => {
+  if (err.graphQLErrors.length) {
+    const errorMessages = err.graphQLErrors.map((gqlError) => {
+      // If it is an internal error, we get the message from internal object
+      // otherwise it is in the message field.
+      return gqlError.extensions?.internal?.error?.message || gqlError.message;
+    });
+
+    return `GraphQL errors: ${errorMessages.join(' | ')}`;
+  }
+
+  if (err.clientErrors.length) {
+    const errorMessages = err.clientErrors.map(
+      (clientError) => clientError.message,
+    );
+
+    return `Client error: ${errorMessages.join(' | ')}`;
+  }
+
+  if (err.networkError) {
+    return `Network error: ${err.networkError.message}`;
+  }
+
+  return err.message;
+};
--- a/ui/src/utils/toastService.tsx
+++ b/ui/src/utils/toastService.tsx
@@ -1,5 +1,7 @@
+import { ApolloError } from '@apollo/client';
 import { toast } from 'react-hot-toast';
 import { Toast, ToastTransition, ToastType } from '../uiComponents';
+import { getApolloErrorMessage } from './apolloErrors';
 
 interface ToastOptions {
   message: string;
@@ -22,11 +24,17 @@ export const showDangerToast = (message: string) => {
   });
 };
 
-export const showDangerToastWithError = (message: string, err: unknown) => {
-  if (err instanceof Error) {
-    return showDangerToast(`${message}: ${err.message}`);
+export const showDangerToastWithError = (
+  messageSubject: string,
+  err: unknown,
+) => {
+  if (err instanceof ApolloError) {
+    return showDangerToast(`${messageSubject}: ${getApolloErrorMessage(err)}`);
   }
-  return showDangerToast(message);
+  if (err instanceof Error) {
+    return showDangerToast(`${messageSubject}: ${err.message}`);
+  }
+  return showDangerToast(messageSubject);
 };
 
 export const showSuccessToast = (message: string) => {


jschummer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala and @jschummer)

___
*[`ui/src/utils/toastService.tsx` line 35 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMj0cM1tRObMwn4a-u:-NBMj0cN3zAjhQSkemsf:bwkmzhu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L35)):*
> ```XML
>   if (err.graphQLErrors.length) {
>     let errorMessages = '';
>     err.graphQLErrors.forEach((gqlError) => {
> ```

Would this look a bit better as `.map((qglError) => ...` ? Could also replace the `let` with a `const` in that case. Not sure though what code conventions are in use in the UI nowadays :-)



_Code quote:_
```XML
    let errorMessages = '';
    err.graphQLErrors.forEach((gqlError) => {
      // If it is an internal error, we get the message from internal object
      // otherwise it is in the message field.
      errorMessages += `${
        gqlError.extensions?.internal?.error?.message || gqlError.message
      }.`;
    });
```
___
*[`ui/src/utils/toastService.tsx` line 43 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMjNZC7noybp5EjCQt:-NBMjNZC7noybp5EjCQu:bz5060) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L43)):*
> ```XML
>     });
> 
>     return `GraphQL error: ${errorMessages}`;
> ```

Should we be prepared here that `errorMessages` is still empty? Or can we be sure that it never is empty?



_Code quote:_
```XML
errorMessages
```
___
*[`ui/src/utils/toastService.tsx` line 49 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMj_ZeB6H9-j-tKPBW:-NBMj_ZeB6H9-j-tKPBX:b-9mren2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L49)):*
> ```XML
>     let errorMessages = '';
>     err.clientErrors.forEach((clientError) => {
>       errorMessages += `${clientError.message}.`;
> ```

Maybe the same comments as above (if applicable at all)



_Code quote:_
```XML
    let errorMessages = '';
    err.clientErrors.forEach((clientError) => {
      errorMessages += `${clientError.message}.`;
    });

    return `Client error: ${errorMessages}`;
```

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala, @jschummer, and @tk-vincit)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMx06B63h4n1u3vzbc):*
Probably there shouldn't be "#" at the beginning of the line?
___
*[`ui/src/utils/toastService.tsx` line 26 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMxqITBP41e2l3O0VA:-NBMxqITBP41e2l3O0VB:busl384) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L26)):*
> ```XML
> };
> 
> /** This will get the actual error message from ApolloError.
> ```

This doesn't belong to `toastService`, should rather be at utils/errors or utils/errors/apolloErrors or something like that. :) 
___
*[`ui/src/utils/toastService.tsx` line 35 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMj0cM1tRObMwn4a-u:-NBMxASr7PF_-fVbny0-:b-c9v4uj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L35)):*

Would this look a bit better as `.map((qglError) => ...` ? Could also replace the `let` with a `const` in that case. Not sure though what code conventions are in use in the UI nowadays :-)
</blockquote></details>

Convention is to avoid `let` as long as possible, of course :wink:

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @jschummer and @tk-vincit)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMx06B63h4n1u3vzbc):*

Probably there shouldn't be "#" at the beginning of the line?
</blockquote></details>

Done.
___
*[`ui/src/utils/toastService.tsx` line 26 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMxqITBP41e2l3O0VA:-NBNOwSTEV8hm2DGFREC:b4vgf3d) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L26)):*

This doesn't belong to `toastService`, should rather be at utils/errors or utils/errors/apolloErrors or something like that. :) 
</blockquote></details>

:+1:  Yeah I went with the `utils/apolloErrors.ts`
___
*[`ui/src/utils/toastService.tsx` line 35 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMj0cM1tRObMwn4a-u:-NBNQszT-UIrmkFi3CtQ:b7v9xt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L35)):*

Convention is to avoid `let` as long as possible, of course :wink: 
</blockquote></details>

Ah yeah, I could use `.map(...)` and then `.join(...)` good point.
___
*[`ui/src/utils/toastService.tsx` line 43 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMjNZC7noybp5EjCQt:-NBNTOUMAv-QnjpCjh0N:b31fkpm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L43)):*

Should we be prepared here that `errorMessages` is still empty? Or can we be sure that it never is empty?
</blockquote></details>

Good point, but I think we don't need to handle that. We are already checking if the array has anything in it `if (err.graphQLErrors.length)`. And otherwise `ApolloError` is typed so that `message` is a required field :+1: 
___
*[`ui/src/utils/toastService.tsx` line 49 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMj_ZeB6H9-j-tKPBW:-NBNVSp6AnR0KbgSQeTs:b4jh4qm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L49)):*

Maybe the same comments as above (if applicable at all)
</blockquote></details>

I can do the mapping for these as well, but this time Im not completely sure if we can be sure about the message being there. Although if the message is missing, I think we would still benefit of seeing that the error is a `Client error` or `Network error`. Or what do you think?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @jschummer and @tk-vincit)

___
*[`ui/src/utils/toastService.tsx` line 35 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBMj0cM1tRObMwn4a-u:-NBNW-VoB7YqQJ-VIA7N:b92zrf6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/04aec1d4a1d74aaef6b275a2a501e1798d053f77/ui/src/utils/toastService.tsx#L35)):*

Ah yeah, I could use `.map(...)` and then `.join(...)` good point.
</blockquote></details>

And btw thanks! I don't like to use `let` at all :D

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala, @jschummer, and @tk-vincit)

___
*[`ui/src/utils/apolloErrors.ts` line 11 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBQHq8Q7bSVagB8z96n:-NBQHq8Q7bSVagB8z96o:b35f279) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/edfb099da279575e2be3e71270958b76ca87dde2/ui/src/utils/apolloErrors.ts#L11)):*
> ```XML
> export const getApolloErrorMessage = (err: ApolloError) => {
>   if (err.graphQLErrors.length) {
>     // let errorMessages = '';
> ```

Please remove this.
___
*[`ui/src/utils/apolloErrors.ts` line 18 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBQJ56j7q30_EfyQ6x4:-NBQJ56j7q30_EfyQ6x5:b5w9qyj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/edfb099da279575e2be3e71270958b76ca87dde2/ui/src/utils/apolloErrors.ts#L18)):*
> ```XML
>     });
> 
>     return `GraphQL errors: ${errorMessages.join(' | ')}`;
> ```

This is improvement over previous implementation so won't add this as blocking comment, but anyway, there are at least 2 problems with this approach:

- The fact that we happen to use GraphQL is implementation detail and end users shouldn't need to know/worry about that.
- These are not even trying to be localized.
The latter one is IMO OK in this case, as exposed error messages from hasura are in english anyway. 

Strictly thinking we shouldn't expose any errors from hasura directly to users, but for sake of convenience it is OK in this case. IMO most useful aspect of this approach is that it allows end users to copy paste these error messages to  bug reports and for developer receiving bug report with something even remotely useful is better than getting just "postgres tx error".

Anyway, I would consider removing these prefixes especially if the error messages in different cases are descriptive enough that developer can figure out if the error comes from grapqhl, client or network.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @jschummer and @tk-vincit)

___
*[`ui/src/utils/apolloErrors.ts` line 11 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBQHq8Q7bSVagB8z96n:-NBQkxwG76bqXcC8lhgb:b9c2275) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/edfb099da279575e2be3e71270958b76ca87dde2/ui/src/utils/apolloErrors.ts#L11)):*

Please remove this.
</blockquote></details>

:man_facepalming: Done
___
*[`ui/src/utils/apolloErrors.ts` line 18 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/314#-NBQJ56j7q30_EfyQ6x4:-NBQl-Z_3or_ma44rxMZ:bnvry8j) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/edfb099da279575e2be3e71270958b76ca87dde2/ui/src/utils/apolloErrors.ts#L18)):*

This is improvement over previous implementation so won't add this as blocking comment, but anyway, there are at least 2 problems with this approach:

- The fact that we happen to use GraphQL is implementation detail and end users shouldn't need to know/worry about that.
- These are not even trying to be localized.
The latter one is IMO OK in this case, as exposed error messages from hasura are in english anyway. 

Strictly thinking we shouldn't expose any errors from hasura directly to users, but for sake of convenience it is OK in this case. IMO most useful aspect of this approach is that it allows end users to copy paste these error messages to  bug reports and for developer receiving bug report with something even remotely useful is better than getting just "postgres tx error".

Anyway, I would consider removing these prefixes especially if the error messages in different cases are descriptive enough that developer can figure out if the error comes from grapqhl, client or network.
</blockquote></details>

This is something that I also thought when implementing this, so I agree that in the end product we shouldn't expose these errors straight to user. But I also thought that in this testing phase it would be beneficial for us as devs and the test users to see a more descriptive error message, since we do not currently have any logger to log this kind of GraphQL errors to a database.

Currently in this test phase I would keep these prefixes.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @jschummer and @tk-vincit)

jschummer (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/314)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jschummer and @tk-vincit)

PR #320, opened 2022-08-30
Hardcode date formats temporarily to avoid test flakiness on CI

tk-vincit:
Seems like github ci runners are not identical, and sometimes
snapshot tests break on ci as date formats change, probably
due to locale-related differences.

Example log of the problem:

```
  ● formatDate() › Should be able to override default locale from i18n

    expect(received).toBe(expected) // Object.is equality

    Expected: "April 20, 2017, 2:32 PM GMT+3"
    Received: "April 20, 2017 at 2:32 PM GMT+3"
```

This fix is not optimal and should hopefully be reverted at
some point, but for now this allows development team to get things done
instead of fighting flaky tests on CI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/320)

### DIFF
--- a/ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap
+++ b/ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap
@@ -123,7 +123,7 @@ exports[`<RouteStopsTable /> Renders the route with stops along its geometry 1`]
               <span
                 data-testid="RouteStopsHeaderRow::lastEdited"
               >
-                !14.2.2017 klo 14.51
+                !14.2.2017 14.51
               </span>
               <svg
                 class="ml-2 inline text-xl text-tweaked-brand"
@@ -331,7 +331,7 @@ exports[`<RouteStopsTable /> Renders the route with stops along its geometry 2`]
               <span
                 data-testid="RouteStopsHeaderRow::lastEdited"
               >
-                !14.2.2017 klo 14.51
+                !14.2.2017 14.51
               </span>
               <svg
                 class="ml-2 inline text-xl text-tweaked-brand"
@@ -450,7 +450,7 @@ exports[`<RouteStopsTable /> Renders the route with stops along its geometry 2`]
               <span
                 data-testid="RouteStopsRow::lastEdited"
               >
-                !14.2.2017 klo 14.51
+                !14.2.2017 14.51
               </span>
               <svg
                 class="ml-2 inline text-xl text-tweaked-brand"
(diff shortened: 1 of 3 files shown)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/320)* status: 0 of 3 files reviewed, all discussions resolved

HenrikHartiala (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/320)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala and @tk-vincit)

___
*[`ui/src/time.spec.ts` line 16 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/320#-NAmTaNr6KkBoIXnRDb7:-NAmTaNr6KkBoIXnRDb8:bm06ovk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1488ce73bdc113a44b93ea47c9d80bcae0fce89a/ui/src/time.spec.ts#L16)):*
> ```XML
>   // depending on ci machine that ran the test.
>   // eslint-disable-next-line jest/no-commented-out-tests
>   /*
> ```

When commenting multiple lines, `//` should be used for each line rather than just `/*` in the start and `*/` in the end.
Or add for example `*` for each line in between, like JSDoc style is doing. 
This is to avoid accidentally broken code (which can't happen in our case since we have checks), but also it makes it much easier to see in review what have been commented out / in.

If using VS code, you can choose the lines and then CMD + K + C to comment them.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/320)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`ui/src/time.spec.ts` line 16 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/320#-NAmTaNr6KkBoIXnRDb7:-NAmYE0RD4PB4urp3rrm:bpaa7ug) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/1488ce73bdc113a44b93ea47c9d80bcae0fce89a/ui/src/time.spec.ts#L16)):*

When commenting multiple lines, `//` should be used for each line rather than just `/*` in the start and `*/` in the end.
Or add for example `*` for each line in between, like JSDoc style is doing. 
This is to avoid accidentally broken code (which can't happen in our case since we have checks), but also it makes it much easier to see in review what have been commented out / in.

If using VS code, you can choose the lines and then CMD + K + C to comment them.

</blockquote></details>

This commenting style was intented. Now I don't have to change every line and thus `git blame` etc will point to the proper commit where those lines were authored also after this commit and even after we (hopefully) revert this at some point.

HenrikHartiala (approved):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/320)* status: all files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

PR #325, opened 2022-09-06
Improve move stop usability

HenrikHartiala:
Resolves HSLdevcom/jore4#733

Move stop functionality and visualisation improved to be more intuitive.
* Stop draggability is now removed. Pressing "move stop" will go to "moveStopMode" and clicking a location will attempt to move the stop there. This now works the same way as when creating a new stop. (Stop marker following the cursor and only one click will decide where to move the stop).
* The stop that is being moved will have the old location visualised with dotted gray circle
* Pressing ESC button will cancel the move
  * This also works when creating stop
* When moving a stop, other modals will be closed / hidden

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/325)

### DIFF
--- a/ui/src/components/map/Map.tsx
+++ b/ui/src/components/map/Map.tsx
@@ -9,6 +9,7 @@ import { Column } from '../../layoutComponents';
 import {
   selectHasDraftRouteGeometry,
   selectIsCreateStopModeEnabled,
+  selectIsMoveStopModeEnabled,
   selectMapEditor,
   selectMapFilter,
   selectSelectedRouteId,
@@ -65,6 +66,7 @@ export const MapComponent = (
   const [showStops, setShowStops] = useState(true);
 
   const isCreateStopModeEnabled = useAppSelector(selectIsCreateStopModeEnabled);
+  const isMoveStopModeEnabled = useAppSelector(selectIsMoveStopModeEnabled);
 
   // TODO: avoid any type
   const editorLayerRef = useRef<ExplicitAny>(null);
@@ -94,10 +96,19 @@ export const MapComponent = (
     }
   };
 
+  const onMoveStop = (e: MapEvent) => {
+    if (!drawingMode) {
+      stopsRef.current?.onMoveStop(e);
+    }
+  };
+
   const onClick = (e: MapEvent) => {
     if (isCreateStopModeEnabled) {
       onCreateStop(e);
       return;
+    }
+    if (isMoveStopModeEnabled) {
+      onMoveStop(e);
     }
 
     // Retrieve all rendered features on the map
--- a/ui/src/components/map/MapFooter.tsx
+++ b/ui/src/components/map/MapFooter.tsx
@@ -9,6 +9,7 @@ import {
   selectHasDraftRouteGeometry,
   selectIsCreateStopModeEnabled,
   selectIsInViewMode,
+  selectIsMoveStopModeEnabled,
   selectMapEditor,
   setIsCreateStopModeEnabledAction,
 } from '../../redux';
@@ -39,14 +40,19 @@ export const MapFooter: React.FC<Props> = ({
 }) => {
   const { t } = useTranslation();
 
-  const { drawingMode, creatingNewRoute, selectedRouteId } =
-    useAppSelector(selectMapEditor);
+  const {
+    drawingMode,
+    creatingNewRoute,
+    isRouteMetadataFormOpen,
+    selectedRouteId,
+  } = useAppSelector(selectMapEditor);
   const hasDraftRouteGeometry = useAppSelector(selectHasDraftRouteGeometry);
 
   const hasChangesInProgress = useAppSelector(selectHasChangesInProgress);
   const isInViewMode = useAppSelector(selectIsInViewMode);
 
   const isCreateStopModeEnabled = useAppSelector(selectIsCreateStopModeEnabled);
+  const isMoveStopModeEnabled = useAppSelector(selectIsMoveStopModeEnabled);
   const setIsCreateStopModeEnabled = useAppAction(
     setIsCreateStopModeEnabledAction,
   );
@@ -60,7 +66,12 @@ export const MapFooter: React.FC<Props> = ({
       <SimpleButton
         testId={testIds.drawRouteButton}
         onClick={onDrawRoute}
-        disabled={!isInViewMode || creatingNewRoute}
+        disabled={
+          !isInViewMode ||
+          creatingNewRoute ||
+          isCreateStopModeEnabled ||
+          isMoveStopModeEnabled
+        }
         inverted={drawingMode !== Mode.Draw}
       >
         {t('map.drawRoute')}
@@ -75,7 +86,7 @@ export const MapFooter: React.FC<Props> = ({
       </SimpleButton>
       <SimpleButton
         onClick={onAddStops}
-        disabled={drawingMode !== undefined}
+        disabled={drawingMode !== undefined || creatingNewRoute}
         inverted={!isCreateStopModeEnabled}
         testId={testIds.addStopButton}
       >
@@ -84,11 +95,11 @@ export const MapFooter: React.FC<Props> = ({
       <SimpleButton
         className="h-full !px-3 text-xl"
         onClick={onDeleteRoute}
-        disabled={!hasChangesInProgress}
+        disabled={!hasChangesInProgress || isRouteMetadataFormOpen}
       >
         <MdDelete aria-label={t('map.deleteRoute')} />
       </SimpleButton>
-      <Visible visible={hasChangesInProgress}>
+      <Visible visible={hasChangesInProgress && !isRouteMetadataFormOpen}>
         <SimpleButton
           containerClassName="!ml-auto"
           onClick={onCancel}
--- a/ui/src/components/map/markers/Circle.tsx
+++ b/ui/src/components/map/markers/Circle.tsx
@@ -6,6 +6,7 @@ interface Props {
   borderWidth?: number;
   fillColor?: string;
   borderColor?: string;
+  strokeDashArray?: number;
   centerDot?: boolean;
   centerDotSize?: number;
   onClick?: () => void;
@@ -17,6 +18,7 @@ const CircleComponent = ({
   borderWidth = 2,
   fillColor = 'white',
   borderColor = 'black',
+  strokeDashArray = 0,
   centerDot = false,
   centerDotSize = 2,
   onClick,
@@ -37,6 +39,7 @@ const CircleComponent = ({
         cy={size / 2}
         r={size / 2 - borderWidth}
         stroke={borderColor}
+        strokeDasharray={strokeDashArray}
         strokeWidth={borderWidth}
         fill={fillColor}
       />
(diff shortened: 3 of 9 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/325)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLXWbR289E_9xcwiXw):*
You had nice description of this change in PR description. Why not to include it in commit message?
___
*[`ui/src/components/map/Map.tsx` line 101 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLXgWt05n1_6Kis0-d:-NBLXgWt05n1_6Kis0-e:b-vppclv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/Map.tsx#L101)):*
> ```XML
> 
>   const onMoveStop = (e: MapEvent) => {
>     if (stopsRef.current && drawingMode === undefined) {
> ```

Do we want to explicitly check if drawingMode is `undefined` or what if it is e.g. `null`? How about just `if (!drawingMode) { stopsRef.current?.onMoveStop(e); }`
___
*[`ui/src/components/map/stops/CreateStopMarker.tsx` line 34 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLYXkx2mucSdiKBDMY:-NBLYXkx2mucSdiKBDMZ:brrqu58) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/stops/CreateStopMarker.tsx#L34)):*
> ```XML
> 
>   const detectKeyDown = (e: KeyboardEvent) => {
>     if (e.key === 'Escape') {
> ```

What happens if user is using touchscreen, are they still able to cancel this operation somehow?

Is it documented that `KeyboardEvent.key` always returns esc key string as 'Escape', or can they also be e.g. 'ESCAPE' or 'escape'? A bit safer way would be to `if (String(e.key).toLowerCase() === 'escape')). But OK, this is a bit easier to read so if this works on chrome I'm ok with this.
___
*[`ui/src/components/map/stops/Stop.tsx` line 54 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBL_Yy-BW1pAOhiN4J2:-NBL_Yy-BW1pAOhiN4J3:b872fuh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/stops/Stop.tsx#L54)):*
> ```XML
>   // If the stop is being moved, we use different styles for the stop
>   // to indicate the placeholder of the old location
>   const isPlaceHolder = selected && isMoveStopModeEnabled;
> ```

isPlaceholder, placeholder is one word.
___
*[`ui/src/components/map/stops/Stops.tsx` line 111 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBL_oIvENF8IXIheLw9:-NBL_oIvENF8IXIheLwA:b-cnmk4e) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/stops/Stops.tsx#L111)):*
> ```XML
>       setIsLoading(false);
>     },
>     onMoveStop: async (e: MapEvent) => {
> ```

I don't see `await` anywhere so does this have to be `async`?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/325)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @tk-vincit)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLXWbR289E_9xcwiXw):*

You had nice description of this change in PR description. Why not to include it in commit message?
</blockquote></details>

Ah, it is not a flow I've used. But good point! I'll try to start adding the descriptions to the commit body :+1: 
___
*[`ui/src/components/map/Map.tsx` line 101 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLXgWt05n1_6Kis0-d:-NBMOJiS2cXb9RJqDnof:b148ta8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/Map.tsx#L101)):*

Do we want to explicitly check if drawingMode is `undefined` or what if it is e.g. `null`? How about just `if (!drawingMode) { stopsRef.current?.onMoveStop(e); }`
</blockquote></details>

Great one, I'll change this.
___
*[`ui/src/components/map/stops/CreateStopMarker.tsx` line 34 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLYXkx2mucSdiKBDMY:-NBMOjQxDXsSRmLPnEyv:bjdh11c) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/stops/CreateStopMarker.tsx#L34)):*

What happens if user is using touchscreen, are they still able to cancel this operation somehow?

Is it documented that `KeyboardEvent.key` always returns esc key string as 'Escape', or can they also be e.g. 'ESCAPE' or 'escape'? A bit safer way would be to `if (String(e.key).toLowerCase() === 'escape')). But OK, this is a bit easier to read so if this works on chrome I'm ok with this.
</blockquote></details>

Yes, the user can also click anywhere on map and then press `Cancel`. The map footer needs some redesign / refactor, because already there are buttons that are not used anywhere else than some special case (for example there is a trash bin that only applies to routes). And within that refactor we could also add a `Cancel` button to the footer while moving a stop.

What comes to the key-event, it should be the same for every browser. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key and https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values
Out of curiosity checked if the virtual keyboard works, and it does :+1: 
___
*[`ui/src/components/map/stops/Stop.tsx` line 54 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBL_Yy-BW1pAOhiN4J2:-NBMQheG8beo3zjKetVe:bxbwcfh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/stops/Stop.tsx#L54)):*

isPlaceholder, placeholder is one word.
</blockquote></details>

Whoops! Thanks.
___
*[`ui/src/components/map/stops/Stops.tsx` line 111 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBL_oIvENF8IXIheLw9:-NBMRcI0ACsWiOb5lYy0:bemivzw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/stops/Stops.tsx#L111)):*

I don't see `await` anywhere so does this have to be `async`?
</blockquote></details>

Hmm, yeah maybe not. I'll remove it

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/325)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @tk-vincit)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLXWbR289E_9xcwiXw):*

Ah, it is not a flow I've used. But good point! I'll try to start adding the descriptions to the commit body :+1: 
</blockquote></details>

I guess in ideal (=ancient?) world there would be some newlines in git commit messages to avoid very long lines, but maybe today's tools can handle also long lines without stupid issues :smile: 

But yeah, in the future I suggest to include these descriptions in commit messages too, since they're easier to find from commit messages than pull requests.
___
*[`ui/src/components/map/stops/CreateStopMarker.tsx` line 34 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBLYXkx2mucSdiKBDMY:-NBMXCH50XpZIs5BpC6k:bz8i9n) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8ffa0ba5af69b90f68c5e5b32e3e4559b7265f13/ui/src/components/map/stops/CreateStopMarker.tsx#L34)):*

Yes, the user can also click anywhere on map and then press `Cancel`. The map footer needs some redesign / refactor, because already there are buttons that are not used anywhere else than some special case (for example there is a trash bin that only applies to routes). And within that refactor we could also add a `Cancel` button to the footer while moving a stop.

What comes to the key-event, it should be the same for every browser. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key and https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values
Out of curiosity checked if the virtual keyboard works, and it does :+1: 
</blockquote></details>

:+1: :ok:

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/325)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`ui/src/components/map/stops/Stop.tsx` line 17 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBROyqhDumSNRC5SKnw:-NBRiK8YEiL6XDbKieAx:b-mgh1n8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5261df74c4e46e26db8a924e8ab50df20bbef10a/ui/src/components/map/stops/Stop.tsx#L17)):*

prop starting with `on` is react convention for naming callbacks (see e.g. onClick or onDragEnd above). Please rename this.
</blockquote></details>

Didn't noticed this was outside of scope of this PR so no need to fix this right now.
___
*[`ui/src/components/map/stops/Stop.tsx` line 33 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/325#-NBRMJhYAILGsVx3I594:-NBRiRPP0uWHVuzaRJWB:b92rdv8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5261df74c4e46e26db8a924e8ab50df20bbef10a/ui/src/components/map/stops/Stop.tsx#L33)):*

This looks suspicious, we have no guarantees that `colors.stops` has key  that corresponds onVehicleRoute.

Also this can got broken _very easily_ by a mistake if someone happens to edit theme and edits `colors.stops`, as theme file has not references to this kind of usage at all!  By very least I would do someting like
```
const vehicleStopColor = colors.stops[onVehicleRoute] || 'black';
return vehicleStopColor;

```
Then you could get rid of ternary and have sensible fallback. There could also be logging:
```
if (onVehicleRoute && !colors.stops[onVehicleRoute]) {
  console.log('Error: expected to find color corresponding ${onVehicleRoute} but none was defined')
}
```
But maybe that's a bit overkill if we have sensible fallback.


</blockquote></details>

OK, this problem was also on original implementation so I'll approve this for now.

PR #332, opened 2022-09-09
Add line and route search tests and resources

sr-vala:
Simple line and route search tests, resources and testIds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/332)

PR #335, opened 2022-09-13
Improve test data builders

tk-vincit:
- Include default values for required fields in buildRoute
- Don't define default validityStart in buildLine, as otherwise default routes can't be included in default lines as route's validity period must lie within line's validity period
- Make parameter optional in db helper utility as it should be
- Simplify test data definitions in routeStops test

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/335)


tk-vincit:
Included & merged within https://github.com/HSLdevcom/jore4-ui/pull/332

PR #346, opened 2022-09-29
Validity check hook tests + small fixes to hook itself

tk-vincit:
Tests included in this PR:
```
 PASS  src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
  useCheckValidityAndPriorityConflicts()
    ✓ Should return conflict on resources with same validity_start (13 ms)
    ✓ Should return conflict on resources with same validity_end (14 ms)
    ✓ Should return conflict on resource within existing instance of resource (20 ms)
    ✓ Should return conflict on resource starting before and ending after existing version (16 ms)
    ✓ Should not allow new version after current version when bounds overlap (13 ms)
    ✓ Should allow new version after current version, bounds not overlapping, validity_end null (13 ms)
    ✓ Should allow new version after current version, bounds not overlapping, validity_end defined (17 ms)
    ✓ Should not allow new version before current version when bounds overlap (15 ms)
    ✓ Should allow new version before current version, bounds not overlapping, validity_start defined (13 ms)
    ✓ Should allow new version before current version, bounds not overlapping, validity_start null (16 ms)
    Indefinite resources (validity_start and validity_end both null)
      ✓ Should return conflict on two resources with indefinite validity (31 ms)
      ✓ Should return conflict on bounded resource within existsting indefinite resource (13 ms)
    Checking conflicts with priorities
      ✓ Should allow creating conflicting version with new priority (17 ms)
      ✓ Should allow creating conflicting version with draft priority (1 ms)
      ✓ A resource should not conflict with itself when being modified (17 ms)
      Routes
        ✓ Does not allow creating conflicting route with same direction (16 ms)
        ✓ Allow creating conflicting route with opposite direction (18 ms)
 ```
 
Resolves HSLdevcom/jore4[#740](https://github.com/HSLdevcom/jore4-ui/issues/740)

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/346)

### DIFF
--- a/ui/src/graphql/client.ts
+++ b/ui/src/graphql/client.ts
@@ -45,44 +45,55 @@ const buildScalarMappingLink = () => {
   return withScalars({ schema, typesMap });
 };
 
-const buildWebSocketLink = (graphqlUrlPath: string) => {
+const getGraphqlUrl = (isTesting: boolean, isWebsocket: boolean) => {
+  const path = '/api/graphql/v1/graphql';
+  if (isTesting) {
+    return `http://localhost:3300${path}`;
+  }
+  if (isWebsocket) {
+    return mapHttpToWs(`${window.location.origin}${path}`);
+  }
+  return path;
+};
+
+const buildWebSocketLink = () => {
   return new WebSocketLink({
     // WebSocketLink doesn't work with relative url's, so we have to
     // turn relative url into absolute.
-    uri: `${mapHttpToWs(window.location.origin)}${graphqlUrlPath}`,
+    uri: getGraphqlUrl(false, true),
     options: {
       reconnect: true,
     },
   });
 };
 
-const buildHttpLink = (graphqlUrlPath: string, isTesting: boolean) => {
+const buildHttpLink = (isTesting: boolean) => {
+  const defaultConfig = {
+    uri: getGraphqlUrl(isTesting, false),
+  };
+
   const httpLinkConfig = isTesting
     ? {
+        ...defaultConfig,
         // `fetch` is not available in test context (with jest) so we have to provide it
         // eslint-disable-next-line global-require, import/no-extraneous-dependencies
         fetch: require('cross-fetch'),
         // logging in directly as admin, only works in e2e environment
         headers: { 'x-hasura-admin-secret': 'hasura' },
-        uri: `http://localhost:3300${graphqlUrlPath}`,
       }
-    : {
-        uri: graphqlUrlPath,
-      };
+    : defaultConfig;
   return new HttpLink(httpLinkConfig);
 };
 
 const buildConnectionLink = (isBrowser: boolean, isTesting: boolean) => {
-  const graphqlUrlPath = '/api/graphql/v1/graphql';
-
   // because next.js might run this on server-side and websockets aren't
   // supported there, we have to check if we are on browser before
   // initializing WebSocket link
-  const httpLink = buildHttpLink(graphqlUrlPath, isTesting);
+  const httpLink = buildHttpLink(isTesting);
   if (isTesting || !isBrowser) {
     return httpLink;
   }
-  const wsLink = buildWebSocketLink(graphqlUrlPath);
+  const wsLink = buildWebSocketLink();
   return split(
     // if running a subscription query, prefer to use wsLink
     ({ query }) => {
--- a/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
+++ b/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
@@ -270,10 +270,7 @@ describe(`${useCheckValidityAndPriorityConflicts.name}()`, () => {
       });
     });
 
-    // TODO: this test should pass, will be fixed in future commit
-    // so that it is easier to see the reason why this didn't.
-    // eslint-disable-next-line jest/no-disabled-tests
-    test.skip('Should allow new version before current version, bounds not overlapping, validity_start null', async () => {
+    test('Should allow new version before current version, bounds not overlapping, validity_start null', async () => {
       await act(async () => {
         const conflicts = await result.current.getConflictingLines(
           buildQuery({
--- a/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
+++ b/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
@@ -126,10 +126,8 @@ describe(`${useCheckValidityAndPriorityConflicts.name}()`, () => {
         expect(conflicts[0].line_id).toEqual(indefiniteLine.line_id);
       });
     });
-    // TODO: this test should pass, will be fixed in future commit
-    // so that it is easier to see the reason why this didn't.
-    // eslint-disable-next-line jest/no-disabled-tests
-    test.skip('Should return conflict on bounded resource within existing indefinite resource', async () => {
+
+    test('Should return conflict on bounded resource within existing indefinite resource', async () => {
       await act(async () => {
         const conflicts = await result.current.getConflictingLines(
           buildQuery({
--- a/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
+++ b/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
@@ -2,7 +2,9 @@
 /* eslint-disable camelcase */
 import {
   buildLine,
+  buildRoute,
   Priority,
+  RouteDirectionEnum,
   RouteLineInsertInput,
 } from '@hsl/jore4-test-db-manager';
 import { act, renderHook } from '@testing-library/react';
@@ -57,6 +59,18 @@ const boundedLine = {
   validity_start: DateTime.fromISO('2022-01-01T13:08:43.315+02:00'),
   validity_end: DateTime.fromISO('2022-01-31T13:08:43.315+02:00'),
 } as TestLine;
+const boundedLine2V1 = {
+  ...buildLine({ label: 'bounded2' }),
+  line_id: '53125ee7-3320-45e6-838d-6e0c2ff1dcd6',
+  validity_start: DateTime.fromISO('2022-01-01T13:08:43.315+02:00'),
+  validity_end: DateTime.fromISO('2022-01-31T13:08:43.315+02:00'),
+} as TestLine;
+const boundedLine2V2 = {
+  ...buildLine({ label: 'bounded2' }),
+  line_id: 'bc6b7767-1f20-41cd-8cf9-1d73e262ed33',
+  validity_start: boundedLine.validity_start?.plus({ months: 1 }),
+  validity_end: boundedLine.validity_end?.plus({ months: 1 }),
+} as TestLine;
 const draftLine = {
   ...buildLine({ label: 'draft' }),
   line_id: 'b3734679-1a15-412f-b3ce-1deeade51e89',
@@ -70,11 +84,23 @@ const lines: TestLine[] = [
   indefiniteStartLine,
   indefiniteEndLine,
   boundedLine,
+  boundedLine2V1,
+  boundedLine2V2,
   draftLine,
 ];
 
+const inboundRoute = {
+  ...buildRoute({ label: 'inboundRoute' }),
+  route_id: '4415f493-f170-4540-8753-fe5c3835ad8a',
+  direction: RouteDirectionEnum.Inbound,
+  on_line_id: indefiniteLine.line_id,
+};
+
+const routes = [inboundRoute];
+
 const dbResources = {
   lines,
+  routes,
 };
 
 const buildQuery = ({
@@ -89,9 +115,7 @@ const buildQuery = ({
   validityEnd,
 });
 
-const deleteCreatedResources = () => {
-  removeFromDbHelper(dbResources);
-};
+const deleteCreatedResources = () => removeFromDbHelper(dbResources);
 
 describe(`${useCheckValidityAndPriorityConflicts.name}()`, () => {
   beforeAll(async () => {
@@ -325,6 +349,48 @@ describe(`${useCheckValidityAndPriorityConflicts.name}()`, () => {
           boundedLine.line_id,
         );
         expect(conflicts.length).toBe(0);
+      });
+    });
+
+    test('A resource should still conflict with existing version if modified', async () => {
+      await act(async () => {
+        const conflicts = await result.current.getConflictingLines(
+          buildQuery({
+            label: boundedLine2V1.label,
+            validityStart: boundedLine2V2.validity_start,
+            validityEnd: boundedLine2V2.validity_end,
+          }),
+          boundedLine2V1.line_id,
+        );
+        expect(conflicts.length).toBe(1);
+        expect(conflicts[0].line_id).toEqual(boundedLine2V2.line_id);
+      });
+    });
+
+    describe('Routes', () => {
+      test('Does not allow creating conflicting route with same direction', async () => {
+        await act(async () => {
+          const conflicts = await result.current.getConflictingRoutes({
+            ...buildQuery({
+              label: inboundRoute.label,
+            }),
+            direction: RouteDirectionEnum.Inbound,
+          });
+          expect(conflicts.length).toBe(1);
+          expect(conflicts[0].route_id).toEqual(inboundRoute.route_id);
+        });
+      });
+
+      test('Allow creating conflicting route with opposite direction', async () => {
+        await act(async () => {
+          const conflicts = await result.current.getConflictingRoutes({
+            ...buildQuery({
+              label: inboundRoute.label,
+            }),
+            direction: RouteDirectionEnum.Outbound,
+          });
+          expect(conflicts.length).toBe(0);
+        });
       });
     });
   });
--- a/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
+++ b/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts
@@ -1,6 +1,10 @@
 // graphql schema uses snake_case instead of camelCase
 /* eslint-disable camelcase */
-import { buildLine, Priority } from '@hsl/jore4-test-db-manager';
+import {
+  buildLine,
+  Priority,
+  RouteLineInsertInput,
+} from '@hsl/jore4-test-db-manager';
 import { act, renderHook } from '@testing-library/react';
 import { DateTime } from 'luxon';
 import { OptionalKeys } from '../types';
@@ -22,17 +26,51 @@ jest.mock('graphql', () => ({
   buildClientSchema: jest.fn(),
 }));
 
-const lines = [
-  {
-    ...buildLine({ label: 'indefinite' }),
-    line_id: 'c23b4134-5098-4cda-9eec-d393001580c3',
-  },
-  {
-    ...buildLine({ label: '65x' }),
-    line_id: 'cb427403-696c-46a6-b1f8-a54ef189fb17',
-    validity_start: DateTime.fromISO('2022-08-11T13:08:43.315+03:00'),
-    validity_end: DateTime.fromISO('2023-08-11T13:08:43.315+03:00'),
-  },
+// custom typings for lines to be used in this test suite to avoid
+// TS errors due to TS interpreting everything in `RouteLineInsertInput`
+// type to be Maybe<> and optional
+type TestLine = RouteLineInsertInput & {
+  label: string;
+  validity_start?: DateTime;
+  validity_end?: DateTime;
+  line_id: UUID;
+  priority: Priority;
+};
+
+const indefiniteLine = {
+  ...buildLine({ label: 'indefinite' }),
+  line_id: 'c23b4134-5098-4cda-9eec-d393001580c3',
+} as TestLine;
+const indefiniteEndLine = {
+  ...buildLine({ label: 'indefiniteEnd' }),
+  line_id: '1ba4e1c8-63b1-4a4f-8658-0c34cfa80b62',
+  validity_start: DateTime.fromISO('2022-08-11T13:08:43.315+02:00'),
+} as TestLine;
+const indefiniteStartLine = {
+  ...buildLine({ label: 'indefiniteStart' }),
+  line_id: '3c6d03a9-9f7f-49dc-b125-506d0eab1854',
+  validity_end: DateTime.fromISO('2022-08-11T13:08:43.315+02:00'),
+} as TestLine;
+const boundedLine = {
+  ...buildLine({ label: 'bounded' }),
+  line_id: 'cb427403-696c-46a6-b1f8-a54ef189fb17',
+  validity_start: DateTime.fromISO('2022-01-01T13:08:43.315+02:00'),
+  validity_end: DateTime.fromISO('2022-01-31T13:08:43.315+02:00'),
+} as TestLine;
+const draftLine = {
+  ...buildLine({ label: 'draft' }),
+  line_id: 'b3734679-1a15-412f-b3ce-1deeade51e89',
+  validity_start: DateTime.fromISO('2022-02-01T13:08:43.315+02:00'),
+  validity_end: DateTime.fromISO('2023-02-28T13:08:43.315+02:00'),
+  priority: Priority.Draft,
+} as TestLine;
+
+const lines: TestLine[] = [
+  indefiniteLine,
+  indefiniteStartLine,
+  indefiniteEndLine,
+  boundedLine,
+  draftLine,
 ];
 
 const dbResources = {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by pallost — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/346)* status: 0 of 3 files reviewed, 15 unresolved discussions (waiting on @pallost and @tk-vincit)

___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.ts` line 52 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlpE8f-Wi7l14X14N1:-NDlpE8f-Wi7l14X14N2:b-81rs7b) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.ts#L52)):*
> ```XML
> const buildCommonGqlFilter = (
>   params: CommonParams,
>   allowUndefinedValidityStart = true,
> ```

I know it's not directly part of this PR, but this flag is never used, so this value is always true. I also don't think we should support it, it's not really in the scope of this hook. Let's just shove all the check conditions from buildValidityStartMissingGqlFilterOrConditions to this buildCommonGqlFilter function to see the full picture.

I also don't see a reason to fork the logic based on if the resource is indefinite. Let's just rather put all these conditions on the same level, connected with an OR operator. Having a comment or a visual representation of the intervals should help a lot!
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 36 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlv0Ym6HERgNHd8rGJ:-NDlv0Ym6HERgNHd8rGK:b-35rce7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L36)):*
> ```XML
> type TestLine = RouteLineInsertInput & {
>   label: string;
>   validity_start?: DateTime;
> ```

I do see the `null` value useful for the validity periods. Undefined means, well, that it's not defined. Null however means that it IS defined and is defined to be infinite.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 116 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm-yBb7yT4ebzKhVdu:-NDm-yBb7yT4ebzKhVdv:b5ikfdu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L116)):*
> ```XML
>   });
> 
>   const { result } = renderHook(
> ```

I'm wondering that is this hook now technically stateful? I understand that for this hook it's not an issue as there's no state within, but I wonder if we should add a comment here that people won't start copy-pasting this pattern to other hook tests but be warned that these hooks are actually stateful and if they need otherwise, they'd need to call the renderHook function within the tests themselves.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 121 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlx2_E9bm8aNxUmJUF:-NDlx2_E9bm8aNxUmJUG:b-c0mni0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L121)):*
> ```XML
>   );
> 
>   describe('Indefinite resources (validity_start and validity_end both null)', () => {
> ```

this text is not true, you have tests within that check for bounded validity periods as well. How about this?
`Checking validity period conflicts (label and priority are the same)`



_Code quote:_
```XML
ndefinite resources (validity_start and validity_end both null)
```
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 125 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlwaaV8NEfud6BmpAm:-NDlwaaV8NEfud6BmpAn:b8qq0ox) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L125)):*
> ```XML
>       await act(async () => {
>         const conflicts = await result.current.getConflictingLines(
>           buildQuery({ label: indefiniteLine.label }),
> ```

let's add the full parameter set for clarity: `{label: indefiniteLine.label, validityStart: null, validityEnd: null}`
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 127 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlxVcBCJC56jdvmQz9:-NDlxVcBCJC56jdvmQzA:bhsq5f2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L127)):*
> ```XML
>           buildQuery({ label: indefiniteLine.label }),
>         );
>         expect(conflicts.length).toBe(1);
> ```

should add an additional expect to every test below to be certain that the correct conflicting resource was found:
`expect(conflicts[0].line_id).toEqual(indefiniteLine.line_id)`
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 130 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlxtIMBWHAVOHhRabm:-NDlxtINAT5NwogcvriH:byd3jki) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L130)):*
> ```XML
>       });
>     });
>     test('Should return conflict on bounded resource within existsting indefinite resource', async () => {
> ```

typo, "existing"
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 144 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlyxQi8EMSja21XtLK:-NDlyxQi8EMSja21XtLL:b-peyqxq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L144)):*
> ```XML
>   });
> 
>   test('Should return conflict on resources with same validity_start', async () => {
> ```

this is an edge case. Would need a more generic test as well where we check that even if we query against (validity_start minus 1 millisecond), the conflict will happen
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 149 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlyqfO2xxeoeRa41GX:-NDlyqfO2xxeoeRa41GY:b-veeoil) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L149)):*
> ```XML
>         buildQuery({
>           label: indefiniteEndLine.label,
>           validityStart: indefiniteEndLine.validity_start,
> ```

let's write explicitly that validity_end = null
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 161 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlzTWo3txQi6vsF1X1:-NDlzTWo3txQi6vsF1X2:bqbeoin) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L161)):*
> ```XML
>         buildQuery({
>           label: indefiniteStartLine.label,
>           validityEnd: indefiniteStartLine.validity_end,
> ```

same as with the previous test: should write explicitly validity_start = null and should also add a test that's not an edge case (validity_end plus 1 millisecond)
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 173 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlzkb04xmn_A8fNv0c:-NDlzkb04xmn_A8fNv0d:bbsotyc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L173)):*
> ```XML
>         buildQuery({
>           label: boundedLine.label,
>           validityStart: boundedLine.validity_start?.plus({ milliseconds: 1 }),
> ```

the "!" operator would be better here instead of the "?" operator. We do know that this is defined.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 186 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlzr_t713pH4HYgcV4:-NDlzr_t713pH4HYgcV5:bq5tvb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L186)):*
> ```XML
>         buildQuery({
>           label: boundedLine.label,
>           validityStart: boundedLine.validity_start?.minus({
> ```

nice! Good that you checked both the inner and the outer intervals!
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 196 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm-Ll7-t4grlhnStGR:-NDm-2GPAfMIBx8IM5_T:b-wog3v3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L196)):*
> ```XML
>   });
> 
>   test('Should not allow new version after current version when bounds overlap', async () => {
> ```

good idea to check this!
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 274 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm-Uaz5fWV963ecGnX:-NDm-Uaz5fWV963ecGnY:b-7a9g5w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L274)):*
> ```XML
>   });
> 
>   describe('Special cases', () => {
> ```

could name this: Checking conflicts with priorities
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 303 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm11so9Uc0RUsDizbo:-NDm11so9Uc0RUsDizbp:b-feee3f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L303)):*
> ```XML
>     });
> 
>     test('Should allow creating conflicting version of the resource itself (=with same id). (Otherwise resources could not be edited. Conflicts do not matter in this case as we are *overwriting* conflicting version)', async () => {
> ```

would it be better to write it like this?
"A resource should not conflict with itself when being modified" and then add a comment below that "but may conflict with other resources"

Also, let's add a test here where the edited bounded line would conflict with another resource

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/346)* status: 0 of 3 files reviewed, 14 unresolved discussions (waiting on @pallost)

___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.ts` line 52 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlpE8f-Wi7l14X14N1:-NE0mt4u7QBzFE8dyZvY:b7turso) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.ts#L52)):*

I know it's not directly part of this PR, but this flag is never used, so this value is always true. I also don't think we should support it, it's not really in the scope of this hook. Let's just shove all the check conditions from buildValidityStartMissingGqlFilterOrConditions to this buildCommonGqlFilter function to see the full picture.

I also don't see a reason to fork the logic based on if the resource is indefinite. Let's just rather put all these conditions on the same level, connected with an OR operator. Having a comment or a visual representation of the intervals should help a lot!
</blockquote></details>

I removed this flag now. 

Refactoring ternary to `_or` operator made some of the tests fail. I can investigate that a bit more tomorrow, but not sure if that improves the readability nor is worth the effort.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 36 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlv0Ym6HERgNHd8rGJ:-NE0m8eC9mK8GiT3krbW:bt0jn8q) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L36)):*

I do see the `null` value useful for the validity periods. Undefined means, well, that it's not defined. Null however means that it IS defined and is defined to be infinite.
</blockquote></details>

Well. I kind of agree with you, but actually `CommonParams` from `ui/src/hooks/useCheckValidityAndPriorityConflicts.ts` have typed these so that validity_start/end are either defined or `undefined`, and this is now symmetrical to that implementation. So if we want to change that, then it would probably make sense to also refactor how `CommonParams` are defined, and that probably causes waterfall effect that requires typings around the whole app adjusted. And that starts to be out of scope of this PR.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 116 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm-yBb7yT4ebzKhVdu:-NE0npmo6DOWIMimEyZm:b-ie6gn9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L116)):*

I'm wondering that is this hook now technically stateful? I understand that for this hook it's not an issue as there's no state within, but I wonder if we should add a comment here that people won't start copy-pasting this pattern to other hook tests but be warned that these hooks are actually stateful and if they need otherwise, they'd need to call the renderHook function within the tests themselves.
</blockquote></details>

Not sure if I fully understand. `useCheckValidityAndPriorityConflicts` isn't stateful, it returns three functions
```
return {
    getConflictingLines,
    getConflictingStops,
    getConflictingRoutes,
  };
```
that are not pure but not "stateful" either. I can add some kind of warning comment if you clarify a bit what you have in mind or suggest the comment, but otherwise I don't see the problem here. I mean yeah, sometimes you should call `renderHook` in each test case, but that's not the case in this test suite.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 121 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlx2_E9bm8aNxUmJUF:-NE0C9qEABggybfwfXn-:bsni1i1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L121)):*

this text is not true, you have tests within that check for bounded validity periods as well. How about this?
`Checking validity period conflicts (label and priority are the same)`
</blockquote></details>

Please note I have only two tests nested under this `describe` block. I don't think your suggestion works here.

Test output is someting like this when writing this comment:

```
  useCheckValidityAndPriorityConflicts()
    ✓ Should return conflict on resources with same validity_start (10 ms)
    ✓ Should return conflict on resources with same validity_end (11 ms)
    ✓ Should return conflict on resource within existing instance of resource (13 ms)
    ✓ Should return conflict on resource starting before and ending after existing version (13 ms)
    ✓ Should not allow new version after current version when bounds overlap (14 ms)
    ✓ Should allow new version after current version, bounds not overlapping, validity_end null (11 ms)
    ✓ Should allow new version after current version, bounds not overlapping, validity_end defined (13 ms)
    ✓ Should not allow new version before current version when bounds overlap (14 ms)
    ✓ Should allow new version before current version, bounds not overlapping, validity_start defined (12 ms)
    ✓ Should allow new version before current version, bounds not overlapping, validity_start null (11 ms)
    Indefinite resources (validity_start and validity_end both null)
      ✓ Should return conflict on two resources with indefinite validity (29 ms)
      ✓ Should return conflict on bounded resource within existsting indefinite resource (11 ms)
    Special cases
      ✓ Should allow creating conflicting version with new priority (10 ms)
      ✓ Should allow creating conflicting version with draft priority (1 ms)
      ✓ Should allow creating conflicting version of the resource itself (=with same id). (Otherwise resources could not be edited. Conflicts do not matter in this case as we are *overwriting* conflicting version) (11 ms)
      Routes
        ✓ Does not allow creating conflicting route with same direction (13 ms)
        ✓ Allow creating conflicting route with opposite direction (9 ms)

```
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 125 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlwaaV8NEfud6BmpAm:-NE0p3xD-K43O5-EKDXN:b-phy6ma) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L125)):*

let's add the full parameter set for clarity: `{label: indefiniteLine.label, validityStart: null, validityEnd: null}`
</blockquote></details>

Yeah, that sound relevant in this case. Done.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 127 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlxVcBCJC56jdvmQz9:-NE0_7F62Yq7_vnLv99H:b-ssybmz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L127)):*

should add an additional expect to every test below to be certain that the correct conflicting resource was found:
`expect(conflicts[0].line_id).toEqual(indefiniteLine.line_id)`
</blockquote></details>

Expects like that added to the tests that had `expect(conflicts.length).toBe(1);` expection.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 130 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlxtIMBWHAVOHhRabm:-NE0RFANCedSczu7Sra7:b-4c5ezp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L130)):*

typo, "existing"
</blockquote></details>

Good catch, done.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 144 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlyxQi8EMSja21XtLK:-NE0lak7BnP_A_c4aH-o:boojmtk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L144)):*

this is an edge case. Would need a more generic test as well where we check that even if we query against (validity_start minus 1 millisecond), the conflict will happen
</blockquote></details>

Does either one of these test cases cover what you asked for?

```
    ✓ Should allow new version before current version, bounds not overlapping, validity_start defined (13 ms)
    ✓ Should allow new version before current version, bounds not overlapping, validity_start null (13 ms)
```
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 149 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlyqfO2xxeoeRa41GX:-NE0lk5lE_T_tV1UHT7T:bj63n0i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L149)):*

let's write explicitly that validity_end = null
</blockquote></details>

Is it relevant for this test? `indefiniteEndLine` variable name already suggests that validityEnd is missing, and that is also default behviour for `buildQuery` function.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 161 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlzTWo3txQi6vsF1X1:-NE0lGch36MfJMRn8NjF:b-kqv5vw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L161)):*

same as with the previous test: should write explicitly validity_start = null and should also add a test that's not an edge case (validity_end plus 1 millisecond)
</blockquote></details>

Same as with the previous test: is validity_start really relevant in this test case, or does it just add unnecessary noise?

Doesn't either one of these test cases already cover what you asked for?

```
    ✓ Should allow new version after current version, bounds not overlapping, validity_end null (10 ms)
    ✓ Should allow new version after current version, bounds not overlapping, validity_end defined (12 ms)
```
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 173 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlzkb04xmn_A8fNv0c:-NE0cH7v3_DGmuXVrVI-:b-hmkwci) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L173)):*

the "!" operator would be better here instead of the "?" operator. We do know that this is defined.
</blockquote></details>

Well, in that case we have to disable eslint, either for each row (which adds unnecessary noise to the tests) or for the whole file or for all test files.

Does it somehow improve test safety if we use `!` instead of `?`? At least this test would fail in either case if validity_start would be undefined.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 186 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlzr_t713pH4HYgcV4:-NE0bt9qBuaRobOI3CCn:b-asvtqm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L186)):*

nice! Good that you checked both the inner and the outer intervals!
</blockquote></details>

Yeah... Just seems like these have to be refactored in the near future to include only date part, without time...
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 196 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm-Ll7-t4grlhnStGR:-NE0_UGg-NKSdbTkHEK2:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L196)):*

good idea to check this!
</blockquote></details>

Done.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 274 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm-Uaz5fWV963ecGnX:-NE0_r7O1-bdtg9sEkEf:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L274)):*

could name this: Checking conflicts with priorities
</blockquote></details>

Done.
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 303 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm11so9Uc0RUsDizbo:-NE0agJX3n9fVYu4F0Pq:b-xja33v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L303)):*

would it be better to write it like this?
"A resource should not conflict with itself when being modified" and then add a comment below that "but may conflict with other resources"

Also, let's add a test here where the edited bounded line would conflict with another resource
</blockquote></details>

Good idea, updated test description.

Not sure if I understand what kind of test you suggest to be added here? Resources can't conflict unless they have same label, and we already have tests covering different cases of operations related to bounded line.

And in this case we are creating version of bounded line that conflicts with existing version of it, but that conflicts is ignored as expalained in the new comment.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/346)* status: 0 of 3 files reviewed, 9 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.ts` line 52 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlpE8f-Wi7l14X14N1:-NE4vEAJ2zZwy98cCm4G:b-jvxw0e) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.ts#L52)):*

I removed this flag now. 

Refactoring ternary to `_or` operator made some of the tests fail. I can investigate that a bit more tomorrow, but not sure if that improves the readability nor is worth the effort.
</blockquote></details>

that is weird... it should work. I don't think that validity_end being undefined should break break anything as you had the same case with validity_start being undefined within buildValidityStartMissingGqlFilterOrConditions

Could you investigate it?
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 36 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlv0Ym6HERgNHd8rGJ:-NE4rvRREobem-RjnRoJ:b32w0x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L36)):*

Well. I kind of agree with you, but actually `CommonParams` from `ui/src/hooks/useCheckValidityAndPriorityConflicts.ts` have typed these so that validity_start/end are either defined or `undefined`, and this is now symmetrical to that implementation. So if we want to change that, then it would probably make sense to also refactor how `CommonParams` are defined, and that probably causes waterfall effect that requires typings around the whole app adjusted. And that starts to be out of scope of this PR.
</blockquote></details>

OK
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 116 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm-yBb7yT4ebzKhVdu:-NE4rzj051gsYMmY_gpN:b-ncqq5k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L116)):*

Not sure if I fully understand. `useCheckValidityAndPriorityConflicts` isn't stateful, it returns three functions
```
return {
    getConflictingLines,
    getConflictingStops,
    getConflictingRoutes,
  };
```
that are not pure but not "stateful" either. I can add some kind of warning comment if you clarify a bit what you have in mind or suggest the comment, but otherwise I don't see the problem here. I mean yeah, sometimes you should call `renderHook` in each test case, but that's not the case in this test suite.
</blockquote></details>

I just mean that IF useCheckValidityAndPriorityConflicts used any variables within (regardless if useState of not), then that state variable would be carried around between the tests, which is to be avoided. How to write this? Well, how about "To make sure that the rendered hook does not carry state between tests, the renderHook function should be called once per test instead of once per suite"?
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 121 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlx2_E9bm8aNxUmJUF:-NE4t9i85Q-LT9xM05Al:bay0w0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L121)):*

Please note I have only two tests nested under this `describe` block. I don't think your suggestion works here.

Test output is someting like this when writing this comment:

```
  useCheckValidityAndPriorityConflicts()
    ✓ Should return conflict on resources with same validity_start (10 ms)
    ✓ Should return conflict on resources with same validity_end (11 ms)
    ✓ Should return conflict on resource within existing instance of resource (13 ms)
    ✓ Should return conflict on resource starting before and ending after existing version (13 ms)
    ✓ Should not allow new version after current version when bounds overlap (14 ms)
    ✓ Should allow new version after current version, bounds not overlapping, validity_end null (11 ms)
    ✓ Should allow new version after current version, bounds not overlapping, validity_end defined (13 ms)
    ✓ Should not allow new version before current version when bounds overlap (14 ms)
    ✓ Should allow new version before current version, bounds not overlapping, validity_start defined (12 ms)
    ✓ Should allow new version before current version, bounds not overlapping, validity_start null (11 ms)
    Indefinite resources (validity_start and validity_end both null)
      ✓ Should return conflict on two resources with indefinite validity (29 ms)
      ✓ Should return conflict on bounded resource within existsting indefinite resource (11 ms)
    Special cases
      ✓ Should allow creating conflicting version with new priority (10 ms)
      ✓ Should allow creating conflicting version with draft priority (1 ms)
      ✓ Should allow creating conflicting version of the resource itself (=with same id). (Otherwise resources could not be edited. Conflicts do not matter in this case as we are *overwriting* conflicting version) (11 ms)
      Routes
        ✓ Does not allow creating conflicting route with same direction (13 ms)
        ✓ Allow creating conflicting route with opposite direction (9 ms)

```
</blockquote></details>

well, this grouping is a bit weird... how about grouping those first 10 tests into a describe block as well ("bounded resources (either validity_start or validity_end are defined)")?
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 303 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDm11so9Uc0RUsDizbo:-NE4uJIO9fUPerYjPwC4:btp54nw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L303)):*

Good idea, updated test description.

Not sure if I understand what kind of test you suggest to be added here? Resources can't conflict unless they have same label, and we already have tests covering different cases of operations related to bounded line.

And in this case we are creating version of bounded line that conflicts with existing version of it, but that conflicts is ignored as expalained in the new comment.
</blockquote></details>

Line1: [yesterday, today]
Line2: [today, tomorrow]
_edit_ Line1 validity_end to be "tomorrow" and expect a conflict

With this, we verify that conflict checking also works also when editing an existing resource, not only when creating one
___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 136 at r14](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NE4tdfh6waOJm7AqQPC:-NE4tdfh6waOJm7AqQPD:bag2ocf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a1bdb64d3bb3c18f0942b2b88afbe426be4e858c/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L136)):*
> ```XML
>     });
> 
>     test('Should return conflict on bounded resource within existsting indefinite resource', async () => {
> ```

existing :)



_Code quote:_
```XML
existsting
```

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/346)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts` line 173 at r9](https://reviewable.io/reviews/hsldevcom/jore4-ui/346#-NDlzkb04xmn_A8fNv0c:-NE4vwTa82N9yuqiEOnU:btdfvnj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/0453715482085e6f800bb814ca3ea09bbcbff591/ui/src/hooks/useCheckValidityAndPriorityConflicts.integration.spec.ts#L173)):*

Well, in that case we have to disable eslint, either for each row (which adds unnecessary noise to the tests) or for the whole file or for all test files.

Does it somehow improve test safety if we use `!` instead of `?`? At least this test would fail in either case if validity_start would be undefined.
</blockquote></details>

that is true, it won't improve type safety... leave it like this.

tk-vincit:
Due to Reasons this was split into two PR's, this one and https://github.com/HSLdevcom/jore4-ui/pull/360. 
Anyway, 360 is now approved and it has same changes as this one, but CI on it doesn't want to run tests.
Will merge this after CI shows green light and that closes both of the PR's.

PR #352, opened 2022-10-05
Define heading styles globally 

tk-vincit:
- Use h1, h2, etc. html tags according to styles

For now on, it should be clear which heading tag
should be used instead of choosing something between
h1-h4 "randomly" and then styling those individually.

Also fixed html tags with text content to be more semantically
correct (e.g. use label tag for observation date inputs instead
of heading tag that was styled to look like label!).

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/352)

### DIFF
--- a/ui/src/components/forms/common/ConfirmSaveForm.tsx
+++ b/ui/src/components/forms/common/ConfirmSaveForm.tsx
@@ -52,11 +52,8 @@ export const ConfirmSaveForm = ({ className = '' }: Props): JSX.Element => {
 
   return (
     <div className={className}>
-      <h2 className="pb-6 text-xl font-bold">
-        {t('saveChangesModal.validityPeriod')}
-      </h2>
-
-      <Row className="mb-4">
+      <h3>{t('saveChangesModal.validityPeriod')}</h3>
+      <Row className="mb-4 pt-6">
         <Column>
           {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
           <label>{t('priority.label')}</label>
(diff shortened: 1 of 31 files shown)


tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/352)* status: 18 of 30 files reviewed, 2 unresolved discussions (waiting on @eleksi)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/352#-NDbpReG5F9tkfa7bPGR):*

Looks good, but please check RoutesAndLinesLists and SearchContainer components for redundant classNames. 
</blockquote></details>

Good catch! Fixed those (+ some other forgotten places).
___
*[`ui/src/styles/globals.css` line 28 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/352#-NDbmHBc9td0YJsqJQaI:-NDbuT1D9Eaw3K2g2sLB:b-4tjehw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/330211fb3580513b0f08f19c3d1fb6c6b4e83fc5/ui/src/styles/globals.css#L28)):*

This is great! If Figma designs are consistent with these, this makes our work easier and code cleaner. But if they are not, it could lead to some confusions.
</blockquote></details>

Figma seems to be pretty consistent so far. But even if its not, that shouldn't be an issue because tailwind (and any other) css classes have higher priority than these global css definitions, so these can be overwritten easily.

eleksi (approved):
> Reviewed 7 of 12 files at r3, 5 of 5 files at r4, 1 of 6 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/352)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tk-vincit)

PR #353, opened 2022-10-05
Check stop draft priority when changing route

culka:
- Check priority for all stops when changing route priority from draft
- Add a confirmation window with options to keep stops or remove them

Resolves HSLdevcom/jore4#839

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/353)

### DIFF
--- a/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx
+++ b/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { DialogWithButtons } from '../../../uiComponents';
+
+interface Props {
+  isOpen: boolean;
+  onConfirm: () => void;
+  onRemoveStops: () => void;
+  onCancel: () => void;
+  routeLabel?: string;
+  stopsLabelsToRemove: string[];
+  className?: string;
+}
+
+export const RouteDraftStopsConfirmationDialog = ({
+  isOpen,
+  onConfirm,
+  onRemoveStops,
+  onCancel,
+  routeLabel,
+  stopsLabelsToRemove,
+  className = '',
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+
+  return (
+    <DialogWithButtons
+      isOpen={isOpen}
+      title={t('confirmRouteDraftStopsDialog.title')}
+      description={t('confirmRouteDraftStopsDialog.description', {
+        routeLabel,
+        stopLabels: stopsLabelsToRemove.join(', '),
+      })}
+      buttons={[
+        { onClick: onCancel, text: t('cancel'), inverted: true },
+        {
+          onClick: onRemoveStops,
+          text: t('confirmRouteDraftStopsDialog.removeStopsText'),
+        },
+        {
+          onClick: onConfirm,
+          text: t('confirmRouteDraftStopsDialog.confirmText'),
+        },
+      ]}
+      onCancel={onCancel}
+      className={className}
+      isWide
+    />
+  );
+};
--- a/ui/src/components/map/stops/DeleteStopConfirmationDialog.tsx
+++ b/ui/src/components/map/stops/DeleteStopConfirmationDialog.tsx
@@ -39,7 +39,7 @@ export const DeleteStopConfirmationDialog: React.FC<Props> = ({
       confirmationTextParts.push(removedRoutesText);
     }
 
-    return confirmationTextParts.join('<br/><br/>');
+    return confirmationTextParts.join('\n\n');
   };
 
   return (
--- a/ui/src/components/map/stops/EditStopConfirmationDialog.tsx
+++ b/ui/src/components/map/stops/EditStopConfirmationDialog.tsx
@@ -39,7 +39,7 @@ export const EditStopConfirmationDialog: React.FC<Props> = ({
       confirmationTextParts.push(removedRoutesText);
     }
 
-    return confirmationTextParts.join('<br/><br/>');
+    return confirmationTextParts.join('\n\n');
   };
 
   return (
--- a/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx
+++ b/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx
@@ -3,12 +3,14 @@ import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 import {
   RouteRoute,
+  ServicePatternScheduledStopPoint,
   useGetRouteDetailsByIdsQuery,
 } from '../../../generated/graphql';
 import { mapRouteResultToRoute } from '../../../graphql';
 import {
   mapRouteToFormState,
   useDeleteRoute,
+  useEditRouteJourneyPattern,
   useEditRouteMetadata,
 } from '../../../hooks';
 import { Container, Row } from '../../../layoutComponents';
@@ -24,6 +26,7 @@ import {
   submitFormByRef,
 } from '../../../utils';
 import { RedirectWithQuery } from '../../common/RedirectWithQuery';
+import { RouteDraftStopsConfirmationDialog } from '../../forms/route/RouteDraftStopsConfirmationDialog';
 import { RoutePropertiesForm } from '../../forms/route/RoutePropertiesForm';
 import { RouteFormState } from '../../forms/route/RoutePropertiesForm.types';
 import {
@@ -41,15 +44,27 @@ const testIds = {
 export const EditRoutePage = (): JSX.Element => {
   const [hasFinishedEditing, setHasFinishedEditing] = useState(false);
   const [isDeleting, setIsDeleting] = useState(false);
+  const [draftStops, setDraftStops] = useState<
+    ServicePatternScheduledStopPoint[]
+  >([]);
+  const [routeFormData, setRouteFormData] = useState<RouteFormState | null>(
+    null,
+  );
 
   const {
     prepareEditRouteMetadata,
+    findDraftStopsOnRoute,
     mapEditRouteMetadataChangesToVariables,
     editRouteMetadata,
     defaultErrorHandler,
   } = useEditRouteMetadata();
   const { deleteRoute, defaultErrorHandler: defaultDeleteErrorHandler } =
     useDeleteRoute();
+  const {
+    prepareDeleteStopFromRoute,
+    mapEditJourneyPatternChangesToVariables,
+    updateRouteGeometryMutation,
+  } = useEditRouteJourneyPattern();
   const [conflicts, setConflicts] = useState<RouteRoute[]>([]);
   const formRef = useRef<ExplicitAny>(null);
   const { id } = useParams<{ id: string }>();
@@ -78,6 +93,41 @@ export const EditRoutePage = (): JSX.Element => {
       const variables = mapEditRouteMetadataChangesToVariables(changes);
       await editRouteMetadata(variables);
       setHasFinishedEditing(true);
+      setRouteFormData(null);
+    } catch (err) {
+      defaultErrorHandler(err);
+    }
+  };
+
+  const onPrepareSubmit = async (form: RouteFormState) => {
+    try {
+      const draftStopsOnRoute = await findDraftStopsOnRoute({
+        routeId: id,
+        oldPriority: route?.priority,
+        form,
+      });
+      if (draftStopsOnRoute.length > 0) {
+        setRouteFormData(form);
+        setDraftStops(draftStopsOnRoute);
+        return;
+      }
+      onSubmit(form);
+    } catch (err) {
+      defaultErrorHandler(err);
+    }
+  };
+
+  const onRemoveStopsFromRoute = async (form: RouteFormState) => {
+    try {
+      if (route) {
+        const changes = prepareDeleteStopFromRoute({
+          route,
+          stopPointLabels: draftStops?.map((stop) => stop.label),
+        });
+        const variables = mapEditJourneyPatternChangesToVariables(changes);
+        await updateRouteGeometryMutation(variables);
+      }
+      onSubmit(form);
     } catch (err) {
       defaultErrorHandler(err);
     }
@@ -123,7 +173,7 @@ export const EditRoutePage = (): JSX.Element => {
                 ref={formRef}
                 className="mb-2 ml-2 p-6"
                 defaultValues={mapRouteToFormState(route)}
-                onSubmit={onSubmit}
+                onSubmit={onPrepareSubmit}
               />
             )}
           </FormContainer>
@@ -169,6 +219,18 @@ export const EditRoutePage = (): JSX.Element => {
         description={t('confirmDeleteRouteDialog.description')}
         confirmText={t('confirmDeleteRouteDialog.confirmText')}
         cancelText={t('cancel')}
+      />
+      <RouteDraftStopsConfirmationDialog
+        isOpen={draftStops?.length > 0}
+        onCancel={() => setDraftStops([])}
+        onRemoveStops={() => {
+          if (routeFormData) onRemoveStopsFromRoute(routeFormData);
+        }}
+        onConfirm={() => {
+          if (routeFormData) onSubmit(routeFormData);
+        }}
+        routeLabel={route?.label}
+        stopsLabelsToRemove={draftStops.map((stop) => stop.label)}
       />
     </div>
   );
--- a/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx
+++ b/ui/src/components/routes-and-lines/line-details/RouteStopsSection.tsx
@@ -69,7 +69,7 @@ export const RouteStopsSection = ({
   const onAddToRoute = async (stopLabel: string) => {
     try {
       const changes = prepareAddStopToRoute({
-        stopPointLabel: stopLabel,
+        stopPointLabels: [stopLabel],
         route,
       });
 
@@ -85,7 +85,7 @@ export const RouteStopsSection = ({
   const onRemoveFromRoute = async (stopLabel: string) => {
     try {
       const changes = prepareDeleteStopFromRoute({
-        stopPointLabel: stopLabel,
+        stopPointLabels: [stopLabel],
         route,
       });
 
--- a/ui/src/generated/graphql.tsx
+++ b/ui/src/generated/graphql.tsx
@@ -11621,42 +11621,58 @@ export type GetRouteDetailsByIdsQuery = {
     priority: number;
     label: string;
     direction: RouteDirectionEnum;
+    infrastructure_links_along_route: Array<{
+      __typename?: 'route_infrastructure_link_along_route';
+      route_id: UUID;
+      infrastructure_link_id: UUID;
+      infrastructure_link_sequence: number;
+      is_traversal_forwards: boolean;
+      infrastructure_link: {
+        __typename?: 'infrastructure_network_infrastructure_link';
+        infrastructure_link_id: UUID;
+        scheduled_stop_points_located_on_infrastructure_link: Array<{
+          __typename?: 'service_pattern_scheduled_stop_point';
+          scheduled_stop_point_id: UUID;
+          label: string;
+          measured_location: GeoJSON.Point;
+          located_on_infrastructure_link_id: UUID;
+          direction: InfrastructureNetworkDirectionEnum;
+          relative_distance_from_infrastructure_link_start: number;
+          closest_point_on_infrastructure_link?:
+            | GeoJSON.Point
+            | null
+            | undefined;
+          validity_start?: luxon.DateTime | null | undefined;
+          validity_end?: luxon.DateTime | null | undefined;
+          priority: number;
+          scheduled_stop_point_in_journey_patterns: Array<{
+            __typename?: 'journey_pattern_scheduled_stop_point_in_journey_pattern';
+            journey_pattern_id: UUID;
+            scheduled_stop_point_label: string;
+            scheduled_stop_point_sequence: number;
+            is_timing_point: boolean;
+            is_via_point: boolean;
+            via_point_name_i18n?: LocalizedString | null | undefined;
+            via_point_short_name_i18n?: LocalizedString | null | undefined;
+            journey_pattern: {
+              __typename?: 'journey_pattern_journey_pattern';
+              journey_pattern_id: UUID;
+              on_route_id: UUID;
+            };
+          }>;
+          vehicle_mode_on_scheduled_stop_point: Array<{
+            __typename?: 'service_pattern_vehicle_mode_on_scheduled_stop_point';
+            vehicle_mode: ReusableComponentsVehicleModeEnum;
+          }>;
+        }>;
+      };
+    }>;
     route_line: {
       __typename?: 'route_line';
       line_id: UUID;
       label: string;
       primary_vehicle_mode: ReusableComponentsVehicleModeEnum;
     };
-    route_journey_patterns: Array<{
-      __typename?: 'journey_pattern_journey_pattern';
-      journey_pattern_id: UUID;
-      on_route_id: UUID;
-      scheduled_stop_point_in_journey_patterns: Array<{
-        __typename?: 'journey_pattern_scheduled_stop_point_in_journey_pattern';
-        journey_pattern_id: UUID;
-        scheduled_stop_point_label: string;
-        scheduled_stop_point_sequence: number;
-        is_timing_point: boolean;
-        is_via_point: boolean;
-        via_point_name_i18n?: LocalizedString | null | undefined;
-        via_point_short_name_i18n?: LocalizedString | null | undefined;
-        scheduled_stop_points: Array<{
-          __typename?: 'service_pattern_scheduled_stop_point';
-          priority: number;
-          direction: InfrastructureNetworkDirectionEnum;
-          scheduled_stop_point_id: UUID;
-          label: string;
-          validity_start?: luxon.DateTime | null | undefined;
-          validity_end?: luxon.DateTime | null | undefined;
-          located_on_infrastructure_link_id: UUID;
-        }>;
-        journey_pattern: {
-          __typename?: 'journey_pattern_journey_pattern';
-          journey_pattern_id: UUID;
-          on_route_id: UUID;
-        };
-      }>;
-    }>;
   }>;
 };
 
@@ -12051,6 +12067,33 @@ export type DeleteRouteMutation = {
       }
     | null
     | undefined;
+};
+
+export type GetScheduledStopsOnRouteQueryVariables = Exact<{
+  routeId: Scalars['uuid'];
+}>;
+
+export type GetScheduledStopsOnRouteQuery = {
+  __typename?: 'query_root';
+  journey_pattern_journey_pattern: Array<{
+    __typename?: 'journey_pattern_journey_pattern';
+    journey_pattern_id: UUID;
+    scheduled_stop_point_in_journey_patterns: Array<{
+      __typename?: 'journey_pattern_scheduled_stop_point_in_journey_pattern';
+      journey_pattern_id: UUID;
+      scheduled_stop_point_sequence: number;
+      scheduled_stop_points: Array<{
+        __typename?: 'service_pattern_scheduled_stop_point';
+        priority: number;
+        direction: InfrastructureNetworkDirectionEnum;
+        scheduled_stop_point_id: UUID;
+        label: string;
+        validity_start?: luxon.DateTime | null | undefined;
+        validity_end?: luxon.DateTime | null | undefined;
+        located_on_infrastructure_link_id: UUID;
+      }>;
+    }>;
+  }>;
 };
 
 export type ScheduledStopPointDefaultFieldsFragment = {
@@ -14558,7 +14601,26 @@ export type GetRouteDetailsByIdQueryResult = Apollo.QueryResult<
 export const GetRouteDetailsByIdsDocument = gql`
   query GetRouteDetailsByIds($route_ids: [uuid!]) {
     route_route(where: { route_id: { _in: $route_ids } }) {
-      ...route_with_journey_pattern_stops
+      ...route_all_fields
+      infrastructure_links_along_route {
+        route_id
+        infrastructure_link_id
+        infrastructure_link_sequence
+        is_traversal_forwards
+        infrastructure_link {
+          infrastructure_link_id
+          scheduled_stop_points_located_on_infrastructure_link {
+            ...scheduled_stop_point_all_fields
+            scheduled_stop_point_in_journey_patterns {
+              ...scheduled_stop_point_in_journey_pattern_all_fields
+              journey_pattern {
+                journey_pattern_id
+                on_route_id
+              }
+            }
+          }
+        }
+      }
       route_line {
         line_id
         label
@@ -14566,7 +14628,9 @@ export const GetRouteDetailsByIdsDocument = gql`
       }
     }
   }
-  ${RouteWithJourneyPatternStopsFragmentDoc}
+  ${RouteAllFieldsFragmentDoc}
+  ${ScheduledStopPointAllFieldsFragmentDoc}
+  ${ScheduledStopPointInJourneyPatternAllFieldsFragmentDoc}
 `;
 
 /**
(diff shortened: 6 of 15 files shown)


culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/353)* status: 0 of 9 files reviewed, 16 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 3 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAUfEl8RN6BRRnqt3A:-NEjmv-UDqM-u8QUgs6g:bxhdsw7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L3)):*

`import { CloseIconButton, SimpleButton } from '../../../uiComponents` (unless there are dependency cycle reasons to use direct imports)
</blockquote></details>

Removed need for these.
___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 11 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAUyS29puT6N7WDRwE:-NEj7mkj1c-luOcawn5d:bkj63rm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L11)):*

This component seems to be highly specific, starting from the name of the component and also looking at the props.
Thus I wonder if it makes any sense to define text contents from outside and pass those as props? I would just put these directly in the component itself, making component interface cleaner.

Or if you want to define actually reusable component, then you probably want to also pass buttons and their onClick callbacks from outside.
</blockquote></details>

Created a DialogWithButtons with variable number of buttons.
___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 42 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAWJlUCHD2qwK7be7V:-NEj7i6-6femaHyZxg0X:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L42)):*

You can avoid defining "text-xl font-bold" classNames by defining this title as `<h3>` tag.
I assume you can write this like this:
`   <Dialog.Title className="flex " as="h3>`
but if that doesn't work, you can just wrap title in h3 tag:

```
<Dialog.Title ...>
<h3>{title}</h3}
</Dialog.Title>

```
</blockquote></details>

Done.
___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 47 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAVlEa5VRhA9RzYSJ4:-NEjKx1b98rwnDbzOugi:b-jvmq2o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L47)):*

Wtf?
Seems like this should be written like this:

```
      <Dialog.Description className="my-5>
        {description}
      </Dialog.Description>
```
</blockquote></details>

I created a new generic DialogWithButtons which inherited this from ConfirmationDialog. Will not change until someone with a better idea how the text creation it should work.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 43 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAXBEZ0WcGily5ToXp:-NEjLPHf5czGQqaymFoK:bbvr834) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L43)):*

Why do these have `has` prefix if these are not boolean flags?

Also wondering why these are defined as arrays? Assuming I parsed this file correctly in my head, seems like these have always either 0 or 1 elements, so array structure doesn't make sense to me.
</blockquote></details>

Renamed to plain `draftStops`. There can be multiple draft stops on the route, so kept as an array.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 79 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAXigr1TE2Qda9uIA8:-NEjLv4N5kqCHWvCJ0_K:bw770f4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L79)):*

`(form: RouteFormState | undefined)` -> `(form?: RouteFormState)`

(Its also a bit confusing why this parameter is optional in first place and what we should do if it is not given.
</blockquote></details>

It should not have been optional at all.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 114 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAZ4bz-zO4JDUAl420:-NEjVlAkCcZaBA4tEbuG:b94lijb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L114)):*

This "onSubmitFinish", "onSubmit" structure doesn't make sense. Seems like "onSubmit" function doesn't actually submit anything, and it then calls "onSubmitFinish" function that **does** submit something!

Seems like this function should be called something like "prepareSubmit". Also wondering if this (and onRemoveStopsFromRoute) logic should be extracted somewhere, e.g.  to`useEditRouteMetadata` or `useEditRouteJourneyPattern` hook, not sure which one these belong to. 
</blockquote></details>

Renamed to the proposed names. This logic was also moved to useEditRouteMetadata. onRemoveStopsFromRoute kept in this file since there is not really logic there, just passing the parameters around, and is done the same way on the geometry edit page.
___
*[`ui/src/graphql/route.ts` line 502 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAahDL97l0Y8dAhzAT:-NEjg1254-DOTZu-nF0B:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/graphql/route.ts#L502)):*

I would rather call this mapper function and not getter.
See this code from this same file as an example.
Please also comment why it is safe to take only first element from arrray and ignore the rest.

```
// The used query has limit: 1 for the result so it can't have more than 1 route_line in result array.
export const mapHighestPriorityLineDetailsWithRoutesResult = (
  result: GqlQueryResult<GetHighestPriorityLineDetailsWithRoutesQuery>,
) =>
  result.data?.route_line.length
    ? (result.data?.route_line[0] as RouteLine)
    : undefined;

```
</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useEditRouteMetadata.ts` line 42 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAcmeZ4wSrmozxISm7:-NEjgEILBcGlcsKj477r:br0b243) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/hooks/routes/useEditRouteMetadata.ts#L42)):*

Kind of same here. Not sure if this interface is even needed at all, or does TS infern return type correctly also without this.
</blockquote></details>

Removed as unnecessary, routeId is not even needed here.
___
*[`ui/src/hooks/routes/useEditRouteMetadata.ts` line 141 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAcvqa-ufrythdqlkF:-NEjhI4c4S5okEF4401B:bt1b19p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/hooks/routes/useEditRouteMetadata.ts#L141)):*

If you can figure out a bit more descriptive name than "mappedResult", that would definitely improve readability of that code.
</blockquote></details>

Renamed.
___
*[`ui/src/hooks/routes/useEditRouteMetadata.ts` line 142 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAcAg819GuOKyfQM3y:-NEjhHAVDzWva0t4_e8d:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/hooks/routes/useEditRouteMetadata.ts#L142)):*

"a" -> "item" or something more descriptive ("stop"? "stopInJourneyPattern"?)
</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useEditRouteMetadata.ts` line 143 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAcJgA3-8XnZ_XVuUF:-NEjhKyj0omY3ZxygiZ2:bqas7wi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/hooks/routes/useEditRouteMetadata.ts#L143)):*

"a" -> "item"
</blockquote></details>

Renamed to a more descriptive one.
___
*[`ui/src/hooks/routes/useEditRouteMetadata.ts` line 183 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAXP44EoPNqN29qpaz:-NEjiPV755ZcYobLwgkG:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/hooks/routes/useEditRouteMetadata.ts#L183)):*

Why do we have this "ForNonDraftRoute" postfix/description? Could and should this rather be something like "findDraftStopsOfRoute"?
</blockquote></details>

Done.
___
*[`ui/src/locales/fi-FI/common.json` line 154 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAbecB9z1VKWkPV_5m:-NEjiRUV1Ux4n4zRQ3dc:b-40eaa6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/locales/fi-FI/common.json#L154)):*

What is "status", does this come from figma or what? If it means "prioriteetti", I would say that it would make much more sense to call it that then.
</blockquote></details>

This was directly from a slack comment on what it should read. But I agree, Proriteetti would be better.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/353)* status: 0 of 13 files reviewed, 12 unresolved discussions (waiting on @culka)

___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 11 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAUyS29puT6N7WDRwE:-NEolWzL4ST1t29j95--:b-ws1b2t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L11)):*

Created a DialogWithButtons with variable number of buttons.
</blockquote></details>

Thanks. Much better.
___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 47 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAVlEa5VRhA9RzYSJ4:-NEoki4U4QoQqmqSf4ej:bxhnwb8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L47)):*

I created a new generic DialogWithButtons which inherited this from ConfirmationDialog. Will not change until someone with a better idea how the text creation it should work.
</blockquote></details>

Moved discussion to the place where this is now, and also mentioned a better idea about how to do this.
___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEokzVZ0MW1JkO-Y2Am:-NEokzV_6lqPeFiTaVl8:bou8rdq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L15)):*
> ```XML
> }
> 
> export const RouteDraftStopsConfirmationDialog: React.FC<Props> = ({
> ```

This doesn't take `children` prop which `React.FC` implicitly have, so should be typed as `JSX.Element` instead of `React.FC`. Didn't verified the syntax, but something like this:
`export const RouteDraftStopsConfirmationDialog: = ( { ... }:Props ): JSX.Element => (`
(Examples available all around the codebase.)
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 50 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEokESP8fj3qoHoRVr5:-NEokESP8fj3qoHoRVr6:b-37jf21) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L50)):*
> ```XML
>     ServicePatternScheduledStopPoint[]
>   >([]);
>   const [hasFormData, setHasFormData] = useState<RouteFormState[]>([]);
> ```

Same question as on previous round: why do these have "has" prefix? That makes it sound like "hasFormData" is boolean flag (or form data of form called "has"), but neither one seems to be the case so these need a better name. Perhaps [routeFormData, setRouteFormData]?

Also do we sometimes have more than one form data stored here, so does this have to be array? From current implementation it seems like this is always populated like this: `   setHasFormData([form]);` and used like `hasFormData[0]`, and if that is the case then this shouldn't be an array at all.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 97 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEofOp86jKYkOcq2bGd:-NEofOp86jKYkOcq2bGe:b-8hg8ju) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L97)):*
> ```XML
>       defaultErrorHandler(err);
>     } finally {
>       setHasFormData([]);
> ```

So, if submitting form fails, we just clear the contents of it? Isn't this a bit annoying for the user? Or did I misunderstood something?
___
*[`ui/src/graphql/route.ts` line 487 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEAa8YcD_lciyPGjNW5:-NEojbOb8CT9fFmCdNBn:bbxl5ll) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e8a632c028757e59ff39b5a42a6044011312d19c/ui/src/graphql/route.ts#L487)):*

I'm surprised if we don't have graphql query for this already, but didn't 
</blockquote></details>

Seems like end of this comment was missing. :smile: Tried to say that I didn't find similar query, so if we really need this then nothing to do here.
___
*[`ui/src/graphql/route.ts` line 514 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEog0yL2lNd2ox1XxCE:-NEog0yL2lNd2ox1XxCF:bfiipvm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/graphql/route.ts#L514)):*
> ```XML
> ) =>
>   result.data?.journey_pattern_journey_pattern[0]?.scheduled_stop_point_in_journey_patterns.flatMap(
>     (sspijp) => sspijp.scheduled_stop_points,
> ```

`sspijp` -> `item`
___
*[`ui/src/hooks/routes/useEditRouteJourneyPattern.ts` line 85 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEogAlYDQ8HDH2li4tt:-NEogAlYDQ8HDH2li4tu:b-av83uu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/hooks/routes/useEditRouteJourneyPattern.ts#L85)):*
> ```XML
> 
>     // Map ServicePatternScheduledStopPoints to RouteStops
>     // const routeStops: RouteStop[] = stopsEligibleForJourneyPattern.map((stop) => {
> ```

Please remove these or explain why we want to have these to version control.



_Code quote:_
```XML
    // Map ServicePatternScheduledStopPoints to RouteStops
    // const routeStops: RouteStop[] = stopsEligibleForJourneyPattern.map((stop) => {
    //  const isStopToEdit = includedStopLabels.includes(stop.label);
```
___
*[`ui/src/uiComponents/ConfirmationDialog.tsx` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEomiaj5cN35ymrHRIo:-NEomiakDGW5LOM9Jaoy:bllpal6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/ConfirmationDialog.tsx#L15)):*
> ```XML
> }
> 
> export const ConfirmationDialog: React.FC<Props> = ({
> ```

Sorry, not part of this PR but also this should be typed as `JSX.Element`.
___
*[`ui/src/uiComponents/ConfirmationDialog.tsx` line 31 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEohJ9_0AMCQU3hhoDz:-NEohJ9_0AMCQU3hhoE-:bpu9kqv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/ConfirmationDialog.tsx#L31)):*
> ```XML
>       description={description}
>       buttons={[
>         { onAction: onCancel, text: cancelText, inverted: true },
> ```

IMO it would improve readability if "text" would be defined before "onAction".
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 7 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEogzCM293_DJg-DVIU:-NEogzCM293_DJg-DVIV:b-g3nxwx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L7)):*
> ```XML
> 
> export interface DialogButton {
>   onAction: () => void;
> ```

onAction -> onClick, buttons have onClick prop and this is passed directly there so it doesn't make sense to invent new name for that here.
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 41 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEojEjy6ngSybaExPLW:-NEojEjy6ngSybaExPLX:b-btu6bx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L41)):*
> ```XML
>     >
>       <div className="flex h-full items-center justify-center">
>         <div
> ```

Do we need both of these divs, or could you just combine these? Now the biggest problem is that we have ambiguous "innerClassName", that is passed to arbitrary div inside this dialog.



_Code quote:_
```XML
      <div className="flex h-full items-center justify-center">
        <div
          className={`w-full rounded-md bg-white p-5 shadow-md ${innerClassName}`}
        >
```
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 50 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEoiDo-9qKB6TfRQzVI:-NEoiDo-9qKB6TfRQzVJ:b-2nkomh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L50)):*
> ```XML
>           <Dialog.Description
>             className="my-5"
>             dangerouslySetInnerHTML={descHtml}
> ```

Still wtf for this... OK, seems like `ui/src/components/map/stops/EditStopConfirmationDialog.tsx` and `ui/src/components/map/stops/DeleteStopConfirmationDialog.tsx` files have used pattern like `return confirmationTextParts.join('<br/><br/>');`, and I assume that is why we have this hack, but probably those could be changed to `return confirmationTextParts.join('\n\n');` and then we could get rid of this... OK, this originates from things not related to this PR so this comment is non-blocking, but still wanted to point this out.
___
*[`ui/src/utils/stops.ts` line 51 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEohfvcCoR23xD1Y8YI:-NEohfvcCoR23xD1Y8YJ:b-pivc9u) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/utils/stops.ts#L51)):*
> ```XML
> ) =>
>   belongsToJourneyPattern
>     ? uniq([...stops, ...stopsToRemove])
> ```

I don't understand this function, at least not anymore. If `belongsToJourneyPattern` is true, we are returning `stopsToRemove` with `stops`?

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/353)* status: 0 of 13 files reviewed, 11 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEokzVZ0MW1JkO-Y2Am:-NF7nEef8rDewI9D1J6i:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/components/forms/route/RouteDraftStopsConfirmationDialog.tsx#L15)):*

This doesn't take `children` prop which `React.FC` implicitly have, so should be typed as `JSX.Element` instead of `React.FC`. Didn't verified the syntax, but something like this:
`export const RouteDraftStopsConfirmationDialog: = ( { ... }:Props ): JSX.Element => (`
(Examples available all around the codebase.)
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 50 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEokESP8fj3qoHoRVr5:-NEpOAbJ-Ky3BJvzGsxI:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L50)):*

Same question as on previous round: why do these have "has" prefix? That makes it sound like "hasFormData" is boolean flag (or form data of form called "has"), but neither one seems to be the case so these need a better name. Perhaps [routeFormData, setRouteFormData]?

Also do we sometimes have more than one form data stored here, so does this have to be array? From current implementation it seems like this is always populated like this: `   setHasFormData([form]);` and used like `hasFormData[0]`, and if that is the case then this shouldn't be an array at all.
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 97 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEofOp86jKYkOcq2bGd:-NEpOCgm07oCMman372D:ba2v7jh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L97)):*

So, if submitting form fails, we just clear the contents of it? Isn't this a bit annoying for the user? Or did I misunderstood something?
</blockquote></details>

Yes, kinda. The formData state holds the form data that is going to be submitted once the user clicks an option in the confirmation dialog. If something goes wrong the interface returns back to the main form where they need to click save again, and then confirm in the dialog.
___
*[`ui/src/graphql/route.ts` line 514 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEog0yL2lNd2ox1XxCE:-NEpPQY_1D5n3AuEGH1O:bl34lxg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/graphql/route.ts#L514)):*

`sspijp` -> `item`
</blockquote></details>

Is that really any better? :slightly_smiling_face: 
___
*[`ui/src/hooks/routes/useEditRouteJourneyPattern.ts` line 85 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEogAlYDQ8HDH2li4tt:-NEpQ3-3EXqilHcDgGgF:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/hooks/routes/useEditRouteJourneyPattern.ts#L85)):*

Please remove these or explain why we want to have these to version control.
</blockquote></details>

Done.
___
*[`ui/src/uiComponents/ConfirmationDialog.tsx` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEomiaj5cN35ymrHRIo:-NEpQhYq9nv4_Tc0lWl7:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/ConfirmationDialog.tsx#L15)):*

Sorry, not part of this PR but also this should be typed as `JSX.Element`.
</blockquote></details>

Done.
___
*[`ui/src/uiComponents/ConfirmationDialog.tsx` line 31 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEohJ9_0AMCQU3hhoDz:-NF7miq04Xvne_QJDmM-:b4ug1nu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/ConfirmationDialog.tsx#L31)):*

IMO it would improve readability if "text" would be defined before "onAction".
</blockquote></details>

Switched them.
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 7 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEogzCM293_DJg-DVIU:-NEpRU1G0X00SiFsEQY7:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L7)):*

onAction -> onClick, buttons have onClick prop and this is passed directly there so it doesn't make sense to invent new name for that here.
</blockquote></details>

Done.
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 41 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEojEjy6ngSybaExPLW:-NEtUSRK1FR6LUPS5iuk:b6gkyms) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L41)):*

Do we need both of these divs, or could you just combine these? Now the biggest problem is that we have ambiguous "innerClassName", that is passed to arbitrary div inside this dialog.
</blockquote></details>

Added a wideDialog boolean property for the element to signify when it should be wide and when not.
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 50 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEoiDo-9qKB6TfRQzVI:-NEt6AHAAFXgmx4zdyFX:b-3wl3ev) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L50)):*

Still wtf for this... OK, seems like `ui/src/components/map/stops/EditStopConfirmationDialog.tsx` and `ui/src/components/map/stops/DeleteStopConfirmationDialog.tsx` files have used pattern like `return confirmationTextParts.join('<br/><br/>');`, and I assume that is why we have this hack, but probably those could be changed to `return confirmationTextParts.join('\n\n');` and then we could get rid of this... OK, this originates from things not related to this PR so this comment is non-blocking, but still wanted to point this out.
</blockquote></details>

Fixed this by changing those string joins and adding `whitespace-pre-line` class to enable line breaks in text. Much better. :slightly_smiling_face: 
___
*[`ui/src/utils/stops.ts` line 51 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEohfvcCoR23xD1Y8YI:-NF-e5KA3nGjgF17HOlm:b-u2zkjr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/utils/stops.ts#L51)):*

I don't understand this function, at least not anymore. If `belongsToJourneyPattern` is true, we are returning `stopsToRemove` with `stops`?
</blockquote></details>

Yeah remove is the wrong name here for the second parameter. The function either adds the provided stops or removes them from the stops -list. Added a proper comment for this function.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/353)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @culka)

___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 41 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEojEjy6ngSybaExPLW:-NF8FRC-6I7vlXyFBL8Z:bdur830) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L41)):*

Added a wideDialog boolean property for the element to signify when it should be wide and when not.
</blockquote></details>

Thanks. Much less confusing now.
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 50 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEoiDo-9qKB6TfRQzVI:-NF8FLgp6PDTu7OL55jo:b-ucfknh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/uiComponents/DialogWithButtons.tsx#L50)):*

Fixed this by changing those string joins and adding `whitespace-pre-line` class to enable line breaks in text. Much better. :slightly_smiling_face: 
</blockquote></details>

Thanks!
___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 20 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NF8IRxK2sgnVfZq3u1v:-NF8IRxK2sgnVfZq3u1w:b-b21uqb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/96470736e26ef2cb7b3ad3abe1d7738d469d1e47/ui/src/uiComponents/DialogWithButtons.tsx#L20)):*
> ```XML
>   onCancel: () => void;
>   className?: string;
>   wideDialog?: boolean;
> ```

I suggest calling this `isWide` to indicate that it is supposed to be boolean flag, or just `wide`. `Dialog` postfix doesn't tell anything IMO but instead makes me wonder what does it mean
___
*[`ui/src/utils/stops.ts` line 51 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NEohfvcCoR23xD1Y8YI:-NF8H_A04BsnATEbsz2D:bu5q3pd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b95dace0be2de9f049b3fc74bf4dd2add7c0ab61/ui/src/utils/stops.ts#L51)):*

Yeah remove is the wrong name here for the second parameter. The function either adds the provided stops or removes them from the stops -list. Added a proper comment for this function.
</blockquote></details>

Thanks. Comment makes this good enough.

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/353)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[`ui/src/uiComponents/DialogWithButtons.tsx` line 20 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/353#-NF8IRxK2sgnVfZq3u1v:-NF8NiSuEKiYCboMqbME:bb6un1j) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/96470736e26ef2cb7b3ad3abe1d7738d469d1e47/ui/src/uiComponents/DialogWithButtons.tsx#L20)):*

I suggest calling this `isWide` to indicate that it is supposed to be boolean flag, or just `wide`. `Dialog` postfix doesn't tell anything IMO but instead makes me wonder what does it mean
</blockquote></details>

Done

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/353)* status: 0 of 15 files reviewed, 1 unresolved discussion

PR #354, opened 2022-10-06
Add route info editing test

sr-vala:
Resolves HSLdevcom/jore4#648

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/354)

### DIFF
--- a/cypress/e2e/editRouteInfo.cy.ts
+++ b/cypress/e2e/editRouteInfo.cy.ts
@@ -1,0 +1,109 @@
+import {
+  buildLine,
+  buildRoute,
+  LineInsertInput,
+  RouteInsertInput,
+  RouteDirectionEnum,
+} from '@hsl/jore4-test-db-manager';
+import { EditRoutePage, LineDetailsPage, TerminusValues } from '../pageObjects';
+import { insertToDbHelper, removeFromDbHelper } from '../utils';
+
+const routeFormTestInputs = {
+  finnishName: 'Muokattu reitin nimi',
+  label: 'Muokattu label',
+  direction: RouteDirectionEnum.Outbound,
+};
+
+const originTestInputs: TerminusValues = {
+  finnishName: 'Muokattu lähtöpaikka FIN',
+  finnishShortName: 'Muokattu lähtöpaikka FIN lyhennys',
+  swedishName: 'Muokattu lähtöpaikka SWE',
+  swedishShortName: 'Muokattu lähtöpaikka SWE lyhennys',
+};
+
+const destinationTestInputs: TerminusValues = {
+  finnishName: 'Muokattu määränpää FIN',
+  finnishShortName: 'Muokattu määränpää FIN lyhennys',
+  swedishName: 'Muokattu määränpää SWE',
+  swedishShortName: 'Muokattu määränpää SWE lyhennys',
+};
+
+const lines: LineInsertInput[] = [
+  {
+    ...buildLine({ label: '3999' }),
+    line_id: '77410a6f-7f6a-4f4f-b303-d03f4a56a38e',
+  },
+];
+
+const routes: RouteInsertInput[] = [
+  {
+    ...buildRoute({ label: '9999' }),
+    route_id: '053a5d5d-5b44-488b-9f52-5839a63c3c21',
+    name_i18n: 'Alkuperäinen nimi',
+    direction: RouteDirectionEnum.Inbound,
+    on_line_id: lines[0].line_id,
+  },
+];
+
+const dbResources = {
+  lines,
+  routes,
+};
+
+const deleteCreatedResources = () => {
+  removeFromDbHelper(dbResources);
+};
+
+describe('Route meta information editing', () => {
+  let editRoutePage: EditRoutePage;
+  let lineDetailsPage: LineDetailsPage;
+
+  before(() => {
+    deleteCreatedResources();
+  });
+  beforeEach(() => {
+    editRoutePage = new EditRoutePage();
+    lineDetailsPage = new LineDetailsPage();
+
+    cy.setupTests();
+    cy.mockLogin();
+    insertToDbHelper(dbResources);
+    editRoutePage.visit(routes[0].route_id);
+  });
+  afterEach(() => {
+    deleteCreatedResources();
+  });
+
+  it("Edits a routes's information", () => {
+    editRoutePage.routePropertiesForm.fillRouteProperties(routeFormTestInputs);
+    editRoutePage.terminusNamesInputs.fillTerminusNameInputsForm(
+      originTestInputs,
+      destinationTestInputs,
+    );
+    editRoutePage.confirmSaveForm.setAsTemporary();
+    editRoutePage.confirmSaveForm.getIndefiniteCheckbox().click();
+    editRoutePage.confirmSaveForm.setStartDate('2022-01-01');
+    editRoutePage.confirmSaveForm.setEndDate('2030-12-31');
+
+    editRoutePage.getSaveRouteButton().click();
+
+    lineDetailsPage.routeStopsTable
+      .getRouteName()
+      .should('contain', routeFormTestInputs.finnishName);
+    lineDetailsPage.routeStopsTable.getRouteHeaderRow(
+      routeFormTestInputs.label,
+    );
+    lineDetailsPage.routeStopsTable.routeDirectionShouldBeOutbound(
+      routeFormTestInputs.label,
+    );
+    lineDetailsPage.routeStopsTable
+      .getRouteValidityPeriod(routeFormTestInputs.label)
+      .should('contain', '1.1.2022 - 31.12.2030');
+
+    editRoutePage.visit(routes[0].route_id);
+    editRoutePage.terminusNamesInputs.verifyOriginValues(originTestInputs);
+    editRoutePage.terminusNamesInputs.verifyDestinationValues(
+      destinationTestInputs,
+    );
+  });
+});
--- a/cypress/e2e/routeStops.cy.ts
+++ b/cypress/e2e/routeStops.cy.ts
@@ -14,7 +14,7 @@ import {
   vehicleSubmodeOnInfrastructureLink,
 } from '@hsl/jore4-test-db-manager';
 import { DateTime } from 'luxon';
-import { LineDetailsPage, Toast } from '../pageObjects';
+import { LineDetailsPage, Toast, RouteStopsTable } from '../pageObjects';
 import { insertToDbHelper, removeFromDbHelper } from '../utils';
 
 const infraLinks: InfraLinkInsertInput[] = [
@@ -122,6 +122,7 @@ const deleteCreatedResources = () => {
 describe('Line details page: stops on route', () => {
   let lineDetailsPage: LineDetailsPage;
   let toast: Toast;
+  let routeStopsTable: RouteStopsTable;
   before(() => {
     deleteCreatedResources();
     insertToDbHelper(dbResources);
@@ -129,6 +130,7 @@ describe('Line details page: stops on route', () => {
   beforeEach(() => {
     lineDetailsPage = new LineDetailsPage();
     toast = new Toast();
+    routeStopsTable = new RouteStopsTable();
 
     cy.setupTests();
     cy.mockLogin();
@@ -137,47 +139,50 @@ describe('Line details page: stops on route', () => {
   after(() => {
     deleteCreatedResources();
   });
+
   it('Verify that stops of route are shown on its list view', () => {
-    lineDetailsPage.toggleRouteSection(routes[0].label);
+    routeStopsTable.toggleRouteSection(routes[0].label);
 
     // verify that stops 0 and 1 are included on route
-    lineDetailsPage.getStopRow(stops[0].label).contains(stops[0].label);
-    lineDetailsPage.getStopRow(stops[1].label).contains(stops[1].label);
+    routeStopsTable.getStopRow(stops[0].label).contains(stops[0].label);
+    routeStopsTable.getStopRow(stops[1].label).contains(stops[1].label);
 
     // stop 2 is not included on route and thus not shown by default
-    lineDetailsPage.getStopRow(stops[2].label).should('not.exist');
+    routeStopsTable.getStopRow(stops[2].label).should('not.exist');
 
     // stop 2 can be shown after toggling unused stops to be visible
-    lineDetailsPage.toggleUnusedStops();
-    lineDetailsPage.getStopRow(stops[2].label).contains(stops[2].label);
+    routeStopsTable.toggleUnusedStops();
+    routeStopsTable.getStopRow(stops[2].label).contains(stops[2].label);
   });
+
   it('User can add stops to route and remove those', () => {
-    lineDetailsPage.toggleRouteSection(routes[0].label);
+    routeStopsTable.toggleRouteSection(routes[0].label);
 
-    lineDetailsPage.toggleUnusedStops();
+    routeStopsTable.toggleUnusedStops();
 
-    lineDetailsPage.addStopToRoute(stops[2].label);
+    routeStopsTable.addStopToRoute(stops[2].label);
     cy.wait('@gqlUpdateRouteJourneyPattern')
       .its('response.statusCode')
       .should('equal', 200);
     // Wait until cache is updated and UI notices that this stop is
     // now included on route
-    lineDetailsPage
+    routeStopsTable
       .getStopRow(stops[2].label)
       .contains('Ei reitin käytössä')
       .should('not.exist');
 
-    lineDetailsPage.removeStopFromRoute(stops[1].label);
+    routeStopsTable.removeStopFromRoute(stops[1].label);
     cy.wait('@gqlUpdateRouteJourneyPattern')
       .its('response.statusCode')
       .should('equal', 200);
 
-    lineDetailsPage.getStopRow(stops[1].label).contains('Ei reitin käytössä');
+    routeStopsTable.getStopRow(stops[1].label).contains('Ei reitin käytössä');
   });
+
   it('User cannot delete too many stops from route', () => {
-    lineDetailsPage.toggleRouteSection(routes[0].label);
+    routeStopsTable.toggleRouteSection(routes[0].label);
     // Route has only 2 stops so user shouldn't be able to remove either of those
-    lineDetailsPage.removeStopFromRoute(stops[0].label);
+    routeStopsTable.removeStopFromRoute(stops[0].label);
     toast
       .getDangerToast()
       .contains('Reitillä on oltava ainakin kaksi pysäkkiä.')
--- a/cypress/pageObjects/EditRouteModal.ts
+++ b/cypress/pageObjects/EditRouteModal.ts
@@ -1,0 +1,13 @@
+export class EditRouteModal {
+  /** Presses the Edit route modal's save button. Can be given forceAction = true
+   * to force the click without waiting for it's actionability (If it's covered
+   * by another element etc.).
+   * https://docs.cypress.io/api/commands/click#Arguments
+   */
+  save(forceAction = false) {
+    return cy
+      .getByTestId('EditRouteModal')
+      .findByTestId('Modal::saveButton')
+      .click({ force: forceAction });
+  }
+}
--- a/cypress/pageObjects/EditRoutePage.ts
+++ b/cypress/pageObjects/EditRoutePage.ts
@@ -1,0 +1,20 @@
+import { UUID } from '../types';
+import { ConfirmSaveForm } from './ConfirmSaveForm';
+import { RoutePropertiesForm } from './RoutePropertiesForm';
+import { TerminusNameInputs } from './TerminusNameInputs';
+
+export class EditRoutePage {
+  routePropertiesForm = new RoutePropertiesForm();
+
+  terminusNamesInputs = new TerminusNameInputs();
+
+  confirmSaveForm = new ConfirmSaveForm();
+
+  visit(routeid: UUID) {
+    cy.visit(`/routes/${routeid}/edit`);
+  }
+
+  getSaveRouteButton() {
+    return cy.getByTestId('EditRoutePage::saveButton');
+  }
+}
--- a/cypress/pageObjects/LineDetailsPage.ts
+++ b/cypress/pageObjects/LineDetailsPage.ts
@@ -1,44 +1,11 @@
 import { UUID } from '../types';
+import { RouteStopsTable } from './RouteStopsTable';
 
 export class LineDetailsPage {
+  routeStopsTable = new RouteStopsTable();
+
   visit(lineId: UUID) {
     cy.visit(`/lines/${lineId}`);
-  }
-
-  toggleUnusedStops() {
-    return cy.getByTestId('show-unused-stops-switch').click();
-  }
-
-  getRouteHeaderRow(routeLabel: string) {
-    return cy.getByTestId(`RouteStopsHeaderRow::${routeLabel}`);
-  }
-
-  toggleRouteSection(routeLabel: string) {
-    return this.getRouteHeaderRow(routeLabel)
-      .getByTestId('RouteStopsHeaderRow::toggleAccordion')
-      .click();
-  }
-
-  getStopRow(stopLabel: string) {
-    return cy.getByTestId(`RouteStopsRow::${stopLabel}`);
-  }
-
-  getStopDropdown(stopLabel: string) {
-    return this.getStopRow(stopLabel).findByTestId('StopActionsDrowdown::menu');
-  }
-
-  addStopToRoute(stopLabel: string) {
-    return this.getStopDropdown(stopLabel)
-      .click()
-      .getByTestId('StopActionsDrowdown::addStopToRouteButton')
-      .click();
-  }
-
-  removeStopFromRoute(stopLabel: string) {
-    return this.getStopDropdown(stopLabel)
-      .click()
-      .getByTestId('StopActionsDrowdown::removeStopFromRouteButton')
-      .click();
   }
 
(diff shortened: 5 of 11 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/354)* status: 0 of 6 files reviewed, 10 unresolved discussions (waiting on @sr-vala and @tk-vincit)

___
*[`cypress/e2e/editRouteInfo.cy.ts` line 41 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgsbEk0L0ejnq1gaTn:-NDgsbEk0L0ejnq1gaTo:b-af4cxq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L41)):*
> ```XML
>   {
>     ...buildLine({ label: '3999' }),
>     line_id: '5dfa82f1-b3f7-4e26-b31d-0d7bd78da0bf',
> ```

This UUID is not unique and might thus cause problems that might be very difficult to reproduce and/or debug e.g. if we happen to run tests in parallel at some point.

Here's unique UUID, please use it instead: `77410a6f-7f6a-4f4f-b303-d03f4a56a38e`
and another one for route_id: `053a5d5d-5b44-488b-9f52-5839a63c3c21`



_Code quote:_
```XML
5dfa82f1-b3f7-4e26-b31d-0d7bd78da0bf
```
___
*[`cypress/e2e/editRouteInfo.cy.ts` line 42 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgt3ugAxCdvaVQqRKG:-NDgt3ugAxCdvaVQqRKH:b-f6ea3m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L42)):*
> ```XML
>     ...buildLine({ label: '3999' }),
>     line_id: '5dfa82f1-b3f7-4e26-b31d-0d7bd78da0bf',
>     validity_start: DateTime.fromISO('2020-05-01 23:11:32Z'),
> ```

These are not valid ISO timestamps and thos can't be parsed as such. https://en.wikipedia.org/wiki/ISO_8601
If tests still suggeeds, I assume these are not needed in the test and should be removed.
If these are really needed, I'd investigate why tests still pass and what is happening here.
___
*[`cypress/e2e/editRouteInfo.cy.ts` line 110 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDh0SJ6Ej3b4y6XCEzL:-NDh0SJ6Ej3b4y6XCEzM:b63acij) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L110)):*
> ```XML
>     lineDetailsPage
>       .getRouteDirection(routeFormTestInputs.label)
>       .should('contain', '1');
> ```

Where does this magical '1' come from and what does it mean? This should be replaced with something more meaningful, `RouteDirectionEnum.Inbound` (assuming 1 means Inbound), or if that doesn't work then `String(RouteDirectionEnum.Inbound)` would still be much easier to understand than this.
___
*[`cypress/e2e/editRouteInfo.cy.ts` line 114 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgxHZu7IaWxBytDX9q:-NDgxHZu7IaWxBytDX9r:b2cmf54) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L114)):*
> ```XML
>       .getRouteValidityPeriod(routeFormTestInputs.label)
>       .should('contain', '1.1.2022 - 31.12.2030');
>     editRoutePage.visit(routes[0].route_id);
> ```

Maybe add newline before this statement, this changes route and thus clearly starts new block in this test. (I was pretty confused at first, since this test jumps between two routes and makes one transition implicitly when saveRouteButton is clicked. I guess that is is OK in this kind of case, but makes the test a bit more difficult to follow. I would even consider adding comments about that to this test case, but not sure if that is needed).
___
*[`cypress/pageObjects/LineDetailsPage.ts` line 67 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgyJbt15S0CTYNSQy6:-NDgyJbt15S0CTYNSQy7:b535ybt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/pageObjects/LineDetailsPage.ts#L67)):*
> ```XML
> 
>   getRouteName() {
>     return cy.getByTestId('RouteStopsHeaderRow::name');
> ```

`RouteStopsHeaderRow` is not direct children of `LineDetailsPage`. Went through parent-children hiearchy in this case and noticed that between `RouteStopsHeaderRow` and `LineDetailsPage` is `RouteStopsTable`, which is used in more than one place, so if these are kept here, we might need to define exactly same rows in some another page object! 

So, I would suggest you to create new pageObject for `RouteStopsTable`, which at least sounds like it forms logical building block of this app. And then move these there. 
`DirectionBadge::value` doesn't sound like it merits its own page object, so in this case I would put also it to `RouteStopsTablePageObject`.

(And if you think that from test's perspective it nicer to have everything accessible from here, you could include RouteStopsTablePageObject to this page object and add `getRouteStopsTable()` function here, so that test itself doesn't get more clumsy.)
___
*[`cypress/pageObjects/RoutePropertiesForm.ts` line 10 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDguv4D1eylofpywuJj:-NDguv4E3xt9M3mcrieW:bbuj37k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/pageObjects/RoutePropertiesForm.ts#L10)):*
> ```XML
> }
> 
> export interface OriginValues {
> ```

This interface is exported, but not used outside of this component..? I guess you'd want to use these in `cypress/e2e/editRouteInfo.cy.ts` to type test data objects.
___
*[`cypress/pageObjects/RoutePropertiesForm.ts` line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgvAj8AAmswmV8rD04:-NDgvAj8AAmswmV8rD05:b9hpy10) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/pageObjects/RoutePropertiesForm.ts#L17)):*
> ```XML
> }
> 
> export interface DestinationValues {
> ```

This interface is exactly same than previous interface. Perhaps you could one of these something like `RouteTerminusInputs" and get rid of another.
___
*[`cypress/pageObjects/RoutePropertiesForm.ts` line 71 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDh-jH12eLpDkw2grpa:-NDh-jH12eLpDkw2grpb:bwnh3gr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/pageObjects/RoutePropertiesForm.ts#L71)):*
> ```XML
> 
>   getOriginNameFinnishInput() {
>     return cy.getByTestId('TerminusNameInputs::origin::finnishNameInput');
> ```

These seem to be used only here, but maybe also these (TerminusNameInputs related things) form their own logical entity and would thus go to their own page object. Then that pageObject could be included in this pageObject and at least fill/verify functions could be exposed from this page object to the end user.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 139 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgwgtM-2oVKb2TLtnc:-NDgwgtM-2oVKb2TLtnd:bqn9f8t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L139)):*
> ```XML
>             onClick={() => setIsDeleting(true)}
>             inverted
>             data-testid={testIds.deleteButton}
> ```

SimpleButton does not support `data-testid` prop.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 148 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgwlLwAabIDLxwvZSG:-NDgwlLwAabIDLxwvZSH:bqn9f8t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L148)):*
> ```XML
>             onClick={onCancel}
>             inverted
>             data-testid={testIds.cancelButton}
> ```

SimpleButton does not support `data-testid` prop.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/354)* status: 0 of 11 files reviewed, 9 unresolved discussions (waiting on @tk-vincit)

___
*[`cypress/e2e/editRouteInfo.cy.ts` line 41 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgsbEk0L0ejnq1gaTn:-NDlvYmoB_ViV66ci73U:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L41)):*

This UUID is not unique and might thus cause problems that might be very difficult to reproduce and/or debug e.g. if we happen to run tests in parallel at some point.

Here's unique UUID, please use it instead: `77410a6f-7f6a-4f4f-b303-d03f4a56a38e`
and another one for route_id: `053a5d5d-5b44-488b-9f52-5839a63c3c21`
</blockquote></details>

Done.
___
*[`cypress/e2e/editRouteInfo.cy.ts` line 42 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgt3ugAxCdvaVQqRKG:-NDlva0H3M-0NKWHx-JH:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L42)):*

These are not valid ISO timestamps and thos can't be parsed as such. https://en.wikipedia.org/wiki/ISO_8601
If tests still suggeeds, I assume these are not needed in the test and should be removed.
If these are really needed, I'd investigate why tests still pass and what is happening here.
</blockquote></details>

Done.
___
*[`cypress/e2e/editRouteInfo.cy.ts` line 110 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDh0SJ6Ej3b4y6XCEzL:-NDlvkqL9FG8VvlT0Uey:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L110)):*

Where does this magical '1' come from and what does it mean? This should be replaced with something more meaningful, `RouteDirectionEnum.Inbound` (assuming 1 means Inbound), or if that doesn't work then `String(RouteDirectionEnum.Inbound)` would still be much easier to understand than this.
</blockquote></details>

Done.
___
*[`cypress/e2e/editRouteInfo.cy.ts` line 114 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgxHZu7IaWxBytDX9q:-NDlvmXb2agcQ78PuLdN:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L114)):*

Maybe add newline before this statement, this changes route and thus clearly starts new block in this test. (I was pretty confused at first, since this test jumps between two routes and makes one transition implicitly when saveRouteButton is clicked. I guess that is is OK in this kind of case, but makes the test a bit more difficult to follow. I would even consider adding comments about that to this test case, but not sure if that is needed).
</blockquote></details>

Done.
___
*[`cypress/pageObjects/LineDetailsPage.ts` line 67 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgyJbt15S0CTYNSQy6:-NDlwkn9879Hrsha7fBX:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/pageObjects/LineDetailsPage.ts#L67)):*

`RouteStopsHeaderRow` is not direct children of `LineDetailsPage`. Went through parent-children hiearchy in this case and noticed that between `RouteStopsHeaderRow` and `LineDetailsPage` is `RouteStopsTable`, which is used in more than one place, so if these are kept here, we might need to define exactly same rows in some another page object! 

So, I would suggest you to create new pageObject for `RouteStopsTable`, which at least sounds like it forms logical building block of this app. And then move these there. 
`DirectionBadge::value` doesn't sound like it merits its own page object, so in this case I would put also it to `RouteStopsTablePageObject`.

(And if you think that from test's perspective it nicer to have everything accessible from here, you could include RouteStopsTablePageObject to this page object and add `getRouteStopsTable()` function here, so that test itself doesn't get more clumsy.)
</blockquote></details>

Done.
___
*[`cypress/pageObjects/RoutePropertiesForm.ts` line 10 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDguv4D1eylofpywuJj:-NDlwooSA1mXPWd5YpkK:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/pageObjects/RoutePropertiesForm.ts#L10)):*

This interface is exported, but not used outside of this component..? I guess you'd want to use these in `cypress/e2e/editRouteInfo.cy.ts` to type test data objects.
</blockquote></details>

Done.
___
*[`cypress/pageObjects/RoutePropertiesForm.ts` line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgvAj8AAmswmV8rD04:-NDlwpaw04J8tZcIPT0-:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/pageObjects/RoutePropertiesForm.ts#L17)):*

This interface is exactly same than previous interface. Perhaps you could one of these something like `RouteTerminusInputs" and get rid of another.
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 139 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgwgtM-2oVKb2TLtnc:-NDlwwqp1StLxzETxiO2:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L139)):*

SimpleButton does not support `data-testid` prop.
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx` line 148 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDgwlLwAabIDLxwvZSG:-NDlwxAX2RmtgkwbMdjw:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/ui/src/components/routes-and-lines/edit-route/EditRoutePage.tsx#L148)):*

SimpleButton does not support `data-testid` prop.
</blockquote></details>

Done.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/354)* status: 0 of 11 files reviewed, 6 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/editRouteInfo.cy.ts` line 110 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NDh0SJ6Ej3b4y6XCEzL:-NE-dqsK9bcJUJhYUtX6:b-an4j9t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a54bb50917e18da16e95538b9c754b68dbd6e251/cypress/e2e/editRouteInfo.cy.ts#L110)):*
<details><summary><i>Previously, sr-vala wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Let's continue this discussion in another thread related to this.
___
*[`cypress/e2e/editRouteInfo.cy.ts` line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-fHbj-z4LFg95LZT-:-NE-fHbj-z4LFg95LZT0:b-3vrso3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/e2e/editRouteInfo.cy.ts#L22)):*
> ```XML
>   label: 'Muokattu label',
>   direction: RouteDirectionEnum.Outbound,
>   line: '3999',
> ```

Property "line" is ambiguous. Should rather be `line_id`, assuming this is supposed to be ID.
Btw, seems like this is not used anywhere maybe this should be just removed.
___
*[`cypress/pageObjects/EditRoutePage.ts` line 2 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-g7_08OmKlrcpjLua:-NE-g7_13VRAB_JKCTfr:b-xzn54m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/pageObjects/EditRoutePage.ts#L2)):*
> ```XML
> import { UUID } from '../types';
> ```

We can discuss a bit more about this in Slack and decide if this is even in the scope of this PR, but anyway: it might be nice for the user to include page objects of descendants of EditRoutePage here, so that it would be easier to write tests for EditRoutePage without needing to include multiple child page objects to each tests.

Something like this;
```
let routesTable;
let anotherChildPage;

constructor() {
  routesTable = new RoutesTablePageObject();
  anotherChildPage = new AnotherChildPagePageObject();
}

```

Then you could do something like this in tests:

```
editRoutePagePageObject.routesTable.someRoutesTableMethod()
```
___
*[`cypress/pageObjects/LineDetailsPage.ts` line 5 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-bQom3To_8v-H_0dx:-NE-bQom3To_8v-H_0dy:b-gh8x0i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/pageObjects/LineDetailsPage.ts#L5)):*
> ```XML
> 
> export class LineDetailsPage {
>   routeStopsTable = new RouteStopsTable();
> ```

Is this used somewhere..?
___
*[`cypress/pageObjects/RoutePropertiesForm.ts` line 9 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-hBi43uAYXzqhYQPf:-NE-hBi43uAYXzqhYQPg:bixe7j7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a25475ea9187ea709bfbc97db8b75f0b1410b2a9/cypress/pageObjects/RoutePropertiesForm.ts#L9)):*
> ```XML
>   line: string;
> }
> ```

Maybe there should rather still be newline.
___
*[`cypress/pageObjects/RouteStopsTable.ts` line 1 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-cyWv9JOthmd4eTqB:-NE-cyWv9JOthmd4eTqC:bhukk8x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/pageObjects/RouteStopsTable.ts#L1)):*
> ```XML
> export class RouteStopsTable {
> ```

This pageObject seems to make sense and wraps functions related to descendants of RouteStopsTable nicely. :+1: 
___
*[`cypress/pageObjects/RouteStopsTable.ts` line 39 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-dEsU9PbSHIcIOBK7:-NE-dEsU9PbSHIcIOBK8:b-id4eui) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/pageObjects/RouteStopsTable.ts#L39)):*
> ```XML
> 
>   getRouteName() {
>     return cy.getByTestId('RouteStopsHeaderRow::name');
> ```

It would make more sense to locate this with rest of the RouteStopsHeaderRow related functions, e.g. between toggleRouteSection and getStopRow, right?
___
*[`cypress/pageObjects/RouteStopsTable.ts` line 55 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-cOL7-jpsJVwGOjcz:-NE-cOL7-jpsJVwGOjd-:b-l927e4) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/pageObjects/RouteStopsTable.ts#L55)):*
I'm quoting my previous comment here:
> Where does this magical '1' come from and what does it mean? This should be replaced with something more meaningful, RouteDirectionEnum.Inbound (assuming 1 means Inbound), or if that doesn't work then String(RouteDirectionEnum.Inbound) would still be much easier to understand than this.

Was there a reason why this could't be written like `String(RouteDirectionEnum.Outbound)`? Same for the similar row below.
___
*[`cypress/pageObjects/RouteStopsTable.ts` line 58 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-e7ly0KeG_n61wwD1:-NE-e7ly0KeG_n61wwD2:b-s5e9wq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/pageObjects/RouteStopsTable.ts#L58)):*
> ```XML
>   }
> 
>   routeDirectionShouldBeInbound(routeLabel: string) {
> ```

How about writing these somehow like this, to avoid duplication:

```
verifyRouteDirectionByRouteLabel(routeLabel: string, direction: RouteDirectionEnum) {
  this.getRouteDirection(routeLabel).should('contain', String(direction));
}
```

(and I would remove String casting if that is not needed)

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/354)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/pageObjects/EditRoutePage.ts` line 7 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE0gbFQ9uJvxGBNnhL1:-NE0gbFRBDIj18MstU9o:b4a52vn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/388ef6f04e9c388ed98ef5309e947518be1b485c/cypress/pageObjects/EditRoutePage.ts#L7)):*
> ```XML
> 
> export class EditRoutePage {
>   routePropertiesForm;
> ```

This syntax looks bad. :see_no_evil: 

Noticed from `cypress/pageObjects/MapItemCreator.ts` that we don't actually need constructor at all.

So you could replace these and costructor with this:

```
routePropertiesForm = new RoutePropertiesForm();
terminusNamesInputs = new TerminusNameInputs();
confirmSaveForm = new ConfirmSaveForm();

```
___
*[`cypress/pageObjects/LineDetailsPage.ts` line 5 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE-bQom3To_8v-H_0dx:-NE0eGso5_Sj_x2nspa5:b-483o82) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/519b1e0df89dc3c0a6ed3fec5db2c07e27b783a5/cypress/pageObjects/LineDetailsPage.ts#L5)):*

Is this used somewhere..?
</blockquote></details>

Sorry, noticed from `cypress/pageObjects/MapItemCreator.ts` that we don't actually need constructor at all.

So you could remove constructor and just write:

```
routeStopsTable = new RouteStopsTable();
```

(Sorry, I know I suggested using constructor in the first place)
___
*[`cypress/pageObjects/RouteStopsTable.ts` line 54 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE0h4uMBbDKxwYlh_E0:-NE0h4uN0hdRFtAmb9wQ:b-n8hvxi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/388ef6f04e9c388ed98ef5309e947518be1b485c/cypress/pageObjects/RouteStopsTable.ts#L54)):*
> ```XML
>   }
> 
>   // TODO: check later if RouteDirectionEnum could be used instead of the numbers 1 and 2
> ```

Thanks for the comment!

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/354)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[`cypress/pageObjects/RouteStopsTable.ts` line 54 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/354#-NE0h4uMBbDKxwYlh_E0:-NE4QDRp5_ew-LbeY0e8:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/388ef6f04e9c388ed98ef5309e947518be1b485c/cypress/pageObjects/RouteStopsTable.ts#L54)):*

Thanks for the comment!
</blockquote></details>

Done.

PR #355, opened 2022-10-06
Extend search filters

HenrikHartiala:
Extend search filters by adding the primaryVehicleMode dropdown to advanced search.

Before that in this PR the GQL filter construction is refactored so that the generated filters from the search criteria are a bit different for `route filter` and `line filter`. Some of the advanced search filters are precisely _line's properties_, but the routes can be filtered with the same criteria, but the GQL filter has to be generated little bit differently. (We need to wrap the GQL filter within `route_line`.

In addition to that had to do some refactoring to EnumDropdown and VehicleModeDropdown to support `AllOption`, so that we can reuse the existing components.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/355)

### DIFF
--- a/ui/src/components/forms/common/EnumDropdown.tsx
+++ b/ui/src/components/forms/common/EnumDropdown.tsx
@@ -10,6 +10,7 @@ const testIds = {
 // this was implemented by processing the Enum values as strings.
 
 export interface EnumDropdownProps<TEnum> extends FormInputProps {
+  id?: string;
   testId?: string;
   // eslint-disable-next-line @typescript-eslint/ban-types
   enumType: Object;
@@ -24,6 +25,7 @@ export interface EnumDropdownProps<TEnum> extends FormInputProps {
  */
 // eslint-disable-next-line @typescript-eslint/ban-types
 export function EnumDropdown<TEnum extends Object>({
+  id,
   testId,
   enumType,
   uiNameMapper,
@@ -54,6 +56,7 @@ export function EnumDropdown<TEnum extends Object>({
 
   return (
     <Listbox
+      id={id}
       testId={testId || testIds.enumDropdown}
       buttonContent={
         value ? uiNameMapper(value as unknown as TEnum) : placeholder
--- a/ui/src/components/forms/common/EnumDropdown.tsx
+++ b/ui/src/components/forms/common/EnumDropdown.tsx
@@ -1,5 +1,5 @@
 import { FormInputProps, Listbox } from '../../../uiComponents';
-import { getEnumValues } from '../../../utils';
+import { AllOptionEnum, getEnumValues } from '../../../utils';
 
 const testIds = {
   enumDropdown: 'EnumDropdown::button',
@@ -15,8 +15,13 @@ export interface EnumDropdownProps<TEnum> extends FormInputProps {
   enumType: Object;
   uiNameMapper: (key: TEnum) => string;
   placeholder: string;
+  includeAllOption?: boolean;
 }
 
+/**
+ * Creates dropdown from enum values. This dropdown can be enrichted with 'All' option by giving
+ * it the includeAllOption flag as true.
+ */
 // eslint-disable-next-line @typescript-eslint/ban-types
 export function EnumDropdown<TEnum extends Object>({
   testId,
@@ -24,9 +29,12 @@ export function EnumDropdown<TEnum extends Object>({
   uiNameMapper,
   placeholder,
   value,
+  includeAllOption,
   ...formInputProps
 }: EnumDropdownProps<TEnum>) {
-  const values = getEnumValues(enumType);
+  const values = getEnumValues(
+    includeAllOption ? { ...AllOptionEnum, ...enumType } : enumType,
+  );
 
   const mapToOption = (item: string) => ({
     key: item,
--- a/ui/src/components/forms/line/VehicleModeDropdown.tsx
+++ b/ui/src/components/forms/line/VehicleModeDropdown.tsx
@@ -6,10 +6,15 @@ import { EnumDropdown } from '../common/EnumDropdown';
 
 interface Props extends FormInputProps {
   testId?: string;
+  includeAllOption?: boolean;
 }
 
+/** Creates VehicleModeDropdown from ReusableComponentsVehicleModeEnum. This Dropdown can be
+ * enrichted with 'All' option by giving it includeAllOption flag as true.
+ */
 export const VehicleModeDropdown = ({
   testId,
+  includeAllOption,
   ...formInputProps
 }: Props): JSX.Element => {
   const { t } = useTranslation();
@@ -20,6 +25,7 @@ export const VehicleModeDropdown = ({
       enumType={ReusableComponentsVehicleModeEnum}
       placeholder={t('lines.chooseVehicleMode')}
       uiNameMapper={mapVehicleModeToUiName}
+      includeAllOption={includeAllOption}
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...formInputProps}
     />
--- a/ui/src/components/forms/line/VehicleModeDropdown.tsx
+++ b/ui/src/components/forms/line/VehicleModeDropdown.tsx
@@ -5,6 +5,7 @@ import { FormInputProps } from '../../../uiComponents';
 import { EnumDropdown } from '../common/EnumDropdown';
 
 interface Props extends FormInputProps {
+  id?: string;
   testId?: string;
   includeAllOption?: boolean;
 }
@@ -13,6 +14,7 @@ interface Props extends FormInputProps {
  * enrichted with 'All' option by giving it includeAllOption flag as true.
  */
 export const VehicleModeDropdown = ({
+  id,
   testId,
   includeAllOption,
   ...formInputProps
@@ -21,6 +23,7 @@ export const VehicleModeDropdown = ({
 
   return (
     <EnumDropdown<ReusableComponentsVehicleModeEnum>
+      id={id}
       testId={testId}
       enumType={ReusableComponentsVehicleModeEnum}
       placeholder={t('lines.chooseVehicleMode')}
--- a/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx
+++ b/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx
@@ -3,6 +3,9 @@ import { useSearch } from '../../../../hooks';
 import { useToggle } from '../../../../hooks/useToggle';
 import { Column, Container, Row, Visible } from '../../../../layoutComponents';
 import { SimpleButton } from '../../../../uiComponents';
+import { AllOptionEnum } from '../../../../utils';
+import { FormRow } from '../../../forms/common';
+import { VehicleModeDropdown } from '../../../forms/line/VehicleModeDropdown';
 import { PriorityCondition } from './PriorityCondition';
 import { SearchConditionToggle } from './SearchConditionsToggle';
 import { SearchInput } from './SearchInput';
@@ -15,6 +18,12 @@ export const SearchContainer = (): JSX.Element => {
   const testIds = {
     searchInput: 'SearchContainer::SearchInput',
   };
+
+  const onChangeVehiclemode = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchCondition('primaryVehicleMode', e.target.value);
+  };
+
+  const vehicleModeDropdownId = 'search.primaryVehicleMode';
 
   return (
     <Container className="py-10">
@@ -38,12 +47,27 @@ export const SearchContainer = (): JSX.Element => {
         </Column>
       </Row>
       <Visible visible={isExpanded}>
-        <div className="border-2 border-background p-10">
+        <div className="space-y-5 border-2 border-background p-10">
           <h2>{t('search.advancedSearchTitle')}</h2>
-          <PriorityCondition
-            onClick={setSearchCondition}
-            priorities={searchConditions.priorities}
-          />
+          <FormRow mdColumns={4}>
+            <Column>
+              <label htmlFor={vehicleModeDropdownId}>
+                {t(`lines.primaryVehicleMode`)}
+              </label>
+              <VehicleModeDropdown
+                id={vehicleModeDropdownId}
+                onChange={onChangeVehiclemode}
+                includeAllOption
+                value={searchConditions.primaryVehicleMode ?? AllOptionEnum.All}
+              />
+            </Column>
+          </FormRow>
+          <FormRow mdColumns={4}>
+            <PriorityCondition
+              onClick={setSearchCondition}
+              priorities={searchConditions.priorities}
+            />
+          </FormRow>
         </div>
         <Row className="flex justify-end bg-background py-4">
           <SimpleButton
--- a/ui/src/hooks/search/useSearchQueryParser.ts
+++ b/ui/src/hooks/search/useSearchQueryParser.ts
@@ -1,9 +1,12 @@
+import { ReusableComponentsVehicleModeEnum } from '../../generated/graphql';
 import { Priority } from '../../types/Priority';
+import { AllOptionEnum } from '../../utils/enum';
 import { useUrlQuery } from '../urlQuery/useUrlQuery';
 
 export type SearchConditions = {
   priorities: Priority[];
   label: string;
+  primaryVehicleMode?: ReusableComponentsVehicleModeEnum | AllOptionEnum;
 };
 
 /** Enum for different search result options */
@@ -31,6 +34,7 @@ export type SearchParameters = {
 export type QueryStringParameters = {
   priorities: string;
   label: string;
+  primaryVehicleMode?: string;
   displayedData: string;
 };
 
@@ -41,6 +45,7 @@ export type QueryStringParameters = {
 export type DeserializedQueryStringParameters = {
   priorities: Priority[];
   label: string;
+  primaryVehicleMode?: ReusableComponentsVehicleModeEnum | AllOptionEnum;
   displayedData: DisplayedSearchResultType;
 };
 
@@ -72,6 +77,22 @@ const mapDisplayedData = (displayedData: string) => {
 };
 
 /**
+ * If primaryVehicleMode query string is not given or is invalid,
+ * default to All option
+ */
+const mapPrimaryVehicleModeToEnum = (primaryVehicleMode?: string) => {
+  if (
+    Object.values(ReusableComponentsVehicleModeEnum).includes(
+      primaryVehicleMode as ReusableComponentsVehicleModeEnum,
+    )
+  ) {
+    return primaryVehicleMode as ReusableComponentsVehicleModeEnum;
+  }
+
+  return AllOptionEnum.All;
+};
+
+/**
  * Returns parsed and validated priorities if priority query string
  * is existing. If the query string is not given, return default
  * priorities
@@ -96,6 +117,9 @@ const deserializeParameters = (
     search: {
       label: queryParams.label || '',
       priorities: getPriorities(queryParams.priorities),
+      primaryVehicleMode: mapPrimaryVehicleModeToEnum(
+        queryParams.primaryVehicleMode,
+      ),
     },
     filter: {
       displayedData: mapDisplayedData(queryParams.displayedData),
--- a/ui/src/i18n/uiNameMappings.ts
+++ b/ui/src/i18n/uiNameMappings.ts
@@ -7,6 +7,7 @@ import {
 import { i18n } from '../i18n';
 import { Priority } from '../types/Priority';
 import { RouteDirection } from '../types/RouteDirection';
+import { AllOptionEnum } from '../utils';
 
 export const mapPriorityToUiName = (key: Priority) => {
   const uiStrings: Record<Priority, string> = {
@@ -18,8 +19,8 @@ export const mapPriorityToUiName = (key: Priority) => {
 };
 
 export const mapVehicleModeToUiName = (
-  key: ReusableComponentsVehicleModeEnum,
-) => i18n.t(`vehicleModeEnum.${key}`);
+  key: ReusableComponentsVehicleModeEnum | AllOptionEnum.All,
+) => i18n.t(key === AllOptionEnum.All ? key : `vehicleModeEnum.${key}`);
 
 export const mapLineTypeToUiName = (key: RouteTypeOfLineEnum) =>
   i18n.t(`lineTypeEnum.${key}`);
--- a/ui/src/locales/en-US/common.json
+++ b/ui/src/locales/en-US/common.json
@@ -164,7 +164,8 @@
     "ferry": "Ferry",
     "metro": "Metro",
     "train": "Train",
-    "tram": "Tram"
+    "tram": "Tram",
+    "all": "All"
   },
   "lineTypeEnum": {
     "regional_rail_service": "Regional rail service",
@@ -259,6 +260,7 @@
   "close": "Close",
   "version": "Version: {{ version }}",
   "hide": "Hide",
+  "all": "All",
   "welcomePage": {
     "heading": "JORE4:n käyttöönotto alkaa!",
     "subheading1": "Mitä?",
--- a/ui/src/locales/fi-FI/common.json
+++ b/ui/src/locales/fi-FI/common.json
@@ -164,7 +164,8 @@
     "ferry": "Lautta",
     "metro": "Metro",
     "train": "Juna",
-    "tram": "Raitiovaunu"
+    "tram": "Raitiovaunu",
+    "all": "Kaikki"
   },
   "lineTypeEnum": {
     "regional_rail_service": "Seudullinen juna",
@@ -259,6 +260,7 @@
   "close": "Sulje",
   "version": "Versio: {{ version }}",
   "hide": "Piilota",
+  "all": "Kaikki",
   "welcomePage": {
     "heading": "JORE4:n käyttöönotto alkaa!",
     "subheading1": "Mitä?",
--- a/ui/src/uiComponents/Listbox.tsx
+++ b/ui/src/uiComponents/Listbox.tsx
@@ -35,7 +35,7 @@ export const dropdownTransition: TransitionClasses = {
 export interface FormInputProps {
   value?: string;
   onChange: ValueFn;
-  onBlur: Noop;
+  onBlur?: Noop;
   fieldState?: ControllerFieldState;
 }
 
--- a/ui/src/utils/enum.ts
+++ b/ui/src/utils/enum.ts
@@ -4,3 +4,10 @@ export function getEnumValues(inputEnum: Object): string[] {
     (value) => typeof value === 'string',
   ) as string[];
 }
+
+/** This enum is used to add the 'All' option to enumDropdowns and also having
+ * the correct types on searchConditions
+ */
+export enum AllOptionEnum {
+  All = 'all',
+}
--- a/ui/src/utils/search.ts
+++ b/ui/src/utils/search.ts
@@ -1,10 +1,17 @@
 import { produce } from 'immer';
+import { ReusableComponentsVehicleModeEnum } from '../generated/graphql';
 import {
   DeserializedQueryStringParameters,
   SearchConditions,
   SearchParameters,
 } from '../hooks/search/useSearchQueryParser';
 import { Priority } from '../types/Priority';
+import { AllOptionEnum } from './enum';
+
+type SearchParameterValueTypes =
+  | string
+  | number[]
+  | ReusableComponentsVehicleModeEnum;
 
 type SearchParametersGqlConfigurations = {
   priority: {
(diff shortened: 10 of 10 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/355)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 51 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaLT54Ro6di3wrnpL:-NDkaLT54Ro6di3wrnpM:byrzz8e) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L51)):*
> ```XML
>         <div className="border-2 border-background p-10">
>           <h2>{t('search.advancedSearchTitle')}</h2>
>           <label htmlFor="lines.primaryVehicleMode">
> ```

is this "lines.primaryVehicleMode" id passed to the input component itself somewhere..?
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 60 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaC8VCijeWOI2Khwo:-NDkaC8VCijeWOI2Khwp:b2fikz5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L60)):*
> ```XML
>                 includeAllOption
>                 value={searchConditions.primaryVehicleMode ?? AllOptionEnum.All}
>                 onBlur={noop} // onBlur is required but not used anywhere
> ```

Maybe it could be also made optional? Both solutions are fine for me.
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 8 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0EX7E6S2fkYGmz2Ye:-NE0EX7F9yeD1eQ9mN6-:brc8ac1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L8)):*
> ```XML
> import { AllOptionEnum } from '../../../../utils';
> import { FormColumn, FormRow } from '../../../forms/common';
> import { VehicleModeDropdown } from '../../../forms/line/VehicleModeDropdown';
> ```

../forms/line? :thinking: 
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 52 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0Ehm-Ae8dXSOOC5wO:-NE0Ehm-Ae8dXSOOC5wP:b-2wnocq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L52)):*
> ```XML
>           <FormColumn>
>             <FormRow mdColumns={4}>
>               <div>
> ```

Is this div needed? No -> please remove
Does this div have semantic meaning? -> Replace with Column(?)
Is this here only because FormRow can't have multiple child elements? -> Replace with <></> or add those in FormRow already.
___
*[`ui/src/hooks/search/useSearchQueryParser.ts` line 83 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0FKmd5okStay2mgfo:-NE0FKme4VKJsvKM6y0r:bll8658) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/hooks/search/useSearchQueryParser.ts#L83)):*
> ```XML
>  * default to All option
>  */
> const mapPrimaryVehicleMode = (primaryVehicleMode?: string) => {
> ```

mapPrimaryVehicleModeToEnum? :shrug: 
___
*[`ui/src/utils/search.ts` line 63 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0KdSg10jI0yo6zoWb:-NE0KdSg10jI0yo6zoWc:brd0m3b) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L63)):*
> ```XML
> };
> 
> const setConfigurations = (queryParams: SearchConditions) => {
> ```

Not part of this PR, but I don't quite understand what does `setConfigurations` mean.
___
*[`ui/src/utils/search.ts` line 128 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0Jz0s1ul63ZOHX32G:-NE0Jz0t33QBFZBoiPX7:b-cwekq9) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L128)):*
> ```XML
>   });
> 
>   // Converting array [{key: value}, ...] into a single object {key: value, ...}
> ```

Do we know that we don't have duplicate keys which would then be overwritten here?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/355)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 51 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaLT54Ro6di3wrnpL:-NE0kjZT4IVhX_uUEkiY:b4d7flq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L51)):*

is this "lines.primaryVehicleMode" id passed to the input component itself somewhere..?
</blockquote></details>

Hmm no. It seems that this was implemented a bit different than I thought. But this seems to be implemented in the `<InputField>` component in the other use case (forms). Not totally sure how this should be done here, but I'll change this to have the `id` as optional parameter and fill it here. And maybe also change this id to `search.primaryVehicleMode`
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 60 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaC8VCijeWOI2Khwo:-NE0j9hPATATEASmf9ME:bdn762d) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L60)):*

Maybe it could be also made optional? Both solutions are fine for me.
</blockquote></details>

Do you mean the `id`? or what could also be made optional?
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 8 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0EX7E6S2fkYGmz2Ye:-NE0i7sL67oehs2OMbRT:b-qaej3z) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L8)):*

../forms/line? :thinking: 
</blockquote></details>

Hmm. Yeah this component seems to live there..
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 52 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0Ehm-Ae8dXSOOC5wO:-NE0hSHxAc8yc4AoLFiB:bh3usir) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L52)):*

Is this div needed? No -> please remove
Does this div have semantic meaning? -> Replace with Column(?)
Is this here only because FormRow can't have multiple child elements? -> Replace with <></> or add those in FormRow already.
</blockquote></details>

The `<FormRow>` will automatically put each child element to own column, and I want to keep the Label on top of the `<VehicleModeDropdown>`. But yeah should Wrap it with `<Column>` rather than `<div>`.
___
*[`ui/src/hooks/search/useSearchQueryParser.ts` line 83 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0FKmd5okStay2mgfo:-NE0dKq20oI4h9ISx-4P:b-wim4uk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/hooks/search/useSearchQueryParser.ts#L83)):*

mapPrimaryVehicleModeToEnum? :shrug: 
</blockquote></details>

Is better yeah, I'll change this. Although this might be removed completely after moving these to `useUrlQuery`
___
*[`ui/src/utils/search.ts` line 63 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0KdSg10jI0yo6zoWb:-NE0n9nC4s5G2JidYpK9:b60fhjt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L63)):*

Not part of this PR, but I don't quite understand what does `setConfigurations` mean.
</blockquote></details>

Yeah this is kind of complex and maybe needs some cleaning / refactoring (and comments) iteration. 
The main thing here is that we have different kind of variables in query parameters. 
Then we have the configurations for each parameter for the actual GQL filter. This configuration structure is currently:
```
[condition]: {
  value
  replaceStar
  opreator
  isLineProperty
}
```
And these configurations are used to make the actual GQL filters by using `transformToFilterAttribute`. This is maybe unnecessarily complex and could be something simpler. This also will not work completely like this when working with dates, because observationDate will need it's own filtering (constructActiveDateGqlFilter). And in the future we will need to have a filter for `Updated during the period of <startDate>  - <endDate>`. 

So this works with most of the filters, but maybe requires some minor rework to simplify this.
___
*[`ui/src/utils/search.ts` line 128 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0Jz0s1ul63ZOHX32G:-NE0TEkJBLK-xjlVqStl:bjqr70) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L128)):*

Do we know that we don't have duplicate keys which would then be overwritten here?
</blockquote></details>

Ah this one. In the next PR there will be duplicate keys, because of the `isLineFilter` which will wrap the filters inside `route_line`. In that PR `Object.Assign` will be replaced with reduce & merge combo which fixes this. I debated with myself that should I already include this change in this PR, but came in to the conclusion that this is already here and it does not affect anything yet.

TL;DR:
No duplicate keys yet, but this will be fixed in the next PR where there will also be duplicate keys.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/355)* status: 0 of 10 files reviewed, 7 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 51 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaLT54Ro6di3wrnpL:-NE4vZ83E_mPHbwqfrVw:b-exz04) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L51)):*

Hmm no. It seems that this was implemented a bit different than I thought. But this seems to be implemented in the `<InputField>` component in the other use case (forms). Not totally sure how this should be done here, but I'll change this to have the `id` as optional parameter and fill it here. And maybe also change this id to `search.primaryVehicleMode`
</blockquote></details>

Yeah. This is important for accessibility reasons, so it is quite bad if we just add `htmlFor` field without corresponding `id` field. Weird that linter does not complain about that, but maybe that is difficult to check with linter.

Btw, I suggest defining this string in a constant that can be given to both label and input component so that they stay in sync and avoid possible typos. See e.g. `ui/src/components/routes-and-lines/line-details/ActionsRow.tsx` for example.
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 60 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaC8VCijeWOI2Khwo:-NE4uIm14o6C-DNW7AwJ:b-o0ytye) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L60)):*

Do you mean the `id`? or what could also be made optional?
</blockquote></details>

Hmm, seems like this comment is now out of place. IIRC this was related to `onBlur={noop}` solution, which seems to be removed now.
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 8 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0EX7E6S2fkYGmz2Ye:-NE4tii_9NW-2AWlUalz:bas6i43) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L8)):*

Hmm. Yeah this component seems to live there..
</blockquote></details>

Did you forget or decided not to fix this?
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 50 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4tyiV5rfjjjr0Y1lh:-NE4tyiV5rfjjjr0Y1li:b-cjxxc6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L50)):*
> ```XML
>         <div className="border-2 border-background p-10">
>           <h2>{t('search.advancedSearchTitle')}</h2>
>           <FormColumn>
> ```

So now we have FormColumn with two FormRows with then again FormColumns? Couldn't you just remove first FormColumn layer?
___
*[`ui/src/locales/fi-FI/common.json` line 168 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4wuoK1maw2j-8xzxh:-NE4wuoK1maw2j-8xzxi:b-vf4fj8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/locales/fi-FI/common.json#L168)):*
> ```JSON
>     "train": "Juna",
>     "tram": "Raitiovaunu",
>     "all": "Kaikki"
> ```

This doesn't actually belong to vehicleModeEnum, so ideally I wouldn't put this here. But if moving this to e.g. root level of this file, e.g. next to "hide" makes your life somehow difficult in the place where this is used, then I guess we can live with this.
___
*[`ui/src/uiComponents/Listbox.tsx` line 38 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4uWJV4agsa9h3XBXF:-NE4uWJV4agsa9h3XBXG:b-d7n723) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/uiComponents/Listbox.tsx#L38)):*
> ```XML
>   value?: string;
>   onChange: ValueFn;
>   onBlur?: Noop;
> ```

Noop is most likely not the correct return type for this function. :thinking: I guess correct type would be `(e: SomeHtmlFormEvent) => void;`. But OK, this is out of scope of this PR so this is non-blocking comment.
___
*[`ui/src/utils/search.ts` line 63 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0KdSg10jI0yo6zoWb:-NE4xUqU3fyPF_sz4k_X:bjywvjf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L63)):*

Yeah this is kind of complex and maybe needs some cleaning / refactoring (and comments) iteration. 
The main thing here is that we have different kind of variables in query parameters. 
Then we have the configurations for each parameter for the actual GQL filter. This configuration structure is currently:
```
[condition]: {
  value
  replaceStar
  opreator
  isLineProperty
}
```
And these configurations are used to make the actual GQL filters by using `transformToFilterAttribute`. This is maybe unnecessarily complex and could be something simpler. This also will not work completely like this when working with dates, because observationDate will need it's own filtering (constructActiveDateGqlFilter). And in the future we will need to have a filter for `Updated during the period of <startDate>  - <endDate>`. 

So this works with most of the filters, but maybe requires some minor rework to simplify this.
</blockquote></details>

OK. By the way, there is also pretty major issue with naming of this function: the function name suggests this SETS the parameters (configurations) to SOMEWHERE (and this doesn't specify where), but actually this seems to rather map (url?) query params to SOMETHING, which we call "configuredParams" later on, but if I understood your explanation, I guess it should rather be called something like "qglSearchCondition".

So, should this function be called something like "mapQueryToSearchCondition" or "mapQueryToSearchParams"? This kind of things have huge impact on readability of code... :) 
___
*[`ui/src/utils/search.ts` line 128 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0Jz0s1ul63ZOHX32G:-NE4yzXs9ItjX2lQTow1:b-9d7a4f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L128)):*

Ah this one. In the next PR there will be duplicate keys, because of the `isLineFilter` which will wrap the filters inside `route_line`. In that PR `Object.Assign` will be replaced with reduce & merge combo which fixes this. I debated with myself that should I already include this change in this PR, but came in to the conclusion that this is already here and it does not affect anything yet.

TL;DR:
No duplicate keys yet, but this will be fixed in the next PR where there will also be duplicate keys.
</blockquote></details>

Ok. Ideally there would maybe then been TODO comment or something, but if this is already being taken care of then I guess this is  OK for now.
___
*[`ui/src/utils/search.ts` line 117 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4zW1KD1oOry_MWzyZ:-NE4zW1KD1oOry_MWzy_:b4arbt1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/utils/search.ts#L117)):*
> ```XML
> /** Transforms the given parameters to GQL filter attributes. Also takes in to
>  * consideration if we are generating route filter. In that case we wrap all
>  * line property filters in to route_lin wrapper.
> ```

Typo: route_lin -> route_line
___
*[`ui/src/utils/search.ts` line 119 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4zFHT9GyZOULWXnQe:-NE4zFHT9GyZOULWXnQf:beduaro) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/utils/search.ts#L119)):*
> ```XML
>  * line property filters in to route_lin wrapper.
>  */
> const transformConfiguredParamsToFilterAttributes = (
> ```

What is difference between "map" and "transform"?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/355)* status: 0 of 10 files reviewed, 7 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 51 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaLT54Ro6di3wrnpL:-NE57lduDuAm9VuxMJTJ:bo3vl6l) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L51)):*

Yeah. This is important for accessibility reasons, so it is quite bad if we just add `htmlFor` field without corresponding `id` field. Weird that linter does not complain about that, but maybe that is difficult to check with linter.

Btw, I suggest defining this string in a constant that can be given to both label and input component so that they stay in sync and avoid possible typos. See e.g. `ui/src/components/routes-and-lines/line-details/ActionsRow.tsx` for example.
</blockquote></details>

Thanks, done!
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 60 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NDkaC8VCijeWOI2Khwo:-NE53tzo7_w-DaOoYA59:b-fm6ekx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/fda6649c2635708e3441c0a0f7e321d62897c55c/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L60)):*

Hmm, seems like this comment is now out of place. IIRC this was related to `onBlur={noop}` solution, which seems to be removed now.
</blockquote></details>

Ah I see, maybe this comment was in a draft state and got only now published when the fix was already done :).
___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 50 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4tyiV5rfjjjr0Y1lh:-NE56taO2EMXso1oFMXE:b-2ky2kv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L50)):*

So now we have FormColumn with two FormRows with then again FormColumns? Couldn't you just remove first FormColumn layer?
</blockquote></details>

That can be removed yeah. Although to keep the styling, added `space-y-5` to the wrapping div which earlier came from the `<FormColumn>`.  And also not sure if my minor fix today was already there before these comments, but the inner `<FormColumn>` was supposed to be `<Column>` like said in the earlier comments. (`<FormColumn>` had the spacing and made the label float 5pixels above the input instead of being right above)
___
*[`ui/src/locales/fi-FI/common.json` line 168 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4wuoK1maw2j-8xzxh:-NE52Qz_4yzZ7MQSvWCr:b-1d62ji) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/locales/fi-FI/common.json#L168)):*

This doesn't actually belong to vehicleModeEnum, so ideally I wouldn't put this here. But if moving this to e.g. root level of this file, e.g. next to "hide" makes your life somehow difficult in the place where this is used, then I guess we can live with this.
</blockquote></details>

I actually had this on the root level at first, because other dropdowns will use this "all" option as well. But then I decided to move it inside the vehicleModeEnums, but it is a good point that this is precisely `vehicleModeEnum` instead of `vehicleModeDropdown`, so I should not have this here. It also means that when getting the translation, we need to make a minor additional check at `uiNameMappings.ts`.
I will do this change now that you pointed out that this is the enum's translation, thanks.
___
*[`ui/src/utils/search.ts` line 63 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0KdSg10jI0yo6zoWb:-NE5Hba13KL-ir2EFOzJ:b3km6tq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L63)):*

OK. By the way, there is also pretty major issue with naming of this function: the function name suggests this SETS the parameters (configurations) to SOMEWHERE (and this doesn't specify where), but actually this seems to rather map (url?) query params to SOMETHING, which we call "configuredParams" later on, but if I understood your explanation, I guess it should rather be called something like "qglSearchCondition".

So, should this function be called something like "mapQueryToSearchCondition" or "mapQueryToSearchParams"? This kind of things have huge impact on readability of code... :) 
</blockquote></details>

Ahh, thank you! That is exactly like you said. Sometimes when creating kind of generic functions on top of already complex stuff, it tends to get overwhelming and then I get confused of my own code and the namings are not easy to do anymore. But nice! These suggestions will make this more readable for sure! I'll try to think some better names for this to make it clear and let's see if those are correct.
I'll also create a separate PR about these namings so we can continue the discussion there, if it's OK?
___
*[`ui/src/utils/search.ts` line 128 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0Jz0s1ul63ZOHX32G:-NE5LMyHEnMFHe-OTdTO:b-dhq4vt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L128)):*

Ok. Ideally there would maybe then been TODO comment or something, but if this is already being taken care of then I guess this is  OK for now.
</blockquote></details>

Yeah, I realised this only when creating the next search component and that is why there is no TODO comments here. But since I want to do some other changes here before the next search component, I will add a TODO.
___
*[`ui/src/utils/search.ts` line 117 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4zW1KD1oOry_MWzyZ:-NE5JDHe9fZ_b5JAneOx:b-8fx2u7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/utils/search.ts#L117)):*

Typo: route_lin -> route_line
</blockquote></details>

:+1: 
___
*[`ui/src/utils/search.ts` line 119 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4zFHT9GyZOULWXnQe:-NE5JNNH8ED8xhKyPXYW:b-w9c769) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/utils/search.ts#L119)):*

What is difference between "map" and "transform"?
</blockquote></details>

Good point: nothing. I will change these two `transform` -> `map`, but I won't touch the rest of the name yet. In the upcoming rename PR I will go through these function names more thoroughly.

tk-vincit (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-:-NE5tB-H0Ci2AYkvSKoP:bg7igkv)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx` line 50 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4tyiV5rfjjjr0Y1lh:-NE5rshA7sWj_92yaEHV:b-x1f4e6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx#L50)):*

That can be removed yeah. Although to keep the styling, added `space-y-5` to the wrapping div which earlier came from the `<FormColumn>`.  And also not sure if my minor fix today was already there before these comments, but the inner `<FormColumn>` was supposed to be `<Column>` like said in the earlier comments. (`<FormColumn>` had the spacing and made the label float 5pixels above the input instead of being right above)
</blockquote></details>

Ok. FormColumn or Column, I don't have strong opinion, but initially suggested Column. And if this had some form layout related meaning then OK, but usually if you start to have multiple column-row-column-row like structures that's sign of something gone wrong. :)
___
*[`ui/src/locales/fi-FI/common.json` line 168 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE4wuoK1maw2j-8xzxh:-NE5sW9g7gEcgyQyJZ-y:brdv30p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/3fb4172ac3f96b2b13fea9bbce4214588278390d/ui/src/locales/fi-FI/common.json#L168)):*

I actually had this on the root level at first, because other dropdowns will use this "all" option as well. But then I decided to move it inside the vehicleModeEnums, but it is a good point that this is precisely `vehicleModeEnum` instead of `vehicleModeDropdown`, so I should not have this here. It also means that when getting the translation, we need to make a minor additional check at `uiNameMappings.ts`.
I will do this change now that you pointed out that this is the enum's translation, thanks.
</blockquote></details>

Yeah, doing that in `uiNameMappings.ts` seems to be good & clear solution. Thanks.
___
*[`ui/src/utils/search.ts` line 63 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/355#-NE0KdSg10jI0yo6zoWb:-NE5sp7dA7bmtlV-eoCA:bxrm2yw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/13e3b32c6e0cc26a401c5d39baf427c624d4ae75/ui/src/utils/search.ts#L63)):*

Ahh, thank you! That is exactly like you said. Sometimes when creating kind of generic functions on top of already complex stuff, it tends to get overwhelming and then I get confused of my own code and the namings are not easy to do anymore. But nice! These suggestions will make this more readable for sure! I'll try to think some better names for this to make it clear and let's see if those are correct.
I'll also create a separate PR about these namings so we can continue the discussion there, if it's OK?
</blockquote></details>

Thats ok for me, just hoping that it won't get forgotten then. :)

PR #356, opened 2022-10-10
Change obserationDate query param init to optional

HenrikHartiala:
Before this the observationDate was initialized automatically to query parameters if the component uses the useObservationDateQueryParam hook. The initialization is unwanted if the component only wants to read the observationDate, and the value is optional. For example in this case the useShowRoutesOnModal hook only uses the observationDate if it is already there. But we do not want to initialize it if it is not there.

We used useShowRoutesOnModal hook with every route and line row, which ment that the observationDate query parameter was initialized n times, where n is the amount of rows rendered..

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/356)

### DIFF
--- a/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts
+++ b/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts
@@ -7,7 +7,17 @@ import { showWarningToast } from '../../utils';
 import { QueryParameterName } from './useMapQueryParams';
 import { useUrlQuery } from './useUrlQuery';
 
-export const useObservationDateQueryParam = () => {
+interface Props {
+  initialize?: boolean;
+}
+
+/**
+ * Query parameter hook for setting and getting observationDate. Initialization
+ * of this query parameter can be set to false if you don't want to initialize it.
+ */
+export const useObservationDateQueryParam = (
+  { initialize }: Props = { initialize: true },
+) => {
   const { getDateTimeFromUrlQuery, setDateTimeToUrlQuery, queryParams } =
     useUrlQuery();
   const { t } = useTranslation();
@@ -68,8 +78,10 @@ export const useObservationDateQueryParam = () => {
   }, [defaultDate, observationDate, queryParams.observationDate]);
 
   useEffect(() => {
-    initializeObservationDate();
-  }, [initializeObservationDate]);
+    if (initialize) {
+      initializeObservationDate();
+    }
+  }, [initialize, initializeObservationDate]);
 
   return {
     observationDate,
--- a/ui/src/hooks/useShowRoutesOnModal.ts
+++ b/ui/src/hooks/useShowRoutesOnModal.ts
@@ -44,7 +44,7 @@ export const useShowRoutesOnModal = () => {
   const dispatch = useAppDispatch();
   const { openMapWithParameters } = useMapQueryParams();
   const { observationDate: listViewObservationDate } =
-    useObservationDateQueryParam();
+    useObservationDateQueryParam({ initialize: false });
 
   const showRoutesOnModal = ({
     viewPortParams,


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/356)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/hooks/urlQuery/useObservationDateQueryParam.ts` line 10 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/356#-NE-i-FLAe_1nkxhKBb_:-NE-i-FLAe_1nkxhKBba:bbaf64k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a10c8b20ac2e3e9a40eec631be4388a4fe22a828/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts#L10)):*
> ```XML
> import { useUrlQuery } from './useUrlQuery';
> 
> export const useObservationDateQueryParam = (
> ```

let's rather use a configuration object instead of just a boolean parameter:
```
type HookParams = {
  initialize: boolean
}

export const useObservationDateQueryParam = ({initialize = true}: HookParams) => {...}
```
___
*[`ui/src/hooks/urlQuery/useObservationDateQueryParam.ts` line 76 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/356#-NE-i_do-Y8uDJvekG6C:-NE-i_do-Y8uDJvekG6D:b-lfijyf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a10c8b20ac2e3e9a40eec631be4388a4fe22a828/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts#L76)):*
> ```XML
> 
>   useEffect(() => {
>     initializeObservationDate();
> ```

the flag checking should happen here:
```
useEffect(() => {
  if (initialize) {
    initializeObservationDate();
  }
});
```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/356)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @pallost)

___
*[`ui/src/hooks/urlQuery/useObservationDateQueryParam.ts` line 10 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/356#-NE-i-FLAe_1nkxhKBb_:-NE-k-mx96DDEJ0dM03D:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a10c8b20ac2e3e9a40eec631be4388a4fe22a828/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts#L10)):*

let's rather use a configuration object instead of just a boolean parameter:
```
type HookParams = {
  initialize: boolean
}

export const useObservationDateQueryParam = ({initialize = true}: HookParams) => {...}
```
</blockquote></details>

Done.
___
*[`ui/src/hooks/urlQuery/useObservationDateQueryParam.ts` line 76 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/356#-NE-i_do-Y8uDJvekG6C:-NE-k0zG-Sh1bpZzla7M:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a10c8b20ac2e3e9a40eec631be4388a4fe22a828/ui/src/hooks/urlQuery/useObservationDateQueryParam.ts#L76)):*

the flag checking should happen here:
```
useEffect(() => {
  if (initialize) {
    initializeObservationDate();
  }
});
```
</blockquote></details>

Done.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/356)* status: 0 of 2 files reviewed, all discussions resolved

PR #358, opened 2022-10-12
Refine namings and add typings to search

HenrikHartiala:
Search functions and variables were lacking some typings and had a lot of misleading namings, which made the code really hard to read.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/358)

### DIFF
--- a/ui/src/hooks/search/useSearchResults.ts
+++ b/ui/src/hooks/search/useSearchResults.ts
@@ -4,7 +4,10 @@ import {
   RouteAllFieldsFragment,
   useSearchLinesAndRoutesQuery,
 } from '../../generated/graphql';
-import { constructGqlFilterObject, mapToVariables } from '../../utils';
+import {
+  constructSearchLinesAndRoutesGqlQueryVariables,
+  mapToVariables,
+} from '../../utils';
 import {
   DisplayedSearchResultType,
   useSearchQueryParser,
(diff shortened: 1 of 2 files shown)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/358)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/utils/search.ts` line 1 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/358#-NEAOp8X398KQm1posEP:-NEAOp8X398KQm1posEQ:b455lu8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/19ac6389a26a83bcf368faca151131487f15b3ca/ui/src/utils/search.ts#L1)):*
> ```XML
> import { produce } from 'immer';
> ```

this is not in the scope of this pull request, but... have you noticed that basically the flow is now:
`query -> SearchCondition[] -> graphql filter`? To me it feels that the `SearchCondition` mapping is kinda unnecessary. Of course, the graphql filter types look weird, but they (necessarily) certainly support all the stuff that you are describing with the SearchCondition type.

Option 1: skip the whole SearchCondition phase and map the query params directly to graphql filters (if possible)
Option 2: as mapping from query -> SearchCondition and SearchCondition -> graphql filter are two separate phases, I'd split these into two separate files.
___
*[`ui/src/utils/search.ts` line 128 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/358#-NEA7Xz74H0pZI_CL2Cq:-NEA7Xz74H0pZI_CL2Cr:bfu1ed2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/19ac6389a26a83bcf368faca151131487f15b3ca/ui/src/utils/search.ts#L128)):*
> ```XML
>     [key]: {
>       [options.operator]: options.replaceStar
>         ? mapToSqlLikeValue(options.value as string) || '%'
> ```

Shouldn't this be `mapToSqlLikeValue(String(options.value))`? The "as string" does not convert it to string just tweaks the typings
___
*[`ui/src/utils/search.ts` line 169 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/358#-NEA8Nik2XGCttbvqe6a:-NEA8Nik2XGCttbvqe6b:b-76ko7v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/19ac6389a26a83bcf368faca151131487f15b3ca/ui/src/utils/search.ts#L169)):*
> ```XML
>   // instead of merging them.
>   // Converting array [{key: value}, ...] into a single object {key: value, ...}
>   return Object.assign({}, ...gqlFilters);
> ```

https://lodash.com/docs/4.17.15#merge ?

PR #361, opened 2022-10-13
Refactor searchquery handling

HenrikHartiala:
Nowadays we have useUrlQuery which is responsible for handling getting and setting properties to URL query parameters. This PR refactors search hooks to use this hook instead of handling the serialization, deserialization and pushing the properties to query parameters itself.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/361)

### DIFF
--- a/ui/src/components/routes-and-lines/search/ExportToolbar.tsx
+++ b/ui/src/components/routes-and-lines/search/ExportToolbar.tsx
@@ -3,7 +3,6 @@ import uniq from 'lodash/uniq';
 import { useTranslation } from 'react-i18next';
 import { pipe } from 'remeda';
 import {
-  DisplayedSearchResultType,
   useAppDispatch,
   useAppSelector,
   useSearchResults,
@@ -16,6 +15,7 @@ import {
   setIsSelectingRoutesForExportAction,
 } from '../../../redux';
 import { SimpleSmallButton } from '../../../uiComponents';
+import { DisplayedSearchResultType } from '../../../utils';
 
 export const ExportToolbar = (): JSX.Element => {
   const dispatch = useAppDispatch();
--- a/ui/src/components/routes-and-lines/search/ResultList.tsx
+++ b/ui/src/components/routes-and-lines/search/ResultList.tsx
@@ -2,8 +2,9 @@ import {
   LineTableRowFragment,
   RouteAllFieldsFragment,
 } from '../../../generated/graphql';
-import { DisplayedSearchResultType, useAppSelector } from '../../../hooks';
+import { useAppSelector } from '../../../hooks';
 import { selectExport } from '../../../redux';
+import { DisplayedSearchResultType } from '../../../utils';
 import { LinesList } from '../main/LinesList';
 import { RoutesList } from '../main/RoutesList';
 
--- a/ui/src/components/routes-and-lines/search/filters/ResultSelector.tsx
+++ b/ui/src/components/routes-and-lines/search/filters/ResultSelector.tsx
@@ -1,11 +1,8 @@
 import { useTranslation } from 'react-i18next';
-import {
-  DisplayedSearchResultType,
-  useAppDispatch,
-  useSearch,
-} from '../../../../hooks';
+import { useAppDispatch, useSearch } from '../../../../hooks';
 import { resetSelectedRoutesAction } from '../../../../redux';
 import { SimpleSmallButton } from '../../../../uiComponents';
+import { DisplayedSearchResultType } from '../../../../utils';
 
 export const ResultSelector = (): JSX.Element => {
   const { t } = useTranslation();
--- a/ui/src/hooks/search/useSearch.ts
+++ b/ui/src/hooks/search/useSearch.ts
@@ -1,13 +1,17 @@
 import { produce } from 'immer';
 import { useState } from 'react';
-import { useHistory } from 'react-router-dom';
 import { Priority } from '../../types/Priority';
-import { createQueryParamString, DisplayedSearchResultType } from '../../utils';
-import { FilterConditions, useSearchQueryParser } from './useSearchQueryParser';
+import { DisplayedSearchResultType } from '../../utils';
+import { QueryParameter, QueryParameterTypes, useUrlQuery } from '../urlQuery';
+import {
+  DeserializedQueryStringParameters,
+  FilterConditions,
+  useSearchQueryParser,
+} from './useSearchQueryParser';
 
 export const useSearch = () => {
-  const history = useHistory();
   const queryParameters = useSearchQueryParser();
+  const { setMultipleParametersToUrlQuery } = useUrlQuery();
 
   const [searchConditions, setSearchConditions] = useState(
     queryParameters.search,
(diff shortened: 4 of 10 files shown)


tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/361)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/hooks/search/useSearch.ts` line 55 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/361#-NEGMFJDCyN5uzTW7grD:-NEGMFJDCyN5uzTW7grE:blv8n6m) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e51ed8b4ce55f2cdbb9df9ccf3f387042afef16a/ui/src/hooks/search/useSearch.ts#L55)):*
> ```XML
>     });
> 
>     const combinedParameters = {
> ```

This looks a bit suspicious as if `queryParameters.search` have same keys as `newParameters.filter`, thise will get overridden.



_Code quote:_
```XML
    const combinedParameters = {
      ...newParameters.filter,
      ...queryParameters.search,
    };
```
___
*[`ui/src/hooks/search/useSearch.ts` line 88 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/361#-NEGMQ892H_hkC3rXBfQ:-NEGMQ892H_hkC3rXBfR:b-l2zojo) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e51ed8b4ce55f2cdbb9df9ccf3f387042afef16a/ui/src/hooks/search/useSearch.ts#L88)):*
> ```XML
>     const combinedParameters = {
>       ...queryParameters.filter,
>       ...queryParameters.search,
> ```

Same problem here as mentioned above.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/361)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/hooks/search/useSearch.ts` line 31 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/361#-NEe-1fFB-hIkR4-1-b6:-NEe-1fFB-hIkR4-1-b7:be1nizs) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/67cf767d77f4824b2a849c461eeb2ae42d15dff5/ui/src/hooks/search/useSearch.ts#L31)):*
> ```XML
> 
>   /**
>    * Maps the search parameters in to key, value pairs for
> ```

At least to me key, value pair sounds like `[[key, value], [key, value]]` like data structure. This seems to map data to ´[{paranName: key, value: value}]` stucture.

Where does this structure come from? Explicit return type for this function could make sense in this case.
___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 154 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/361#-NEe0MKcAddPCyASHBO9:-NEe0MKcAddPCyASHBOA:bedyzwd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/67cf767d77f4824b2a849c461eeb2ae42d15dff5/ui/src/hooks/urlQuery/useUrlQuery.ts#L154)):*
> ```XML
>    * used to combine with further deserialization functions.
>    */
>   const getArrayAsStringFromUrlQuery = (paramName: string) => {
> ```

getStringArrayFromUrlQuery? This doesn't return "array as string" as name suggests, or am I missing something?

Explicit return type could be considerer, for consistency with the rest of the similar get functions in this file.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/361)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/hooks/search/useSearch.ts` line 31 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/361#-NEe-1fFB-hIkR4-1-b6:-NEeXBof8b2DhKBwkHou:b7gsyiv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/67cf767d77f4824b2a849c461eeb2ae42d15dff5/ui/src/hooks/search/useSearch.ts#L31)):*

At least to me key, value pair sounds like `[[key, value], [key, value]]` like data structure. This seems to map data to ´[{paranName: key, value: value}]` stucture.

Where does this structure come from? Explicit return type for this function could make sense in this case.
</blockquote></details>

I added some more typings :+1: 
___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 154 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/361#-NEe0MKcAddPCyASHBO9:-NEeXw35DzFAMjLxM3LJ:bqnoxy1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/67cf767d77f4824b2a849c461eeb2ae42d15dff5/ui/src/hooks/urlQuery/useUrlQuery.ts#L154)):*

getStringArrayFromUrlQuery? This doesn't return "array as string" as name suggests, or am I missing something?

Explicit return type could be considerer, for consistency with the rest of the similar get functions in this file.
</blockquote></details>

Hmm yeah how should I name this. The point is that this function will get the Array from URL, but the type is string. If the type is number, or enum or what ever, we still need to cast it to the wanted type. So that is what I tried to express here.

Maybe I will not do this chaining:
```
/** Returns a query parameter in Priority array type */
  const getPriorityArrayFromUrlQuery = (paramName: string): Priority[] => {
    return getArrayAsStringFromUrlQuery(paramName)
      ?.map((p) => parseInt(p, 10))
      ?.filter((p) => Object.values(Priority).includes(p));
  };
```

So this was the point, I can first use the "get as string" and then cast it to the wanted type. But this is maybe unnecessary chain and I can just do a function for each type separately from the start.

TL;DR: It was maybe an unnecessary overengineering (saving no lines :D) with bad naming and I'll modify this.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/361)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 154 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/361#-NEe0MKcAddPCyASHBO9:-NEejYIb9Takrz6Hq11c:bbho259) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/67cf767d77f4824b2a849c461eeb2ae42d15dff5/ui/src/hooks/urlQuery/useUrlQuery.ts#L154)):*

Hmm yeah how should I name this. The point is that this function will get the Array from URL, but the type is string. If the type is number, or enum or what ever, we still need to cast it to the wanted type. So that is what I tried to express here.

Maybe I will not do this chaining:
```
/** Returns a query parameter in Priority array type */
  const getPriorityArrayFromUrlQuery = (paramName: string): Priority[] => {
    return getArrayAsStringFromUrlQuery(paramName)
      ?.map((p) => parseInt(p, 10))
      ?.filter((p) => Object.values(Priority).includes(p));
  };
```

So this was the point, I can first use the "get as string" and then cast it to the wanted type. But this is maybe unnecessary chain and I can just do a function for each type separately from the start.

TL;DR: It was maybe an unnecessary overengineering (saving no lines :D) with bad naming and I'll modify this.
</blockquote></details>

I see. :+1:

PR #365, opened 2022-10-18
Fix MultiLineString error in stop to infra line

culka:
Turn any possible MultiLineStrings into LineStrings when drawing a line from a stop to the nearest infralink.

Resolves https://github.com/HSLdevcom/jore4/issues/960

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -24,6 +24,7 @@
     "@nebula.gl/edit-modes": "0.23.8",
     "@reduxjs/toolkit": "^1.8.0",
     "@seznam/compose-react-refs": "^1.0.6",
+    "@turf/flatten": "^6.5.0",
     "@turf/distance": "^6.5.0",
     "@turf/helpers": "^6.5.0",
     "@turf/nearest-point-on-line": "^6.5.0",
--- a/ui/src/utils/map/lineFromStopToInfraLink.ts
+++ b/ui/src/utils/map/lineFromStopToInfraLink.ts
@@ -1,3 +1,4 @@
+import flatten from '@turf/flatten';
 import { Feature, LineString, point } from '@turf/helpers';
 import nearestPointOnLine from '@turf/nearest-point-on-line';
 import pointToLineDistance from '@turf/point-to-line-distance';
@@ -37,6 +38,15 @@ const findFeaturesForLayerWithRadius = (
     );
   }
   return [];
+};
+
+// Turn any possible MultiLineStrings in the feature into LineStrings
+const getLineStringFromFeature = (feature: Feature<Geometry>) => {
+  const flattened = flatten(feature);
+  if (flattened.features.length > 0) {
+    return flattened.features[0];
+  }
+  return feature.geometry;
 };
 
 export const removeLineFromStopToInfraLink = (map: MaplibreGLMap) => {
@@ -95,7 +105,13 @@ export const drawLineToClosestRoad = (map: MaplibreGLMap, coords: Coords) => {
     // pair each network connection with the distance from the coordinates
     const linesWithDistance: FeatureAndDistance[] = features.map((line) => ({
       feature: line,
-      distance: pointToLineDistance(cursorLocation, line, { units: 'meters' }),
+      distance: pointToLineDistance(
+        cursorLocation,
+        getLineStringFromFeature(line),
+        {
+          units: 'meters',
+        },
+      ),
     }));
 
     // find the nearest network connection


tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/365)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @culka)

___
*[`ui/src/utils/map/lineFromStopToInfraLink.ts` line 47 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/365#-NEdrKcI1lFHrge82Z_d:-NEdrKcI1lFHrge82Z_e:b-hvgohl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f2b0724bbdd810c17290c07190483bce82f8144e/ui/src/utils/map/lineFromStopToInfraLink.ts#L47)):*
> ```XML
>   const flattened = flatten(feature);
>   if (flattened.features.length > 0) {
>     return flattened.features[0];
> ```

Seeing something like this makes me always wonder what should happen if array happens to have more than 1 element? Why are we ignoring the rest and do we have a bug here?

jarkkoka (commented):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/365)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @culka and @tk-vincit)

___
*[`ui/src/utils/map/lineFromStopToInfraLink.ts` line 97 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/365#-NEdsOZE-61KvSKIuxt0:-NEdsOZE-61KvSKIuxt1:bkjhtqy) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f2b0724bbdd810c17290c07190483bce82f8144e/ui/src/utils/map/lineFromStopToInfraLink.ts#L97)):*
> ```XML
>     map,
>     ROAD_LAYER_ID,
>     coords,
> ```

Taking that I understand this code correctly, I am a bit surprised to see that Digiroad MVT layer is used as a data source for UI logic. Up until now, I have designed the MVT layer (in jore4-digiroad-import repository) as a pure visualisation layer on top of base background map. The data contents of the MVT layer are subject to change in the future when we later add trams and other vehicle modes and types. Though it mitigates the issue that only the geometry information is extracted from the MVT layer.

I am still wondering why the closest infrastructure links are not fetched via GraphQP API like any other data in the app? Is it so much faster to query the MVT layer or why? There could also be an unpredictable delay while loading the MVT tiles. Is this code correctly synchronised to MVT loading?



_Code quote:_
```XML
  const features: Feature<LineString>[] = findFeaturesForLayerWithRadius(
    map,
    ROAD_LAYER_ID,
    coords,
    SEARCH_RADIUS_IN_PIXELS,
  );
```

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/365)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jarkkoka and @tk-vincit)

___
*[`ui/src/utils/map/lineFromStopToInfraLink.ts` line 47 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/365#-NEdrKcI1lFHrge82Z_d:-NEdxFrE7Guh5K9ZkQoY:b-3n7gef) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f2b0724bbdd810c17290c07190483bce82f8144e/ui/src/utils/map/lineFromStopToInfraLink.ts#L47)):*

Seeing something like this makes me always wonder what should happen if array happens to have more than 1 element? Why are we ignoring the rest and do we have a bug here?
</blockquote></details>

Changing this to check every possible feature.
___
*[`ui/src/utils/map/lineFromStopToInfraLink.ts` line 97 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/365#-NEdsOZE-61KvSKIuxt0:-NEdvK7j4s7j_5_bLcV-:b-wk8r76) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f2b0724bbdd810c17290c07190483bce82f8144e/ui/src/utils/map/lineFromStopToInfraLink.ts#L97)):*

Taking that I understand this code correctly, I am a bit surprised to see that Digiroad MVT layer is used as a data source for UI logic. Up until now, I have designed the MVT layer (in jore4-digiroad-import repository) as a pure visualisation layer on top of base background map. The data contents of the MVT layer are subject to change in the future when we later add trams and other vehicle modes and types. Though it mitigates the issue that only the geometry information is extracted from the MVT layer.

I am still wondering why the closest infrastructure links are not fetched via GraphQP API like any other data in the app? Is it so much faster to query the MVT layer or why? There could also be an unpredictable delay while loading the MVT tiles. Is this code correctly synchronised to MVT loading?
</blockquote></details>

This calculation is done in real time when the user is creating a new stop and moving the cursor on screen, so it can be tens of times per second. It could have been done via the API, but it didn't feel appropriate to do so due to the amount of queries needed. The API is queried for the closest point after the stop has been placed.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/365)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @culka and @tk-vincit)

___
*[`ui/src/utils/map/lineFromStopToInfraLink.ts` line 47 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/365#-NEdrKcI1lFHrge82Z_d:-NEe8UW51foJjXgGm4IO:b-7fjxxo) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f2b0724bbdd810c17290c07190483bce82f8144e/ui/src/utils/map/lineFromStopToInfraLink.ts#L47)):*

Changing this to check every possible feature.
</blockquote></details>

I was wondering the same thing. Looked into Turf/Meta and Turf/Flatten library source codes but did not get certainty about the issue with a quick glance.
___
*[`ui/src/utils/map/lineFromStopToInfraLink.ts` line 97 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/365#-NEdsOZE-61KvSKIuxt0:-NEe8m3ZCG3MqwMUwO00:bvt42b7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/f2b0724bbdd810c17290c07190483bce82f8144e/ui/src/utils/map/lineFromStopToInfraLink.ts#L97)):*

This calculation is done in real time when the user is creating a new stop and moving the cursor on screen, so it can be tens of times per second. It could have been done via the API, but it didn't feel appropriate to do so due to the amount of queries needed. The API is queried for the closest point after the stop has been placed.
</blockquote></details>

Ok. Sounds like a fair reason. Though I consider this a little bit fragile solution. Sometimes there could be delays while loading MVT tiles. Some tiles may be missing e.g. if rapidly zooming in and out in the map (because a great number of tile load requests is sent). If the infra link layer is not visible, then user might not notice if a tile is missing. But I don't know whether a tile loaded from a previous zoom level could serve the purpose.

But if there haven't arised any issues or hiccups within thorough testing then this probably is ok.

PR #367, opened 2022-10-18
Refactor query parameter to gql flow

HenrikHartiala:
Remove unnecessary step from creating GQL filters from query parameters. There were these predefined options that we used to construct the GQL filters dynamically, but they are pretty much unnecessary and we can just use the constructors on 'utils/gql.ts'.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/367)

### DIFF
--- a/ui/src/utils/gql.ts
+++ b/ui/src/utils/gql.ts
@@ -1,5 +1,8 @@
 import { DateTime } from 'luxon';
-import { GeographyComparisonExp } from '../generated/graphql';
+import {
+  GeographyComparisonExp,
+  ReusableComponentsVehicleModeEnum,
+} from '../generated/graphql';
 import { Viewport } from '../redux/types/modalMap';
 import { Priority } from '../types/Priority';
 
@@ -51,6 +54,13 @@ export const constructLabelGqlFilter = (label?: string) => ({
   label: { _eq: label },
 });
 
+/** Constructs an object for gql to filter by label using the '_like' operator.
+ * This will means that all the '%' in the label are considered as 'any'
+ */
+export const constructLabelLikeGqlFilter = (label?: string) => ({
+  label: { _like: label },
+});
+
 /** Constructs an object for gql to filter route by line label */
 export const constructRouteLineLabelGqlFilter = (label: string) => ({
   route_line: constructLabelGqlFilter(label),
@@ -65,5 +75,14 @@ export const constructWithinViewportGqlFilter = (
       coordinates: [viewport.longitude, viewport.latitude],
     },
     distance: viewport.radius,
+  },
+});
+
+/** Constructs an object for gql to filter by primary_vehicle_mode */
+export const constructPrimaryVehicleModeGqlFilter = (
+  primaryVehicleMode: ReusableComponentsVehicleModeEnum,
+) => ({
+  primary_vehicle_mode: {
+    _eq: primaryVehicleMode,
   },
 });
--- a/ui/src/utils/search.ts
+++ b/ui/src/utils/search.ts
@@ -1,4 +1,3 @@
-import { produce } from 'immer';
 import {
   OrderBy,
   ReusableComponentsVehicleModeEnum,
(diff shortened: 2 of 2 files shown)


tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/367)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/utils/gql.ts` line 54 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEeqqKXCVEORm2P7aOE:-NEeqqKXCVEORm2P7aOF:bfjs795) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/gql.ts#L54)):*
> ```XML
> /** Constructs an object for gql to filter by label */
> export const constructLabelGqlFilter = (label?: string) => ({
>   label: { _eq: label },
> ```

I'm not 100% sure if this kind of trivial eq filter constructors really help with the readability of the code. :)
___
*[`ui/src/utils/search.ts` line 70 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEer7Lf2nM6GPE0Kvlz:-NEer7Lg6Q83vjQGzPtN:b-3i1l7t) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/search.ts#L70)):*
> ```XML
>     // Construct all the generic filters.
>     ...constructLabelLikeGqlFilter(
>       mapToSqlLikeValue(searchConditions.label) || '%',
> ```

this condition is a bit confusing - why `|| '%'`?
___
*[`ui/src/utils/search.ts` line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEerRwV03t3_mRNdeIh:-NEerRwV03t3_mRNdeIi:by1rth8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/search.ts#L77)):*
> ```XML
>     ...handleLinePropertyGqlFilters({
>       properties: {
>         ...constructOptionalSearchConditionGqlFilter<ReusableComponentsVehicleModeEnum>(
> ```

Seems like term "construct" has already been used in this file, but started to wonder if it means the same as "build"? :thinking:

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/367)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/utils/gql.ts` line 54 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEeqqKXCVEORm2P7aOE:-NEetlwh9VMPB4kozoJV:b-hux5rv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/gql.ts#L54)):*

I'm not 100% sure if this kind of trivial eq filter constructors really help with the readability of the code. :)
</blockquote></details>

Hmm, true. We could just type the simple ones by hand. But I was thinking that if we use these constructors for all, it would also be kind of consistent.
```
  const lineFilters = {
    label: {
      _eq: line?.label
    },
    ...constructActiveDateGqlFilter(observationDate),
    ...constructDraftPriorityGqlFilter(line?.priority),
  };
```
I think this might be confusing as well to have them mixed up. 

Good thing is that these are easy to change if it becomes too hard to read etc.
___
*[`ui/src/utils/search.ts` line 70 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEer7Lf2nM6GPE0Kvlz:-NEesDsE0sSVLIFwoN5L:bp6q2qn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/search.ts#L70)):*

this condition is a bit confusing - why `|| '%'`?
</blockquote></details>

It is because we want to search for "all" if there is nothing typed in the search input, but  hmm, should this actually be inside like:
`mapToSqlLikeValue(searchConditions.label || '%')` And maybe comment this as well.
___
*[`ui/src/utils/search.ts` line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEerRwV03t3_mRNdeIh:-NEes_msCJt7mNpjq66M:b-h8qfeb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/search.ts#L77)):*

Seems like term "construct" has already been used in this file, but started to wonder if it means the same as "build"? :thinking: 
</blockquote></details>

Maybe? I think I haven't seen "build" or registered if seen. It is not mapping, because it uses other information than the input as well to create the end result. So I've just used a prefix for this. 

Oh wait, right. The cypress tests uses the `build` word for the same purpose I think.

tk-vincit (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/367)* status: all files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/utils/gql.ts` line 54 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEeqqKXCVEORm2P7aOE:-NEey9Pe3HedgyfpUCiF:bktsxuw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/gql.ts#L54)):*

Hmm, true. We could just type the simple ones by hand. But I was thinking that if we use these constructors for all, it would also be kind of consistent.
```
  const lineFilters = {
    label: {
      _eq: line?.label
    },
    ...constructActiveDateGqlFilter(observationDate),
    ...constructDraftPriorityGqlFilter(line?.priority),
  };
```
I think this might be confusing as well to have them mixed up. 

Good thing is that these are easy to change if it becomes too hard to read etc.
</blockquote></details>

Yeah, your point is good too. I don't have strong opinions on this one, just wanted to point this out so that also you could consider this. :)
___
*[`ui/src/utils/search.ts` line 70 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEer7Lf2nM6GPE0Kvlz:-NEexnaFBhl1V3zVJvBI:bwowwgi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/search.ts#L70)):*

It is because we want to search for "all" if there is nothing typed in the search input, but  hmm, should this actually be inside like:
`mapToSqlLikeValue(searchConditions.label || '%')` And maybe comment this as well.
</blockquote></details>

That could be at least a bit easier to understand, but still not perfect. :thinking: 

In the lack of better ideas, how about if we comment like this?
`      // If there is no label filter given, search for all with '%' operator.`
Generally I don't like comments that explain too much _what_ code is doing, but in this case it could make this easier to understand.


Or in fact, when I think about this a bit more, does this logic even belong here or could we somehow improve this a bit?
If I start to phrase this a bit differently:

"If there is no label filter given, search for all."
-> "If there is no label filter given, don't filter by it"
-> "ignore label filter if it is not given"
-> "do nothing if label filter is not given
-> "filter by label if label filter is given"

So, would it make more sense to do something like this?

```
...  searchConditions.label ? constructLabelLikeGqlFilter(searchConditions.label) : {}
```
I mean, do we even need fallback to `label: { _like: % },` if label is not given? That filter sounds like it matches everything anyway.
___
*[`ui/src/utils/search.ts` line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEerRwV03t3_mRNdeIh:-NEewtSD08an3Dmtq-uY:blazw1g) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/search.ts#L77)):*

Maybe? I think I haven't seen "build" or registered if seen. It is not mapping, because it uses other information than the input as well to create the end result. So I've just used a prefix for this. 

Oh wait, right. The cypress tests uses the `build` word for the same purpose I think. 
</blockquote></details>

Just did global search with "build" in our codebase, it seems to have used in many instances, e.g. in `buildCommonGqlFilterOrConditions`, which seems to be doing pretty much same thing than we are doing here. See e.g. `ui/src/hooks/useCheckValidityAndPriorityConflicts.ts` for more examples.

As "construct" and "build" seem to have same semantic meaning here, I would prefer "build" as it is shorter, but I guess this is a bit matter of opinion. And for consistency's sake I would prefer to use only either one of those.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/367)* status: all files reviewed, 3 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/utils/search.ts` line 77 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/367#-NEerRwV03t3_mRNdeIh:-NEf1nU87TTm-pzbQAhy:b-rwncbn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/612ebfb5af8d4cd29f825bb41d7e20e9d9eba2b3/ui/src/utils/search.ts#L77)):*

Just did global search with "build" in our codebase, it seems to have used in many instances, e.g. in `buildCommonGqlFilterOrConditions`, which seems to be doing pretty much same thing than we are doing here. See e.g. `ui/src/hooks/useCheckValidityAndPriorityConflicts.ts` for more examples.

As "construct" and "build" seem to have same semantic meaning here, I would prefer "build" as it is shorter, but I guess this is a bit matter of opinion. And for consistency's sake I would prefer to use only either one of those.
</blockquote></details>

Yeah, I agree and it also doesn't matter to me which one we use, so after these PR's I can make a separate PR to change all these `construct` prefixes to `build` :+1:

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/367)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

PR #370, opened 2022-10-19
Add observationdate searchfilter

HenrikHartiala:
Add observationDate search filter and also a minor change to SearchInput for consistency

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/370)

### DIFF
--- a/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx
+++ b/ui/src/components/routes-and-lines/search/conditions/SearchContainer.tsx
@@ -20,6 +20,10 @@ export const SearchContainer = (): JSX.Element => {
     searchInput: 'SearchContainer::SearchInput',
   };
 
+  const onChangeLabel = (value: string) => {
+    setSearchCondition(SearchQueryParameterNames.Label, value);
+  };
+
   const onChangeVehiclemode = (e: React.ChangeEvent<HTMLInputElement>) => {
     setSearchCondition(
       SearchQueryParameterNames.PrimaryVehicleMode,
@@ -46,7 +50,7 @@ export const SearchContainer = (): JSX.Element => {
               testId={testIds.searchInput}
               value={searchConditions.label}
               onSearch={handleSearch}
-              onChange={(value) => setSearchCondition('label', value)}
+              onChange={onChangeLabel}
             />
             <SearchConditionToggle
               isToggled={isExpanded}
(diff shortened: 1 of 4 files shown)


tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/370)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`ui/src/components/routes-and-lines/search/conditions/SearchInput.tsx` line 6 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/370#-NEjchAL4Hed74WF9nWY:-NEjchAL4Hed74WF9nWZ:b-qk6997) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de9bbba13c1b2b0ae7c648ff3510673d9a542d9d/ui/src/components/routes-and-lines/search/conditions/SearchInput.tsx#L6)):*
> ```XML
> type Props = {
>   value?: string;
>   onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
> ```

Generally speaking I'm not fan of this change. Now we are exposing implementation detail (= change event) of this component via this api to the user of this component.

OK, this is not so bad as this component wraps native html input element, but generally it is _very rare_ that user needs anything else than `event.target.value` from event like this, so I would prefer to keep this kind of complexity as close of the input elements of possible and thus keep props apis simple.

OK, I see the point in consistency, but maybe this pattern should be followed also in other places rather than passing react change events all around the app.

If you make changes to this, please update also commit message then.



_Code quote:_
```XML
  onChange: (value: string) => void;
```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/370)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[`ui/src/components/routes-and-lines/search/conditions/SearchInput.tsx` line 6 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/370#-NEjchAL4Hed74WF9nWY:-NEjsX1A2TXaWcrYQCKA:b-f6jeqp) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/de9bbba13c1b2b0ae7c648ff3510673d9a542d9d/ui/src/components/routes-and-lines/search/conditions/SearchInput.tsx#L6)):*

Generally speaking I'm not fan of this change. Now we are exposing implementation detail (= change event) of this component via this api to the user of this component.

OK, this is not so bad as this component wraps native html input element, but generally it is _very rare_ that user needs anything else than `event.target.value` from event like this, so I would prefer to keep this kind of complexity as close of the input elements of possible and thus keep props apis simple.

OK, I see the point in consistency, but maybe this pattern should be followed also in other places rather than passing react change events all around the app.

If you make changes to this, please update also commit message then.
</blockquote></details>

Hmm, yeah maybe this should be the other way around. I'll retrace this change and add a note to check the others as well.

PR #373, opened 2022-10-24
Reset map state when opening map

eleksi:
Resolves HSLdevcom/jore4#997

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/373)

### DIFF
--- a/ui/src/components/routes-and-lines/main/RoutesAndLinesPage.tsx
+++ b/ui/src/components/routes-and-lines/main/RoutesAndLinesPage.tsx
@@ -5,11 +5,7 @@ import {
   useMapQueryParams,
 } from '../../../hooks';
 import { Container, Row } from '../../../layoutComponents';
-import {
-  FilterType,
-  resetMapRouteEditorStateAction,
-  resetModalMapAction,
-} from '../../../redux';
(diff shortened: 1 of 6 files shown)


tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/373)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @eleksi)

___
*[`ui/src/redux/thunks/map.ts` line 6 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/373#-NF8e-Zi5dQSdxV7Ypsi:-NF8e-Zi5dQSdxV7Ypsj:bua0cgg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/9efcb2b9b4ae8000809eca02d6b537540575fec3/ui/src/redux/thunks/map.ts#L6)):*
> ```XML
> import { resetModalMapAction } from '../slices/modalMap';
> 
> export const resetMapStateAction = createAsyncThunk(
> ```

Hmm... As this is not "Action" like non-async slice-related actions, not sure if we want to call this "Action". Maybe this could be something like "onResetMapState". on "resetMapState"?

PR #386, opened 2022-10-28
Use component specific attribute fetching for table rows

HenrikHartiala:
We fetched ...route_all_fields and ...line_all_fields for component that does not need all the fields. It is better to use the component specific fragments. This will also make it easier to re-use this component when we specify only the fields that are needed. Otherwise we would have to fetch all the fields every time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/386)

### DIFF
--- a/cypress/pageObjects/EditRouteModal.ts
+++ b/cypress/pageObjects/EditRouteModal.ts
@@ -5,9 +5,11 @@ export class EditRouteModal {
    * https://docs.cypress.io/api/commands/click#Arguments
    */
   save(forceAction = false) {
-    return cy
-      .getByTestId('EditRouteModal')
+    cy.getByTestId('EditRouteModal')
       .findByTestId('Modal::saveButton')
       .click({ force: forceAction });
+
+    // After save the edit route modal should close
+    cy.getByTestId('EditRouteModal').should('not.exist');
   }
 }
--- a/ui/src/components/routes-and-lines/main/LineTableRow.tsx
+++ b/ui/src/components/routes-and-lines/main/LineTableRow.tsx
@@ -20,7 +20,10 @@ interface Props {
 
 const GQL_LINE_TABLE_ROW = gql`
   fragment line_table_row on route_line {
-    ...line_all_fields
+    name_i18n
+    short_name_i18n
+    priority
+    ...line_information_for_map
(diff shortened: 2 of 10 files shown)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/386)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/pageObjects/MapItemCreator.ts` line 162 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/386#-NFrzqvx8_chAZyfAoYI:-NFrzqvx8_chAZyfAoYJ:bixgzrd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b55211c1e48282933dec7a2e37b9d92eaa9f4eeb/cypress/pageObjects/MapItemCreator.ts#L162)):*
> ```XML
>     this.editRouteModal.save();
> 
>     cy.getByTestId('EditRouteModal').should('not.exist');
> ```

I wonder: would it make sense to bake this line into the editRouteModal.save() function? That way every test that uses that save() function will automatically wait for the modal to be hidden.

Also, let's put a comment here that why this addition was needed

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/386)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[`cypress/pageObjects/MapItemCreator.ts` line 162 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/386#-NFrzqvx8_chAZyfAoYI:-NFsVGLy0UoZDydRljfm:b-kb87tl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b55211c1e48282933dec7a2e37b9d92eaa9f4eeb/cypress/pageObjects/MapItemCreator.ts#L162)):*

I wonder: would it make sense to bake this line into the editRouteModal.save() function? That way every test that uses that save() function will automatically wait for the modal to be hidden.

Also, let's put a comment here that why this addition was needed
</blockquote></details>

Good point, I'll move this inside the save :+1:

PR #393, opened 2022-11-04
Fix route creation with open validity end

jschummer:
When setting the validity end for a route to "open end", it is set to 'null'. This was accidentally caught by a check, which was supposed to verify that the route data was set at all, because it tested for falsiness instead of 'undefined'.

This commit fixes the check to explicitly test for '=== undefined'.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/393)

### DIFF
--- a/ui/src/components/map/routes/DrawRouteLayer.tsx
+++ b/ui/src/components/map/routes/DrawRouteLayer.tsx
@@ -159,9 +159,9 @@ const DrawRouteLayerComponent = (
       // (when editing, fetch it from graphql; when creating, filled in through form)
       if (
         !editedRouteData.metaData ||
-        !editedRouteData.metaData.validityStart ||
-        !editedRouteData.metaData.validityEnd ||
-        !editedRouteData.metaData.priority
+        editedRouteData.metaData.validityStart === undefined ||
+        editedRouteData.metaData.validityEnd === undefined ||
+        editedRouteData.metaData.priority === undefined
       ) {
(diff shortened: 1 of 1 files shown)


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/393)* status: 0 of 1 files reviewed, all discussions resolved

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/393)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 162 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/393#-NG1dK8U3ZeYHE60EVFk:-NG1dK8U3ZeYHE60EVFl:bzbauya) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/2f678af7169612b94c4b05c593462a6ab8ccd61d/ui/src/components/map/routes/DrawRouteLayer.tsx#L162)):*
> ```XML
>       if (
>         !editedRouteData.metaData ||
>         editedRouteData.metaData.validityStart === undefined ||
> ```

I think it is pretty rare case when these differ (falsyness vs undefined), so I think this would benefit from a comment why we specifically check for only `undefined` but not falsyness.

PR #408, opened 2022-11-21
Optimize infrastructure link snap point picking for route editing

eleksi:
Resolves https://github.com/HSLdevcom/jore4/issues/978

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)


HenrikHartiala (changes requested):
> Reviewed 7 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: 7 of 8 files reviewed, 6 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 208 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7WgcKARSxOFYgW0po:-NI7WgcKARSxOFYgW0pp:bdll6dg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L208)):*
> ```XML
> };
> 
> const mapShapeToFeature = (shape: LineString, shouldReverseGeometry: boolean) =>
> ```

This could benefit from a JSDoc comment explaining what this does. 
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 222 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7WwDQ4STQf0Xvp8_-:-NI7WwDQ4STQf0Xvp8_0:b-fbuma0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L222)):*
> ```XML
> // Minimun length (meters) for infrastructure links,
> // from which feature points are picked from
> const MIN_INFRA_LINK_POINT_PICK_LENGTH = 15;
> ```

In my opinion, all CONSTANTS should be in the top-part of the file. So in this case before or after the `GQL_..` consts.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 233 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7Yu-G4MIADbH4PRRe:-NI7Yu-G4MIADbH4PRRf:b-92w9i2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L233)):*
> ```XML
>       is_traversal_forwards: isTraversalForwards,
>     } = link;
>     // Infrastructure link geometry as feature
> ```

I'm not sure if this comment is in the right place. Or at least I don't understand it
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 234 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7Y9tk3kPa3RUoipmC:-NI7Y9tk3kPa3RUoipmD:b-yhyku1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L234)):*
> ```XML
>     } = link;
>     // Infrastructure link geometry as feature
>     const reverseGeometry = !!shape.coordinates.length && !isTraversalForwards;
> ```

I would name this with a prefix of "should" (if this determines whether we reverse the geometry or not). Without the prefix (like it is now) one could think that the actual reversed geometry is in this variable. 

It is actually already named like this in the mapShapeToFeature function.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 249 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7_3hm3G_uZXCOaxYr:-NI7_3hm3G_uZXCOaxYs:b-ss6tf5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L249)):*
> ```XML
> 
>     // Distances in percentages how far along the infrastructure link the desired point is located
>     let relativeDistancesAlongLink: number[] = [];
> ```

To avoid using `let` you could extract the following if-else block to it's own function.
___
*[`ui/src/utils/gis.ts` line 45 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7ac7FCvGcSYbQp_Vq:-NI7ac7FCvGcSYbQp_Vr:b-e7lt1f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/utils/gis.ts#L45)):*
> ```XML
> 
> /**
>  *
> ```

Nitpicking, but there is an extra * -line here :D
Although I do know that these are not consistent in all parts of the code.
sometimes I've started the actual comment from the first line like this: `/** description already starts here ` and sometimes I've started the actual comment from the second line like this:
```
/**
/* Description starts here
```

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: 5 of 8 files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 208 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7WgcKARSxOFYgW0po:-NI7mLwI8YUP-y8TRzmn:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L208)):*

This could benefit from a JSDoc comment explaining what this does. 
</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 222 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7WwDQ4STQf0Xvp8_-:-NI7g4Mm5zUGv3zNRHhh:b-lg21dr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L222)):*

In my opinion, all CONSTANTS should be in the top-part of the file. So in this case before or after the `GQL_..` consts.
</blockquote></details>

I was thinking about this as well but I think this is kind of special case as the file is full of exported constant functions anyway and moving this to top of this file does really not add any value, just takes it further from its usage context. But I don't really care where it is, I can move it to the top. :) Modern IDE will find the constant and its usages no matter where it is.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 233 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7Yu-G4MIADbH4PRRe:-NI7mlFS5Lw4-0ljBCLt:bq9fr4j) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L233)):*

I'm not sure if this comment is in the right place. Or at least I don't understand it
</blockquote></details>

It is trying to explain what happens in the rows following it, but as we don't want to add empty lines between the the comment line and the lines following it, it seems like it has something to do with the following row. I changed it a bit, hope it is easier to read now... Or what should I do if I want to comment multiple lines of code to explain what is happening in the next few lines..?
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 234 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7Y9tk3kPa3RUoipmC:-NI7nXoQCKgcnwXkmWJV:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L234)):*

I would name this with a prefix of "should" (if this determines whether we reverse the geometry or not). Without the prefix (like it is now) one could think that the actual reversed geometry is in this variable. 

It is actually already named like this in the mapShapeToFeature function.
</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 249 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7_3hm3G_uZXCOaxYr:-NI7nYXTDG3AKvYhZicV:b8s9o7h) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L249)):*

To avoid using `let` you could extract the following if-else block to it's own function.
</blockquote></details>

Done. This came to my mind also when I did this but there was so much state that would have to be passed as parameters to the function I decided it's not worth it. But I did the change now and I guess it is ok.
___
*[`ui/src/utils/gis.ts` line 45 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7ac7FCvGcSYbQp_Vq:-NIC5FhiC3QutZHDqOX2:bnbbw3x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/utils/gis.ts#L45)):*

Nitpicking, but there is an extra * -line here :D
Although I do know that these are not consistent in all parts of the code.
sometimes I've started the actual comment from the first line like this: `/** description already starts here ` and sometimes I've started the actual comment from the second line like this:
```
/**
/* Description starts here
```
</blockquote></details>

Yep. This one row is extra, but I don't like if the comment starts from the first row, too little space.

HenrikHartiala (commented):
> Reviewed 1 of 8 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: 7 of 8 files reviewed, all discussions resolved (waiting on @eleksi)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 208 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7WgcKARSxOFYgW0po:-NICHSqHCQMHVVojdKgP:b-j669jk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L208)):*

Done.
</blockquote></details>

Hmm, still a bit confused what this does. So the name is `mapShapeToFeature` and the comment is that it maps the shape to GeoJSON Feature, but then inside this function it calls `mapGeoJSONtoFeature`.

So is GeoJSON actually the input, not the output?
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 222 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7WwDQ4STQf0Xvp8_-:-NICIeeKBv9nL4IhnC85:b-chxuxi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L222)):*

I was thinking about this as well but I think this is kind of special case as the file is full of exported constant functions anyway and moving this to top of this file does really not add any value, just takes it further from its usage context. But I don't really care where it is, I can move it to the top. :) Modern IDE will find the constant and its usages no matter where it is.
</blockquote></details>

No, but when we have multiple CONSTANTS (just the ones that are written in caps and only there to give names to numbers for example), they are easy to adjust if needed when they are on top of the file. 

Also if it is used in multiple places, it is nice also to have it on top of the file. It is true that the IDE finds it anywhere, but it also is a bit noice in the middle of the code.
But this is good now, thanks!
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 233 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7Yu-G4MIADbH4PRRe:-NICJibz9kB0tR3Wgg5W:bmtyql0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L233)):*

It is trying to explain what happens in the rows following it, but as we don't want to add empty lines between the the comment line and the lines following it, it seems like it has something to do with the following row. I changed it a bit, hope it is easier to read now... Or what should I do if I want to comment multiple lines of code to explain what is happening in the next few lines..?
</blockquote></details>

Hmm, sorry I might have been wrong about the empty lines in multiline commenting. (Although if you need to comment multiple lines, that might also be a hint to have a function for that part, not saying that these should). 

So maybe the empty line is the way to go for multiline comments. I think I just haven't commented multiple lines myself so that might be why it hilighted as strange for me earlier. 

Anyway, this comment is better now! Thanks
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 249 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7_3hm3G_uZXCOaxYr:-NIClwv2ACwrD9Glm0sS:b-mb8f2w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L249)):*

Done. This came to my mind also when I did this but there was so much state that would have to be passed as parameters to the function I decided it's not worth it. But I did the change now and I guess it is ok.
</blockquote></details>

Thanks, I think it is worth it because the actual logic how those are determined are not relevant in the perspective of `mapInfraLinksToFeature`. And btw I think you can narrow down the parameters to 4, if you pass the index as parameter and determine `isFirstLink` and `isLastLink` inside the function. Actually you could narrow it down even to 3 if you also determine the featureLength and isLoopLink inside the function. So just pass in linkFeature, shape and direction to the function.
Or what do you think?
___
*[`ui/src/utils/gis.ts` line 45 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7ac7FCvGcSYbQp_Vq:-NICng7TBPel53aG_I3g:b2h0aso) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/utils/gis.ts#L45)):*

Yep. This one row is extra, but I don't like if the comment starts from the first row, too little space.
</blockquote></details>

Agreed, need to change the ones that starts the description from the first row already.

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @eleksi)

jarkkoka (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: all files reviewed, 3 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 258 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIDQzSl0Rvkd_b46cfd:-NIDQzSl0Rvkd_b46cfe:b-8hwvqd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L258)):*
> ```XML
>     // In case link is first of route's infrastructure links, replace 33% point with starting point.
>     // In case link is last of route's infrastructure links, replace 66% point with endpoint.
>     return [isFirstLink ? 0 : 1 / 3, isLastLink ? 1 : 2 / 3];
> ```

When a route consists of only one link, which is a bi-directional closed loop, this will cause a failure when editing the route. Of course, real-life routes consist of multiple links, so I don't see this as much of a problem. However, there is a bigger problem with this `if` block. If the first link in the route is a bi-directional closed loop, then this would yield an array of `[0, 2/3]` which causes the map-matching service to traverse the link backwards, since the subsection `[2/3, 1]` of the link is shorter than `[0, 2/3]`. A similar problem occurs when the last link is a bi-directional closed loop.

The issue could be fixed by returning simply:

```javascript
return [1/3, 2/3];
```

Also, it is not necessary to return strictly 0/1 for the first/last link, so it is safe to return always `[1/3, 2/3]` from this  `if` block.



_Code quote:_
```XML
return [isFirstLink ? 0 : 1 / 3, isLastLink ? 1 : 2 / 3]
```
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 262 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIDRUZ_4bWifdvdUFlg:-NIDRUZ_4bWifdvdUFlh:b57cox2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L262)):*
> ```XML
>   if (isOnlyLink) {
>     // If this link is the only link, return first and last points.
>     return [0, 1];
> ```

When a route consists of only one link, which is a one-way closed loop, this will cause a failure when editing the route. Of course, real-life routes consist of several links, but it's better to fix this e.g. by setting the array to [0.1, 0.9] or using other "magic" values between zero and one. Ideally, the locations of the first/last stop point on the route would be taken into account, but that's too cumbersome.

E.g. the following could work.

```javascript
if (isOnlyLink) {
  return isClosedLoop ? [0.1, 0.9] : [0, 1];
}
```

This simpler version would also work:

```javascript
if (isOnlyLink) {
  return [0.1, 0.9];
}
```

The fact is that for non-closed-loop links, any valid subrange between [0, 1] will result in the entire link geometry being returned in the map-matching service response.



_Code quote:_
```XML
  if (isOnlyLink) {
    // If this link is the only link, return first and last points.
    return [0, 1];
  }
```

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: all files reviewed, 3 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 262 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIDRUZ_4bWifdvdUFlg:-NIDhNE24Pa-GyF7LSUA:b-nz4z8o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L262)):*

When a route consists of only one link, which is a one-way closed loop, this will cause a failure when editing the route. Of course, real-life routes consist of several links, but it's better to fix this e.g. by setting the array to [0.1, 0.9] or using other "magic" values between zero and one. Ideally, the locations of the first/last stop point on the route would be taken into account, but that's too cumbersome.

E.g. the following could work.

```javascript
if (isOnlyLink) {
  return isClosedLoop ? [0.1, 0.9] : [0, 1];
}
```

This simpler version would also work:

```javascript
if (isOnlyLink) {
  return [0.1, 0.9];
}
```

The fact is that for non-closed-loop links, any valid subrange between [0, 1] will result in the entire link geometry being returned in the map-matching service response.
</blockquote></details>

Fix to previous message:

Wrote: "The fact is that for non-closed-loop links..."

Should have been:

"For single-link routes, for both unidirectional closed loops and non-closed-loops, any valid subrange between `[0, 1]` will result in the entire link geometry being returned in the map-matching service response."

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: all files reviewed, 3 unresolved discussions (waiting on @eleksi and @jarkkoka)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 258 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIDQzSl0Rvkd_b46cfd:-NIGcmTp7Ry3hixMpSTj:b55etit) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L258)):*

When a route consists of only one link, which is a bi-directional closed loop, this will cause a failure when editing the route. Of course, real-life routes consist of multiple links, so I don't see this as much of a problem. However, there is a bigger problem with this `if` block. If the first link in the route is a bi-directional closed loop, then this would yield an array of `[0, 2/3]` which causes the map-matching service to traverse the link backwards, since the subsection `[2/3, 1]` of the link is shorter than `[0, 2/3]`. A similar problem occurs when the last link is a bi-directional closed loop.

The issue could be fixed by returning simply:

```javascript
return [1/3, 2/3];
```

Also, it is not necessary to return strictly 0/1 for the first/last link, so it is safe to return always `[1/3, 2/3]` from this  `if` block.
</blockquote></details>

Nice catch :ok_hand: . We should document this in the comments as well :+1:

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: all files reviewed, 5 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 266 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIGr9gPBn_bwZduHUtT:-NIGr9gR1f_YUAyGzXML:bnffj11) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L266)):*
> ```XML
>   if (isFirstLink) {
>     // Otherwise,
>     // If link is first of route's infrastructure links, add link's starting point.
> ```

In case of closed-loops links (bi-directional or not), 0 or 1 will not work.



_Code quote:_
```XML
  if (isFirstLink) {
    // Otherwise,
    // If link is first of route's infrastructure links, add link's starting point.
    return [0];
  }
```
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 276 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIGrJ_H1kqJ-VdzasZb:-NIGrJ_H1kqJ-VdzasZc:bnffj11) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L276)):*
> ```XML
>   if (isLastLink) {
>     // If link is last of route's infrastructure links, add link's endpoint.
>     return [1];
> ```

In case of closed-loops links (bi-directional or not), 0 or 1 will not work.



_Code quote:_
```XML
  if (isLastLink) {
    // If link is last of route's infrastructure links, add link's endpoint.
    return [1];
  }
```

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: all files reviewed, 5 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 258 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIDQzSl0Rvkd_b46cfd:-NIGrQBM-iHxEnGKINlO:bknen0k) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L258)):*

Nice catch :ok_hand: . We should document this in the comments as well :+1: 
</blockquote></details>

Yeah. I would actually suggest replacing the start of the `if` block sequence with the following code where closed loops are dealt all at once:

```javascript
  if (isLoopLink) {
    if (direction === InfrastructureNetworkDirectionEnum.Bidirectional || isOnlyLink) {
      // If closed-loop link is bidirectional or only link along route, add two points at fractional
      // locations (33%, 66%) of the link geometry.
      return [1 / 3, 2 / 3];
    }
    return [0.5];
  }
  if (isOnlyLink) {
    // If this link is the only link, return first and last points.
    return [0, 1];
  }
  // Other if blocks as they are now...
  ...
```

The above code is simple and concise and should cover all cases.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: all files reviewed, 5 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 262 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIDRUZ_4bWifdvdUFlg:-NIGtuw169Ko0WzwOI8Q:b-y4gcuo) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L262)):*

Fix to previous message:

Wrote: "The fact is that for non-closed-loop links..."

Should have been:

"For single-link routes, for both unidirectional closed loops and non-closed-loops, any valid subrange between `[0, 1]` will result in the entire link geometry being returned in the map-matching service response."
</blockquote></details>

Forget this change request to `if (isOnlyLink) {...}` block. I updated the suggestion for the first `if` block. It also fixes this problem.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 266 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIGr9gPBn_bwZduHUtT:-NIGuINn6svYGQEBNFBu:b-3i5f9w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L266)):*

In case of closed-loops links (bi-directional or not), 0 or 1 will not work.
</blockquote></details>

See my suggestion with regard to the first `if` block of the series.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 276 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIGrJ_H1kqJ-VdzasZb:-NIGuMvnD2rMyQ6wyULd:b-3i5f9w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/c0570a91017113f80ebcfc8fb6a192821f35c335/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L276)):*

In case of closed-loops links (bi-directional or not), 0 or 1 will not work.
</blockquote></details>

See my suggestion with regard to the first `if` block of the series.

jarkkoka (commented):
Basically, looks good to me. Left two comments. Consider what to do with them.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-:-NIX3-Gr3Q5hLp-iN2HD:b-7sixwd)* status: all files reviewed, 8 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 242 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIX-yTHADcehAIX7D00:-NIX-yTHADcehAIX7D01:b2u1qwg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L242)):*
> ```XML
>   isBetweenLink: boolean;
>   isOnlyLink: boolean;
>   isBidirectional: boolean;
> ```

Good abstraction clarifying the code. :+1: 



_Code quote:_
```XML
interface SnapPointCalculateParams {
  isFirstLink: boolean;
  isLastLink: boolean;
  isLoopLink: boolean;
  isBetweenLink: boolean;
  isOnlyLink: boolean;
  isBidirectional: boolean;
  isLongEnough: boolean;
}
```
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 248 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIX1FWV1h3dWf-e3KSo:-NIX1FWV1h3dWf-e3KSp:b-wllm2s) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L248)):*
> ```XML
> const getSnapPointCalculateParamsForInfraLink = (
>   link: RouteInfraLink,
>   linkFeature: GeoJSON.Feature<LineString>,
> ```

It seems to me that:
- The only property used from `link` is its `direction`. Hence, `link` parameter could be replaced by `linkDirection`, i.e. the call-site passes the direction instead of the whole `link` object.

This would make the code a bit more readable and better abstracted.



_Code quote:_
```XML
  link: RouteInfraLink,
  linkFeature: GeoJSON.Feature<LineString>,
```
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 311 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIWzusOEySlMyIBYxek:-NIWzusOEySlMyIBYxel:b-c5pqdt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L311)):*
> ```XML
>   if (isBetweenLink && isLongEnough) {
>     // If link is in between route's other infrastructure links (is not first or last link),
>     // add link's center point (if infrastructure link is long enough).
> ```

Could not figure out what was the bug and the reason for trimming the arrays for the first and last link?...

Could this be further simplified with the following code?

```javascript
if (isFirstLink || isLastLink || isLongEnough) {
  return [0.5];
}
```



_Code quote:_
```XML
  if (isFirstLink) {

    // If link is first of route's infrastructure links, add link's starting point and center point.
    return [0, 0.5];
  }
  if (isBetweenLink && isLongEnough) {
    // If link is in between route's other infrastructure links (is not first or last link),
    // add link's center point (if infrastructure link is long enough).
    return [0.5];
  }
  if (isLastLink) {
    // If link is last of route's infrastructure links, add link's center point and endpoint.
    return [0.5, 1];
```

HenrikHartiala (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: all files reviewed, 7 unresolved discussions (waiting on @eleksi and @jarkkoka)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 208 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7WgcKARSxOFYgW0po:-NIX3G7c6xtINtcZ68qy:b-tbntzb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L208)):*

Technically both input and output are GeoJSON, input is GeoJSON.Geometry (LineString) and output is GeoJSON.Feature. What we need to do is take the geometry, reverse it if needed and then put it inside a Feature object.

I didn't want to name this function anything like mapGeoJSONGeometryToGeoJSONFeature. This is purely just for the purpose of splitting `mapInfraLinksToFeature` logic to smaller pieces, so the parameter is called shape instead of trying to be very generic like "geometry". I could add information about LineString being GeoJSON to the comments if that makes it more clear, but I didn't really think documenting this one-off function with such precision was necessary.

Anyway, after changes it is now completely different function.
</blockquote></details>

Okay, Understood. The result seems nice and clear now!
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 249 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NI7_3hm3G_uZXCOaxYr:-NIX4BAU3oDgyeWaIG9s:b-wf2ww) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/371c0075f84f536c1d9bcd2b38cdf5e6317333e7/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L249)):*

`isFirstLink` needs `index`, `isLastLink` needs `index` and `infraLinks.length`.

Well, I could write it so that we need linkFeature, direction, index and something I can use to determine if the link is the last one (e.g. infraLinks.length), so there would be 4 parameters. But I don't think getting 1 parameter out of the function really helps us any way. I would like to keep `getRelativeSnapPointDistancesAlongLink` as simple as possible and give it only self explanatory parameters.

Narrowing it down to 3 parameters would probably require passing link, infralinks and index, which the I don't think is smart.

Now there is a completely new suggestion which I think makes makes at least this function more clear.
</blockquote></details>

Ah, missed the `infraLinks.length`. But this function is very nice and clear now! :ok_hand: 
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 236 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIWzAhn8tFh9dc_TklO:-NIWzAhn8tFh9dc_TklP:b-let42i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L236)):*
> ```XML
> };
> 
> interface SnapPointCalculateParams {
> ```

Calculation*
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 246 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIWzNcHBcoqJrCPrTtQ:-NIWzNcHBcoqJrCPrTtR:b-let42i) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L246)):*
> ```XML
> }
> 
> const getSnapPointCalculateParamsForInfraLink = (
> ```

Calculation*
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 266 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIX-OZC9_L0ku_SIkL_:-NIX-OZC9_L0ku_SIkLa:bqz5dip) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L266)):*
> ```XML
>   const isBidirectional =
>     link.direction === InfrastructureNetworkDirectionEnum.Bidirectional;
>   const isLongEnough = featureLength > MIN_INFRA_LINK_POINT_PICK_LENGTH;
> ```

Hmm, `isLongEnough` kind of masks the determination of what is `enough`. I'm thinking of either not using a variable here at all, or rename this to something like "linkLengthIsValid" or "isValidLength" etc. 

Although not sure if this is a problem at all, not sure if the renaming of the variable makes difference here. Maybe the only thing that would make a difference was to remove the variable and just use the `featureLength > MIN_INFRA_LINK_POINT_PICK_LENGTH`.

But all this said, it is also fine with me to leave it as it is :+1:

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/408)* status: 7 of 8 files reviewed, 7 unresolved discussions (waiting on @jarkkoka)

___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 242 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIX-yTHADcehAIX7D00:-NIX8tlrDbH4Gh5JKyxS:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L242)):*

Good abstraction clarifying the code. :+1: 
</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 248 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIX1FWV1h3dWf-e3KSo:-NIX8w--CcmPn338knxV:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L248)):*

It seems to me that:
- The only property used from `link` is its `direction`. Hence, `link` parameter could be replaced by `linkDirection`, i.e. the call-site passes the direction instead of the whole `link` object.

This would make the code a bit more readable and better abstracted.

</blockquote></details>

Done.
___
*[`ui/src/hooks/routes/useExtractRouteFromFeature.ts` line 311 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/408#-NIWzusOEySlMyIBYxek:-NIX93mv7whT9f_K3UgV:b8kyqvs) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/91a08a7b85309d82c08a3686600640cea930ec73/ui/src/hooks/routes/useExtractRouteFromFeature.ts#L311)):*

Could not figure out what was the bug and the reason for trimming the arrays for the first and last link?...

Could this be further simplified with the following code?

```javascript
if (isFirstLink || isLastLink || isLongEnough) {
  return [0.5];
}
```
</blockquote></details>

As discussed in slack, first and last point are there for first and last link to make it look better in UI. Also the bug was related to open loop link as first or last link.

PR #409, opened 2022-11-23
Disable adding new route for past lines in line view

eleksi:
Resolves https://github.com/HSLdevcom/jore4/issues/809

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/409)

### DIFF
--- a/ui/src/components/routes-and-lines/line-details/CreateRouteBox.tsx
+++ b/ui/src/components/routes-and-lines/line-details/CreateRouteBox.tsx
@@ -5,7 +5,7 @@ import { SimpleButton } from '../../../uiComponents';
 
 interface Props {
   className?: string;
-  onCreateRoute: () => void;
+  onCreateRoute?: () => void;
 }
 
 export const CreateRouteBox: React.FC<Props> = ({
@@ -18,14 +18,20 @@ export const CreateRouteBox: React.FC<Props> = ({
     <Column
       className={`items-center border border-light-grey bg-background p-8 ${className}`}
     >
-      <SimpleButton
-        id="create-route-button"
-        containerClassName="mb-4"
-        onClick={onCreateRoute}
-      >
-        {t('lines.createNewRoute')}
-      </SimpleButton>
-      <span>{t('lines.createNewRouteInstructions')}</span>
+      {onCreateRoute && (
+        <SimpleButton
+          id="create-route-button"
+          containerClassName="mb-4"
+          onClick={onCreateRoute}
+        >
+          {t('lines.createNewRoute')}
+        </SimpleButton>
+      )}
+      {onCreateRoute ? (
+        <span>{t('lines.createNewRouteInstructions')}</span>
+      ) : (
+        <span>{t('lines.cannotCreateRoute')}</span>
+      )}
     </Column>
   );
 };
--- a/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
@@ -1,3 +1,4 @@
+import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
 import { LineAllFieldsFragment } from '../../../generated/graphql';
 import {
@@ -13,6 +14,7 @@ import {
   setLineInfoAction,
 } from '../../../redux';
 import { Priority } from '../../../types/Priority';
+import { isPastEntity } from '../../../utils';
 import { PageHeader } from '../common/PageHeader';
 import { ViaModal } from '../via/ViaModal';
 import { ActionsRow } from './ActionsRow';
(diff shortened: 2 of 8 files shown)


HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/409)* status: all files reviewed, 2 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/utils/validity.ts` line 29 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/409#-NHZAeNB7nQTrOrqMafy:-NHZAeNB7nQTrOrqMafz:b-xlhf50) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a3a563a0fe2d00448b385451f8c97f274fb82e9b/ui/src/utils/validity.ts#L29)):*
> ```XML
> };
> 
> export const isPastEntity = (
> ```

These are nice! :+1:

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/409)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/routes-and-lines/line-details/CreateRouteBox.tsx` line 21 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/409#-NHZ72lm6_l7Dymht3g_:-NHZvd1b9LHvg3aiwSU4:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a3a563a0fe2d00448b385451f8c97f274fb82e9b/ui/src/components/routes-and-lines/line-details/CreateRouteBox.tsx#L21)):*

Just making sure, is onCreateRoute false _only_ if the line is in the past? And if so, should it be named to something that indicates that. Practically `IsLineNotInThePast` <- this but something which does not include the negation :D 

Ah now I see that it is also the onClick function. So I would not use the same for both of these, I would definitely add another parameter for determining if the line is in the past or not. The problem with using this onCreatroute here, is that it is hard to understand what is the connection between it and line's validity and also if the onCreateRoute will be changed in the future to have other conditions than the validity time, we still show the "cannotCreateRouteForPastLine" text, even though it might not be the reason anymore.
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/line-details/CreateRouteBox.tsx` line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/409#-NHZ87XwCNvayg2Gxw8x:-NHZUExH3fomn5VGFsZe:b3hh0ia) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a3a563a0fe2d00448b385451f8c97f274fb82e9b/ui/src/components/routes-and-lines/line-details/CreateRouteBox.tsx#L22)):*

If you want to avoid the empty fragment, you could use `<Visible>` component for the `<SimpleButton>` and then use the ternary to decide which text to include in the span.
</blockquote></details>

As discussed in Slack, ts unfortunately does not realize onCreateRoute cannot be undefined if `<Visible>` is used.
___
*[`ui/src/components/routes-and-lines/line-details/LineTitle.tsx` line 52 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/409#-NHZ9ydg3YAEuySudOO1:-NHZTfdzAB2jVwEmMd4N:b3hh0ia) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a3a563a0fe2d00448b385451f8c97f274fb82e9b/ui/src/components/routes-and-lines/line-details/LineTitle.tsx#L52)):*

Could also use the `<Visible>` component here, but either is fine with me.
</blockquote></details>

As discussed in Slack, ts unfortunately does not realize onCreateRoute cannot be undefined if `<Visible>` is used.
___
*[`ui/src/utils/validity.ts` line 15 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/409#-NHYvzV3-E96pf9tXkYU:-NHZTaYoCozqF0ol1C5P:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/57067a8c157b5321262fb89a76939e221ed2d482/ui/src/utils/validity.ts#L15)):*

These functions would benefit from JSDoc comments explaining the function. Maybe include the comments below in this JSDoc comment and also explain the "entity". If I understand this correctly, the entity can be anything which has `validity_start` and `validity_end` properties.

Although Im not sure if it would be better to have these parameters in the form of object and spread the validity attributes. Like this:
```
export const isFutureEntity = ({
  observationDate: DateTime,
  validity_start: DateTime,
  validity_end: DateTime
})
```
But this is quite handy the way you implemented it so maybe this is nice as the way it is, so if you just add the JSDoc comments, this would be bueno :+1: 
</blockquote></details>

Done.

HenrikHartiala (approved):
:lgtm: Great work!

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/409#-:-NHcJcWm7q_LLQXBjex6:bolbkdq)* status: all files reviewed, 2 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/utils/validity.ts` line 21 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/409#-NHcJXiu1a0Exlf6nlrM:-NHcJXiu1a0Exlf6nlrN:bcv502w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/176b587d8a0c689baf05e7a0d99cfd22ddcaef07/ui/src/utils/validity.ts#L21)):*
> ```XML
>  * @param entity Object with validity period, e.g. stop, line or route
>  * @returns True if `entity`'s validity start is in the future
>  */
> ```

Nice :+1:

PR #411, opened 2022-11-24
Add test that creates a stop with an end time

sr-vala:
Resolves HSLdevcom/jore4#1021

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)

### DIFF
--- a/cypress/e2e/createStop.cy.ts
+++ b/cypress/e2e/createStop.cy.ts
@@ -3,12 +3,17 @@ import {
   ReusableComponentsVehicleSubmodeEnum,
   VehicleSubmodeOnInfraLinkInsertInput,
 } from '@hsl/jore4-test-db-manager';
+import { ConfirmSaveForm, Map } from '../pageObjects';
 import { FilterPanel } from '../pageObjects/FilterPanel';
 import { MapItemCreator } from '../pageObjects/MapItemCreator';
 import { insertToDbHelper, removeFromDbHelper } from '../utils';
 import { deleteStopsByLabel } from './utils';
 
-const testStopLabels = ['T0001', 'TManual'];
+const testStopLabels = {
+  testLabel1: 'T0001',
+  manualCoordinatesLabel: 'TManual',
+  endDateLabel: 'TEndDate',
+};
 
 const testInfraLinks = {
   1: 'c63b749f-5060-4710-8b07-ec9ac017cb5f',
@@ -32,13 +37,15 @@ const dbResources = {
 };
 
 const clearDatabase = () => {
-  deleteStopsByLabel(testStopLabels);
+  deleteStopsByLabel(Object.values(testStopLabels));
   removeFromDbHelper(dbResources);
 };
 
 describe('Stop creation tests', () => {
   let mapItemCreator: MapItemCreator;
   let mapFilterPanel: FilterPanel;
+  let map: Map;
+  let confirmSaveForm: ConfirmSaveForm;
 
   before(() => {
     cy.fixture('infraLinks/infraLinks.sql').then((infraLinksQuery) => {
@@ -52,13 +59,17 @@ describe('Stop creation tests', () => {
 
     mapItemCreator = new MapItemCreator();
     mapFilterPanel = new FilterPanel();
+    map = new Map();
+    confirmSaveForm = new ConfirmSaveForm();
 
     cy.setupTests();
     cy.mockLogin();
 
-    cy.visit(
-      '/routes?mapOpen=true&lng=24.93021804533524&lat=60.164074274478054&z=15',
-    );
+    map.visit({
+      zoom: 15,
+      lat: 60.164074274478054,
+      lng: 24.93021804533524,
+    });
   });
 
   after(() => {
@@ -71,7 +82,7 @@ describe('Stop creation tests', () => {
     { scrollBehavior: 'bottom', defaultCommandTimeout: 10000 },
     () => {
       mapItemCreator.createStopAtLocation({
-        stopFormInfo: { label: testStopLabels[0] },
+        stopFormInfo: { label: testStopLabels.testLabel1 },
         clickRelativePoint: {
           xPercentage: 43.5,
           yPercentage: 53,
@@ -79,15 +90,14 @@ describe('Stop creation tests', () => {
         validityStartISODate: '2022-01-01',
       });
 
-      // TODO: Currently 'toast.checkStopSubmitSuccess()' doesn't work because
-      // cypress on chrome can't find the toast messages for some reason. Find out why this is and add
-      // assertion for the toast message
-      cy.wait('@gqlInsertStop').its('response.statusCode').should('equal', 200);
+      mapItemCreator.gqlStopShouldBeCreatedSuccessfully();
+
+      mapItemCreator.checkStopSubmitSuccessToast();
 
       mapFilterPanel.toggleShowStops(ReusableComponentsVehicleModeEnum.Bus);
 
       cy.getByTestId(
-        `Map::Stops::stopMarker::${testStopLabels[0]}_Standard`,
+        `Map::Stops::stopMarker::${testStopLabels.testLabel1}_Standard`,
       ).should('exist');
     },
   );
@@ -99,7 +109,7 @@ describe('Stop creation tests', () => {
       // Create stop
       mapItemCreator.createStopAtLocation({
         stopFormInfo: {
-          label: testStopLabels[1],
+          label: testStopLabels.manualCoordinatesLabel,
           // Actual coordinates will be on Topeliuksenkatu
           latitude: '60.18083637150667',
           longitude: '24.9215054260969',
@@ -111,20 +121,53 @@ describe('Stop creation tests', () => {
         validityStartISODate: '2022-01-01',
       });
 
-      // TODO: Currently 'toast.checkStopSubmitSuccess()' doesn't work because
-      // cypress on chrome can't find the toast messages for some reason. Find out why this is and add
-      // assertion for the toast message
-      cy.wait('@gqlInsertStop').its('response.statusCode').should('equal', 200);
+      mapItemCreator.gqlStopShouldBeCreatedSuccessfully();
+
+      mapItemCreator.checkStopSubmitSuccessToast();
+
       // Change map position to created stop location
-      cy.visit(
-        '/routes?lat=60.1805636468358&lng=24.918451016960763&z=15.008647482331973&mapOpen=true',
-      );
+      map.visit({
+        zoom: 15,
+        lat: 60.1805636468358,
+        lng: 24.918451016960763,
+      });
 
       mapFilterPanel.toggleShowStops(ReusableComponentsVehicleModeEnum.Bus);
 
       cy.getByTestId(
-        `Map::Stops::stopMarker::${testStopLabels[1]}_Standard`,
+        `Map::Stops::stopMarker::${testStopLabels.manualCoordinatesLabel}_Standard`,
       ).should('exist');
+    },
+  );
+
+  it(
+    'Should create stop with end time on map',
+    // Map opening seems to take time, so we increase the timeout
+    { scrollBehavior: 'bottom', defaultCommandTimeout: 10000 },
+    () => {
+      mapItemCreator.createStopAtLocation({
+        stopFormInfo: { label: testStopLabels.endDateLabel },
+        clickRelativePoint: {
+          xPercentage: 43.5,
+          yPercentage: 53,
+        },
+        validityStartISODate: '2022-01-01',
+        validityEndISODate: '2040-12-31',
+      });
+
+      mapItemCreator.gqlStopShouldBeCreatedSuccessfully();
+
+      mapItemCreator.checkStopSubmitSuccessToast();
+
+      mapFilterPanel.toggleShowStops(ReusableComponentsVehicleModeEnum.Bus);
+
+      cy.getByTestId(
+        `Map::Stops::stopMarker::${testStopLabels.endDateLabel}_Standard`,
+      ).click();
+
+      map.stopPopUp.getEditButton().click();
+
+      confirmSaveForm.getEndDateInput().should('have.value', '2040-12-31');
     },
   );
 });
--- a/cypress/pageObjects/ConfirmSaveForm.ts
+++ b/cypress/pageObjects/ConfirmSaveForm.ts
@@ -20,13 +20,14 @@ export class ConfirmSaveForm {
       .type(isoDate);
   }
 
+  getEndDateInput() {
+    return cy.getByTestId('ConfirmSaveForm::endDateInput');
+  }
+
   setEndDate(isoDate: string) {
     // This invoke is a workaround to
     // prevent map from zooming out when typing '-' value to the date input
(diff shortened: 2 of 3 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/createStop.cy.ts` line 124 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxkUh64pYJ9r5ICdsP:-NHxkUh64pYJ9r5ICdsQ:b-q2o5ev) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L124)):*
> ```XML
>       // Change map position to created stop location
>       map.visit({
>         zoom: 15.008647482331973,
> ```

Is it really relevant to define zoom level with this many digits? :raised_eyebrow: 
___
*[`cypress/e2e/createStop.cy.ts` line 138 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxk8rI4wL_f1bjpu0f:-NHxk8rI4wL_f1bjpu0g:bwzs7kj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L138)):*
> ```XML
> 
>   it(
>     'Should create stop with end time on map',
> ```

By the way, we are not verifying that end time is actually saved to the db.
This tests tests "something" even now so this is non-blocking comment, but maybe we should either check graphql response or then verify via UI that end date actually ends up somewhere?
___
*[`cypress/e2e/createStop.cy.ts` line 143 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxjOwVC_xJsanpKcUv:-NHxjOwVC_xJsanpKcUw:beqgomw) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L143)):*
> ```XML
>     () => {
>       mapItemCreator.createStopAtLocation({
>         stopFormInfo: { label: testStopLabels[2] },
> ```

`testStopLabels[2]` is only used within this test case.
Instead of defining it in shared array and then reading it from there it would be better to create new string variable to be used only in this this test case (inside this `it` block):

```
const testStopLabel = '...';
...
```

Also existing items in `testStopLabels` array seems to be used only in one test case each, so I would remove the whole array and do this same thing for existing tests too.

HenrikHartiala (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @sr-vala and @tk-vincit)

___
*[`cypress/e2e/createStop.cy.ts` line 138 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxhr3eEjYlwlf7v1uK:-NHxhr3eEjYlwlf7v1uL:b-qswqjb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L138)):*
> ```XML
> 
>   it(
>     'Should create stop with end time on map',
> ```

So is the main point of this test to check that a stop creation is successful with the endDate given? Because if it is, then I think we could just input the end date to the existing stop creation test.
Otherwise I would rather test that the endDate "works". So either check that the endDate is saved (open the stop modify again and verify this) or change the observationDate and check if the stop's endDate is corresponding to that correctly.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`cypress/e2e/createStop.cy.ts` line 124 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxkUh64pYJ9r5ICdsP:-NHxrMMN6GOaBzvsbaqG:bq76af6) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L124)):*

Is it really relevant to define zoom level with this many digits? :raised_eyebrow: 
</blockquote></details>

I did not change the zoom level. If it was originally somehow wrong, let's handle that separately.
___
*[`cypress/e2e/createStop.cy.ts` line 143 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxjOwVC_xJsanpKcUv:-NHxwOB76YR_jTn60Has:b4gingd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L143)):*

`testStopLabels[2]` is only used within this test case.
Instead of defining it in shared array and then reading it from there it would be better to create new string variable to be used only in this this test case (inside this `it` block):

```
const testStopLabel = '...';
...
```

Also existing items in `testStopLabels` array seems to be used only in one test case each, so I would remove the whole array and do this same thing for existing tests too.
</blockquote></details>

It is used in deleteStopsByLabel, which is used in clearDatabase which is used in beforeEach.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala and @sr-vala)

___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 2 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NI1ECEk22ISsbQmtoZK:-NI1ECEk22ISsbQmtoZL:b7sjfdv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e085fce0118774b4ca10f74e3a49b9a071d8c1bb/cypress/pageObjects/ConfirmSaveForm.ts#L2)):*
> ```XML
> export class ConfirmSaveForm {
>   getEndDateInput() {
> ```

This should be located right before `setEndDate` function definition. This should also be used there to prevent having to define testid twice.
___
*[`cypress/e2e/createStop.cy.ts` line 124 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxkUh64pYJ9r5ICdsP:-NI1DBrgC9rszXeO8zxd:bw7knnx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L124)):*

I did not change the zoom level. If it was originally somehow wrong, let's handle that separately.
</blockquote></details>

Yet there was still some other refactorings done in this commit. I don't understand why not to fix this now, this is literally one line of code and really trivial change.  Probably would have been faster to fix that than to write this commit, as there was other changes to do anyway. But whatever then.
___
*[`cypress/e2e/createStop.cy.ts` line 143 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxjOwVC_xJsanpKcUv:-NI1EX2-67HpT04MTTpY:b407mzb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L143)):*

It is used in deleteStopsByLabel, which is used in clearDatabase which is used in beforeEach.
</blockquote></details>

True, didn't notice that, sorry.



_Code quote:_
```XML
      mapItemCreator.createStopAtLocation({
```

HenrikHartiala (changes requested):
> Reviewed 1 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)* status: all files reviewed, 2 unresolved discussions (waiting on @sr-vala and @tk-vincit)

___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 2 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NI1ECEk22ISsbQmtoZK:-NI1GHtOAIh0l7niNpbZ:b-ul4ih3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e085fce0118774b4ca10f74e3a49b9a071d8c1bb/cypress/pageObjects/ConfirmSaveForm.ts#L2)):*

This should be located right before `setEndDate` function definition. This should also be used there to prevent having to define testid twice.
</blockquote></details>

Good one!
___
*[`cypress/e2e/createStop.cy.ts` line 12 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NI1EyyM4_YjeWMY25-e:-NI1EyyM4_YjeWMY25-f:b-adjokv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e085fce0118774b4ca10f74e3a49b9a071d8c1bb/cypress/e2e/createStop.cy.ts#L12)):*
> ```XML
> import { deleteStopsByLabel } from './utils';
> 
> const testStopLabels = ['T0001', 'TManual', 'TEndDate'];
> ```

Not part of this PR, but it _might_ make these more readable, if they were in separate constants. It is easier to match names than numbers.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala and @tk-vincit)

___
*[`cypress/pageObjects/ConfirmSaveForm.ts` line 2 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NI1ECEk22ISsbQmtoZK:-NI1N6CP07_Nfhdw36Ra:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e085fce0118774b4ca10f74e3a49b9a071d8c1bb/cypress/pageObjects/ConfirmSaveForm.ts#L2)):*

Good one!
</blockquote></details>

Done.
___
*[`cypress/e2e/createStop.cy.ts` line 138 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxk8rI4wL_f1bjpu0f:-NI1MtQ_194PBUd01G0j:bhoxoxe) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L138)):*

By the way, we are not verifying that end time is actually saved to the db.
This tests tests "something" even now so this is non-blocking comment, but maybe we should either check graphql response or then verify via UI that end date actually ends up somewhere?
</blockquote></details>

Graphql check was already there I think. Added check via UI.

tk-vincit (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-:-NI1UQiY9NAPtF_y7CWe:b-uwkowd)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/e2e/createStop.cy.ts` line 138 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NHxk8rI4wL_f1bjpu0f:-NI1SoQQDlIsMPTE2sgR:b-ucfknh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/78877cad59b8c424e3b3c4c05b133b42a9b8f513/cypress/e2e/createStop.cy.ts#L138)):*

Graphql check was already there I think. Added check via UI.
</blockquote></details>

Thanks!
___
*[`cypress/e2e/createStop.cy.ts` line 12 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NI1EyyM4_YjeWMY25-e:-NI1Ssj24Pi6biLLf4g2:b4q8wko) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e085fce0118774b4ca10f74e3a49b9a071d8c1bb/cypress/e2e/createStop.cy.ts#L12)):*

Not part of this PR, but it _might_ make these more readable, if they were in separate constants. It is easier to match names than numbers.
</blockquote></details>

I agree with this.

Actually currently used implementation of deleteStopsByLabel is naive and thus it could also make sense to define these just one per each test case and then beging the test by removing stop by label there. Drawback of that approach would be that it would then clutter up test cases, so I would not recommend it in fact.

```
export const deleteStopsByLabel = (labels: string[]) => {
  labels.forEach((label) => {
    const query =
      'DELETE FROM "service_pattern"."scheduled_stop_point" WHERE label=?';
    cy.task('executeRawDbQuery', { query, bindings: label });
  });
};

```

Yet another alternative could be something like this:
```
const testStopLabels = {
 someDescriptiveName: 'T0001',
}
```

Then these could be referred in test cases by `testStopLabels.someDescriptiveName` instead of arbitrary index numbers, and these all could still be removed in beforeEach easily like `deleteStopsByLabel(Object.values(testStopLabels))`. So maybe this would be the preferred approach.

Anyway, these are just ideas, right now we have only 3 items here so this is not too difficult to decode when needed.

HenrikHartiala (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/411)* status: all files reviewed, 1 unresolved discussion (waiting on @sr-vala)

___
*[`cypress/e2e/createStop.cy.ts` line 12 at r3](https://reviewable.io/reviews/hsldevcom/jore4-ui/411#-NI1EyyM4_YjeWMY25-e:-NI1Y30ED3Rd0UUEeaOL:bqrwyvb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e085fce0118774b4ca10f74e3a49b9a071d8c1bb/cypress/e2e/createStop.cy.ts#L12)):*

I agree with this.

Actually currently used implementation of deleteStopsByLabel is naive and thus it could also make sense to define these just one per each test case and then beging the test by removing stop by label there. Drawback of that approach would be that it would then clutter up test cases, so I would not recommend it in fact.

```
export const deleteStopsByLabel = (labels: string[]) => {
  labels.forEach((label) => {
    const query =
      'DELETE FROM "service_pattern"."scheduled_stop_point" WHERE label=?';
    cy.task('executeRawDbQuery', { query, bindings: label });
  });
};

```

Yet another alternative could be something like this:
```
const testStopLabels = {
 someDescriptiveName: 'T0001',
}
```

Then these could be referred in test cases by `testStopLabels.someDescriptiveName` instead of arbitrary index numbers, and these all could still be removed in beforeEach easily like `deleteStopsByLabel(Object.values(testStopLabels))`. So maybe this would be the preferred approach.

Anyway, these are just ideas, right now we have only 3 items here so this is not too difficult to decode when needed.
</blockquote></details>

Yeah, I was also thinking about the object approach. But then Couldn't think of the difference between that and just plain constants. But very good point that it would clutter up the tests if we need to remove them one by one. So the object approach would be nice solution to that also :+1:

PR #414, opened 2022-11-29
Adjust stop filters on map according to stop changes

HenrikHartiala:
Resolves HSLdevcom/jore4#879
When editing or creating a new stop, we make sure that after the edit / creation the stop won't dissapear due to map stop filtering. So if we for example create a stop with draft priority and the drafts are not currently shown on map -> change the filters so that drafts are also shown on map.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/414)

### DIFF
--- a/ui/src/components/map/stops/EditStopLayer.tsx
+++ b/ui/src/components/map/stops/EditStopLayer.tsx
@@ -15,6 +15,7 @@ import {
   useCreateStop,
   useDeleteStop,
   useEditStop,
+  useFilterStops,
   useLoader,
   useObservationDateQueryParam,
 } from '../../../hooks';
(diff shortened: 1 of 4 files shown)


HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/414)* status: all files reviewed, 1 unresolved discussion (waiting on @eleksi)

___
*[`ui/src/hooks/stops/useFilterStops.ts` line 221 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/414#-NI6dtTj7yWzbI_4SuQ0:-NI6zVVmBZgla218oY10:b-pdzoxg) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/a6989771a3bee8a0ef47e9c4c3d1d5830cadcc34/ui/src/hooks/stops/useFilterStops.ts#L221)):*

Just wondering if this mapping could be avoided but didn't figure out anything immediately. Somehow this bothers me but I don't know if it is that bad..
</blockquote></details>

Yep, bothered me also. But it is not easily refactored, so the whole useFilterStops structure would need a bit of refactoring, and that is something that deserves it's own PR.

PR #416, opened 2022-11-30
Do not load arrow image multiple times

eleksi:
- Previously right-arrow.png was loaded tons of times because loadMapAssets function was called on every map update.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/416)

### DIFF
--- a/ui/src/utils/map/mapUtils.ts
+++ b/ui/src/utils/map/mapUtils.ts
@@ -24,12 +24,29 @@ export const createGeometryLineBetweenPoints = (
 export const loadMapAssets = (mapRef: React.RefObject<MapRef>) => {
   const map = mapRef.current?.getMap();
 
-  map?.loadImage('/img/arrow-right.png', (error: unknown, image: unknown) => {
-    if (error) throw error;
+  const imageAssets: { name: string; fileUrl: string }[] = [
+    { name: 'arrow', fileUrl: '/img/arrow-right.png' },
+  ];
 
-    // Enable sdf to make enable icon coloring.
-    // https://docs.mapbox.com/help/troubleshooting/using-recolorable-images-in-mapbox-maps/
-    if (!map.hasImage('arrow')) map.addImage('arrow', image, { sdf: true });
+  imageAssets.forEach((asset) => {
+    if (map?.hasImage(asset.name)) {
+      return;
+    }
+
+    map?.loadImage(asset.fileUrl, (error: unknown, image: unknown) => {
+      if (error) {
+        throw error;
+      }
+
+      // Even though we have already checked if the map has the image,
+      // we need to check it again here because this code is inside event callback
+      // which could be called multiple times before the map actually has the image
+      if (!map.hasImage(asset.name)) {
+        // Enable Signed Distance Fields (sdf) to make enable icon coloring.
+        // https://docs.mapbox.com/help/troubleshooting/using-recolorable-images-in-mapbox-maps/
+        map.addImage(asset.name, image, { sdf: true });
+      }
+    });
   });
 };
 


HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/416)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @eleksi)

___
*[`ui/src/utils/map/mapUtils.ts` line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/416#-NI7deVi1fJJn9KI7UNl:-NI7deVjAGgp3x5EHm_3:br7dlkx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5d48ac342b59129b166fdc5525019e01380dfce2/ui/src/utils/map/mapUtils.ts#L27)):*
> ```XML
>   const map = mapRef.current?.getMap();
> 
>   if (map?.hasImage('arrow')) return;
> ```

Two things:
I personally don't like the if-statements which do not have explicit block. It is harder to read and more prone to errors. So I would prefer this as 
```
if(map?.hasImage('arrow')) {
  return;
}
```
Same applies to all the other if-statements in this function. I mean that I'm okay if these are used, I understand it but just saying this if after the explanation it makes sense for you to use the explicit block as well.

Now to the other thing:
This function is named `loadMapAssets`, now we are returning if the 'arrow' is already loaded, but what if we wanted to load something else as well? Practically it should still work because then we could _maybe_ determine that this function is already ran, but still it is not the safe option.

So I would enclose the loading of the image in an if block with a negated if-clause: like so:
```
if (!map.hasImage('arrow)) {
    map?.loadImage('/img/arrow-right.png', (error: unknown, image: unknown) => {
    if (error) throw error;

    // Enable sdf to make enable icon coloring.
    // https://docs.mapbox.com/help/troubleshooting/using-recolorable-images-in-mapbox-maps/
    map.addImage('arrow', image, { sdf: true });
  });
}
```
If I understood the code correctly, you can also remove the `if` from the `map.addImage...`

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/416)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`ui/src/utils/map/mapUtils.ts` line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/416#-NI7deVi1fJJn9KI7UNl:-NICDgEy8GUzG1bCYlBm:b-8ncmyv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5d48ac342b59129b166fdc5525019e01380dfce2/ui/src/utils/map/mapUtils.ts#L27)):*

Two things:
I personally don't like the if-statements which do not have explicit block. It is harder to read and more prone to errors. So I would prefer this as 
```
if(map?.hasImage('arrow')) {
  return;
}
```
Same applies to all the other if-statements in this function. I mean that I'm okay if these are used, I understand it but just saying this if after the explanation it makes sense for you to use the explicit block as well.

Now to the other thing:
This function is named `loadMapAssets`, now we are returning if the 'arrow' is already loaded, but what if we wanted to load something else as well? Practically it should still work because then we could _maybe_ determine that this function is already ran, but still it is not the safe option.

So I would enclose the loading of the image in an if block with a negated if-clause: like so:
```
if (!map.hasImage('arrow)) {
    map?.loadImage('/img/arrow-right.png', (error: unknown, image: unknown) => {
    if (error) throw error;

    // Enable sdf to make enable icon coloring.
    // https://docs.mapbox.com/help/troubleshooting/using-recolorable-images-in-mapbox-maps/
    map.addImage('arrow', image, { sdf: true });
  });
}
```
If I understood the code correctly, you can also remove the `if` from the `map.addImage...`
</blockquote></details>

As discussed in Slack, I ended up refactoring the entire thing to make it more welcoming to possible future asset additions.

HenrikHartiala (approved):
Nice work!

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/416#-:-NICGDww5Rh4HBhAkveu:bdtx8r6)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @eleksi)

PR #421, opened 2022-12-08
Implement route visibility toggle in line details page

eleksi:
Resolves https://github.com/HSLdevcom/jore4/issues/1039

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/421)

### DIFF
--- a/ui/src/components/common/LineDetailsButton.tsx
+++ b/ui/src/components/common/LineDetailsButton.tsx
@@ -10,18 +10,20 @@ const testIds = {
 interface Props {
   testId?: string;
   lineId: UUID;
+  routeLabel?: string;
   className?: string;
 }
 
 export const LineDetailsButton = ({
   testId,
   lineId,
+  routeLabel,
   className = '',
 }: Props): JSX.Element => {
   const history = useHistory();
   const onClick = () => {
     history.push({
-      pathname: routeDetails[Path.lineDetails].getLink(lineId),
+      pathname: routeDetails[Path.lineDetails].getLink(lineId, routeLabel),
     });
   };
   return (
--- a/ui/src/components/common/LineTimetablesButton.tsx
+++ b/ui/src/components/common/LineTimetablesButton.tsx
@@ -10,6 +10,7 @@ interface Props {
   testId?: string;
   disabled?: boolean;
   lineId: UUID;
+  routeLabel?: string;
   className?: string;
 }
 
@@ -17,13 +18,14 @@ export const LineTimetablesButton = ({
   testId,
   disabled,
   lineId,
+  routeLabel,
   className = '',
 }: Props): JSX.Element => {
   const disabledStyle = '!bg-background opacity-70 pointer-events-none';
   const history = useHistory();
   const onClick = () => {
     history.push({
-      pathname: routeDetails[Path.lineTimetables].getLink(lineId),
+      pathname: routeDetails[Path.lineTimetables].getLink(lineId, routeLabel),
     });
   };
   return (
--- a/ui/src/components/common/RouteLineTableRow.tsx
+++ b/ui/src/components/common/RouteLineTableRow.tsx
@@ -4,6 +4,7 @@ import {
   LineTableRowFragment,
   RouteTableRowFragment,
 } from '../../generated/graphql';
+import { isRoute } from '../../graphql';
 import { useAlertsAndHighLights } from '../../hooks';
 import { Column, Row, Visible } from '../../layoutComponents';
 import { Path, routeDetails } from '../../router/routeDetails';
@@ -17,13 +18,15 @@ export enum RouteLineTableRowVariant {
   RoutesAndLines,
 }
 
+type RowItem = LineTableRowFragment | RouteTableRowFragment;
+
 interface Props {
   className?: string;
   lineId: UUID;
   hasTimetables?: boolean;
   onLocatorButtonClick?: () => void;
   locatorButtonTestId: string;
-  rowItem: LineTableRowFragment | RouteTableRowFragment;
+  rowItem: RowItem;
   rowVariant: RouteLineTableRowVariant;
   isSelected?: boolean;
   onSelectChanged?: (event: React.ChangeEvent<HTMLInputElement>) => void;
@@ -35,8 +38,11 @@ const yBorderClassnames = 'border-y border-y-light-grey';
 const getDisplayInformation = (
   routeLineTableRowVariant: RouteLineTableRowVariant,
   lineId: UUID,
+  rowItem: RowItem,
   hasTimetables?: boolean,
 ) => {
+  const routeLabel = isRoute(rowItem) ? rowItem.label : undefined;
+
   switch (routeLineTableRowVariant) {
     case RouteLineTableRowVariant.Timetables:
       return {
@@ -47,8 +53,10 @@ const getDisplayInformation = (
             }`}
           />
         ),
-        alternativeRowActionButton: <LineDetailsButton lineId={lineId} />,
-        linkTo: routeDetails[Path.lineTimetables].getLink(lineId),
+        alternativeRowActionButton: (
+          <LineDetailsButton lineId={lineId} routeLabel={routeLabel} />
+        ),
+        linkTo: routeDetails[Path.lineTimetables].getLink(lineId, routeLabel),
         isDisabled: !hasTimetables,
       };
     case RouteLineTableRowVariant.RoutesAndLines:
@@ -56,9 +64,13 @@ const getDisplayInformation = (
       return {
         rowIcon: <i className="icon-bus-alt text-2xl text-tweaked-brand" />,
         alternativeRowActionButton: (
-          <LineTimetablesButton disabled={!hasTimetables} lineId={lineId} />
+          <LineTimetablesButton
+            disabled={!hasTimetables}
+            lineId={lineId}
+            routeLabel={routeLabel}
+          />
         ),
-        linkTo: routeDetails[Path.lineDetails].getLink(lineId),
+        linkTo: routeDetails[Path.lineDetails].getLink(lineId, routeLabel),
         isDisabled: false,
       };
   }
@@ -88,6 +100,7 @@ export const RouteLineTableRow = ({
   const displayInformation = getDisplayInformation(
     rowVariant,
     lineId,
+    rowItem,
     hasTimetables,
   );
 
--- a/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx
@@ -1,11 +1,14 @@
 import { DateTime } from 'luxon';
+import { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
+import { pipe, uniq } from 'remeda';
 import { LineAllFieldsFragment } from '../../../generated/graphql';
 import {
   useAppDispatch,
   useAppSelector,
   useGetLineDetails,
   useMapQueryParams,
+  useRouteLabelsQueryParam,
 } from '../../../hooks';
 import { Column, Container, Row, Visible } from '../../../layoutComponents';
 import {
@@ -29,6 +32,8 @@ export const LineDetailsPage = (): JSX.Element => {
 
   const dispatch = useAppDispatch();
   const { addMapOpenQueryParameter } = useMapQueryParams();
+  const { displayedRouteLabels, setDisplayedRoutesToUrl } =
+    useRouteLabelsQueryParam();
 
   const { line } = useGetLineDetails();
 
@@ -55,6 +60,26 @@ export const LineDetailsPage = (): JSX.Element => {
     ? () => createRoute(line)
     : undefined;
 
+  const uniqueLineRouteLabels = pipe(
+    line?.line_routes,
+    (routes) => routes?.map((route) => route.label) || [],
+    (routeLabels) => uniq(routeLabels),
+  );
+
+  const displayedRoutes =
+    line?.line_routes?.filter((route) =>
+      displayedRouteLabels?.includes(route.label),
+    ) || [];
+
+  // If no route has been initially selected to display, show all line's routes
+  // Set the default value to query params if route labels query param doesn't exist
+  useEffect(() => {
+    if (!displayedRouteLabels && uniqueLineRouteLabels.length !== 0) {
+      setDisplayedRoutesToUrl(uniqueLineRouteLabels);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [uniqueLineRouteLabels]);
+
   return (
     <div>
       <PageHeader className={getHeaderBorderClassName()}>
@@ -74,7 +99,7 @@ export const LineDetailsPage = (): JSX.Element => {
             <Column className="w-full">
               <h1 className="mt-8">{t('lines.routes')}</h1>
               {line.line_routes?.length > 0 ? (
-                <RouteStopsTable routes={line.line_routes} />
+                <RouteStopsTable routes={displayedRoutes} />
               ) : (
                 <CreateRouteBox onCreateRoute={onCreateRoute} />
               )}
--- a/ui/src/components/routes-and-lines/line-details/LineTitle.tsx
+++ b/ui/src/components/routes-and-lines/line-details/LineTitle.tsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { AiFillPlusCircle } from 'react-icons/ai';
+import { uniqBy } from 'remeda';
 import { RouteLine } from '../../../generated/graphql';
+import { useRouteLabelsQueryParam } from '../../../hooks';
 import { Column, Row } from '../../../layoutComponents';
 import { IconButton, SimpleButton } from '../../../uiComponents';
 import { LineValidityPeriod } from './LineValidityPeriod';
@@ -24,11 +26,10 @@ export const LineTitle: React.FC<Props> = ({
   onCreateRoute,
 }) => {
   const { t } = useTranslation();
+  const { toggleDisplayedRoute, displayedRouteLabels } =
+    useRouteLabelsQueryParam();
 
-  const onToggleRoute = () => {
-    // eslint-disable-next-line no-console
-    console.log('TODO');
-  };
+  const lineRoutes = uniqBy(line.line_routes, (route) => route.label);
 
   return (
     <Column>
@@ -37,13 +38,15 @@ export const LineTitle: React.FC<Props> = ({
           {t('lines.line', { label: line.label })}
         </h1>
         <span>
-          {line.line_routes?.length > 0 &&
-            line.line_routes.map((item) => (
+          {lineRoutes?.length > 0 &&
+            lineRoutes.map((item) => (
               <SimpleButton
                 key={item.route_id}
                 containerClassName="mr-2"
-                className="w-20 !rounded bg-tweaked-brand !px-0 !py-0 !text-sm !font-normal text-white"
-                onClick={onToggleRoute}
+                className="!mx-0 w-20 !rounded !px-0 !py-0 !text-sm !font-normal"
+                invertedClassName="!text-black"
+                onClick={() => toggleDisplayedRoute(item.label)}
+                inverted={!displayedRouteLabels?.includes(item.label)}
               >
                 {item.label}
               </SimpleButton>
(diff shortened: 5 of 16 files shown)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/421)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @eleksi)

___
*[`ui/src/components/common/search/RoutesList.tsx` line 33 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5ZDPm2lsN_4ZiIM4V:-NJ5ZDPn6umUbEMQMSoM:bxhvkov) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/components/common/search/RoutesList.tsx#L33)):*
> ```XML
>         <RouteTableRow
>           key={item.route_id}
>           linkTo={getRouteDetailsLink(item.on_line_id, item.label)}
> ```

why is the linkTo defined here and not within the RouteTableRow component? I don't see a reason to extract it outside of the component

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/421)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 70 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5__QQ7_7Kc7TJH_xm:-NJ5__QQ7_7Kc7TJH_xn:betkp9x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L70)):*
> ```XML
> 
>   const shownRoutes =
>     line?.line_routes?.filter((route) =>
> ```

instead of filtering it here, let's include this filtering within useGetLineDetails hook where we are already filtering by the observation date. There should not be any performance issues with rerunning the backend query with every route enable/disable click as it can reuse the data in the apollo cache
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 33 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5aRsc-9kYHKjbmTfM:-NJ5aRsc-9kYHKjbmTfN:buwu20z) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L33)):*
> ```XML
>     if (displayedRouteLabels.includes(routeLabel)) {
>       setDisplayedRoutesToUrl(
>         [...displayedRouteLabels].filter((label) => label !== routeLabel),
> ```

the filter function returns a new array instance, no need to clone the array like this: `[...displayedRouteLabels]`
___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 162 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5awhWA4_TDW563uQJ:-NJ5awhX1zgOVMbhSWjl:bz9miyz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/hooks/urlQuery/useUrlQuery.ts#L162)):*
> ```XML
>   /** Returns a query parameter in array type */
>   const getArrayFromUrlQuery = (paramName: string): string[] | undefined => {
>     return (queryParams[paramName] as string)?.split(',');
> ```

this is fine just makes me wonder why didn't we use `qs.parse` at the first place :)

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/421)* status: 0 of 11 files reviewed, 12 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 63 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJAAxpz9BN9KaHBXvWl:-NJAAn_c6lx_ISBizBez:bfz2h6s) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L63)):*
> ```XML
>     : undefined;
> 
>   const lineRouteLabels = pipe(
> ```

maybe uniqueLineRouteLabels would be a better name
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 69 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJAB2pv07udW-5aOE3-:-NJAB2pv07udW-5aOE30:b8vpprk) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L69)):*
> ```XML
>   );
> 
>   const shownRoutes =
> ```

the other variable is called displayedRouteLabels. So let's call this displayedRoutes as well
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 75 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJA9WUB3UNOeS8Jmsqb:-NJA9WUB3UNOeS8Jmsqc:b96y7du) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L75)):*
> ```XML
> 
>   // If no route has been initially selected to display, show all line's routes
>   useEffect(() => {
> ```

could you extend this comment that the aim here is actually to set a default value to the query if it has nothing? Also, should it be displayedRouteLabels === undefined instead of displayedRouteLabels.length === 0?
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 80 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJABHAvDRtY-SlujMFE:-NJABHAvDRtY-SlujMFF:bxxfbw2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L80)):*
> ```XML
>     }
>     // eslint-disable-next-line react-hooks/exhaustive-deps
>   }, [lineRouteLabels]);
> ```

shouldn't this also list displayedRouteLabels as a dependency?
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 101 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJAALZqAGToRmVqPTHr:-NJAALZqAGToRmVqPTHs:bkvcal1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L101)):*
> ```XML
>               <h1 className="mt-8">{t('lines.routes')}</h1>
>               {line.line_routes?.length > 0 ? (
>                 <RouteStopsTable routes={shownRoutes} />
> ```

what happens if the line has routes but the user somehow manages to toggle all of them off. Do we show the CreateRouteBox? Or just... nothing?
Or will the useEffect then automagically just display all the routes as displayedRouteLabels will be an empty array?
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 70 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5__QQ7_7Kc7TJH_xm:-NJA8_vD-wrr75DxMHX7:b-3w1rjo) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L70)):*

instead of filtering it here, let's include this filtering within useGetLineDetails hook where we are already filtering by the observation date. There should not be any performance issues with rerunning the backend query with every route enable/disable click as it can reuse the data in the apollo cache
</blockquote></details>

you had a good point in Slack that the LineTitle component still wants to know about all the routes, so we should not do the filtering within the useGetLineDetails.
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 2 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJABr398oRgD34lh-lC:-NJABr398oRgD34lh-lD:b-qsyi34) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L2)):*
> ```XML
> import { useMemo } from 'react';
> import { QueryParameterName } from './useMapQueryParams';
> ```

the query params hooks should not point at each other. Please move the QueryParameterName enum to the parent hook, useUrlQuery
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 8 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJACF1k3ueIjxKei65P:-NJACF1k3ueIjxKei65Q:b-2az1pb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L8)):*
> ```XML
>  * Query parameter hook for setting and getting displayed routes.
>  */
> export const useDisplayedRoutesQueryParam = () => {
> ```

should this rather be "useRouteLabelsQueryParam"? At least the query parameter name is "routeLabels". We don't have the "displayed" verb for other query param names either
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJACqxUEyCtgDMq8LIW:-NJACqxUEyCtgDMq8LIX:bkwgdiz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L27)):*
> ```XML
>   // Memoize the actual value to prevent unnecessary updates
>   const displayedRouteLabels = useMemo(() => {
>     return getArrayFromUrlQuery(QueryParameterName.RouteLabels) || [];
> ```

this should not default to empty array in my opinion. There's a difference between "undefined" (=filter not set) and "empty array" (=filter is set and no results are to be returned)

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/421)* status: 0 of 12 files reviewed, 12 unresolved discussions (waiting on @pallost)

___
*[`ui/src/components/common/search/RoutesList.tsx` line 33 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5ZDPm2lsN_4ZiIM4V:-NJ5f1yuAzyOwQohNOyH:b-v2dyqt) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/components/common/search/RoutesList.tsx#L33)):*

why is the linkTo defined here and not within the RouteTableRow component? I don't see a reason to extract it outside of the component
</blockquote></details>

Idk. But this part is under heavy changes in another PR by Henkka so I would rather not touch it here.
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 63 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJAAxpz9BN9KaHBXvWl:-NJAGS5aEAuOl4a5FUn1:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L63)):*

maybe uniqueLineRouteLabels would be a better name
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 69 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJAB2pv07udW-5aOE3-:-NJAGVyz1CoSjnSymQEC:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L69)):*

the other variable is called displayedRouteLabels. So let's call this displayedRoutes as well
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 75 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJA9WUB3UNOeS8Jmsqb:-NJAH0a65bWUhTEsG_2o:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L75)):*

could you extend this comment that the aim here is actually to set a default value to the query if it has nothing? Also, should it be displayedRouteLabels === undefined instead of displayedRouteLabels.length === 0?
</blockquote></details>

Done.
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 80 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJABHAvDRtY-SlujMFE:-NJAH1lZBKQP5CkDVTbC:b-ew5yx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L80)):*

shouldn't this also list displayedRouteLabels as a dependency?
</blockquote></details>

We only need to run this on component first render. But because `lineRouteLabels` could be still undefined at first render as line details have not been loaded, we need to have it in the dependency array so we set the default value only after we know what routes line has. We do not want to track `displayedRouteLabels` changes as this hook is just for default value. 
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 101 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJAALZqAGToRmVqPTHr:-NJAHpkE6IcFm0jeN2ah:bj027x1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L101)):*

what happens if the line has routes but the user somehow manages to toggle all of them off. Do we show the CreateRouteBox? Or just... nothing?
Or will the useEffect then automagically just display all the routes as displayedRouteLabels will be an empty array?
</blockquote></details>

I think if the user intentionally hides all the routes (which very well is possible by just toggling them off at the top of the page), we want to just show them nothing...? That is how it works right now and I think if the user really wants to see no routes, that is what they will see. Deselecting all routes is so explicit "I don't want to see anything" that I don't think we really need to handle this case any special way.
___
*[`ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx` line 70 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5__QQ7_7Kc7TJH_xm:-NJAGXUK-HcOxudDoyDg:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/components/routes-and-lines/line-details/LineDetailsPage.tsx#L70)):*

you had a good point in Slack that the LineTitle component still wants to know about all the routes, so we should not do the filtering within the useGetLineDetails.
</blockquote></details>

Done.
___
*[`ui/src/hooks/urlQuery/useUrlQuery.ts` line 162 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5awhWA4_TDW563uQJ:-NJ5gHna93FYB7u-imvs:bl43lls) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/hooks/urlQuery/useUrlQuery.ts#L162)):*

this is fine just makes me wonder why didn't we use `qs.parse` at the first place :)
</blockquote></details>

Yep, maybe we should have. 
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 2 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJABr398oRgD34lh-lC:-NJAIOXy5xBFwvq7B_Er:b-z972wi) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L2)):*

the query params hooks should not point at each other. Please move the QueryParameterName enum to the parent hook, useUrlQuery
</blockquote></details>

Good point! It definitely was in the wrong place.
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 8 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJACF1k3ueIjxKei65P:-NJAJWKj1vD65lZR9L8X:b-y931b5) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L8)):*

should this rather be "useRouteLabelsQueryParam"? At least the query parameter name is "routeLabels". We don't have the "displayed" verb for other query param names either
</blockquote></details>

Yep I guess this works better.
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJACqxUEyCtgDMq8LIW:-NJAJkXn7CM8q2Ds-vPF:b-xa0u9p) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b99f9aa9b1eae027a3eee24495cc0ce8602d067f/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L27)):*

this should not default to empty array in my opinion. There's a difference between "undefined" (=filter not set) and "empty array" (=filter is set and no results are to be returned)
</blockquote></details>

That is a good point. Undefined definitely has a meaning here.
___
*[`ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts` line 33 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJ5aRsc-9kYHKjbmTfM:-NJ9agoc4BP33vnJwGlO:birx69r) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/8b95e50c405535c9716bdca9c2a1c02b732358c6/ui/src/hooks/urlQuery/useDisplayedRoutesQueryParam.ts#L33)):*

the filter function returns a new array instance, no need to clone the array like this: `[...displayedRouteLabels]`
</blockquote></details>

True, oops :sweat_smile:

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/421)* status: 0 of 12 files reviewed, all discussions resolved

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/421)* status: 0 of 17 files reviewed, all discussions resolved (waiting on @pallost)

___
*[`ui/src/router/routeDetails.ts` line 25 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/421#-NJFSdZZ9jcf5VZRzPnj:-NJFSdZZ9jcf5VZRzPnk:b-b4nifv) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/88b48706b07f50720c18eea3ec15603f1af50e5c/ui/src/router/routeDetails.ts#L25)):*
> ```XML
> }
> 
> const getLineIdRouteLabelLink = (
> ```

This is not the most descriptive name, although not sure what would be good. One suggestion is to separate the path.replace(:id, lineId) and the searchQuery determination, and then the searchQueryDetermination. But I don't have a straight suggestion for this. 

Maybe this is OK as well, up to you.

PR #426, opened 2022-12-13
Replace journey pattern stops instead of journey pattern in route edit

eleksi:
Resolves HSLdevcom/jore4#1056

### DIFF
--- a/ui/src/components/map/routes/EditRouteMetadataLayer.tsx
+++ b/ui/src/components/map/routes/EditRouteMetadataLayer.tsx
@@ -10,7 +10,7 @@ import {
   resetRouteCreatingAction,
   selectEditedRouteData,
   selectMapRouteEditor,
-  setDraftRouteJourneyPatternStopsAction,
+  setDraftRouteJourneyPatternAction,
   setLineInfoAction,
   setRouteMetadataFormOpenAction,
 } from '../../../redux';
@@ -78,7 +78,9 @@ export const EditRouteMetadataLayer: React.FC = () => {
           })),
       );
 
-      dispatch(setDraftRouteJourneyPatternStopsAction(newJourneyPatternStops));
+      dispatch(
+        setDraftRouteJourneyPatternAction({ stops: newJourneyPatternStops }),
+      );
     }
   };
 
--- a/ui/src/components/map/routes/RouteEditor.tsx
+++ b/ui/src/components/map/routes/RouteEditor.tsx
@@ -24,7 +24,7 @@ import {
   selectEditedRouteData,
   selectMapRouteEditor,
   selectSelectedRouteId,
-  setDraftRouteJourneyPatternStopsAction,
+  setDraftRouteJourneyPatternAction,
   setLineInfoAction,
   setRouteMetadataAction,
   setSelectedRouteIdAction,
@@ -64,7 +64,7 @@ const RouteEditorComponent = (
     infraLinks,
     stopsEligibleForJourneyPattern,
     includedStopLabels,
-    journeyPatternStops,
+    journeyPattern,
     metaData: routeDetails,
   } = useAppSelector(selectEditedRouteData);
 
@@ -98,8 +98,11 @@ const RouteEditorComponent = (
       routeId,
       stopsEligibleForJourneyPattern,
       includedStopLabels,
-      journeyPatternStops,
+      journeyPatternStops: journeyPattern.stops,
       infraLinksAlongRoute: infraLinks || [],
+      // If editing route, journeyPattern id is defined
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      journeyPatternId: journeyPattern.id!,
     });
 
     const variables = mapEditGeometryChangesToVariables(changes);
@@ -116,7 +119,7 @@ const RouteEditorComponent = (
       form: routeDetails as RouteFormState,
       stopsEligibleForJourneyPattern,
       includedStopLabels,
-      journeyPatternStops,
+      journeyPatternStops: journeyPattern.stops,
       infraLinksAlongRoute: infraLinks || [],
     });
     if (changes.conflicts?.length) {
@@ -197,12 +200,14 @@ const RouteEditorComponent = (
         ),
       );
 
+      const editedRouteJourneyPattern =
+        routeDetailsResult.data.route_route_by_pk.route_journey_patterns[0];
+
       /**
        * Preserve journey pattern stop metadata (e.g. via info)
        */
       const newJourneyPatternStops = pipe(
-        routeDetailsResult.data.route_route_by_pk.route_journey_patterns[0]
-          .scheduled_stop_point_in_journey_patterns,
+        editedRouteJourneyPattern.scheduled_stop_point_in_journey_patterns,
         (stopsInJourneyPattern) =>
           stopsInJourneyPattern.map((stopInJourneyPattern) => ({
             ...stopInJourneyPattern,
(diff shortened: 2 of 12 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/426)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/map/routes/RouteEditor.tsx` line 103 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFGWQrDJGxY3bAEx1b:-NJFGWQrDJGxY3bAEx1c:bxrmgm0) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/components/map/routes/RouteEditor.tsx#L103)):*
> ```XML
>       journeyPatternStops: journeyPattern.stops,
>       infraLinksAlongRoute: infraLinks || [],
>       // If editing route, journeyPattern id is defined
> ```

How about: "journeyPattern id is defined when editing route". 

"If editing route" is confusing, it raises the question: "what if we are NOT editing route". Anyway, name of this function seems to be "editRoute" so no need to think about if-else.
___
*[`ui/src/components/map/routes/RouteEditor.tsx` line 203 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFGuUb8h_ETbnV5QvI:-NJFGuUb8h_ETbnV5QvJ:b-m8u1ec) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/components/map/routes/RouteEditor.tsx#L203)):*
> ```XML
>       );
> 
>       const editedRouteJourneyPattern =
> ```

Please comment why it is safe to expect that array has one element. I'm also wondering what happens if there is more than one element, should we handle that case too?
___
*[`ui/src/graphql/client.ts` line 144 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFHDJ3-vYZxk7Ccpnd:-NJFHDJ3-vYZxk7Ccpne:b-naibml) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/graphql/client.ts#L144)):*
> ```XML
>           scheduled_stop_point_in_journey_patterns: {
>             merge(existing, incoming) {
>               // Just return incoming contents without trying to do any merging
> ```

So is this "good solution" or are we just silencing console warning for now? So should there be "TODO: " prefix and explanation about what we should do in the future? Same for the next merge statement in this PR.
___
*[`ui/src/hooks/routes/useEditRouteJourneyPattern.ts` line 119 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFHyJD15GMPknDhRhK:-NJFHyJD15GMPknDhRhL:brw0a94) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/hooks/routes/useEditRouteJourneyPattern.ts#L119)):*
> ```XML
> 
>     const changes: UpdateJourneyPatternChanges = {
>       journeyPatternId: route.route_journey_patterns[0].journey_pattern_id,
> ```

Why it is safe to assume that array has at least one element? What if there are multiple elements?

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/426)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/map/routes/RouteEditor.tsx` line 103 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFGWQrDJGxY3bAEx1b:-NJFcqsg4zr2lQVRZXAX:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/components/map/routes/RouteEditor.tsx#L103)):*

How about: "journeyPattern id is defined when editing route". 

"If editing route" is confusing, it raises the question: "what if we are NOT editing route". Anyway, name of this function seems to be "editRoute" so no need to think about if-else.

</blockquote></details>

Done.
___
*[`ui/src/components/map/routes/RouteEditor.tsx` line 203 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFGuUb8h_ETbnV5QvI:-NJFd2YM8-wrPTW1Z3OF:b-qvr20n) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/components/map/routes/RouteEditor.tsx#L203)):*

Please comment why it is safe to expect that array has one element. I'm also wondering what happens if there is more than one element, should we handle that case too?
</blockquote></details>

We have built everything on the assumption that route <-> journeyPattern is 1:1-relation.
___
*[`ui/src/graphql/client.ts` line 144 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFHDJ3-vYZxk7Ccpnd:-NJFdRLhDZ9H16LoEpWA:byolw46) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/graphql/client.ts#L144)):*

So is this "good solution" or are we just silencing console warning for now? So should there be "TODO: " prefix and explanation about what we should do in the future? Same for the next merge statement in this PR.
</blockquote></details>

This is the default functionality, but somehow silencing the warning can be done by just explicitly defining the function.. :thinking: 

I added more comment to clarify this.
___
*[`ui/src/hooks/routes/useEditRouteJourneyPattern.ts` line 119 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFHyJD15GMPknDhRhK:-NJFeNnN9-heweueB0nA:ba7qi0w) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/hooks/routes/useEditRouteJourneyPattern.ts#L119)):*

Why it is safe to assume that array has at least one element? What if there are multiple elements?
</blockquote></details>

As mentioned before, there is only one journey pattern right now for each route. If there were multiple journey patterns, we would have to change tons of different places in the code. Also we don't even know how these should function with multiple journey patterns so I don't think we need to prepare for unlikely future situation that we don't know anything about.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/426)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/map/routes/RouteEditor.tsx` line 203 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFGuUb8h_ETbnV5QvI:-NJKErSs86or38P-Siqr:buiknqc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/components/map/routes/RouteEditor.tsx#L203)):*

We have built everything on the assumption that route <-> journeyPattern is 1:1-relation.
</blockquote></details>

Well, could then have said that "In our data route has always exactly one journey pattern" for extra clarity :wink: 
___
*[`ui/src/graphql/client.ts` line 144 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/426#-NJFHDJ3-vYZxk7Ccpnd:-NJKEjQh4H1JKsagVZa2:b-jm5frf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/73cb154f05a633c51e5347a6f72649c589e972fc/ui/src/graphql/client.ts#L144)):*

This is the default functionality, but somehow silencing the warning can be done by just explicitly defining the function.. :thinking: 

I added more comment to clarify this.
</blockquote></details>

Thanks, much clearer now. :+1:

PR #441, opened 2023-01-04
Use Hastus microservice to export routes

eleksi:
- Note: Hastus import/export microservice is not yet in docker-compose setup. It has to be started separately from its repository for this to work until in is added to docker-compose setup.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/441)

### DIFF
--- a/ui/server.js
+++ b/ui/server.js
@@ -40,6 +40,13 @@ const devProxy = {
     },
     target: 'http://localhost:3005',
   },
+  '/api/hastus': {
+    // Proxy hastus importer / exporter requests to hastus import / export microservice
+    pathRewrite: {
+      '^/api/hastus': '', // remove path.
+    },
+    target: 'http://localhost:3008',
+  },
 };
 
 const port = parseInt(process.env.PORT, 10) || 3300;
--- a/ui/src/api/hastus.ts
+++ b/ui/src/api/hastus.ts
@@ -1,0 +1,45 @@
+import axios from 'axios';
+import { DateTime } from 'luxon';
+import { Priority } from '../types/Priority';
+
+interface CommonExportParams {
+  readonly labels: string[];
+  readonly priority: Priority;
+}
+
+interface ExportBody extends CommonExportParams {
+  readonly observationDate: string;
+}
+
+interface ExportParams extends CommonExportParams {
+  readonly observationDate: DateTime;
+}
+
+const apiClient = axios.create({
+  baseURL: '/api/hastus',
+});
+
+const exportRoutes = (payload: ExportBody) =>
+  apiClient.post('/export/routes', payload, {
+    responseType: 'blob',
+    headers: {
+      // TODO: Authenticate properly
+      'x-hasura-admin-secret': 'hasura',
+      'x-hasura-role': 'admin',
+    },
+  });
+
+export const exportRoutesToHastus = async ({
+  labels,
+  priority,
+  observationDate,
+}: ExportParams) => {
+  const request: ExportBody = {
+    labels,
+    priority,
+    observationDate: observationDate.toISODate(),
+  };
+  const response = await exportRoutes(request);
+
+  return response;
+};
--- a/ui/src/components/routes-and-lines/line-details/LineTitle.tsx
+++ b/ui/src/components/routes-and-lines/line-details/LineTitle.tsx
@@ -5,7 +5,7 @@ import { uniqBy } from 'remeda';
 import { RouteLine } from '../../../generated/graphql';
 import { useRouteLabelsQueryParam } from '../../../hooks';
 import { Column, Row } from '../../../layoutComponents';
-import { IconButton, SimpleButton } from '../../../uiComponents';
+import { IconButton, SimpleSmallButton } from '../../../uiComponents';
 import { LineValidityPeriod } from './LineValidityPeriod';
 
 const testIds = {
@@ -37,25 +37,21 @@ export const LineTitle: React.FC<Props> = ({
         <h1 className="mr-4" data-testid={testIds.heading}>
           {t('lines.line', { label: line.label })}
         </h1>
-        <span>
+        <span className="space-x-2">
           {lineRoutes?.length > 0 &&
             lineRoutes.map((item) => (
-              <SimpleButton
+              <SimpleSmallButton
                 key={item.route_id}
-                containerClassName="mr-2"
-                className="!mx-0 w-20 !rounded !px-0 !py-0 !text-sm !font-normal"
-                invertedClassName="!text-black"
                 onClick={() => toggleDisplayedRoute(item.label)}
                 inverted={!displayedRouteLabels?.includes(item.label)}
-              >
-                {item.label}
-              </SimpleButton>
+                label={item.label}
+              />
             ))}
         </span>
         {onCreateRoute && (
           <IconButton
             testId={testIds.createRouteButton}
-            icon={<AiFillPlusCircle className="text-3xl text-brand" />}
+            icon={<AiFillPlusCircle className="ml-2 text-3xl text-brand" />}
             onClick={onCreateRoute}
           />
         )}
--- a/ui/src/components/routes-and-lines/search/ExportToolbar.tsx
+++ b/ui/src/components/routes-and-lines/search/ExportToolbar.tsx
@@ -5,6 +5,7 @@ import { pipe } from 'remeda';
 import {
   useAppDispatch,
   useAppSelector,
+  useExportRoutes,
   useSearchResults,
 } from '../../../hooks';
 import { Row, Visible } from '../../../layoutComponents';
@@ -24,6 +25,7 @@ export const ExportToolbar = (): JSX.Element => {
   const { resultCount, resultType, routes, lines } = useSearchResults();
   const { isSelectingRoutesForExport, selectedRouteLabels } =
     useAppSelector(selectExport);
+  const { canExport, exportRoutesToHastus } = useExportRoutes();
 
   const searchResultRouteLabels = pipe(
     resultType === DisplayedSearchResultType.Routes
@@ -51,11 +53,17 @@ export const ExportToolbar = (): JSX.Element => {
     dispatch(resetSelectedRoutesAction());
   };
 
-  // If no route is selected, export all routes by default
-  const isExportingAll = selectedRouteLabels.length === 0;
+  // If no route is selected, export all routes in search results by default
+  const isExportingAllSearchResults = selectedRouteLabels.length === 0;
 
-  // TODO: Implement exporting
-  const exportRoutes = () => null;
+  const exportRoutes = () => {
+    exportRoutesToHastus(
+      // If we are "exporting all", export all routes in search results
+      isExportingAllSearchResults
+        ? searchResultRouteLabels
+        : selectedRouteLabels,
+    );
+  };
 
   return (
     <Row className="items-center gap-3">
@@ -85,9 +93,12 @@ export const ExportToolbar = (): JSX.Element => {
       <Visible visible={isSelectingRoutesForExport}>
         <SimpleSmallButton
           inverted
+          disabled={!canExport}
           onClick={exportRoutes}
           label={t(
-            isExportingAll ? 'export.exportAll' : 'export.exportSelected',
+            isExportingAllSearchResults
+              ? 'export.exportAll'
+              : 'export.exportSelected',
           )}
           className="w-auto rounded-full px-4"
         />
--- a/ui/src/components/routes-and-lines/search/ExportToolbar.tsx
+++ b/ui/src/components/routes-and-lines/search/ExportToolbar.tsx
@@ -88,7 +88,7 @@ export const ExportToolbar = (): JSX.Element => {
             ? 'export.quitSelecting'
             : 'export.startSelecting',
         )}
-        className="w-auto rounded-full px-4"
+        className="!rounded-full"
       />
       <Visible visible={isSelectingRoutesForExport}>
         <SimpleSmallButton
@@ -100,7 +100,7 @@ export const ExportToolbar = (): JSX.Element => {
               ? 'export.exportAll'
               : 'export.exportSelected',
           )}
-          className="w-auto rounded-full px-4"
+          className="!rounded-full"
         />
       </Visible>
     </Row>
--- a/ui/src/hooks/routes/index.ts
+++ b/ui/src/hooks/routes/index.ts
@@ -3,6 +3,7 @@ export * from './useDeleteRoute';
 export * from './useEditRouteGeometry';
 export * from './useEditRouteJourneyPattern';
 export * from './useEditRouteMetadata';
+export * from './useExportRoutes';
 export * from './useExtractRouteFromFeature';
 export * from './useGetDisplayedRoutes';
 export * from './useRouteInfo';
--- a/ui/src/hooks/routes/useExportRoutes.ts
+++ b/ui/src/hooks/routes/useExportRoutes.ts
@@ -1,0 +1,35 @@
+import { exportRoutesToHastus as exportToHastus } from '../../api/hastus';
+import { downloadFile } from '../../utils';
+import { useSearchQueryParser } from '../search';
+import { useObservationDateQueryParam } from '../urlQuery';
+
+export const useExportRoutes = () => {
+  const { observationDate } = useObservationDateQueryParam();
+  const { search } = useSearchQueryParser();
+
+  const { priorities } = search;
+
+  // Routes can be exported to Hastus only when there is only 1 priority selected
+  const canExport = priorities?.length === 1;
+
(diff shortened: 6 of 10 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/441)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/api/hastus.ts` line 25 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLK_UU11aASVkMf75JO:-NLK_UU11aASVkMf75JP:b-embjgd) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/api/hastus.ts#L25)):*
> ```XML
>   });
> 
> export const exportToHastus = async (
> ```

"exportRoutesToHastus" for clarity. (You can easily rename this when importing if you want to define "exportRoutesToHastus" name again: `import { exportRoutesToHastus:  exportToHastus } from '../../api/hastus';`
___
*[`ui/src/api/hastus.ts` line 28 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKXsPv1zSYfNzHBcTs:-NLKXsPv1zSYfNzHBcTt:b-er6hpn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/api/hastus.ts#L28)):*
> ```XML
>   labels: string[],
>   priority: Priority,
>   observationDate: DateTime,
> ```

So, we have `ExportBody` interface defined, which is used for one purpose, and then we have this very similar inline definition, which has 2/3 same fields and is defined inline? 
This could be fixed by multiple ways, e.g.
- creating common interface and extending it where needed
- changing this api to take `observationDate` as ISO date (and also renaming it to `observationISODate` for clarity)



_Code quote:_
```XML
  labels: string[],
  priority: Priority,
  observationDate: DateTime,
```
___
*[`ui/src/components/routes-and-lines/search/ExportToolbar.tsx` line 61 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKaNUl3auqv_k_5jsF:-NLKaNUl3auqv_k_5jsG:b-i4oix3) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/components/routes-and-lines/search/ExportToolbar.tsx#L61)):*
> ```XML
>   const exportRoutes = () => {
>     exportRoutesToHastus(
>       isExportingAll ? searchResultRouteLabels : selectedRouteLabels,
> ```

This looks suspicious because neither `searchResultRouteLabels` nor `selectedRouteLabels` sounds like they include "all" routes. What is happening here?
___
*[`ui/src/hooks/routes/useExportRoutes.ts` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKZs1y9eWukoGeVPem:-NLKZs1y9eWukoGeVPen:byrvd93) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/hooks/routes/useExportRoutes.ts#L15)):*
> ```XML
>   const canExport = priorities?.length === 1;
> 
>   const exportRoutesToHastus = async (routeLabels: string[]) => {
> ```

You could make this less error-prone and easier to debug/maintain by this:
```
if (!canExport ) {
throw new Error(`Can't export routes, expected exactly 1 priority but got "${priorities}"`)
}
```
___
*[`ui/src/uiComponents/SimpleSmallButton.tsx` line 21 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKad4c4bOGUmsPF57S:-NLKad4c4bOGUmsPF57T:b-qpazcm) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/uiComponents/SimpleSmallButton.tsx#L21)):*
> ```XML
>     ? 'border-grey bg-white text-gray-900 hover:border-brand active:border-brand'
>     : 'border-brand bg-brand text-white hover:bg-opacity-50 active:bg-opacity-50';
>   const disabledClassNames = disabled ? 'pointer-events-none opacity-70' : '';
> ```

Do we really need this pointer-events-none className? Usually browsers should handle this kind of things with buttons automatically and if they won't do that, then we have accessibility error.

Btw, disabled button should most likely have `cursor: not allowed` cursor and NOT "pointer-events-none"

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/441)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/api/hastus.ts` line 25 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLK_UU11aASVkMf75JO:-NLKoyN11P46l4sCS0Cf:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/api/hastus.ts#L25)):*

"exportRoutesToHastus" for clarity. (You can easily rename this when importing if you want to define "exportRoutesToHastus" name again: `import { exportRoutesToHastus:  exportToHastus } from '../../api/hastus';`
</blockquote></details>

Done.
___
*[`ui/src/api/hastus.ts` line 28 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKXsPv1zSYfNzHBcTs:-NLKpdjb-ohdG55m9FFI:b-uum4p8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/api/hastus.ts#L28)):*

So, we have `ExportBody` interface defined, which is used for one purpose, and then we have this very similar inline definition, which has 2/3 same fields and is defined inline? 
This could be fixed by multiple ways, e.g.
- creating common interface and extending it where needed
- changing this api to take `observationDate` as ISO date (and also renaming it to `observationISODate` for clarity)

</blockquote></details>

Well the second one was not inline interface definition but just separate function params (not object). But I have changed it now that both the functions take an object as a parameter and created a common interface which is extended for both purposes. I cannot change the Hastus API so the ISO format Date is still called `observationDate`.
___
*[`ui/src/components/routes-and-lines/search/ExportToolbar.tsx` line 61 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKaNUl3auqv_k_5jsF:-NLKppnO8MAUFG6fl6b3:bleb854) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/components/routes-and-lines/search/ExportToolbar.tsx#L61)):*

This looks suspicious because neither `searchResultRouteLabels` nor `selectedRouteLabels` sounds like they include "all" routes. What is happening here?
</blockquote></details>

Well by "all" we mean all search result routes, without having to select any. I renamed the variable and added a comment for clarity.
___
*[`ui/src/hooks/routes/useExportRoutes.ts` line 15 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKZs1y9eWukoGeVPem:-NLKqKPY11SeGioQG94x:b498c3v) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/hooks/routes/useExportRoutes.ts#L15)):*

You could make this less error-prone and easier to debug/maintain by this:
```
if (!canExport ) {
throw new Error(`Can't export routes, expected exactly 1 priority but got "${priorities}"`)
}
```
</blockquote></details>

This function should not be called if there are multiple priorities selected, as the button is disabled in that case. But maybe there will be bug somewhere or this function is called in some other context, so this error could be thrown. So I added it.
___
*[`ui/src/uiComponents/SimpleSmallButton.tsx` line 21 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKad4c4bOGUmsPF57S:-NLKrKVc6PPXsV8-Np_f:bpj1q41) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/uiComponents/SimpleSmallButton.tsx#L21)):*

Do we really need this pointer-events-none className? Usually browsers should handle this kind of things with buttons automatically and if they won't do that, then we have accessibility error.

Btw, disabled button should most likely have `cursor: not allowed` cursor and NOT "pointer-events-none"
</blockquote></details>

This is just implemented the same way SimpleButton is. The button is disabled even without the styles, but we also want to make it look disabled.

The reason we have `pointer-events-none` here is because we want to disable e.g. hover effects for this button when it is disabled. If we pass only `cursor: not allowed`, we would still see button corners turn blue when hovering over it. Is we pass both classes, `cursor: not allowed` does nothing as pointer effects are disabled.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/441)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/api/hastus.ts` line 28 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKXsPv1zSYfNzHBcTs:-NLL9g-u4N2rsdbvoxbi:bpwasox) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/api/hastus.ts#L28)):*

Well the second one was not inline interface definition but just separate function params (not object). But I have changed it now that both the functions take an object as a parameter and created a common interface which is extended for both purposes. I cannot change the Hastus API so the ISO format Date is still called `observationDate`.
</blockquote></details>

What do you mean "cannot change hastus api"? You don't have to, you can use whatever name in our own code and then rename it just in time before sending it to api.
```
{
 observationDate: observationISODate.
}
```

But OK, as long as `ExportParams` is used just in this file, there is not much room for confusion (at least so far) so this is good enough for now.
___
*[`ui/src/uiComponents/SimpleSmallButton.tsx` line 21 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKad4c4bOGUmsPF57S:-NLLAMYQ6WOx7i4xPDHK:bqszqzq) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/uiComponents/SimpleSmallButton.tsx#L21)):*

This is just implemented the same way SimpleButton is. The button is disabled even without the styles, but we also want to make it look disabled.

The reason we have `pointer-events-none` here is because we want to disable e.g. hover effects for this button when it is disabled. If we pass only `cursor: not allowed`, we would still see button corners turn blue when hovering over it. Is we pass both classes, `cursor: not allowed` does nothing as pointer effects are disabled.
</blockquote></details>

Tried to google this a bit, and seems like `cursor: not-allowed !important` might do the trick: https://stackoverflow.com/questions/50349017/how-can-i-change-cursor-for-disabled-button-or-a-in-bootstrap-4
Anyway, proper solution could actually be then extracting hover styles to own function:

```
const getHoverStyles = (disabled, inverted) => {
if (disabled) {
 return ''
}
return inverted ? ...;

const colorClassNames = `... ${getHoverStyles(disabled, inverted)}`;
```

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/441)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/api/hastus.ts` line 28 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKXsPv1zSYfNzHBcTs:-NLLFZNFCBfrHpkOG37Q:b-e4s2lb) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/api/hastus.ts#L28)):*

What do you mean "cannot change hastus api"? You don't have to, you can use whatever name in our own code and then rename it just in time before sending it to api.
```
{
 observationDate: observationISODate.
}
```

But OK, as long as `ExportParams` is used just in this file, there is not much room for confusion (at least so far) so this is good enough for now.
</blockquote></details>

I know we can use whatever name we want, but `ExportBody` is representing the Hastus API endpoint body schema, and `exportRoutes` was accepting the request body, aka. `ExportBody` as only parameter. So if we want to use `observationISODate` field, we cannot use it in this particular interface.

Of course we could change the function responsibilities so that exportRoutes takes something else than the request body as parameter and then renames the field from `observationISODate` to `observationDate`, but renaming / type converting is what `exportRoutesToHastus` is responsible for right now. I think it is nice to keep the actual HTTP call function as simple as possible so it does not have to rename object fields / convert types etc. before calling, even though in this case it probably wouldn't be too bad idea to combine `exportRoutes` and `exportRoutesToHastus` into one function either.

Anyway, good that this will suffice for now.
___
*[`ui/src/uiComponents/SimpleSmallButton.tsx` line 21 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKad4c4bOGUmsPF57S:-NLLUR1S9VpDcH2eHpFb:b2xa34l) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/uiComponents/SimpleSmallButton.tsx#L21)):*

Tried to google this a bit, and seems like `cursor: not-allowed !important` might do the trick: https://stackoverflow.com/questions/50349017/how-can-i-change-cursor-for-disabled-button-or-a-in-bootstrap-4
Anyway, proper solution could actually be then extracting hover styles to own function:

```
const getHoverStyles = (disabled, inverted) => {
if (disabled) {
 return ''
}
return inverted ? ...;

const colorClassNames = `... ${getHoverStyles(disabled, inverted)}`;
```
</blockquote></details>

Yes. I noticed also that there is quite a lot of code duplication and string concatenation in `SimpleButton` and `SimpleSmallButton` when it comes to classNames, and it has come quite hard to follow how `classNames` is constructed and which style is applied and when, especially in `SimpleButton`. 

So, I tried to refactor button classname generation a little bit and make it reusable. Please note that adding separate hover styles function as you suggested above would have made the original code even more lengthy, which is not visible straight away from the diffs.

I really can't decide whether or not this change takes it to a good or bad direction, there are pros and cons. What do you think?

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/441)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @eleksi)

___
*[`ui/src/uiComponents/SimpleButton.tsx` line 28 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLQ6HBoCXtNt5tfdVEe:-NLQ6HBoCXtNt5tfdVEf:b7a90py) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/b13a33ec4937d9d474d3bbca1af714a8e6ff42dd/ui/src/uiComponents/SimpleButton.tsx#L28)):*
> ```XML
> 
> export const commonHoverStyle = 'hover:border-2 hover:border-tweaked-brand';
> const hoverStyle = `${commonHoverStyle} m-px hover:m-0`;
> ```

Could this be defined inside `getHoverStyles`? Its a bit confusing that we have both "hoverStyle" constant and also "getHoverStyles" function.
___
*[`ui/src/uiComponents/SimpleSmallButton.tsx` line 21 at r2](https://reviewable.io/reviews/hsldevcom/jore4-ui/441#-NLKad4c4bOGUmsPF57S:-NLQ6Win0VCs7lGRQ4iZ:b-2jdvxz) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/55fa2516cbf1371a25812b5af853cf2b7a80a74b/ui/src/uiComponents/SimpleSmallButton.tsx#L21)):*

Yes. I noticed also that there is quite a lot of code duplication and string concatenation in `SimpleButton` and `SimpleSmallButton` when it comes to classNames, and it has come quite hard to follow how `classNames` is constructed and which style is applied and when, especially in `SimpleButton`. 

So, I tried to refactor button classname generation a little bit and make it reusable. Please note that adding separate hover styles function as you suggested above would have made the original code even more lengthy, which is not visible straight away from the diffs.

I really can't decide whether or not this change takes it to a good or bad direction, there are pros and cons. What do you think?
</blockquote></details>

After couple of iteration rounds this starts to look OK. :+1:

eleksi:
e2e-tests that fail in CI work when run locally. They fail in CI apparently because of map-matching -response times out, which does not happen when test are run locally. Therefore, this PR doesn't seem to break any tests, the CI pipe is just broken, thus this PR can be merged.

PR #431, opened 2022-12-20
Add test that changes a route's shape

sr-vala:
Resolves HSLdevcom/jore4#760

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/431)

PR #442, opened 2023-01-04
Add test that creates a route from a template

sr-vala:
Resolves HSLdevcom/jore4#995

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)

### DIFF
--- a/cypress/e2e/createRoute.cy.ts
+++ b/cypress/e2e/createRoute.cy.ts
@@ -1,18 +1,22 @@
 import {
   buildLine,
+  buildRoute,
   buildStop,
+  buildStopsInJourneyPattern,
+  InfraLinkAlongRouteInsertInput,
+  JourneyPatternInsertInput,
   LineInsertInput,
   Priority,
   ReusableComponentsVehicleModeEnum,
   ReusableComponentsVehicleSubmodeEnum,
   RouteDirectionEnum,
+  RouteInsertInput,
   StopInsertInput,
   VehicleSubmodeOnInfraLinkInsertInput,
 } from '@hsl/jore4-test-db-manager';
 import { DateTime } from 'luxon';
-import { Map, RouteEditor } from '../pageObjects';
+import { ModalMap, RouteEditor } from '../pageObjects';
 import { FilterPanel } from '../pageObjects/FilterPanel';
-import { ModalMap } from '../pageObjects/ModalMap';
 import { RouteStopsOverlay } from '../pageObjects/RouteStopsOverlay';
 import { insertToDbHelper, removeFromDbHelper } from '../utils';
 import { deleteRoutesByLabel } from './utils';
@@ -22,26 +26,27 @@ const testRouteLabels = {
   label2: 'T-reitti 2',
   label3: 'T-reitti 3',
   label4: 'Indefinite end time route',
+  label5: 'Template route',
 };
 
 // These infralink IDs exist in the 'infraLinks.sql' test data file.
 // These form a straight line on Eerikinkatu in Helsinki.
 // Coordinates are partial since they are needed only for the stop creation.
 
-const testInfraLinks = {
-  1: {
+const testInfraLinks = [
+  {
     id: '73bc2df9-f5af-4c38-a1dd-5ed1f71c90a8',
-    coordinates: [24.92492146851626, 60.1634759878872, 0],
+    coordinates: [24.926699622176628, 60.164181083308065, 10.0969999999943],
   },
-  2: {
+  {
     id: 'ea69415a-9c54-4327-8836-f38b36d8fa99',
     coordinates: [24.92904198486008, 60.16490775039894, 0],
   },
-  3: {
+  {
     id: '13de61c2-3fc9-4255-955f-0a2350c389e1',
     coordinates: [24.932072417514647, 60.166003223527824, 0],
   },
-};
+];
 
 const lines: LineInsertInput[] = [
   {
@@ -66,10 +71,22 @@ const stops: StopInsertInput[] = [
   {
     ...buildStop({
       label: 'Test stop 1',
-      located_on_infrastructure_link_id: testInfraLinks[1].id,
+      located_on_infrastructure_link_id: testInfraLinks[0].id,
     }),
     validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
     scheduled_stop_point_id: '3f23a4c5-f527-4395-bd9f-bbc398f837df',
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[0].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: 'Test stop 2',
+      located_on_infrastructure_link_id: testInfraLinks[1].id,
+    }),
+    validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
+    scheduled_stop_point_id: '431a6791-f1f5-45d4-8c9d-9e154a2531e0',
     measured_location: {
       type: 'Point',
       coordinates: testInfraLinks[1].coordinates,
@@ -77,32 +94,55 @@ const stops: StopInsertInput[] = [
   },
   {
     ...buildStop({
-      label: 'Test stop 2',
-      located_on_infrastructure_link_id: testInfraLinks[2].id,
-    }),
-    validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
-    scheduled_stop_point_id: '431a6791-f1f5-45d4-8c9d-9e154a2531e0',
-    measured_location: {
-      type: 'Point',
-      coordinates: testInfraLinks[2].coordinates,
-    },
-  },
-  {
-    ...buildStop({
       label: 'Test stop 3',
-      located_on_infrastructure_link_id: testInfraLinks[3].id,
+      located_on_infrastructure_link_id: testInfraLinks[2].id,
     }),
     validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
     scheduled_stop_point_id: '6c09b8d9-5952-4ee3-92b9-a7b4847517d3',
     measured_location: {
       type: 'Point',
-      coordinates: testInfraLinks[3].coordinates,
+      coordinates: testInfraLinks[2].coordinates,
     },
+  },
+];
+
+const routes: RouteInsertInput[] = [
+  {
+    ...buildRoute({ label: testRouteLabels.label1 }),
+    route_id: '7961d12f-26cc-4e0f-b6a7-845bc334df63',
+    on_line_id: lines[0].line_id,
+    validity_start: DateTime.fromISO('2022-08-11T13:08:43.315+03:00'),
+    validity_end: DateTime.fromISO('2032-08-11T13:08:43.315+03:00'),
+  },
+];
+
+const infraLinksAlongRoute: InfraLinkAlongRouteInsertInput[] = [
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: testInfraLinks[0].id,
+    infrastructure_link_sequence: 0,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: testInfraLinks[1].id,
+    infrastructure_link_sequence: 1,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: testInfraLinks[2].id,
+    infrastructure_link_sequence: 2,
+    is_traversal_forwards: true,
   },
 ];
 
 const vehicleSubmodeOnInfrastructureLink: VehicleSubmodeOnInfraLinkInsertInput[] =
   [
+    {
+      infrastructure_link_id: testInfraLinks[0].id,
+      vehicle_submode: ReusableComponentsVehicleSubmodeEnum.GenericBus,
+    },
     {
       infrastructure_link_id: testInfraLinks[1].id,
       vehicle_submode: ReusableComponentsVehicleSubmodeEnum.GenericBus,
@@ -111,16 +151,29 @@ const vehicleSubmodeOnInfrastructureLink: VehicleSubmodeOnInfraLinkInsertInput[]
       infrastructure_link_id: testInfraLinks[2].id,
       vehicle_submode: ReusableComponentsVehicleSubmodeEnum.GenericBus,
     },
-    {
-      infrastructure_link_id: testInfraLinks[3].id,
-      vehicle_submode: ReusableComponentsVehicleSubmodeEnum.GenericBus,
-    },
   ];
+
+const journeyPatterns: JourneyPatternInsertInput[] = [
+  {
+    journey_pattern_id: '6cae356b-20f4-4e04-a969-097999b351f0',
+    on_route_id: routes[0].route_id,
+  },
+];
+
+const stopsInJourneyPattern = buildStopsInJourneyPattern(
+  stops.map((item) => item.label),
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  journeyPatterns[0].journey_pattern_id!,
+);
 
 const dbResources = {
   vehicleSubmodeOnInfrastructureLink,
   lines,
   stops,
+  routes,
+  infraLinksAlongRoute,
+  journeyPatterns,
+  stopsInJourneyPattern,
 };
 
 const stopTestIds = {
@@ -136,7 +189,6 @@ const clearDatabase = () => {
 
 describe('Route creation', () => {
   let modalMap: ModalMap;
-  let map: Map;
   let routeStopsOverlay: RouteStopsOverlay;
   let routeEditor: RouteEditor;
 
@@ -153,7 +205,6 @@ describe('Route creation', () => {
     clearDatabase();
     insertToDbHelper(dbResources);
 
-    map = new Map();
     routeStopsOverlay = new RouteStopsOverlay();
     routeEditor = new RouteEditor();
     cy.setupTests();
@@ -162,7 +213,7 @@ describe('Route creation', () => {
     // Location where all test stops and routes are visible.
     const mapLocation = { lng: 24.929689228090112, lat: 60.16495016651525 };
 
-    map.visit({
+    modalMap.map.visit({
       zoom: 15,
       lat: mapLocation.lat,
       lng: mapLocation.lng,
@@ -170,7 +221,7 @@ describe('Route creation', () => {
 
     mapFilterPanel.toggleShowStops(ReusableComponentsVehicleModeEnum.Bus);
 
-    map.waitForLoadToComplete();
+    modalMap.map.waitForLoadToComplete();
   });
 
   afterEach(() => {
@@ -321,6 +372,44 @@ describe('Route creation', () => {
       routeEditor.checkRouteSubmitSuccessToast();
 
       routeStopsOverlay.routeShouldBeSelected(routeName);
+    },
+  );
+
+  it(
+    'Should create a new route using an existing route as a template',
+    { scrollBehavior: 'bottom' },
+    () => {
+      modalMap.createRoute({
+        routeFormInfo: {
+          finnishName: 'Reitin pohjalta luotu reitti',
+          label: testRouteLabels.label5,
+          direction: RouteDirectionEnum.Outbound,
+          line: String(lines[0].label),
+          templateRoute: {
+            templateRouteSelectorInfo: {
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              priority: routes[0].priority!,
+              label: routes[0].label,
+            },
+          },
+          validityStartISODate: String(routes[0].validity_start),
+          validityEndISODate: String(routes[0].validity_end),
+          priority: Priority.Standard,
+        },
+      });
+
+      routeEditor.gqlRouteShouldBeCreatedSuccessfully();
+
+      routeEditor.checkRouteSubmitSuccessToast();
+
+      routeStopsOverlay.routeShouldBeSelected(testRouteLabels.label5);
+
+      // Verify that the stops from the template route are included in the new route
+      // and that the stop count is correct
+      routeStopsOverlay.stopsShouldBeIncludedInRoute(
+        stops.map((item) => item.label),
+      );
+      routeStopsOverlay.assertRouteStopCount(3);
     },
   );
 });
--- a/cypress/e2e/createStop.cy.ts
+++ b/cypress/e2e/createStop.cy.ts
@@ -5,9 +5,8 @@ import {
   timingPlaces,
   VehicleSubmodeOnInfraLinkInsertInput,
 } from '@hsl/jore4-test-db-manager';
-import { ChangeValidityForm, Map, StopForm } from '../pageObjects';
+import { ChangeValidityForm, ModalMap } from '../pageObjects';
 import { FilterPanel } from '../pageObjects/FilterPanel';
-import { ModalMap } from '../pageObjects/ModalMap';
 import { insertToDbHelper, removeFromDbHelper } from '../utils';
 import { deleteStopsByLabel } from './utils';
 
@@ -48,9 +47,7 @@ const clearDatabase = () => {
 describe('Stop creation tests', () => {
   let modalMap: ModalMap;
   let mapFilterPanel: FilterPanel;
-  let map: Map;
   let changeValidityForm: ChangeValidityForm;
-  let stopForm: StopForm;
 
   before(() => {
     cy.fixture('infraLinks/infraLinks.sql').then((infraLinksQuery) => {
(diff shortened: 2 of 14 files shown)


HenrikHartiala (changes requested):
> Reviewed 16 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 16 of 17 files reviewed, 7 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 1 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3-BC8L4UP7b7ElIA:-NLL3-BC8L4UP7b7ElIB:b-s4jls7) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L1)):*
> ```XML
> import {
> ```

Hmm, I'm a little bit confused, shouldn't this `editRouteShape` be in some other PR and not in this one? Isn't this PR only about the route template.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 254 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3FFY-_MVZj02u4L6:-NLL3FFY-_MVZj02u4L7:b-ot98wn) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L254)):*
> ```XML
> 
>   it(
>     'Should create a new route using an existing route as a template',
> ```

This test seems to be the right one for this PR, but I think this is in a wrong file or?
___
*[`cypress/e2e/editRouteShape.cy.ts` line 272 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3h6B0-abg1UTd75N:-NLL3h6B0-abg1UTd75O:b-o7ybkc) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L272)):*
> ```XML
>       });
> 
>       // map.getLoader().should('exist');
> ```

Commented code
___
*[`cypress/e2e/editRouteShape.cy.ts` line 282 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL6UkHCaafPqmYoWqT:-NLL6UkHCaafPqmYoWqU:b-8mc41x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L282)):*
> ```XML
>           line: String(lines[0].label),
>           templateRoute: {
>             templateRouteSelectorInfo: templateInfo,
> ```

I think it makes this test more readable if we spread the input here rather than have a separate variable. We are not using the variable anywhere else. 
___
*[`cypress/e2e/editRouteShape.cy.ts` line 283 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL72pi4FhiZ3Bh4Mq1:-NLL72pjB66adCGk7wR-:b-xijp0o) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L283)):*
> ```XML
>           templateRoute: {
>             templateRouteSelectorInfo: templateInfo,
>             moveRouteEditHandleInfo: moveHandleInfo,
> ```

Also this moveHandleInfo should be spread here.

Second thing about this, if I understood correctly, we are now creating a route using a template and then we edit the route template by using this moveRouteEditHandleInfo. Im not sure if we should have a separate test for this, but in my opinion we should first have a test to check that the route tempalte works as intended. (Comment about this below)
___
*[`cypress/e2e/editRouteShape.cy.ts` line 299 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3zPkCKaTomU4ErIW:-NLL3zPkCKaTomU4ErIX:bks5gol) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L299)):*
> ```XML
>       );
> 
>       routeStopsOverlay.stopsShouldNotBeIncludedInRoute([stops[1].label]);
> ```

I think we should rather check that the route matches the original route. Now we only check that one random stop should not be included in route. 
___
*[`cypress/pageObjects/ModalMap.ts` line 94 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLLAQxr6eOLifI5rs-L:-NLLAQxr6eOLifI5rs-M:bi07q21) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/pageObjects/ModalMap.ts#L94)):*
> ```XML
>     this.map.getLoader().should('not.exist');
> 
>     if (routeFormInfo.templateRoute?.moveRouteEditHandleInfo) {
> ```

This somehow feels wrong to have this here, since it is also hard coded to be able to only change one route point. I am not sure if we need to have the possibility to change the shape during create route IF it is template route and IF we gave the parameter to change it. 

I would maybe think that do we need to test that route shape changing during `create route from template` is working, because we are already checking that the shape changing is working in another test. 

And if we decide that we want to test this, then think that do we want to create another method for this. Now it seems that we will use this part of the code only in one test. 

I'm not putting this comment as `Blocking` for now, but maybe this is more clear after other changes.
___
*[`cypress/pageObjects/RouteStopsOverlay.ts` line 20 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL2A6M2Zzlr5V9QNvA:-NLL2A6M2Zzlr5V9QNvB:bwcmbag) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/pageObjects/RouteStopsOverlay.ts#L20)):*
> ```XML
>   }
> 
>   routeShouldBeSelected(routeName: string) {
> ```

Nice, great change! Now this is much more understandable.

HenrikHartiala (changes requested):
> Reviewed 1 of 17 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: all files reviewed, 4 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 1 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3-BC8L4UP7b7ElIA:-NLQfZxbCGjqUQh0G19O:b-s11ery) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L1)):*

It is present also in https://github.com/HSLdevcom/jore4-ui/pull/431 that has unresolved discussions. The functionality was needed here so in this one it's just copy-pasted code.
</blockquote></details>

Ah, if there is some code that is present on another branch and it is needed in some other branch, then my tip is to create the branch on top of the other branch which has the dependent code in it. And then create the PR after the other PR is merged.

If you copy-paste the same code to two different PR's, then 
1. It is hard to keep track what changes are in the scope of the current PR (like just here happened)
2. If you need to make changes, you need to do the change to both PR's. 
3. Possible conflicts when these PR's needs to be merged.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 254 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3FFY-_MVZj02u4L6:-NLQh0T39-3sUxte1nFm:bipq414) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L254)):*

We can separate this to its own file, but I initially added this here since the setup for this test would be identical.
</blockquote></details>

Yeah, I think that this `Should create a new route using an existing route as a template` should be in `createRoute.ts` and just test that the result's route matches the route used as template.

EDIT:
So I was thinking that should we have the test I described in the other file, which does not modify the route. Just creates new route using template and checks that it matches the route that was used as template. And then we have this test you created but the name would be something like `Should edit route shape correctly when creating new route with template` or something.
What do you think?
___
*[`cypress/e2e/editRouteShape.cy.ts` line 283 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL72pi4FhiZ3Bh4Mq1:-NLQhJu33-9Ef-VXhlxJ:b6ek74f) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L283)):*

How was it intended to work? I know the functionality but not the use case.
</blockquote></details>

I mean that I think we should have this test check that the route created here matches completely the route used as template. So if we pick route '65' as template, and route '65' has stops E001, E002 and E006 for example. Then we should check that this created route has those same stops.

The way you do this here is the actual use case (you choose a template route and then modify it a little bit and save), but as a test, I think we first need to check that the template works and matches the original route.

Hmm, on the first glance I thought that this test was a little bit different (like explained), but if you think we can be sure with this test that the result route is what we want it to be, then I think this test is fine like it is. 
___
*[`cypress/e2e/editRouteShape.cy.ts` line 299 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3zPkCKaTomU4ErIW:-NLQkAIM8s5ZJvSay_kS:bilm1vf) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L299)):*

Done.
</blockquote></details>

Maybe add an assert to check that there are only 2 stops on this route? 
Then we can be sure that it has: 
* exactly 2 stops
* The stops are the ones we want them to be (matching the original route)
* The route modify has changed the route so that one of the stops is not included anymore (`stops[1]`)
___
*[`cypress/pageObjects/ModalMap.ts` line 94 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLLAQxr6eOLifI5rs-L:-NLQlNw251hAwms6mpsa:b-vfr5gr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/pageObjects/ModalMap.ts#L94)):*

If you ask me do we need to test something, I will probably answer yes most of the time. I'm not a developer and have no built-in certainty that some functionality will work if it is tested in only one context. Trust but verify.
</blockquote></details>

Yeah, it is true that this does deserve it's own test. But was just thinking if there was a better approach for the logic inside this if-block. But we can also iterate later if we come up with a better solution, so it is fine :+1:

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 15 of 17 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 1 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3-BC8L4UP7b7ElIA:-NLUlqhVD-nbZpp36B3U:bkxd2jr) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L1)):*

Ah, if there is some code that is present on another branch and it is needed in some other branch, then my tip is to create the branch on top of the other branch which has the dependent code in it. And then create the PR after the other PR is merged.

If you copy-paste the same code to two different PR's, then 
1. It is hard to keep track what changes are in the scope of the current PR (like just here happened)
2. If you need to make changes, you need to do the change to both PR's. 
3. Possible conflicts when these PR's needs to be merged.



</blockquote></details>

Yep, didn't really know which was the better way, but I'll try to remember this.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 299 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3zPkCKaTomU4ErIW:-NLW4lkXC0XjR0Xr1oOz:b-7ttaj1) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L299)):*

Maybe add an assert to check that there are only 2 stops on this route? 
Then we can be sure that it has: 
* exactly 2 stops
* The stops are the ones we want them to be (matching the original route)
* The route modify has changed the route so that one of the stops is not included anymore (`stops[1]`)
</blockquote></details>

I could not come up with a way to reliably check the stop count, but I think that it's a bit redundant since we control the amount of stops that are created and included in the route, and can check that some of those are/are not included in the route.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 15 of 17 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 254 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3FFY-_MVZj02u4L6:-NLtbotl0lBdP317hmR7:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L254)):*

Yeah, I think that this `Should create a new route using an existing route as a template` should be in `createRoute.ts` and just test that the result's route matches the route used as template.

EDIT:
So I was thinking that should we have the test I described in the other file, which does not modify the route. Just creates new route using template and checks that it matches the route that was used as template. And then we have this test you created but the name would be something like `Should edit route shape correctly when creating new route with template` or something.
What do you think?
</blockquote></details>

Done.

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 5 of 18 files reviewed, 4 unresolved discussions (waiting on @sr-vala)

___
*[`-- commits` line 2 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NMD4ukf0HDu2k-yIDH0):*
Commit message is wrong

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 5 of 18 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 2 at r4](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NMD4ukf0HDu2k-yIDH0):*

Commit message is wrong
</blockquote></details>

Done.

HenrikHartiala (commented):
> Reviewed 7 of 13 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 12 of 18 files reviewed, 1 unresolved discussion (waiting on @sr-vala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 299 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3zPkCKaTomU4ErIW:-NMD7LQg-jj-0GK9koTO:blfiaie) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L299)):*

I could not come up with a way to reliably check the stop count, but I think that it's a bit redundant since we control the amount of stops that are created and included in the route, and can check that some of those are/are not included in the route.
</blockquote></details>

I was thinking that do we control the stops in this tests enough to be sure that they can't be tampered with from somewhere else. But now I checked the code again and you are right, the stops are created at the top of the file. This is again something that I do not like that much in these tests that it is not obvious what is the setup for a particular test and what we are or should be asserting.
Well, not sure if those stops could be even created inside the test sensibly, but anyway it is not in the scope of this PR so this is fine by me.
___
*[`cypress/pageObjects/Map.ts` line 76 at r5](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NMDEWuJC6YIQIMtqKEl:-NMDEWuK3PE6lT7RBU8C:b-blo0x) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/5aec99b1aa78cf20712a0266ce2d3e3298a781f2/cypress/pageObjects/Map.ts#L76)):*
> ```XML
>           })}`,
>         ),
>         // Some operations might fail if performed too quickly after opening the map
> ```

Just pondering, not in the scope of this PR, but I would really like to come up with another solution to these. Because in e2e tests we should not need to wait for things that are not visible to user. And in addition to that, we are having these troubles of the tests performing actions before the app is ready. So maybe we need to discuss about this in slack for example to add loading spinners everywhere or some other solution.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/442)* status: 11 of 18 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 299 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/442#-NLL3zPkCKaTomU4ErIW:-NMDh_rD93y7luAkWD-f:b2q3p9x) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L299)):*

I was thinking that do we control the stops in this tests enough to be sure that they can't be tampered with from somewhere else. But now I checked the code again and you are right, the stops are created at the top of the file. This is again something that I do not like that much in these tests that it is not obvious what is the setup for a particular test and what we are or should be asserting.
Well, not sure if those stops could be even created inside the test sensibly, but anyway it is not in the scope of this PR so this is fine by me.
</blockquote></details>

We could research dividing the tests up to separate contexts in the same spec file, and having the setup for each test separately, but this does not make sense if the context for all tests in the file is actually the same. https://stackoverflow.com/a/56193026

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 11 of 18 files reviewed, 1 unresolved discussion (waiting on @sr-vala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 299 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NLL3zPkCKaTomU4ErIW:-NMSZwtS7osPR4PklNJh:b-1k4ah8) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/e807b2f0918b91a0094861644dfab2414620a8e2/cypress/e2e/editRouteShape.cy.ts#L299)):*
Hmm, not sure how context would help us. https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Test-Structure
It just seems to be another word for structuring tests.
> context() is identical to describe() and specify() is identical to it(), so choose whatever terminology works best for you.

HenrikHartiala (changes requested):
> Reviewed 4 of 13 files at r4, 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: all files reviewed, 3 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/createRoute.cy.ts` line 416 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NNetzE3ACre1vF2G8KH:-NNetzE3ACre1vF2G8KI:b-vcr1rx) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bf1aca7fbc5317ca63baef84e543ba3f248a90a8/cypress/e2e/createRoute.cy.ts#L416)):*
> ```XML
>       routeStopsOverlay.routeShouldBeSelected(testRouteLabels.label5);
> 
>       routeStopsOverlay.stopsShouldBeIncludedInRoute(
> ```

Maybe add a comment here that the stops included in `stops` variable, are the same ones than in the original route?

Also I would add an assert here that the amount of the stops also match, because now we are only checking that all the necessary stops are included, but think of a case where a bug occurs and ALL the existing stops are included in this route. This test would still pass.

EDIT: Before anything, check the other comment about this as well.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 297 at r7](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NNevdnn0E64JHAP-dLC:-NNevdnn0E64JHAP-dLD:b9dtneh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/bf1aca7fbc5317ca63baef84e543ba3f248a90a8/cypress/e2e/editRouteShape.cy.ts#L297)):*
> ```XML
>       ]);
> 
>       routeStopsOverlay.stopsShouldNotBeIncludedInRoute([stops[1].label]);
> ```

Same thing here, I think we want to check that the amount of the stops match. 

EDIT: Ah, we have already gone this discussion once. I went through that discussion and I'm actually now leaning towards another end. Because even though we are controlling the stops and we "know" that there could not be anymore stops, then I would say that we would need to add more stops to seed data that won't be included in the route. This way we get more out of this test.
Again the same example: what if a bug occurs which would include more stops (or all the stops) to the route than intended? This test would still pass.

Then also, would something like this work for counting the stops?
.find('RouteStopsOverlayRow::').should('have.length', 4)

We might need to modify the code to get the amount of the stops. If the counting of the stops is not trivial here, then I would maybe add a TODO-comment about this.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/442)* status: 15 of 18 files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/createRoute.cy.ts` line 416 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/442#-NNetzE3ACre1vF2G8KH:-NNfv994AhJNsqHQ7Zin:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bf1aca7fbc5317ca63baef84e543ba3f248a90a8/cypress/e2e/createRoute.cy.ts#L416)):*

Maybe add a comment here that the stops included in `stops` variable, are the same ones than in the original route?

Also I would add an assert here that the amount of the stops also match, because now we are only checking that all the necessary stops are included, but think of a case where a bug occurs and ALL the existing stops are included in this route. This test would still pass.

EDIT: Before anything, check the other comment about this as well.
</blockquote></details>

Done.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 297 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/442#-NNevdnn0E64JHAP-dLC:-NNfv9kt4X7gvu71iBz8:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bf1aca7fbc5317ca63baef84e543ba3f248a90a8/cypress/e2e/editRouteShape.cy.ts#L297)):*

Same thing here, I think we want to check that the amount of the stops match. 

EDIT: Ah, we have already gone this discussion once. I went through that discussion and I'm actually now leaning towards another end. Because even though we are controlling the stops and we "know" that there could not be anymore stops, then I would say that we would need to add more stops to seed data that won't be included in the route. This way we get more out of this test.
Again the same example: what if a bug occurs which would include more stops (or all the stops) to the route than intended? This test would still pass.

Then also, would something like this work for counting the stops?
.find('RouteStopsOverlayRow::').should('have.length', 4)

We might need to modify the code to get the amount of the stops. If the counting of the stops is not trivial here, then I would maybe add a TODO-comment about this. 
</blockquote></details>

Done.

HenrikHartiala (approved):
> Reviewed 1 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: 16 of 18 files reviewed, 3 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 292 at r8](https://reviewable.io/reviews/hsldevcom/jore4-ui/442#-NNgN3c8EDepxZ29QbEz:-NNgN3c8EDepxZ29QbF-:b-hwdxeh) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/963bf64baf2914160561fb6e186c0eac8fd50ac3/cypress/e2e/editRouteShape.cy.ts#L292)):*
> ```XML
>       );
> 
>       // Verify that stop with label 'E2E002' is not included in the route
> ```

"Verify that the edited route shape is excluding the second stop" would maybe be more descriptive of why we want it to be excluded. Also I would put the assertions in the same order as in the other test. So assert the count last. 

But up to you if you want to change this.

HenrikHartiala (approved):
> Reviewed 1 of 3 files at r8, 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/442)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sr-vala)

PR #444, opened 2023-01-05
Add hidden variant to route creation test

sr-vala:
Resolves HSLdevcom/jore4#1090

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/444)

### DIFF
--- a/cypress/e2e/createRoute.cy.ts
+++ b/cypress/e2e/createRoute.cy.ts
@@ -10,10 +10,15 @@ import {
   VehicleSubmodeOnInfraLinkInsertInput,
 } from '@hsl/jore4-test-db-manager';
 import { DateTime } from 'luxon';
-import { Map, RouteEditor } from '../pageObjects';
-import { FilterPanel } from '../pageObjects/FilterPanel';
-import { ModalMap } from '../pageObjects/ModalMap';
-import { RouteStopsOverlay } from '../pageObjects/RouteStopsOverlay';
+import {
+  FilterPanel,
+  LineDetailsPage,
+  ModalMap,
+  RouteEditor,
+  RoutePropertiesForm,
+  RouteStopsTable,
+  TerminusNameInputs,
+} from '../pageObjects';
 import { insertToDbHelper, removeFromDbHelper } from '../utils';
 import { deleteRoutesByLabel } from './utils';
 
@@ -28,20 +33,20 @@ const testRouteLabels = {
 // These form a straight line on Eerikinkatu in Helsinki.
 // Coordinates are partial since they are needed only for the stop creation.
 
-const testInfraLinks = {
-  1: {
+const testInfraLinks = [
+  {
     id: '73bc2df9-f5af-4c38-a1dd-5ed1f71c90a8',
     coordinates: [24.92492146851626, 60.1634759878872, 0],
   },
-  2: {
+  {
     id: 'ea69415a-9c54-4327-8836-f38b36d8fa99',
     coordinates: [24.92904198486008, 60.16490775039894, 0],
   },
-  3: {
+  {
     id: '13de61c2-3fc9-4255-955f-0a2350c389e1',
     coordinates: [24.932072417514647, 60.166003223527824, 0],
   },
-};
+];
 
 const lines: LineInsertInput[] = [
   {
@@ -66,10 +71,22 @@ const stops: StopInsertInput[] = [
   {
     ...buildStop({
       label: 'Test stop 1',
-      located_on_infrastructure_link_id: testInfraLinks[1].id,
+      located_on_infrastructure_link_id: testInfraLinks[0].id,
     }),
     validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
     scheduled_stop_point_id: '3f23a4c5-f527-4395-bd9f-bbc398f837df',
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[0].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: 'Test stop 2',
+      located_on_infrastructure_link_id: testInfraLinks[1].id,
+    }),
+    validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
+    scheduled_stop_point_id: '431a6791-f1f5-45d4-8c9d-9e154a2531e0',
     measured_location: {
       type: 'Point',
       coordinates: testInfraLinks[1].coordinates,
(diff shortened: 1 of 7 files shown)


HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/444)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/createRoute.cy.ts` line 189 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/444#-NLLPGTu3krwSUxGKYnU:-NLLPGTu3krwSUxGKYnV:b-w021g2) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/516488c12ad18ff65d32144c724fd07c8d5c3063/cypress/e2e/createRoute.cy.ts#L189)):*
> ```XML
>     () => {
>       const routeName = 'Testireitti 1';
>       const routeInfo: RouteFormInfo = {
> ```

This might be a personal preference, because both of these approaches has their ups and downs. But I personally prefer to not have this kind of variables within the test. 

I think that the upside for this approach is that you don't need to repeat the values in the end and also that if you change the input, you don't need to change the assertion as well. But, that is also a downside in tests. Because if we repeat the values (for assign and assert) we have a double check that the test is testing what we want it to test.

Example: If we have an imagined test named `Foo priority should be Standard`
then our code inside the test looks something like this:
```
const fooInfo = {
  priority: Priority.Standard,
  ...
}

bar.createFoo(fooInfo);

bar.verifyFooInfo(fooInfo);
```

This works, but it also works if we change `fooInfo` to 
```
const fooInfo = {
  priority: Priority.Draft
  ...
}
```

Compare this to the next:
```
bar.createFoo({
  priority: Priority.Standard
})

bar.verifyFoo({
  priority: Priority.Standard,
  ...
})
```
Now it is very clear what we are assigning in the creation, what we are asserting in the end and also, now if you make the following change:
```
bar.createFoo({
  priority: Priority.Draft,
  ...
})
```
The test will fail like it should.
Would like to hear your thoughts about this. Does this make sense?

I would also like to hear what @tk-vincit thinks about this
___
*[`cypress/e2e/createRoute.cy.ts` line 230 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/444#-NLLSOrP4yzWnWJvy6xZ:-NLLSOrP4yzWnWJvy6x_:b-q19hia) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/516488c12ad18ff65d32144c724fd07c8d5c3063/cypress/e2e/createRoute.cy.ts#L230)):*
> ```XML
>       routeStopsTable.getEditRouteButton(routeName).click();
> 
>       modalMap.routePropertiesForm.verifyRouteFormInfo(routeInfo);
> ```

In addition to the long comment above, this method also kind of hides what we are actually asserting in this test. It is hard to check that do we actually assert the correct things.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/444)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`cypress/e2e/createRoute.cy.ts` line 189 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/444#-NLLPGTu3krwSUxGKYnU:-NLPbCBz--_e1YAVma80:b8y45yj) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/516488c12ad18ff65d32144c724fd07c8d5c3063/cypress/e2e/createRoute.cy.ts#L189)):*

This might be a personal preference, because both of these approaches has their ups and downs. But I personally prefer to not have this kind of variables within the test. 

I think that the upside for this approach is that you don't need to repeat the values in the end and also that if you change the input, you don't need to change the assertion as well. But, that is also a downside in tests. Because if we repeat the values (for assign and assert) we have a double check that the test is testing what we want it to test.

Example: If we have an imagined test named `Foo priority should be Standard`
then our code inside the test looks something like this:
```
const fooInfo = {
  priority: Priority.Standard,
  ...
}

bar.createFoo(fooInfo);

bar.verifyFooInfo(fooInfo);
```

This works, but it also works if we change `fooInfo` to 
```
const fooInfo = {
  priority: Priority.Draft
  ...
}
```

Compare this to the next:
```
bar.createFoo({
  priority: Priority.Standard
})

bar.verifyFoo({
  priority: Priority.Standard,
  ...
})
```
Now it is very clear what we are assigning in the creation, what we are asserting in the end and also, now if you make the following change:
```
bar.createFoo({
  priority: Priority.Draft,
  ...
})
```
The test will fail like it should.
Would like to hear your thoughts about this. Does this make sense?

I would also like to hear what @tk-vincit thinks about this
</blockquote></details>

No strong feelings about this one, but I removed the variables now.
___
*[`cypress/e2e/createRoute.cy.ts` line 230 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/444#-NLLSOrP4yzWnWJvy6xZ:-NLPbNjWBA4DYX8rUDNE:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/516488c12ad18ff65d32144c724fd07c8d5c3063/cypress/e2e/createRoute.cy.ts#L230)):*

In addition to the long comment above, this method also kind of hides what we are actually asserting in this test. It is hard to check that do we actually assert the correct things.
</blockquote></details>

Done.

HenrikHartiala (changes requested):
> Reviewed 4 of 8 files at r1, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/444)* status: all files reviewed, 1 unresolved discussion (waiting on @sr-vala)

___
*[`cypress/e2e/createRoute.cy.ts` line 230 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/444#-NLLSOrP4yzWnWJvy6xZ:-NM3v9tq3Im6etDal-Y7:bbllvtl) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/516488c12ad18ff65d32144c724fd07c8d5c3063/cypress/e2e/createRoute.cy.ts#L230)):*

Done.
</blockquote></details>

I more like meant that this verifyRouteFormInfo as a whole hides what is actually checked. We now only see that what it takes as parameter, but can't see straight away if they are all checked properly. So my suggestion here is to not have this kind of bunched up assertions behind methods and rather have them inside this test.

There are exceptions to this kind of extracted asserts, for example the `assertPriority`because what it really does, is to wrap up a bit more complex asserting logic due to the nature of the component. But this here is a whole form and we do not benefit for it to be extracted as it's own method.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/444)* status: 4 of 10 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/e2e/createRoute.cy.ts` line 230 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/444#-NLLSOrP4yzWnWJvy6xZ:-NM8MEH989T-2rcQFBia:bqijszy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/516488c12ad18ff65d32144c724fd07c8d5c3063/cypress/e2e/createRoute.cy.ts#L230)):*

I more like meant that this verifyRouteFormInfo as a whole hides what is actually checked. We now only see that what it takes as parameter, but can't see straight away if they are all checked properly. So my suggestion here is to not have this kind of bunched up assertions behind methods and rather have them inside this test.

There are exceptions to this kind of extracted asserts, for example the `assertPriority`because what it really does, is to wrap up a bit more complex asserting logic due to the nature of the component. But this here is a whole form and we do not benefit for it to be extracted as it's own method.
</blockquote></details>

Refactored some code to show the inputs in the tests.

sr-vala:
There's a bit too much stuff in this PR currently. I'm closing this and opening a new one for just the hidden variant changes. Let's handle test data related changes separately.

PR #446, opened 2023-01-09
Add test that deletes a route

sr-vala:
Resolves HSLdevcom/jore4#1064

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/446)

### DIFF
--- a/cypress/e2e/editRoute.cy.ts
+++ b/cypress/e2e/editRoute.cy.ts
@@ -5,7 +5,14 @@ import {
   RouteInsertInput,
   RouteDirectionEnum,
 } from '@hsl/jore4-test-db-manager';
-import { EditRoutePage, LineDetailsPage, TerminusValues } from '../pageObjects';
+import {
+  EditRoutePage,
+  LineDetailsPage,
+  RoutesAndLinesPage,
+  SearchResultsPage,
+  TerminusValues,
+  Toast,
+} from '../pageObjects';
 import { insertToDbHelper, removeFromDbHelper } from '../utils';
 
 const routeFormTestInputs = {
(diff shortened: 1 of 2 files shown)


eleksi (commented):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/446)* status: all files reviewed, 1 unresolved discussion (waiting on @sr-vala)

___
*[`cypress/e2e/editRoute.cy.ts` line 123 at r1](https://reviewable.io/reviews/hsldevcom/jore4-ui/446#-NLuhH0p45e-4Z1Ih_Dw:-NLuhH0p45e-4Z1Ih_Dx:bi4iefu) ([raw file](https://github.com/hsldevcom/jore4-ui/blob/617175175a336cc34b30bed60d11a89805ae702c/cypress/e2e/editRoute.cy.ts#L123)):*
> ```XML
>     toast.checkSuccessToastHasMessage('Reitti poistettu');
>     editRoutePage.visit(routes[0].route_id);
>     editRoutePage.routePropertiesForm.getForm().should('not.exist');
> ```

Just wondering if our UI should show some 404 page or something like that in case the route does not exist. Now it doesn't show the form but only the buttons for deletion, cancel and save, which does not make sense. Well, that is not really related to this test but the UI itself.

I was thinking if just checking the form not existing is enough proof of the route not existing... Maybe check the delete request response code like in `gqlRouteShouldBeCreatedSuccessfully` or do a separate query to graphql to check if a route is returned when fetched with the route id.

eleksi (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-ui/446)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sr-vala)

PR #468, opened 2023-01-23
Throw error when Db Helper operations fail

sr-vala:
Resolves HSLdevcom/jore4#1095

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/468)

### DIFF
--- a/cypress/utils/db.ts
+++ b/cypress/utils/db.ts
@@ -29,9 +29,33 @@ import {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const responseLogger = (message: string, res: any) => {
   if (res.errors) {
-    // eslint-disable-next-line no-console
-    console.log(`${message}:`, res);
+    cy.log(`${message}:`, res);
   }
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const throwOnErrorResponse = (message: string, res: any) => {
+  if (res.errors) {
+    if (res.errors[0].message) {
+      throw new Error(
+        `Graphql request error when ${message}: ${res.errors[0].message}`,
+      );
+    }
+    if (res.errors[0].extensions.internal.error.message) {
+      throw new Error(
+        `Graphql request error when ${message}: ${res.errors[0].extensions.internal.error.message}`,
+      );
+    }
+    if (!res.errors[0]) {
+      throw new Error(`Graphql request error when ${message}: ${res.errors}`);
+    }
+  }
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const handleResponse = (message: string, res: any) => {
+  throwOnErrorResponse(message, res);
+  responseLogger(message, res);
 };
 
 interface SupportedResources {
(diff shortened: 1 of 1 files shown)


tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/468)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/utils/db.ts` line 32 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/468#-NMYtG8n1m2jWoVtkOTK:-NMYtG8n1m2jWoVtkOTL:b-2sjohv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/56dc584707b16c314c1c135b783df9e36d74bab0/cypress/utils/db.ts#L32)):*
> ```XML
> const responseLogger = (message: string, res: any) => {
>   if (res.errors) {
>     cy.log(`${message}:${res.errors[0].extensions.internal.error.message}`);
> ```

Seems like we can have error messages at least in two different paths:
- `res.errors[0].message`
- `res.errors[0].extensions.internal.error.message`

As it seems like we did want to take that into account when throwing error messages, perharps that should have been done also here?

Or as alternative, just leave this as this was, with current implementation we won't end up here by default anyway as we are throwing before this gets executed. Anyway, when debugging one might want to disable error throwing and then logger can be useful.

So:
`console.log(`${message}:`, res);`
OR
`cy.log(`${message}:`, res);`
___
*[`cypress/utils/db.ts` line 49 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/468#-NMYuEUX7-Xjxl7VhlzO:-NMYuEUX7-Xjxl7VhlzP:b-c9bxkr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/56dc584707b16c314c1c135b783df9e36d74bab0/cypress/utils/db.ts#L49)):*
> ```XML
>       );
>     }
>   }
> ```

What if res.errors[0] is not defined, do we still want to throw some generic error like:
` throw new Error(
        `Graphql request error when ${message}: ${res.errors}`,
      );
This could be added right after `  if (res.errors[0].extensions.internal.error.message) {} ` code block, inside `if (res.errors)`

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/468)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[`cypress/utils/db.ts` line 32 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/468#-NMYtG8n1m2jWoVtkOTK:-NMcttj893P8HjG6IRPd:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/56dc584707b16c314c1c135b783df9e36d74bab0/cypress/utils/db.ts#L32)):*

Seems like we can have error messages at least in two different paths:
- `res.errors[0].message`
- `res.errors[0].extensions.internal.error.message`

As it seems like we did want to take that into account when throwing error messages, perharps that should have been done also here?

Or as alternative, just leave this as this was, with current implementation we won't end up here by default anyway as we are throwing before this gets executed. Anyway, when debugging one might want to disable error throwing and then logger can be useful.

So:
`console.log(`${message}:`, res);`
OR
`cy.log(`${message}:`, res);`
</blockquote></details>

Done.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/468)* status: 0 of 2 files reviewed, all discussions resolved

PR #492, opened 2023-02-01
Import timetables with preview

eleksi:
Resolves https://github.com/HSLdevcom/jore4/issues/1087

Test data for imported timetables can be generated with branch `seed-staging-timetables` seed commands.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)

### DIFF
--- a/ui/src/components/map/MapLoader.tsx
+++ b/ui/src/components/map/MapLoader.tsx
@@ -2,10 +2,11 @@ import { useAppSelector, useMapQueryParams } from '../../hooks';
 import { selectIsMapOperationLoading } from '../../redux';
 import { LoadingOverlay } from '../../uiComponents';
 
-export const MapLoader = () => {
-  const testIds = {
-    loader: 'MapLoader::loader',
-  };
+const testIds = {
+  loader: 'MapLoader::loader',
+};
+
+export const MapLoader = (): JSX.Element => {
   const { isMapOpen } = useMapQueryParams();
   const isLoading = useAppSelector(selectIsMapOperationLoading);
 
--- a/ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap
+++ b/ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap
@@ -182,7 +182,7 @@ exports[`<RouteStopsTable /> Renders the route with stops along its geometry 1`]
               type="button"
             >
               <svg
-                class="text-3xl text-3xl text-tweaked-brand inline text-center"
+                class="text-3xl text-tweaked-brand text-3xl  inline text-center"
                 fill="currentColor"
                 height="1em"
                 stroke="currentColor"
@@ -390,7 +390,7 @@ exports[`<RouteStopsTable /> Renders the route with stops along its geometry 2`]
               type="button"
             >
               <svg
-                class="text-3xl text-3xl text-tweaked-brand inline text-center"
+                class="text-3xl text-tweaked-brand text-3xl inline text-center"
                 fill="currentColor"
                 height="1em"
                 stroke="currentColor"
@@ -798,7 +798,7 @@ exports[`<RouteStopsTable /> Renders the route with stops along its geometry 3`]
               type="button"
             >
               <svg
-                class="text-3xl text-3xl text-tweaked-brand inline text-center"
+                class="text-3xl text-tweaked-brand text-3xl inline text-center"
                 fill="currentColor"
                 height="1em"
                 stroke="currentColor"
--- a/ui/src/components/routes-and-lines/search/SearchContainer.tsx
+++ b/ui/src/components/routes-and-lines/search/SearchContainer.tsx
@@ -1,7 +1,10 @@
 import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
-import { SearchQueryParameterNames, useSearch } from '../../../hooks';
-import { useToggle } from '../../../hooks/useToggle';
+import {
+  SearchQueryParameterNames,
+  useSearch,
+  useToggle,
+} from '../../../hooks';
 import { Column, Container, Row, Visible } from '../../../layoutComponents';
 import { ChevronToggle, SimpleButton } from '../../../uiComponents';
 import { AllOptionEnum } from '../../../utils';
--- a/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx
+++ b/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx
@@ -1,0 +1,111 @@
+import { VehicleJourneyWithRouteInfoFragment } from '../../../generated/graphql';
+import { useToggle } from '../../../hooks';
+import { parseI18nField } from '../../../i18n/utils';
+import { Row, Visible } from '../../../layoutComponents';
+import { mapDurationToShortTime } from '../../../time';
+import { AccordionButton } from '../../../uiComponents';
+import { getRouteLabelVariantText } from '../../../utils/route';
+import { DirectionBadge } from '../../routes-and-lines/line-details/DirectionBadge';
+
+const testIds = {
+  table: 'BlockVehicleJourneysTable::table',
+  toggleShowTable: 'BlockVehicleJourneysTable::toggleShowTable',
+};
+
+const VehicleJourneyRow = ({
+  vehicleJourney,
+}: {
+  vehicleJourney: VehicleJourneyWithRouteInfoFragment;
+}): JSX.Element => {
+  const route =
+    vehicleJourney.journey_pattern_ref.journey_pattern_instance
+      ?.journey_pattern_route;
+
+  // This should never happen as route info is fetched in the same query as the vehicle journey itself.
+  // Do this to tell typescript that the route exists.
+  if (!route) {
+    // eslint-disable-next-line no-console
+    console.error('No route found for vehicle journey!');
+    return <></>;
+  }
+
+  return (
+    <tr className="odd:bg-white [&>td]:border [&>td]:border-light-grey [&>td]:px-5 [&>td]:py-2">
+      <td>
+        <Row className="items-center font-bold">
+          <DirectionBadge
+            direction={route.direction}
+            className="mr-2 h-5 w-5 text-base"
+          />
+          {getRouteLabelVariantText(route)}
+        </Row>
+      </td>
+      <td>
+        {parseI18nField(
+          vehicleJourney.block.vehicle_service.day_type.name_i18n,
+        )}
+      </td>
+      <td>
+        {vehicleJourney.start_time
+          ? mapDurationToShortTime(vehicleJourney.start_time)
+          : ''}
+      </td>
+      <td>
+        {vehicleJourney.end_time
+          ? mapDurationToShortTime(vehicleJourney.end_time)
+          : ''}
+      </td>
+    </tr>
+  );
+};
+
+interface Props {
+  vehicleJourneys: VehicleJourneyWithRouteInfoFragment[];
+  title: string;
+}
+
+export const BlockVehicleJourneysTable = ({
+  vehicleJourneys,
+  title,
+}: Props): JSX.Element => {
+  const [isOpen, toggleIsOpen] = useToggle();
+
+  return (
+    <table
+      className="border-brand-gray w-full border"
+      data-testid={testIds.table}
+    >
+      <thead>
+        <tr className="[&>th]:border [&>th]:border-light-grey">
+          <th className="w-1/6 bg-brand py-2 px-6 text-left text-white">
+            {title}
+          </th>
+          <th
+            className="border-l border-l-white bg-white py-2 px-4 text-left text-hsl-dark-80"
+            colSpan={4}
+          >
+            <Row className="flex-1 items-center justify-between font-normal">
+              <p>!Kalustotyyppi 3 - Normaalibussi</p>
+              <AccordionButton
+                testId={testIds.toggleShowTable}
+                isOpen={isOpen}
+                onToggle={toggleIsOpen}
+                iconClassName="text-brand text-[50px]"
+              />
+            </Row>
+          </th>
+        </tr>
+      </thead>
+      <Visible visible={isOpen}>
+        <tbody>
+          {vehicleJourneys.map((vehicleJourney) => (
+            <VehicleJourneyRow
+              key={vehicleJourney.vehicle_journey_id}
+              vehicleJourney={vehicleJourney}
+            />
+          ))}
+        </tbody>
+      </Visible>
+    </table>
+  );
+};
--- a/ui/src/components/timetables/import/ConfirmPreviewedTimetablesImportForm.tsx
+++ b/ui/src/components/timetables/import/ConfirmPreviewedTimetablesImportForm.tsx
@@ -1,0 +1,47 @@
+import { zodResolver } from '@hookform/resolvers/zod';
+import React from 'react';
+import { FormProvider, useForm } from 'react-hook-form';
+import {
+  PriorityForm,
+  PriorityFormState,
+  priorityFormSchema,
+} from '../../forms/common';
+
+const schema = priorityFormSchema;
+
+export type FormState = PriorityFormState;
+
+interface Props {
+  defaultValues?: Partial<FormState>;
+  onSubmit: (state: FormState) => void;
+}
+
+export const ConfirmPreviewedTimetablesImportFormComponent = (
+  { defaultValues, onSubmit }: Props,
+  externalRef: ExplicitAny,
+): JSX.Element => {
+  const methods = useForm<FormState>({
+    defaultValues,
+    resolver: zodResolver(schema),
+  });
+
+  const { handleSubmit } = methods;
+
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    <FormProvider {...methods}>
+      <form
+        id="save-timetables-form"
+        onSubmit={handleSubmit(onSubmit)}
+        ref={externalRef}
+      >
+        <PriorityForm showLabel={false} />
+      </form>
+    </FormProvider>
+  );
+};
+
+export const ConfirmPreviewedTimetablesImportForm = React.forwardRef<
+  ExplicitAny,
+  Props
+>(ConfirmPreviewedTimetablesImportFormComponent);
--- a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx
+++ b/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { useConfirmTimetablesImport } from '../../../hooks/timetables-import/useConfirmTimetablesImport';
-import { TimetablePriority } from '../../../types/enums';
+import { useConfirmTimetablesImportUIAction } from '../../../hooks';
 import { Modal, ModalBody, ModalHeader } from '../../../uiComponents';
 import {
   ConfirmTimetablesImportForm,
@@ -20,12 +19,11 @@ export const ConfirmTimetablesImportModal: React.FC<Props> = ({
   className = '',
 }) => {
   const { t } = useTranslation();
-  const { confirmTimetablesImport } = useConfirmTimetablesImport();
+  const { onConfirmTimetablesImport } = useConfirmTimetablesImportUIAction();
 
   const onSave = async (state: FormState) => {
-    await confirmTimetablesImport(
-      state.priority as unknown as TimetablePriority,
-    );
+    await onConfirmTimetablesImport(state.priority);
+
     onClose();
   };
 
--- a/ui/src/components/timetables/import/ImportContentsView.tsx
+++ b/ui/src/components/timetables/import/ImportContentsView.tsx
@@ -1,0 +1,24 @@
+import { useTranslation } from 'react-i18next';
+import { VehicleScheduleFrameWithRouteInfoFragment } from '../../../generated/graphql';
+import { VehicleScheduleFrameBlocksView } from './VehicleScheduleFrameBlocksView';
+
+interface Props {
+  vehicleScheduleFrames: VehicleScheduleFrameWithRouteInfoFragment[];
+}
+
+export const ImportContentsView = ({
+  vehicleScheduleFrames,
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+  return (
+    <div className="space-y-3">
+      <h2>{t('timetablesPreview.fileContent')}</h2>
+      {vehicleScheduleFrames.map((frame) => (
+        <VehicleScheduleFrameBlocksView
+          key={frame.vehicle_schedule_frame_id}
+          vehicleScheduleFrame={frame}
+        />
+      ))}
+    </div>
+  );
+};
--- a/ui/src/components/timetables/import/ImportTimetablesPage.tsx
+++ b/ui/src/components/timetables/import/ImportTimetablesPage.tsx
@@ -12,10 +12,10 @@ const testIds = {
 
 export const ImportTimetablesPage = (): JSX.Element => {
   const { t } = useTranslation();
-  const { vehicleJourneyCount } = useConfirmTimetablesImport();
+  const { vehicleJourneys } = useConfirmTimetablesImport();
   const [isSavingImport, setIsSavingImport] = useState(false);
 
-  const importedTimetablesExist = vehicleJourneyCount > 0;
+  const importedTimetablesExist = vehicleJourneys.length > 0;
 
   return (
     <Container>
--- a/ui/src/components/timetables/import/ImportTimetablesPage.tsx
+++ b/ui/src/components/timetables/import/ImportTimetablesPage.tsx
@@ -1,7 +1,8 @@
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useConfirmTimetablesImport } from '../../../hooks/timetables-import/useConfirmTimetablesImport';
-import { Container } from '../../../layoutComponents';
+import { Container, Row } from '../../../layoutComponents';
+import { Path } from '../../../router/routeDetails';
 import { SimpleButton } from '../../../uiComponents';
 import { ConfirmTimetablesImportModal } from './ConfirmTimetablesImportModal';
 
@@ -11,21 +12,30 @@ const testIds = {
 
 export const ImportTimetablesPage = (): JSX.Element => {
   const { t } = useTranslation();
-  const { vehicleServiceCount } = useConfirmTimetablesImport();
+  const { vehicleJourneyCount } = useConfirmTimetablesImport();
   const [isSavingImport, setIsSavingImport] = useState(false);
 
-  const importedTimetablesExist = vehicleServiceCount > 0;
+  const importedTimetablesExist = vehicleJourneyCount > 0;
 
   return (
     <Container>
       <h1>{t('timetables.importTimetablesFromHastus')}</h1>
-      <SimpleButton
-        testId={testIds.saveButton}
-        onClick={() => setIsSavingImport(true)}
-        disabled={!importedTimetablesExist}
-      >
-        {t('save')}
-      </SimpleButton>
+      <Row className="space-x-3">
+        <SimpleButton
+          testId={testIds.saveButton}
+          onClick={() => setIsSavingImport(true)}
+          disabled={!importedTimetablesExist}
+        >
+          {t('save')}
+        </SimpleButton>
+        <SimpleButton
+          testId={testIds.saveButton}
+          href={Path.timetablesImportPreview}
+          disabled={!importedTimetablesExist}
+        >
+          {t('timetables.openPreview')}
+        </SimpleButton>
+      </Row>
       <ConfirmTimetablesImportModal
         isOpen={isSavingImport}
         onClose={() => setIsSavingImport(false)}
--- a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
+++ b/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
@@ -1,6 +1,4 @@
-import { zodResolver } from '@hookform/resolvers/zod';
 import { useRef } from 'react';
-import { FormProvider, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router';
 import {
@@ -13,15 +11,10 @@ import { Path, routeDetails } from '../../../router/routeDetails';
 import { AccordionButton, SimpleButton } from '../../../uiComponents';
 import { submitFormByRef } from '../../../utils';
 import {
-  PriorityForm,
-  PriorityFormState,
-  priorityFormSchema,
-} from '../../forms/common';
+  ConfirmPreviewedTimetablesImportForm,
+  FormState,
+} from './ConfirmPreviewedTimetablesImportForm';
 import { ImportContentsView } from './ImportContentsView';
-
-const schema = priorityFormSchema;
-
-type FormState = PriorityFormState;
 
 const testIds = {
   toggleShowStagingTimetables:
@@ -43,13 +36,6 @@ export const PreviewTimetablesPage = (): JSX.Element => {
   const importedTimetablesExist = vehicleJourneyCount > 0;
 
   const formRef = useRef<ExplicitAny>(null);
-
-  const methods = useForm<FormState>({
-    defaultValues: undefined,
-    resolver: zodResolver(schema),
-  });
-
-  const { handleSubmit } = methods;
 
   const onSave = () => {
     submitFormByRef(formRef);
@@ -95,16 +81,10 @@ export const PreviewTimetablesPage = (): JSX.Element => {
         </Row>
         <Row className="items-center space-x-14 py-9 px-16">
           <h3>{t('timetablesPreview.contentUsage')}</h3>
-          {/* eslint-disable-next-line react/jsx-props-no-spreading */}
-          <FormProvider {...methods}>
-            <form
-              id="save-timetables-form"
-              onSubmit={handleSubmit(onSubmit)}
-              ref={formRef}
-            >
-              <PriorityForm showLabel={false} />
-            </form>
-          </FormProvider>
+          <ConfirmPreviewedTimetablesImportForm
+            ref={formRef}
+            onSubmit={onSubmit}
+          />
         </Row>
         <Visible visible={showStagingTimetables}>
           <div className="items-center space-x-14 rounded-b-sm bg-hsl-neutral-blue py-9 px-16">
--- a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
+++ b/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
@@ -1,0 +1,126 @@
+import { zodResolver } from '@hookform/resolvers/zod';
+import { useReducer, useRef } from 'react';
+import { FormProvider, useForm } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
+import { useConfirmTimetablesImport } from '../../../hooks/timetables-import/useConfirmTimetablesImport';
+import { Container, Row, Visible } from '../../../layoutComponents';
+import { Path } from '../../../router/routeDetails';
+import { TimetablePriority } from '../../../types/Priority';
+import { AccordionButton, SimpleButton } from '../../../uiComponents';
+import { submitFormByRef } from '../../../utils';
+import {
+  PriorityForm,
+  PriorityFormState,
+  priorityFormSchema,
+} from '../../forms/common';
+
+const schema = priorityFormSchema;
+
+export type FormState = PriorityFormState;
+
+const testIds = {
+  toggleShowStagingTimetables:
+    'PreviewTimetablesPage::toggleShowStagingTimetables',
+  saveButton: 'PreviewTimetablesPage::saveButton',
+  cancelButton: 'PreviewTimetablesPage::cancelButton',
+};
+
+export const PreviewTimetablesPage = (): JSX.Element => {
+  const { t } = useTranslation();
+
+  const { vehicleJourneyCount, confirmTimetablesImport } =
+    useConfirmTimetablesImport();
+  const [showStagingTimetables, toggleShowStagingTimetables] = useReducer(
+    (value) => !value,
+    false,
+  );
+
+  const importedTimetablesExist = vehicleJourneyCount > 0;
+
+  const formRef = useRef<ExplicitAny>(null);
+
+  const methods = useForm<FormState>({
+    defaultValues: undefined,
+    resolver: zodResolver(schema),
+  });
+
+  const { handleSubmit } = methods;
+
+  const onSave = () => {
+    submitFormByRef(formRef);
+  };
+
+  const onSubmit = async (state: FormState) => {
+    await confirmTimetablesImport(
+      state.priority as unknown as TimetablePriority,
+    );
+  };
+
+  return (
+    <Container>
+      <h1>{t('timetablesPreview.preview')}</h1>
+      <div className="overflow-none mt-9 rounded border border-grey">
+        <Row className="justify-between rounded-t-sm border-brand bg-brand pl-16 pr-8 text-white">
+          <div className="py-2">
+            <h2>
+              {t('timetablesPreview.departures')} {vehicleJourneyCount}
+            </h2>
+          </div>
+          <div className="flex items-center">
+            <button
+              onClick={toggleShowStagingTimetables}
+              type="button"
+              className="font-bold"
+            >
+              {showStagingTimetables
+                ? t('timetablesPreview.closeContent')
+                : t('timetablesPreview.showContent')}
+            </button>
+            <AccordionButton
+              testId={testIds.toggleShowStagingTimetables}
+              isOpen={showStagingTimetables}
+              onToggle={toggleShowStagingTimetables}
+              iconClassName="text-white text-[50px]"
+            />
+          </div>
+        </Row>
+        <Row className="items-center space-x-14 py-9 px-16">
+          <h3>{t('timetablesPreview.contentUsage')}</h3>
+          {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+          <FormProvider {...methods}>
+            <form
+              id="save-timetables-form"
+              onSubmit={handleSubmit(onSubmit)}
+              ref={formRef}
+            >
+              <PriorityForm showLabel={false} />
+            </form>
+          </FormProvider>
+        </Row>
+        <Visible visible={showStagingTimetables}>
+          <div className="rounded-b-sm bg-hsl-neutral-blue">
+            TODO: Implement
+          </div>
+        </Visible>
+      </div>
+      <div className="pt-10">
+        <Row className="justify-end space-x-4">
+          <SimpleButton
+            inverted
+            testId={testIds.cancelButton}
+            href={Path.timetablesImport}
+          >
+            {t('cancel')}
+          </SimpleButton>
+          <SimpleButton
+            testId={testIds.saveButton}
+            onClick={onSave}
+            disabled={!importedTimetablesExist}
+          >
+            {t('timetablesPreview.save')}
+          </SimpleButton>
+        </Row>
+      </div>
+    </Container>
+  );
+};
--- a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
+++ b/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
@@ -1,7 +1,8 @@
 import { zodResolver } from '@hookform/resolvers/zod';
-import { useReducer, useRef } from 'react';
+import { useRef } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
+import { useToggle } from '../../../hooks';
 import { useConfirmTimetablesImport } from '../../../hooks/timetables-import/useConfirmTimetablesImport';
 import { Container, Row, Visible } from '../../../layoutComponents';
 import { Path } from '../../../router/routeDetails';
@@ -13,10 +14,11 @@ import {
   PriorityFormState,
   priorityFormSchema,
 } from '../../forms/common';
+import { ImportContentsView } from './ImportContentsView';
 
 const schema = priorityFormSchema;
 
-export type FormState = PriorityFormState;
+type FormState = PriorityFormState;
 
 const testIds = {
   toggleShowStagingTimetables:
@@ -28,13 +30,11 @@ const testIds = {
 export const PreviewTimetablesPage = (): JSX.Element => {
   const { t } = useTranslation();
 
-  const { vehicleJourneyCount, confirmTimetablesImport } =
+  const { confirmTimetablesImport, vehicleJourneys, vehicleScheduleFrames } =
     useConfirmTimetablesImport();
-  const [showStagingTimetables, toggleShowStagingTimetables] = useReducer(
-    (value) => !value,
-    false,
-  );
+  const [showStagingTimetables, toggleShowStagingTimetables] = useToggle(true);
 
+  const vehicleJourneyCount = vehicleJourneys?.length || 0;
   const importedTimetablesExist = vehicleJourneyCount > 0;
 
   const formRef = useRef<ExplicitAny>(null);
@@ -63,7 +63,9 @@ export const PreviewTimetablesPage = (): JSX.Element => {
         <Row className="justify-between rounded-t-sm border-brand bg-brand pl-16 pr-8 text-white">
           <div className="py-2">
             <h2>
-              {t('timetablesPreview.departures')} {vehicleJourneyCount}
+              {t('timetablesPreview.departures', {
+                count: vehicleJourneyCount,
+              })}
             </h2>
           </div>
           <div className="flex items-center">
@@ -98,8 +100,8 @@ export const PreviewTimetablesPage = (): JSX.Element => {
           </FormProvider>
         </Row>
         <Visible visible={showStagingTimetables}>
-          <div className="rounded-b-sm bg-hsl-neutral-blue">
-            TODO: Implement
+          <div className="items-center space-x-14 rounded-b-sm bg-hsl-neutral-blue py-9 px-16">
+            <ImportContentsView vehicleScheduleFrames={vehicleScheduleFrames} />
           </div>
         </Visible>
       </div>
--- a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
+++ b/ui/src/components/timetables/import/PreviewTimetablesPage.tsx
@@ -2,11 +2,14 @@ import { zodResolver } from '@hookform/resolvers/zod';
 import { useRef } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
-import { useToggle } from '../../../hooks';
-import { useConfirmTimetablesImport } from '../../../hooks/timetables-import/useConfirmTimetablesImport';
+import { useHistory } from 'react-router';
+import {
+  useConfirmTimetablesImport,
+  useConfirmTimetablesImportUIAction,
+  useToggle,
+} from '../../../hooks';
 import { Container, Row, Visible } from '../../../layoutComponents';
-import { Path } from '../../../router/routeDetails';
-import { TimetablePriority } from '../../../types/Priority';
+import { Path, routeDetails } from '../../../router/routeDetails';
 import { AccordionButton, SimpleButton } from '../../../uiComponents';
 import { submitFormByRef } from '../../../utils';
 import {
@@ -29,10 +32,12 @@ const testIds = {
 
 export const PreviewTimetablesPage = (): JSX.Element => {
   const { t } = useTranslation();
+  const history = useHistory();
 
-  const { confirmTimetablesImport, vehicleJourneys, vehicleScheduleFrames } =
+  const { vehicleJourneys, vehicleScheduleFrames } =
     useConfirmTimetablesImport();
   const [showStagingTimetables, toggleShowStagingTimetables] = useToggle(true);
+  const { onConfirmTimetablesImport } = useConfirmTimetablesImportUIAction();
 
   const vehicleJourneyCount = vehicleJourneys?.length || 0;
   const importedTimetablesExist = vehicleJourneyCount > 0;
@@ -51,9 +56,11 @@ export const PreviewTimetablesPage = (): JSX.Element => {
   };
 
   const onSubmit = async (state: FormState) => {
-    await confirmTimetablesImport(
-      state.priority as unknown as TimetablePriority,
-    );
+    await onConfirmTimetablesImport(state.priority);
+
+    history.push({
+      pathname: routeDetails[Path.timetablesImport].getLink(),
+    });
   };
 
   return (
--- a/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx
+++ b/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx
@@ -1,0 +1,64 @@
+import { useTranslation } from 'react-i18next';
+import { VehicleScheduleFrameWithRouteInfoFragment } from '../../../generated/graphql';
+import { useToggle } from '../../../hooks';
+import { Row, Visible } from '../../../layoutComponents';
+import { AccordionButton } from '../../../uiComponents';
+import { BlockVehicleJourneysTable } from './BlockVehicleJourneysTable';
+
+const testIds = {
+  table: 'VehicleScheduleFrameBlocksView::table',
+  toggleShowTable: 'VehicleScheduleFrameBlocksView::toggleShowTable',
+};
+
+interface Props {
+  vehicleScheduleFrame: VehicleScheduleFrameWithRouteInfoFragment;
+}
+
+export const VehicleScheduleFrameBlocksView = ({
+  vehicleScheduleFrame,
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+  const [isOpen, toggleIsOpen] = useToggle(true);
+
+  const blocks = vehicleScheduleFrame.vehicle_services.flatMap((service) =>
+    // TODO: Get Block label from vehicle service after it has been implemented to data model
+    service.blocks.map((block) => ({ label: '!Block label', block })),
+  );
+
+  return (
+    <div>
+      <Row className="border-brand-gray w-full border text-white">
+        <p className="flex w-1/6 items-center border border-light-grey bg-brand py-2 px-6 font-bold">
+          !Aikataulun nimi
+        </p>
+        <Row className="flex-1 border border-l border-light-grey border-l-white bg-brand py-2 px-4">
+          <Row className="flex-1 items-center justify-between font-normal">
+            <p>
+              {t('timetablesPreview.blockCount', {
+                count: blocks.length,
+              })}
+            </p>
+            <AccordionButton
+              testId={testIds.toggleShowTable}
+              isOpen={isOpen}
+              onToggle={toggleIsOpen}
+              iconClassName="text-white text-[50px]"
+            />
+          </Row>
+        </Row>
+      </Row>
+
+      <Visible visible={isOpen}>
+        <div className="my-4 space-y-4">
+          {blocks.map((block) => (
+            <BlockVehicleJourneysTable
+              key={block.block.block_id}
+              title={block.label}
+              vehicleJourneys={block.block.vehicle_journeys}
+            />
+          ))}
+        </div>
+      </Visible>
+    </div>
+  );
+};
--- a/ui/src/components/timetables/vehicle-schedule-details/VehicleRouteTimetableSection.tsx
+++ b/ui/src/components/timetables/vehicle-schedule-details/VehicleRouteTimetableSection.tsx
@@ -1,7 +1,6 @@
 import { useTranslation } from 'react-i18next';
 import { RouteRoute } from '../../../generated/graphql';
-import { useGetTimetables } from '../../../hooks';
-import { useToggle } from '../../../hooks/useToggle';
+import { useGetTimetables, useToggle } from '../../../hooks';
 import { Row, Visible } from '../../../layoutComponents';
 import { mapToShortDate } from '../../../time';
 import { AccordionButton } from '../../../uiComponents';
--- a/ui/src/generated/graphql.tsx
+++ b/ui/src/generated/graphql.tsx
@@ -16151,6 +16151,50 @@ export type GetScheduledStopPointsInJourneyPatternsUsedAsTimingPointsQuery = {
   }>;
 };
 
+export type VehicleJourneyWithRouteInfoFragment = {
+  __typename?: 'timetables_vehicle_journey_vehicle_journey';
+  start_time: luxon.Duration;
+  end_time: luxon.Duration;
+  vehicle_journey_id: UUID;
+  journey_pattern_ref: {
+    __typename?: 'timetables_journey_pattern_journey_pattern_ref';
+    journey_pattern_ref_id: UUID;
+    journey_pattern_instance?: {
+      __typename?: 'journey_pattern_journey_pattern';
+      journey_pattern_id: UUID;
+      journey_pattern_route?: {
+        __typename?: 'route_route';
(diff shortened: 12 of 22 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 17 files reviewed, 23 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/common/Loader.tsx` line 5 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCcIb5MlB4bIkVYPR:-NNCCcIb5MlB4bIkVYPS:b-14ld46) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L5)):*
> ```XML
> import { LoadingOverlay } from '../../uiComponents';
> 
> export const Loader = () => {
> ```

Should be typed as JSX.Element
___
*[`ui/src/components/common/Loader.tsx` line 6 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCgp69oi7fQdd2rw3:-NNCCgp69oi7fQdd2rw4:by626y2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L6)):*
> ```XML
> 
> export const Loader = () => {
>   const testIds = {
> ```

Should be defined outside of the component. Now it is impossible to e.g. import these to test file if needed.

By the way, is this even used anywhere?
___
*[`ui/src/components/common/Loader.tsx` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCsyw4YuLMkCpW2gm:-NNCCsyw4YuLMkCpW2gn:b-s45c8y) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L10)):*
> ```XML
>   };
>   const { isMapOpen } = useMapQueryParams();
>   const isLoading = useAppSelector(selectIsOperationLoading);
> ```

"operationLoading" sounds a bit vague :thinking: 
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 9 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCDib39NvoMIKl69-1:-NNCDib39NvoMIKl69-2:b-sz8k6e) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L9)):*
> ```XML
> import { DirectionBadge } from '../../routes-and-lines/line-details/DirectionBadge';
> 
> interface Props {
> ```

Please define this right before the component that actually used this.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 15 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCDPV20gqfQntQoW9J:-NNCDPV20gqfQntQoW9K:bkdlrkh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L15)):*
> ```XML
> 
> const testIds = {
>   table: 'RouteVehicleJourneysTable::table',
> ```

BlockVehicleJourneysTable. Same for next row.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 19 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCD_tgEsSnj0Y4fmYX:-NNCD_tgEsSnj0Y4fmYY:b28bem6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L19)):*
> ```XML
> };
> 
> const Td: React.FC = ({ children }) => (
> ```

No need to define this as own component, just extract `tdClassName` to own variable.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 28 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCDtdB05rLR6cLRA6N:-NNCDtdB05rLR6cLRA6O:b-q8cll8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L28)):*
> ```XML
> }
> 
> const Th: React.FC<ThProps> = ({ children, colSpan = 1, className = '' }) => (
> ```

No need to define this as own component. Now this just adds layer of complexity to the code, without making it any clearer.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 34 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCE7mc14vxdpDsN0gU:-NNCE7mc14vxdpDsN0gV:b-14ld46) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L34)):*
> ```XML
> );
> 
> const VehicleJourneyRow = ({
> ```

Should be typed as JSX.Element
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 44 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCEAzzAsZpUJUykgIN:-NNCEAzzAsZpUJUykgIO:b-4ttulp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L44)):*
> ```XML
> 
>   if (!route) {
>     throw new Error('No route found for vehicle journey!');
> ```

Don't throw from UI component unless defining error boundary. I don't know how this situation should be handled from user's perspective, but crashing the UI is not the way to go.

Rather e.g.:
- log error to console / show error toast
- render error message?
- render empty component?
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 60 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFBDDAmG8piUneXST:-NNCFBDDAmG8piUneXSU:b-qkgrb6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L60)):*
> ```XML
>       <Td>{vehicleJourney.block.vehicle_service.day_type.name_i18n.fi_FI}</Td>
>       <Td>{mapDurationToShortTime(vehicleJourney.start_time)}</Td>
>       <Td>{mapDurationToShortTime(vehicleJourney.end_time)}</Td>
> ```

Did you checked what happens if either one of these is undefined? Just wondering if these should have `|| ''`



_Code quote:_
```XML
      <Td>{mapDurationToShortTime(vehicleJourney.start_time)}</Td>
      <Td>{mapDurationToShortTime(vehicleJourney.end_time)}</Td>
```
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 69 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFKD034M2gvR5lmzd:-NNCFKD16A61S6HMu6-v:bc95pyc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L69)):*
> ```XML
>   title,
> }: Props): JSX.Element => {
>   const [isOpen, setIsOpen] = useReducer((value) => !value, true);
> ```

We have useToggle component hook for this: `ui/src/hooks/useToggle.ts`
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 78 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFVqrECdikZO3CRYN:-NNCFVqrECdikZO3CRYO:b-v2fzfq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L78)):*
> ```XML
>       <thead>
>         <tr>
>           <Th className="w-64 bg-brand py-2 px-6 text-left text-white">
> ```

Do we have to use explicit widht? This won't work well on small screens.
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 34 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFm5q77FA8E52Wc7M:-NNCFm5q77FA8E52Wc7N:bvxgjzp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L34)):*
> ```XML
>     try {
>       await confirmTimetablesImport(
>         state.priority as unknown as TimetablePriority,
> ```

Not part of this pr so this is optional commment, but why do we need as unknown casting here? Could we fix typings in `state`?
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 23 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCHI7EBtGJwvXk3999:-NNCHI7EBtGJwvXk399A:bb76va3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L23)):*
> ```XML
> const schema = priorityFormSchema;
> 
> export type FormState = PriorityFormState;
> ```

Wtf?
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 38 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCHSktExij3ApWbLwj:-NNCHSktExij3ApWbLwk:b-muwtka) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L38)):*
> ```XML
>   const { confirmTimetablesImport, vehicleJourneys, vehicleScheduleFrames } =
>     useConfirmTimetablesImport();
>   const [showStagingTimetables, toggleShowStagingTimetables] = useReducer(
> ```

useToggle
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 12 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIBiK2hr2k8uhpEcP:-NNCIBiK2hr2k8uhpEcQ:b-aq8h1w) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L12)):*
> ```XML
> }
> 
> const testIds = {
> ```

Please define this before Props definition. Props are logically tied to the component definition and they are used directly on the first lines of component definition, so it makes sense to couple those by aligning those next to each other.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 21 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIRjF7XDgfFgXod6a:-NNCIRjF7XDgfFgXod6b:b-muwtka) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L21)):*
> ```XML
> }: Props): JSX.Element => {
>   const [isOpen, setIsOpen] = useReducer((value) => !value, true);
> ```

useToggle
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 31 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIWeNDbLsWxwujkh1:-NNCIWeNDbLsWxwujkh2:b-vbcvhr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L31)):*
> ```XML
>     <div>
>       <Row className="border-brand-gray w-full border text-white">
>         <Row className="w-64 items-center border border-light-grey bg-brand py-2 px-6 font-bold">
> ```

Do we really have to define explicit width?
By the way, please use semantic tag like `p` for text content. If nothing else, it improves accessibility.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 36 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIiq51MxSLk6TpLNH:-NNCIiq51MxSLk6TpLNI:b61b0rf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L36)):*
> ```XML
>         <Row className="flex-1 border border-l border-light-grey border-l-white bg-brand py-2 px-4">
>           <Row className="flex-1 items-center justify-between font-normal">
>             {t('timetablesPreview.blockCount', {
> ```

Please use semantic tag for text content.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 43 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCJ6zm1_IL2jspQzsc:-NNCJ6zm1_IL2jspQzsd:b-q6ok3x) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L43)):*
> ```XML
>               isOpen={isOpen}
>               onToggle={setIsOpen}
>               iconClassName="text-white text-[50px]"
> ```

I think this is third time in this PR I'm seeing this text-[50px] definition, which is usually intented for one-off fixes. I would consider defining component for this.
___
*[`ui/src/locales/en-US/common.json` line 301 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCGJti-5YXs0NIM_BT:-NNCGJti-5YXs0NIM_BU:bp0929z) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/locales/en-US/common.json#L301)):*
> ```JSON
>     "preview": "Preview",
>     "save": "Save timetables",
>     "departures": "Departures:",
> ```

Departures: {{ count }}
___
*[`ui/src/locales/en-US/common.json` line 304 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCGG32A4Nb3OqW2AZk:-NNCGG32A4Nb3OqW2AZl:bgfwysy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/locales/en-US/common.json#L304)):*
> ```JSON
>     "showContent": "Show content",
>     "closeContent": "Close content",
>     "contentUsage": "Contant usage",
> ```

typo, "Contant"
___
*[`ui/src/redux/selectors/index.ts` line 111 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCGiozCw9zAVE3R14p:-NNCGiozCw9zAVE3R14q:bqqbyz1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/redux/selectors/index.ts#L111)):*
> ```XML
>   selectLoader,
>   (loaders) =>
>     Object.entries(loaders)
> ```

Seems like you need only `key` from Object.entries, so Object.keys should be used instead.
Seems like this file has also `selectIsMapOperationLoading` implemented like this, could you please refactor it aswell?

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 22 files reviewed, 23 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/common/Loader.tsx` line 5 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCcIb5MlB4bIkVYPR:-NNGBlus6j5UzumJMR91:b7ie0vg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L5)):*

Should be typed as JSX.Element
</blockquote></details>

This could be added to linter rules if we want to enforce it. I fixed this to another component as well.
___
*[`ui/src/components/common/Loader.tsx` line 6 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCgp69oi7fQdd2rw3:-NNGBtHrDeDQZx80jySx:b-bqy3ux) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L6)):*

Should be defined outside of the component. Now it is impossible to e.g. import these to test file if needed.

By the way, is this even used anywhere?
</blockquote></details>

Done. It is used
___
*[`ui/src/components/common/Loader.tsx` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCsyw4YuLMkCpW2gm:-NNGC-GfCYnFpQjG1wYz:bco4zjy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L10)):*

"operationLoading" sounds a bit vague :thinking: 
</blockquote></details>

It is vague but I just cannot figure out what to name some operation that is happening outside map. We have `mapOperationLoading`, but do we want `notMapOperationLoading` :D, or do we want something like `listViewOperationLoading`...? I just cannot figure it out..
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 9 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCDib39NvoMIKl69-1:-NNGCuMw8Xniid9F7Bdi:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L9)):*

Please define this right before the component that actually used this.
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 15 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCDPV20gqfQntQoW9J:-NNGChuD3_SXR2iB0Exq:b-oq6h9b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L15)):*

BlockVehicleJourneysTable. Same for next row.
</blockquote></details>

Good spot!
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 19 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCD_tgEsSnj0Y4fmYX:-NNGCwV37VzZItSs1rpz:b-8fbzkm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L19)):*

No need to define this as own component, just extract `tdClassName` to own variable.
</blockquote></details>

I followed a pattern used earlier, but as discussed this will be fixed when new version of tailwindcss is available.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 28 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCDtdB05rLR6cLRA6N:-NNGD65A968B-ccLBUk7:b-kuzsez) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L28)):*

No need to define this as own component. Now this just adds layer of complexity to the code, without making it any clearer.
</blockquote></details>

Same as above. Will be fixed when tailwindcss new version is available
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 34 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCE7mc14vxdpDsN0gU:-NNGDL_p4UZtMHrbqt0G:b-zddnt3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L34)):*

Should be typed as JSX.Element
</blockquote></details>

Done. Can you make sure we get linter rule from this? There is probably a few more places in the code base where this is not followed.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 44 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCEAzzAsZpUJUykgIN:-NNGDW274y6JPl1o7-w-:bef7lwr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L44)):*

Don't throw from UI component unless defining error boundary. I don't know how this situation should be handled from user's perspective, but crashing the UI is not the way to go.

Rather e.g.:
- log error to console / show error toast
- render error message?
- render empty component?
</blockquote></details>

This should never happen if we don't have a bug in the code, it is pretty much just a guard for ts to know route is defined. Fixed it a little bit though.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 60 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFBDDAmG8piUneXST:-NNGG-DdCi2Uq8ZueokM:b-jvn20g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L60)):*

Did you checked what happens if either one of these is undefined? Just wondering if these should have `|| ''`
</blockquote></details>

Good point. Added empty strings for safety.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 69 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFKD034M2gvR5lmzd:-NNGG5JAExw-9qtaiKPv:bhsxype) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L69)):*

We have useToggle component hook for this: `ui/src/hooks/useToggle.ts`
</blockquote></details>

Yes :) I can use it then, even though I think this one liner is not that bad either :)
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 78 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFVqrECdikZO3CRYN:-NNGGhwfAr_-BJGTwy-m:b-twmye2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L78)):*

Do we have to use explicit widht? This won't work well on small screens.
</blockquote></details>

Changed it to relative width and it seems to work ok as well.
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 34 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFm5q77FA8E52Wc7M:-NNGH4Oe2x88OhMpz0sz:b-61qjkt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L34)):*

Not part of this pr so this is optional commment, but why do we need as unknown casting here? Could we fix typings in `state`?
</blockquote></details>

"Conversion of type 'Priority' to type 'TimetablePriority' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first."

I'm not sure how to fix this. I don't think enums can extend each other. TS does not realize `TimetablePriority` contains all possible values of `Priority`.
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 23 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCHI7EBtGJwvXk3999:-NNGHPzU3thYskxkpU0e:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L23)):*

Wtf?
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 38 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCHSktExij3ApWbLwj:-NNGHVIZ6JHnu41EMaA3:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L38)):*

useToggle
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 12 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIBiK2hr2k8uhpEcP:-NNGH_nb5CpcYfJ_ulgp:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L12)):*

Please define this before Props definition. Props are logically tied to the component definition and they are used directly on the first lines of component definition, so it makes sense to couple those by aligning those next to each other.
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 21 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIRjF7XDgfFgXod6a:-NNGHoBaAM3X2_r_73FA:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L21)):*

useToggle
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 31 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIWeNDbLsWxwujkh1:-NNGI5hW4C4QzR-CvgiI:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L31)):*

Do we really have to define explicit width?
By the way, please use semantic tag like `p` for text content. If nothing else, it improves accessibility.
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 36 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIiq51MxSLk6TpLNH:-NNGI6jR2Dl3CUcCQJT6:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L36)):*

Please use semantic tag for text content.
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 43 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCJ6zm1_IL2jspQzsc:-NNGIAr4EKVH-bU2zRNR:b26ffds) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L43)):*

I think this is third time in this PR I'm seeing this text-[50px] definition, which is usually intented for one-off fixes. I would consider defining component for this.
</blockquote></details>

What would the component contain? I just want the icon to be 50px high, there is nothing else that is special with that component. I don't see why this should be only considered as one-off fix.
___
*[`ui/src/locales/en-US/common.json` line 301 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCGJti-5YXs0NIM_BT:-NNGLdIB6vwb0hkW9ZDD:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/locales/en-US/common.json#L301)):*

Departures: {{ count }}
</blockquote></details>

Done.
___
*[`ui/src/locales/en-US/common.json` line 304 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCGG32A4Nb3OqW2AZk:-NNGLfFg2Ms2oSjH31lH:b-oq6h9b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/locales/en-US/common.json#L304)):*

typo, "Contant"
</blockquote></details>

Good spot!
___
*[`ui/src/redux/selectors/index.ts` line 111 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCGiozCw9zAVE3R14p:-NNGM8SsEKh311WK1Hc2:b-1sr3rb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/redux/selectors/index.ts#L111)):*

Seems like you need only `key` from Object.entries, so Object.keys should be used instead.
Seems like this file has also `selectIsMapOperationLoading` implemented like this, could you please refactor it aswell?
</blockquote></details>

Hmm, I do need both of them.. :thinking:

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 22 files reviewed, 11 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/common/Loader.tsx` line 5 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCcIb5MlB4bIkVYPR:-NNLQHlH8-Ap-a4I__WH:bxxdu0p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L5)):*

This could be added to linter rules if we want to enforce it. I fixed this to another component as well.
</blockquote></details>

That would indeed be useful, but haven't seen suitable eslint rule for that and didn't found one either when tried briefly to find rule for that.
___
*[`ui/src/components/common/Loader.tsx` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCsyw4YuLMkCpW2gm:-NNLcugg4hn4KTbku-Qa:bym1d16) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L10)):*

It is vague but I just cannot figure out what to name some operation that is happening outside map. We have `mapOperationLoading`, but do we want `notMapOperationLoading` :D, or do we want something like `listViewOperationLoading`...? I just cannot figure it out..
</blockquote></details>

Umm... When I dug a bit deeper onto this, I don't see any real benefits of separating MapLoader and Loader components just now. I mean, their state in redux is stored in literally same slice, and the loader components are ~identical but one is shown when map is open and another when map is not open.

So OK, this name could be fine but then we probably don't need separate MapLoader concept. Or what do you think about this?
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 19 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCD_tgEsSnj0Y4fmYX:-NNLccMn9lnj6AEGb7Vs:by36q0b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L19)):*

I followed a pattern used earlier, but as discussed this will be fixed when new version of tailwindcss is available.
</blockquote></details>

I just approved PR for tailwindcss update, so I assume it is in main branch when you are reading this.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 69 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFKD034M2gvR5lmzd:-NNLRw5g-pyEPWuGfqP5:b12njm1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L69)):*

Yes :) I can use it then, even though I think this one liner is not that bad either :)
</blockquote></details>

useToggle(true)
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 98 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLeKvAB11EXRlEpgLR:-NNLeKvAB11EXRlEpgLS:b302mpo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L98)):*
> ```XML
>           >
>             <Row className="flex-1 items-center justify-between font-normal">
>               <div>!Kalustotyyppi 3 - Normaalibussi</div>
> ```

Why div instead of p?
If font-normal style is not intented to be shared across multiple elements, it could be defined directly for this tag. It is easier to understand styles if they are defined directly to the component that they are supposed to affect to. 
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 30 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLgZoX3n0Mph4Zwz07:-NNLgZoX3n0Mph4Zwz08:b-fpwzvz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L30)):*
> ```XML
> 
>   const onSave = async (state: FormState) => {
>     setIsLoading(true);
> ```

Seems like this is intented for handling redux state during api call. Why this is not done in `confirmtTimetablesImport` hook?
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 36 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLfz5t2c2f0Q-oHiQr:-NNLfz5t2c2f0Q-oHiQs:b-h2nxzp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L36)):*
> ```XML
>         state.priority as unknown as TimetablePriority,
>       );
>       showSuccessToast(t('timetables.importSuccess'));
> ```

Should `confirmTimetablesImport` rather handle showing the success toast? If it is something that we always want to do after saving the changes, it would make more sense to define it there. Also it shouldn't be responsibility of UI component.
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 50 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLh97TEC8aowuCVAZ6:-NNLh97TEC8aowuCVAZ7:bv4a7ih) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L50)):*
> ```XML
>         state.priority as unknown as TimetablePriority,
>       );
>       showSuccessToast(t('timetables.importSuccess'));
> ```

Ok, this is second time UI component is handling this after calling `confirmTimetablesImport`, so this definitely belongs to the `confirmTimetablesImport` itself. It could most likely handle setting isLoading flag too.
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 78 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLhSYJ5mqRuXZ2RuHK:-NNLhSYJ5mqRuXZ2RuHL:b8r9v88) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L78)):*
> ```XML
>               className="font-bold"
>             >
>               {showStagingTimetables
> ```

Semantic tag for text content
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 21 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIRjF7XDgfFgXod6a:-NNLRilu65lHGUutcvpx:b-lrs5n2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L21)):*

Done.
</blockquote></details>

Now you also changed initial state from true tof alse. `useToggle(true)`
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 43 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCJ6zm1_IL2jspQzsc:-NNLhd8j5tf119zHyE67:b9b2np3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L43)):*

What would the component contain? I just want the icon to be 50px high, there is nothing else that is special with that component. I don't see why this should be only considered as one-off fix.
</blockquote></details>

Well, this sounds kind of bad for maintainability because one would assume that if font size in one of these is changed, the others should be changed too. But OK, this is fine for me.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 55 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNL_kLZ6x2TzTCYyi4d:-NNL_kL_5EmYFnTyhyhr:b-pqep72) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L55)):*
> ```XML
>           {blocks.map((block) => (
>             <BlockVehicleJourneysTable
>               key={block.label}
> ```

Are these guaranteed to be unique? At least I got key uniqueness violation error when testing locally. Was that only because of limitations of seed data, or could this be replaced e.g. with `block.block.block_id`?
___
*[`ui/src/hooks/index.ts` line 16 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLSD44C58zszAN-XDD:-NNLSD44C58zszAN-XDE:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/hooks/index.ts#L16)):*
> ```XML
> export * from './usePagination';
> export * from './useShowRoutesOnModal';
> export * from './useToggle';
> ```

:+1:

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 22 files reviewed, 10 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/common/Loader.tsx` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCsyw4YuLMkCpW2gm:-NNf0rYP7gMspQHYo31e:b7u77tc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L10)):*

Umm... When I dug a bit deeper onto this, I don't see any real benefits of separating MapLoader and Loader components just now. I mean, their state in redux is stored in literally same slice, and the loader components are ~identical but one is shown when map is open and another when map is not open.

So OK, this name could be fine but then we probably don't need separate MapLoader concept. Or what do you think about this?
</blockquote></details>

When the loader was separated for map and other views, the assumption was made that these components would differ from each other even visually, probably different position or something. Now that I implemented the first loader for non-map-operation, I was lazy and just copy pasted this. I think this is something that should be checked with design and then act according to that.

If we realize they are the same, there probably is no reason to keep them separate. I'm just wondering if e.g. something done in map view could trigger loading event in underlying other view and then the loading spinner would be shown with no reason. E.g. if we change observation date in map, the query is changed, would e.g. list view start loading new routes also in the background showing the spinner with no reason over map view.

I don't know if that made any sense explained in textual format, I can elaborate in some other forum.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 19 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCD_tgEsSnj0Y4fmYX:-NNf8dcsCQO1mvXfw5dl:b-2s1tgq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L19)):*

I just approved PR for tailwindcss update, so I assume it is in main branch when you are reading this.
</blockquote></details>

Ahh it is so nice to get rid of these.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 69 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCFKD034M2gvR5lmzd:-NNf92V0EEiGaIijV3gT:b-48szhd) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L69)):*

useToggle(true)
</blockquote></details>

Now that I think about this, I'm not sure if we want to open these by default. There could be quite a lot of these when the view is opened, and a lot of scrolling would be needed. Maybe default value false is ok after all.
___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 98 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLeKvAB11EXRlEpgLR:-NNf9Q2q1kQ-GRJwS8gw:bikx91h) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L98)):*

Why div instead of p?
If font-normal style is not intented to be shared across multiple elements, it could be defined directly for this tag. It is easier to understand styles if they are defined directly to the component that they are supposed to affect to. 
</blockquote></details>

I think there could be more content added in here, e.g. day type.
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 30 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLgZoX3n0Mph4Zwz07:-NNfATbk39WTsgM1YH5n:b-dtufp7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L30)):*

Seems like this is intented for handling redux state during api call. Why this is not done in `confirmtTimetablesImport` hook?
</blockquote></details>

I kind of disagree. I don't think we want to mix UI state (showing loader) with the business logic that the hook is for. At least for now, loading state and success toasts are something that UI components are responsible for in all other places in our codebase, not the business logic hooks.
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 36 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLfz5t2c2f0Q-oHiQr:-NNfAkffBHjwTVVEw2wG:bsse9xe) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L36)):*

Should `confirmTimetablesImport` rather handle showing the success toast? If it is something that we always want to do after saving the changes, it would make more sense to define it there. Also it shouldn't be responsibility of UI component.
</blockquote></details>

I kind of disagree. I don't think we want to mix UI things with the business logic that the hook is for. At least for now, loading state and success toasts are something that UI components are responsible for in all other places in our codebase, not the business logic hooks.
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 50 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLh97TEC8aowuCVAZ6:-NNfAwWc2x-ge0dqMe3q:b-k7nfm1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L50)):*

Ok, this is second time UI component is handling this after calling `confirmTimetablesImport`, so this definitely belongs to the `confirmTimetablesImport` itself. It could most likely handle setting isLoading flag too.
</blockquote></details>

I kind of disagree. I don't think we want to mix UI things with the business logic that the hook is for. At least for now, loading state and success toasts are something that UI components are responsible for in all other places in our codebase, not the business logic hooks.

I agree that we do this similar thing multiple times and it seems a little stupid, but what if at some point `confirmTimetablesImport` is called in some other context where we don't want to show the loader or success toast?
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 78 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLhSYJ5mqRuXZ2RuHK:-NNfCd_Y6NESBPEqyjr9:b86ohs4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L78)):*

Semantic tag for text content
</blockquote></details>

But this is a button :thinking: , do we want to put `<p>` inside a button?
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 21 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCIRjF7XDgfFgXod6a:-NNfD-0pBMLpmKihb2B3:b3zz7ua) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L21)):*

Now you also changed initial state from true tof alse. `useToggle(true)`
</blockquote></details>

As explained above, maybe it is better this way, otherwise the point of the accordions is kind of nothing.
___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 55 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNL_kLZ6x2TzTCYyi4d:-NNfDIEI3yxHhtFpcIHW:bwq6fog) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L55)):*

Are these guaranteed to be unique? At least I got key uniqueness violation error when testing locally. Was that only because of limitations of seed data, or could this be replaced e.g. with `block.block.block_id`?
</blockquote></details>

Hmm the reason you are getting this is probably because of seed data limitations, but now that I think about this there could be multiple blocks with the same label if they are grouped in the same vehicle task. I changed it. :+1:

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 22 files reviewed, 6 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 42 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNkYTdO00M_NkQZo3tS:-NNkYTdO00M_NkQZo3tT:bvk8k4p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/f5b5f5abbd588e5627bba183977ac9645d2d110d/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L42)):*
> ```XML
>         </Row>
>       </td>
>       <td>{vehicleJourney.block.vehicle_service.day_type.name_i18n.fi_FI}</td>
> ```

See `ui/src/i18n/utils.ts` and use `parseI18nField` from it to show correct language version instead of hardcoding to fi_Fi.
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 30 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLgZoX3n0Mph4Zwz07:-NNkdFdMDkedOvdvhVRw:bnag14r) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L30)):*

I kind of disagree. I don't think we want to mix UI state (showing loader) with the business logic that the hook is for. At least for now, loading state and success toasts are something that UI components are responsible for in all other places in our codebase, not the business logic hooks.
</blockquote></details>

Replied to discussion below, but keeping this open until the discussion is finished as it concerns also this place
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 36 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLfz5t2c2f0Q-oHiQr:-NNkdM-LDuUK6AyhU717:bnag14r) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L36)):*

I kind of disagree. I don't think we want to mix UI things with the business logic that the hook is for. At least for now, loading state and success toasts are something that UI components are responsible for in all other places in our codebase, not the business logic hooks.
</blockquote></details>

Replied to discussion below, but keeping this open until the discussion is finished as it concerns also this place
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 50 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLh97TEC8aowuCVAZ6:-NNk_CQt9f42ncLBWzT2:bvz3v93) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L50)):*
>but what if at some point confirmTimetablesImport is called in some other context where we don't want to show the loader or success toast?

~Why we would want to be inconsistent with giving feedback to the user, especially when doing undoable operation like this?~

OK, so far we have kept all showSuccessToast calls as responsibility of UI components itself, so keeping it here is OK for consistency, but IMO it just doesn't make sense.

But regarding loading state, it makes even less sense. We are setting loading state to redux, instead of local state of this component, and we don't even utilize the loading state in this component, so why it should be done here? It just bloats this component up, an which is even worse, in this case this duplicates almost same logic in two different components.

If we want to separate this kind of stuff from the hook which makes the api call, then we could intorude yet another layer of hooks which would combine
- hook for doing api call
- hook for setting loading state
- hook for showing toast message 

That option would
- keep api hook "pure" from this kind of side effects so that it could still be used in the unlikely case were we don't want toast messages or spinners
- keep unrelated logics out of ui components
- improve maintanability by reducing code duplciation
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 78 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLhSYJ5mqRuXZ2RuHK:-NNk_YtEBRQXX7t2l9bS:b-3aqhqk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L78)):*

But this is a button :thinking: , do we want to put `<p>` inside a button?
</blockquote></details>

Sorry, missed the button tag for some reason. Yeah, no other tags inside button.

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 24 files reviewed, 6 unresolved discussions (waiting on @tk-vincit)

___
*[`ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx` line 42 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNkYTdO00M_NkQZo3tS:-NNlNGtz8P42XHVIB8B6:b7zy2ce) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/f5b5f5abbd588e5627bba183977ac9645d2d110d/ui/src/components/timetables/import/BlockVehicleJourneysTable.tsx#L42)):*

See `ui/src/i18n/utils.ts` and use `parseI18nField` from it to show correct language version instead of hardcoding to fi_Fi.
</blockquote></details>

Thanks! Didn't really know of this.
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 30 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLgZoX3n0Mph4Zwz07:-NNlTOvt-sJR9520-w9s:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L30)):*

Replied to discussion below, but keeping this open until the discussion is finished as it concerns also this place
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx` line 36 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLfz5t2c2f0Q-oHiQr:-NNlTPHq5ahVqJYYxD8A:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/ConfirmTimetablesImportModal.tsx#L36)):*

Replied to discussion below, but keeping this open until the discussion is finished as it concerns also this place
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 50 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLh97TEC8aowuCVAZ6:-NNlTROB2BHo09wmNEEo:b-jligct) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L50)):*

>but what if at some point confirmTimetablesImport is called in some other context where we don't want to show the loader or success toast?

~Why we would want to be inconsistent with giving feedback to the user, especially when doing undoable operation like this?~

OK, so far we have kept all showSuccessToast calls as responsibility of UI components itself, so keeping it here is OK for consistency, but IMO it just doesn't make sense.

But regarding loading state, it makes even less sense. We are setting loading state to redux, instead of local state of this component, and we don't even utilize the loading state in this component, so why it should be done here? It just bloats this component up, an which is even worse, in this case this duplicates almost same logic in two different components.

If we want to separate this kind of stuff from the hook which makes the api call, then we could intorude yet another layer of hooks which would combine
- hook for doing api call
- hook for setting loading state
- hook for showing toast message 

That option would
- keep api hook "pure" from this kind of side effects so that it could still be used in the unlikely case were we don't want toast messages or spinners
- keep unrelated logics out of ui components
- improve maintanability by reducing code duplciation
</blockquote></details>

I tried to do something for this. How does it look?

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 24 files reviewed, 4 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/common/Loader.tsx` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCsyw4YuLMkCpW2gm:-NNowFSl6STskIpx6m2_:b-f3gdw6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L10)):*

When the loader was separated for map and other views, the assumption was made that these components would differ from each other even visually, probably different position or something. Now that I implemented the first loader for non-map-operation, I was lazy and just copy pasted this. I think this is something that should be checked with design and then act according to that.

If we realize they are the same, there probably is no reason to keep them separate. I'm just wondering if e.g. something done in map view could trigger loading event in underlying other view and then the loading spinner would be shown with no reason. E.g. if we change observation date in map, the query is changed, would e.g. list view start loading new routes also in the background showing the spinner with no reason over map view.

I don't know if that made any sense explained in textual format, I can elaborate in some other forum.
</blockquote></details>

As discussed in slack, blocking full screen spinner is antipattern and shouldn't be used unless there is _real_ need for it. If I understood correctly, in this case there is no need for that, so this should be removed and we should rather show spinner in disabled "save" button during the operation.
___
*[`ui/src/components/timetables/import/PreviewTimetablesPage.tsx` line 50 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNLh97TEC8aowuCVAZ6:-NNovwoL99ntstmL2FRE:bcs1g51) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ee754f6cf7710154f3360d0d17cf4182c055b98a/ui/src/components/timetables/import/PreviewTimetablesPage.tsx#L50)):*

I tried to do something for this. How does it look?
</blockquote></details>

IMO much better now. :+1: 
___
*[`ui/src/hooks/timetables-import/useTimetablesImportUIActions.ts` line 6 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNouxU44XKupxEu5dQE:-NNouxU44XKupxEu5dQF:b-w8lxno) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77130a85ae0810387ede22ecc391a0e94e85832b/ui/src/hooks/timetables-import/useTimetablesImportUIActions.ts#L6)):*
> ```XML
> import { showSuccessToast } from '../../utils';
> import { useLoader } from '../ui';
> import { useConfirmTimetablesImport } from './useConfirmTimetablesImport';
> ```

The "original" hook file is called "useConfirmTimetablesImport", so it could make sense to name this file according to it.
"UIActions" postfix is OK for me, so the final name could then be "useConfirmTimetablesImportUIAction".

Thumbs up for organizing these to the same folder already, that helps too. :+1: 
___
*[`ui/src/hooks/timetables-import/useTimetablesImportUIActions.ts` line 24 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNotyNh5NMJdgZ_PJrU:-NNotyNh5NMJdgZ_PJrV:b-r1nh30) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77130a85ae0810387ede22ecc391a0e94e85832b/ui/src/hooks/timetables-import/useTimetablesImportUIActions.ts#L24)):*
> ```XML
>   };
> 
>   return { onTimetablesImportConfirm };
> ```

:raised_eyebrow: , "onConfirmTimetablesImport" would sound better. OK, hmm, the name is already used original hook, but still this doesn't make any sense. :) 
Maybe you could either:
- Import it `as` another name `import { useConfirmTimetablesImport as confirmImport } from './useConfirmTimetablesImport';` so that you could re-export it as "onConfirmTimetablesImport"
- Invent new naming scheme for these actions, "onComfirmTimetablesImportWithFeedback" or something.
___
*[`ui/src/hooks/timetables-import/useTimetablesImportUIActions.ts` line 24 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNovsI24S2lw9u6XLO6:-NNovsI24S2lw9u6XLO7:bl515gt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77130a85ae0810387ede22ecc391a0e94e85832b/ui/src/hooks/timetables-import/useTimetablesImportUIActions.ts#L24)):*
> ```XML
>   };
> 
>   return { onTimetablesImportConfirm };
> ```

I could also re-consider the whole "confirm timetables import" naming, I found it very confusing on first time. Should we rather talk about "saving draft timetables" or something like that? "on confirm" sounds like we are showing some kind of confirmation dialog before actually saving something. But as this naming pattern is probably used even before this PR, you can decide that this it out of scope of this PR.

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492)* status: 0 of 24 files reviewed, 3 unresolved discussions (waiting on @eleksi)

___
*[`ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx` line 25 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNpg86426SGHYr5N9Mk:-NNpg86426SGHYr5N9Ml:b4tg1np) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6f90ea9b380c90424b4c1b378c138ac2c1bc9263/ui/src/components/timetables/import/VehicleScheduleFrameBlocksView.tsx#L25)):*
> ```XML
>   const blocks = vehicleScheduleFrame.vehicle_services.flatMap((service) =>
>     // TODO: Get Block label from vehicle service after it has been implemented to data model
>     service.blocks.map((block) => ({ label: '!Block label', block })),
> ```

By the way, did you rather wanted to: `({ label: '...', ...block})`?
___
*[`ui/src/redux/selectors/index.ts` line 108 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNpgUfrAQyySlxkMbN4:-NNpgUfrAQyySlxkMbN5:b9p6es1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6f90ea9b380c90424b4c1b378c138ac2c1bc9263/ui/src/redux/selectors/index.ts#L108)):*
> ```XML
> );
> 
> export const selectIsOperationLoading = createSelector(
> ```

Not sure if we have use case for this anymore.
___
*[`ui/src/components/common/Loader.tsx` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-NNCCsyw4YuLMkCpW2gm:-NNpc4i3BaS4OBzrDZgT:b-bgmhia) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a26b4572c6ae49dbe63660c44e641bfbd7c63684/ui/src/components/common/Loader.tsx#L10)):*

Spinner removed, ticket created.
</blockquote></details>

If this loader is not used anymore, please remove this.

tk-vincit (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/492#-:-NNpnzFF-EKSJmtOk2Bq:b-uwkowd)* status: 0 of 24 files reviewed, all discussions resolved

PR #496, opened 2023-02-02
Add builder utils for easier timetables seed data generation

eleksi:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/496)


eleksi:
We don't want this heavy OOP solution, rather something lighter and closer to GraphQL original schema. This task was moved to backlog, as this PR's code is pretty much useless when creating the new solution, even though the principles might be the same.

https://github.com/HSLdevcom/jore4/issues/1191

PR #501, opened 2023-02-03
Timetable versions page

HenrikHartiala:
Adds the timetable versions page first iteration. This has some logic that will be moved to SQL function and also we need to rework some logic after we get reference days implemented.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "test:e2e:open": "yarn ws:db build && yarn ws:e2e cypress open",
     "test:e2e:list": "yarn ws:e2e print-tests 'e2e/**.cy.ts'",
     "qa": "yarn --frozen-lockfile && yarn ws:db build && yarn ts:check && yarn prettier:check && yarn lint && yarn test && yarn test:integration && yarn test:e2e",
-    "qa:fast": "yarn:qa --config-file cypress.config.fast.ts"
+    "qa:fast": "yarn qa --config-file cypress.config.fast.ts"
   },
   "dependencies": {},
   "devDependencies": {
--- a/ui/src/components/forms/common/ObservationDateInput.tsx
+++ b/ui/src/components/forms/common/ObservationDateInput.tsx
@@ -1,5 +1,6 @@
 import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
+import { Column } from '../../../layoutComponents';
 
 interface Props {
   value: DateTime;
@@ -21,7 +22,7 @@ export const ObservationDateInput = ({
   const { t } = useTranslation();
   const dateInputId = 'observation-date-input';
   return (
-    <>
+    <Column className={className}>
       <label htmlFor={dateInputId}>{t('filters.observationDate')}</label>
       <input
         type="date"
@@ -33,6 +34,6 @@ export const ObservationDateInput = ({
         required={required}
         disabled={disabled}
       />
-    </>
+    </Column>
   );
 };
--- a/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
+++ b/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
@@ -1,19 +1,12 @@
-import qs from 'qs';
 import { useTranslation } from 'react-i18next';
-import { RouteLine } from '../../../generated/graphql';
-import { useGetLineDetails } from '../../../hooks';
+import {
+  useGetLineDetails,
+  useRoutesAndLinesDraftReturnToQueryParam,
+} from '../../../hooks';
 import { Column, Container } from '../../../layoutComponents';
-import { Path, routeDetails } from '../../../router/routeDetails';
 import { SimpleButton } from '../../../uiComponents';
 import { ObservationDateControl } from '../../common/ObservationDateControl';
 import { FormRow } from '../../forms/common';
-
-const getDraftsUrlWithReturnToQueryString = (line: RouteLine) => {
-  const draftUrl = routeDetails[Path.lineDrafts].getLink(line.label);
-  const returnToQueryString = qs.stringify({ returnTo: line.line_id });
-
-  return `${draftUrl}?${returnToQueryString}`;
-};
 
 export const ActionsRow = ({
   className = '',
@@ -23,6 +16,7 @@ export const ActionsRow = ({
   const { t } = useTranslation();
 
   const { line } = useGetLineDetails();
+  const { getDraftsUrl } = useRoutesAndLinesDraftReturnToQueryParam();
 
   return (
     <Container className={className}>
@@ -32,7 +26,7 @@ export const ActionsRow = ({
           <Column className="items-end justify-end">
             <SimpleButton
               inverted
-              href={getDraftsUrlWithReturnToQueryString(line)}
+              href={getDraftsUrl(line.label, line.line_id)}
             >
               {t('lines.showDrafts')}
             </SimpleButton>
--- a/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
+++ b/ui/src/components/routes-and-lines/line-details/ActionsRow.tsx
@@ -27,9 +27,7 @@ export const ActionsRow = ({
   return (
     <Container className={className}>
       <FormRow mdColumns={2}>
-        <Column className="w-1/4">
-          <ObservationDateControl className="flex-1" />
-        </Column>
+        <ObservationDateControl className="max-w-max" />
         {line && (
           <Column className="items-end justify-end">
             <SimpleButton
--- a/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
+++ b/ui/src/components/routes-and-lines/line-drafts/LineDraftsPage.tsx
@@ -1,30 +1,18 @@
 import { useTranslation } from 'react-i18next';
-import { useHistory, useParams } from 'react-router';
-import { useUrlQuery } from '../../../hooks';
+import { useParams } from 'react-router';
+import { useRoutesAndLinesDraftReturnToQueryParam } from '../../../hooks';
 import { useGetLineDraftDetails } from '../../../hooks/line-drafts/useGetLineDraftDetails';
 import { Column, Container, Row } from '../../../layoutComponents';
-import { Path, routeDetails } from '../../../router/routeDetails';
 import { CloseIconButton } from '../../../uiComponents';
 import { ObservationDateControl } from '../../common/ObservationDateControl';
 import { RouteStopsTable } from '../line-details/RouteStopsTable';
 
 export const LineDraftsPage = (): JSX.Element => {
   const { t } = useTranslation();
-  const history = useHistory();
   const { label } = useParams<{ label: string }>();
-  const { queryParams } = useUrlQuery();
   const { routes } = useGetLineDraftDetails();
 
-  const onClose = () => {
-    // If there is no returnTo set, we return to 'routes and lines' page
-    const pathname = queryParams.returnTo
-      ? routeDetails[Path.lineDetails].getLink(queryParams.returnTo as string)
-      : routeDetails[Path.routes].getLink();
-
-    history.push({
-      pathname,
-    });
-  };
+  const { onClose } = useRoutesAndLinesDraftReturnToQueryParam();
 
   return (
     <Container>
--- a/ui/src/components/routes-and-lines/search/SearchContainer.tsx
+++ b/ui/src/components/routes-and-lines/search/SearchContainer.tsx
@@ -102,13 +102,14 @@ export const SearchContainer = (): JSX.Element => {
               priorities={searchConditions.priorities}
             />
             <Column className="justify-self-end">
-              <ObservationDateInput
-                value={searchConditions.observationDate}
-                onChange={onChangeDate}
-                className="flex-1"
-                required
-                testId={testIds.observationDateInput}
-              />
+              <Row className="flex-1">
+                <ObservationDateInput
+                  value={searchConditions.observationDate}
+                  onChange={onChangeDate}
+                  required
+                  testId={testIds.observationDateInput}
+                />
+              </Row>
             </Column>
           </FormRow>
         </div>
--- a/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx
+++ b/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx
@@ -1,9 +1,14 @@
 import { gql } from '@apollo/client';
 import { useTranslation } from 'react-i18next';
 import { useGetVehicleJourneysQuery } from '../../../generated/graphql';
-import { useGetLineDetails } from '../../../hooks';
-import { Column, Container, Row } from '../../../layoutComponents';
+import {
+  useGetLineDetails,
+  useTimetableVersionsReturnToQueryParam,
+} from '../../../hooks';
+import { Container } from '../../../layoutComponents';
+import { SimpleButton } from '../../../uiComponents';
 import { ObservationDateControl } from '../../common/ObservationDateControl';
+import { FormColumn, FormRow } from '../../forms/common';
 import { PageHeader } from '../../routes-and-lines/common/PageHeader';
 import { PassingTimesByStopTable } from '../passing-times-by-stop';
 import { VehicleRouteTimetables } from './VehicleRouteTimetables';
@@ -23,6 +28,8 @@ export const VehicleScheduleDetailsPage = (): JSX.Element => {
 
   const { line } = useGetLineDetails();
 
+  const { getVersionsUrl } = useTimetableVersionsReturnToQueryParam();
+
   // Just get all vehicle journeys from back end as test data
   // TODO: Use PassingTimesByStopTable in the right place and fetch correct data
   const vehicleJourneysResult = useGetVehicleJourneysQuery();
@@ -39,11 +46,19 @@ export const VehicleScheduleDetailsPage = (): JSX.Element => {
         </h1>
       </PageHeader>
       <div className="mx-12 my-8">
-        <Row className="mb-8">
-          <Column>
-            <ObservationDateControl />
-          </Column>
-        </Row>
+        <FormRow mdColumns={2} className="mb-8 ">
+          <ObservationDateControl className="max-w-max" />
+          <FormColumn className="items-end justify-end">
+            {line && (
+              <SimpleButton
+                inverted
+                href={getVersionsUrl(line.label, line.line_id)}
+              >
+                {t('timetables.showVersions')}
+              </SimpleButton>
+            )}
+          </FormColumn>
+        </FormRow>
         <VehicleRouteTimetables routes={line?.line_routes || []} />
       </div>
       <Container className="space-y-10 pt-10">
--- a/ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTable.tsx
+++ b/ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTable.tsx
@@ -31,7 +31,7 @@ export const VehicleServiceTable = ({
 }: Props): JSX.Element => {
   const { t } = useTranslation();
 
-  const getHeadingBgColor = (key: TimetablePriority) => {
+  const getTimetableHeadingBgColor = (key: TimetablePriority) => {
     const bgColors: Record<TimetablePriority, string> = {
       [TimetablePriority.Standard]: 'bg-hsl-dark-green',
       [TimetablePriority.Temporary]: 'bg-city-bicycle-yellow',
@@ -75,7 +75,7 @@ export const VehicleServiceTable = ({
   return (
     <div>
       <Row
-        className={`mb-4 rounded-md bg-opacity-50 px-4 py-1 text-hsl-dark-80 ${getHeadingBgColor(
+        className={`mb-4 rounded-md bg-opacity-50 px-4 py-1 text-hsl-dark-80 ${getTimetableHeadingBgColor(
           priority,
         )}`}
       >
--- a/ui/src/components/timetables/versions/TimetableVersionTable.tsx
+++ b/ui/src/components/timetables/versions/TimetableVersionTable.tsx
@@ -1,0 +1,65 @@
+import { useTranslation } from 'react-i18next';
+import { TimetablesVersionsRowData as TimetableVersionsRowData } from '../../../hooks';
+import { TimetableVersionTableRow } from './TimetableVersionTableRow';
+
+interface Props {
+  data: TimetableVersionsRowData[];
+  className: string;
+}
+
+export const TimetableVersionTable = ({
+  data,
+  className = '',
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+
+  // NOTE: These widths are given to prevent width jumping depending on data
+  // It will also keep the headings in line when there are multiple tables
+  // on same page. This is not by any means optimal solution, so feel free to
+  // change if a better solution comes up.
+  return (
+    <table className={`${className}`}>
+      <thead className="text-center [&>th]:p-4">
+        <tr>
+          <th className="w-[10%]">
+            {t('timetableVersionsTableHeaders.status')}
+          </th>
+          <th className="w-[15%]">
+            {t('timetableVersionsTableHeaders.dayType')}
+          </th>
+          <th className="w-[10%]">
+            {t('timetableVersionsTableHeaders.validityStart')}
+          </th>
+          <th className="w-[10%]">
+            {t('timetableVersionsTableHeaders.validityEnd')}
+          </th>
+          <th className="w-[7%]">{t('timetableVersionsTableHeaders.label')}</th>
+          <th className="w-[20%]">
+            {t('timetableVersionsTableHeaders.version')}
+          </th>
+          <th className="w-[12.5%]">
+            {t('timetableVersionsTableHeaders.modified')}
+          </th>
+          <th className="w-[12.5%]">
+            {t('timetableVersionsTableHeaders.modifier')}
+          </th>
+          <th className="invisible w-[3%]">Actions</th>
+        </tr>
+      </thead>
+      <tbody>
+        {data.length ? (
+          data.map((row) => (
+            <TimetableVersionTableRow
+              key={row.id + row.routeLabelAndVariant}
+              data={row}
+            />
+          ))
+        ) : (
+          <tr>
+            <td>{t('timetables.noTimetables')}</td>
+          </tr>
+        )}
+      </tbody>
+    </table>
+  );
+};
--- a/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx
+++ b/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx
@@ -1,0 +1,64 @@
+import { useTranslation } from 'react-i18next';
+import { TimetablesVersionsRowData } from '../../../hooks';
+import { mapTimetablePriorityToUiName } from '../../../i18n/uiNameMappings';
+import { parseI18nField } from '../../../i18n/utils';
+import { mapToShortDate } from '../../../time';
+import { TimetablePriority } from '../../../types/enums';
+
+const getStatusClassName = ({
+  priority,
+  inEffect,
+}: {
+  priority: TimetablePriority;
+  inEffect?: boolean;
+}) => {
+  if (inEffect) {
+    return 'bg-hsl-dark-green text-white';
+  }
+  const statusClassNames: Record<TimetablePriority, string> = {
+    [TimetablePriority.Standard]: 'bg-brand text-white',
+    [TimetablePriority.Temporary]: 'bg-city-bicycle-yellow',
+    [TimetablePriority.Special]: 'bg-hsl-light-purple',
+    [TimetablePriority.Draft]: 'bg-background',
+    [TimetablePriority.Staging]: 'bg-hsl-red',
+  };
+  return statusClassNames[priority];
+};
+
+interface Props {
+  data: TimetablesVersionsRowData;
+}
+
+export const TimetableVersionTableRow = ({ data }: Props): JSX.Element => {
+  const { t } = useTranslation();
+
+  const statusClassName = ` ${getStatusClassName({
+    priority: data.vehicleScheduleFrame.priority,
+    inEffect: data.inEffect,
+  })}`;
+
+  const statusText = data.inEffect
+    ? t('timetables.inEffect')
+    : mapTimetablePriorityToUiName(data.vehicleScheduleFrame.priority);
+
+  // TODO: After we get special days implemented, we need to determine this className
+  // depending on wheter the row is from vehicleService or from special day.
+  // The className for special days is: 'bg-city-bicycle-yellow bg-opacity-25'
+  const dayTypeClassName = 'bg-hsl-dark-green bg-opacity-25';
+
+  return (
+    <tr className="text-center [&>td]:border [&>td]:border-light-grey [&>td]:p-4">
+      <td className={statusClassName}>{statusText}</td>
+      <td className={dayTypeClassName}>
+        {parseI18nField(data.dayType.nameI18n)}
+      </td>
+      <td>{mapToShortDate(data.vehicleScheduleFrame.validityStart)}</td>
+      <td>{mapToShortDate(data.vehicleScheduleFrame.validityEnd)}</td>
+      <td>{data.routeLabelAndVariant}</td>
+      <td>{parseI18nField(data.vehicleScheduleFrame.nameI18n)}</td>
+      <td>!Muokkaaja</td>
+      <td>!Muokattu</td>
+      <td>...</td>
+    </tr>
+  );
+};
--- a/ui/src/components/timetables/versions/TimetableVersionsPage.tsx
+++ b/ui/src/components/timetables/versions/TimetableVersionsPage.tsx
@@ -1,0 +1,48 @@
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
+import {
+  useGetLineTimetableVersions,
+  useTimetableVersionsReturnToQueryParam,
+} from '../../../hooks';
+import { Container } from '../../../layoutComponents';
+import { CloseIconButton } from '../../../uiComponents';
+import { FormColumn, FormRow } from '../../forms/common';
+import { TimetableVersionTable } from './TimetableVersionTable';
+
+export const TimetableVersionsPage = (): JSX.Element => {
+  const { t } = useTranslation();
+  const { label } = useParams<{ label: string }>();
+
+  const { timetablesExcludingDrafts, onlyDraftTimetables } =
+    useGetLineTimetableVersions({ label });
+  const { onClose } = useTimetableVersionsReturnToQueryParam();
+
+  return (
+    <Container>
+      <FormRow mdColumns={2}>
+        <h2>{`${t('timetables.versionsTitle')} | ${t('lines.line', {
+          label,
+        })}`}</h2>
+        <FormColumn className="items-end">
+          <CloseIconButton
+            label={t('close')}
+            className="text-base font-bold text-brand"
+            onClick={onClose}
+          />
+        </FormColumn>
+      </FormRow>
+      <Container>
+        <h3>{t('timetables.operatingCalendar')}</h3>
+        <TimetableVersionTable
+          className="mb-8 w-full"
+          data={timetablesExcludingDrafts}
+        />
+        <h3>{t('timetables.drafts')}</h3>
+        <TimetableVersionTable
+          className="mb-8 w-full"
+          data={onlyDraftTimetables}
+        />
+      </Container>
+    </Container>
+  );
+};
--- a/ui/src/components/timetables/versions/index.ts
+++ b/ui/src/components/timetables/versions/index.ts
@@ -1,0 +1,3 @@
+export * from './TimetableVersionsPage';
+export * from './TimetableVersionTable';
+export * from './TimetableVersionTableRow';
--- a/ui/src/generated/graphql.tsx
+++ b/ui/src/generated/graphql.tsx
@@ -16604,6 +16604,54 @@ export type TimingPlaceForComboboxFragment = {
   description?: any | null;
 };
 
+export type GetTimetableVersionsByLineLabelQueryVariables = Exact<{
+  routeFilters?: InputMaybe<RouteRouteBoolExp>;
+}>;
+
+export type GetTimetableVersionsByLineLabelQuery = {
+  __typename?: 'query_root';
+  route_route: Array<{
+    __typename?: 'route_route';
+    route_id: UUID;
+    label: string;
+    variant?: number | null;
+    route_journey_patterns: Array<{
+      __typename?: 'journey_pattern_journey_pattern';
+      journey_pattern_id: UUID;
+      journey_pattern_refs: Array<{
+        __typename?: 'timetables_journey_pattern_journey_pattern_ref';
+        journey_pattern_ref_id: UUID;
+        vehicle_journeys: Array<{
+          __typename?: 'timetables_vehicle_journey_vehicle_journey';
+          vehicle_journey_id: UUID;
+          block: {
+            __typename?: 'timetables_vehicle_service_block';
+            block_id: UUID;
+            vehicle_service: {
+              __typename?: 'timetables_vehicle_service_vehicle_service';
+              vehicle_service_id: UUID;
+              vehicle_schedule_frame: {
+                __typename?: 'timetables_vehicle_schedule_vehicle_schedule_frame';
+                vehicle_schedule_frame_id: UUID;
+                priority: number;
+                validity_start?: luxon.DateTime | null;
+                validity_end?: luxon.DateTime | null;
+                name_i18n: any;
+              };
+              day_type: {
+                __typename?: 'timetables_service_calendar_day_type';
+                day_type_id: UUID;
+                label: string;
+                name_i18n: any;
+              };
+            };
+          };
+        }>;
+      }>;
+    }>;
+  }>;
+};
+
 export type RouteInformationForMapFragment = {
   __typename?: 'route_route';
   route_id: UUID;
@@ -20997,6 +21045,94 @@ export type GetSelectedTimingPlaceDetailsByIdQueryResult = Apollo.QueryResult<
   GetSelectedTimingPlaceDetailsByIdQuery,
   GetSelectedTimingPlaceDetailsByIdQueryVariables
 >;
+export const GetTimetableVersionsByLineLabelDocument = gql`
+  query GetTimetableVersionsByLineLabel($routeFilters: route_route_bool_exp) {
+    route_route(where: $routeFilters) {
+      route_id
+      label
+      variant
+      route_journey_patterns {
+        journey_pattern_id
+        journey_pattern_refs {
+          journey_pattern_ref_id
+          vehicle_journeys {
+            vehicle_journey_id
+            block {
+              block_id
+              vehicle_service {
+                vehicle_service_id
+                vehicle_schedule_frame {
+                  vehicle_schedule_frame_id
+                  priority
+                  validity_start
+                  validity_end
+                  priority
+                  name_i18n
+                }
+                day_type {
+                  day_type_id
+                  label
+                  name_i18n
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+/**
+ * __useGetTimetableVersionsByLineLabelQuery__
+ *
+ * To run a query within a React component, call `useGetTimetableVersionsByLineLabelQuery` and pass it any options that fit your needs.
+ * When your component renders, `useGetTimetableVersionsByLineLabelQuery` returns an object from Apollo Client that contains loading, error, and data properties
+ * you can use to render your UI.
+ *
+ * @param baseOptions options that will be passed into the query, supported options are listed on: https://www.apollographql.com/docs/react/api/react-hooks/#options;
+ *
+ * @example
+ * const { data, loading, error } = useGetTimetableVersionsByLineLabelQuery({
+ *   variables: {
+ *      routeFilters: // value for 'routeFilters'
+ *   },
+ * });
+ */
+export function useGetTimetableVersionsByLineLabelQuery(
+  baseOptions?: Apollo.QueryHookOptions<
+    GetTimetableVersionsByLineLabelQuery,
+    GetTimetableVersionsByLineLabelQueryVariables
+  >,
+) {
+  const options = { ...defaultOptions, ...baseOptions };
+  return Apollo.useQuery<
+    GetTimetableVersionsByLineLabelQuery,
+    GetTimetableVersionsByLineLabelQueryVariables
+  >(GetTimetableVersionsByLineLabelDocument, options);
+}
+export function useGetTimetableVersionsByLineLabelLazyQuery(
+  baseOptions?: Apollo.LazyQueryHookOptions<
+    GetTimetableVersionsByLineLabelQuery,
+    GetTimetableVersionsByLineLabelQueryVariables
+  >,
+) {
+  const options = { ...defaultOptions, ...baseOptions };
+  return Apollo.useLazyQuery<
+    GetTimetableVersionsByLineLabelQuery,
+    GetTimetableVersionsByLineLabelQueryVariables
+  >(GetTimetableVersionsByLineLabelDocument, options);
+}
+export type GetTimetableVersionsByLineLabelQueryHookResult = ReturnType<
+  typeof useGetTimetableVersionsByLineLabelQuery
+>;
+export type GetTimetableVersionsByLineLabelLazyQueryHookResult = ReturnType<
+  typeof useGetTimetableVersionsByLineLabelLazyQuery
+>;
+export type GetTimetableVersionsByLineLabelQueryResult = Apollo.QueryResult<
+  GetTimetableVersionsByLineLabelQuery,
+  GetTimetableVersionsByLineLabelQueryVariables
+>;
 export const GetTimetablesForOperationDayDocument = gql`
   query GetTimetablesForOperationDay(
     $journey_pattern_id: uuid!
@@ -21451,6 +21587,15 @@ export function useGetSelectedTimingPlaceDetailsByIdAsyncQuery() {
 }
 export type GetSelectedTimingPlaceDetailsByIdAsyncQueryHookResult = ReturnType<
   typeof useGetSelectedTimingPlaceDetailsByIdAsyncQuery
+>;
+export function useGetTimetableVersionsByLineLabelAsyncQuery() {
+  return useAsyncQuery<
+    GetTimetableVersionsByLineLabelQuery,
+    GetTimetableVersionsByLineLabelQueryVariables
+  >(GetTimetableVersionsByLineLabelDocument);
+}
+export type GetTimetableVersionsByLineLabelAsyncQueryHookResult = ReturnType<
+  typeof useGetTimetableVersionsByLineLabelAsyncQuery
 >;
 export function useGetTimetablesForOperationDayAsyncQuery() {
   return useAsyncQuery<
--- a/ui/src/hooks/index.ts
+++ b/ui/src/hooks/index.ts
@@ -12,6 +12,7 @@ export * from './ui';
 export * from './urlQuery';
 export * from './useAsyncQuery';
 export * from './useCheckValidityAndPriorityConflicts';
+export * from './useGetLineTimetableVersions';
 export * from './usePagination';
 export * from './useShowRoutesOnModal';
 export * from './useToggle';
--- a/ui/src/hooks/stops/useCreateStop.ts
+++ b/ui/src/hooks/stops/useCreateStop.ts
@@ -8,10 +8,10 @@ import {
 import { StopWithLocation } from '../../graphql';
 import { OptionalKeys } from '../../types';
 import {
+  getRouteLabelVariantText,
   IncompatibleWithExistingRoutesError,
   removeFromApolloCache,
 } from '../../utils';
-import { getRouteLabelVariantText } from '../../utils/route';
 import { useCheckValidityAndPriorityConflicts } from '../useCheckValidityAndPriorityConflicts';
 import { BrokenRouteCheckParams, useEditStop } from './useEditStop';
 import { useGetStopLinkAndDirection } from './useGetStopLinkAndDirection';
--- a/ui/src/hooks/stops/useValidateTimingSettings.ts
+++ b/ui/src/hooks/stops/useValidateTimingSettings.ts
@@ -1,7 +1,9 @@
 import { gql } from '@apollo/client';
 import { useGetScheduledStopPointsInJourneyPatternsUsedAsTimingPointsAsyncQuery } from '../../generated/graphql';
-import { TimingPlaceRequiredError } from '../../utils';
-import { getRouteLabelVariantText } from '../../utils/route';
+import {
+  getRouteLabelVariantText,
+  TimingPlaceRequiredError,
+} from '../../utils';
 
 const GQL_GET_SCHEDULED_STOP_POINTS_IN_JOURNEY_PATTERNS_USED_AS_TIMING_POINTS = gql`
   query GetScheduledStopPointsInJourneyPatternsUsedAsTimingPoints(
--- a/ui/src/hooks/ui/index.ts
+++ b/ui/src/hooks/ui/index.ts
@@ -2,3 +2,5 @@ export * from './useAlertsAndHighLights';
 export * from './useChooseLineDropdown';
 export * from './useChooseRouteDropdown';
 export * from './useLoader';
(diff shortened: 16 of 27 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 0 of 17 files reviewed, 10 unresolved discussions (waiting on @HenrikHartiala)

___
*[`package.json` line 26 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk1dja7XNHpai2Vd80:-NNk1dja7XNHpai2Vd81:b-v7kss8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/package.json#L26)):*
> ```JSON
>     "test:e2e:list": "yarn ws:e2e print-tests 'e2e/**.cy.ts'",
>     "qa": "yarn --frozen-lockfile && yarn ws:db build && yarn ts:check && yarn prettier:check && yarn lint && yarn test && yarn test:integration && yarn test:e2e",
>     "qa:fast": "yarn qa --config-file cypress.config.fast.ts"
> ```

By the way, this command works mostly "by luck" as `--` flag (which is intented for cypress) gets passedt to last ommand in `yarn qa` script, and if someone happens to add more commands there, things explode...
Thus I would just remove this command (assuming that no one used it as it was broken), but if you find this useful, we can keep it.
___
*[`ui/src/components/forms/common/ObservationDateInput.tsx` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk2W12Bbx8sTbUfYbn:-NNk2W12Bbx8sTbUfYbo:bav5ygl) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/forms/common/ObservationDateInput.tsx#L20)):*
> ```XML
>   return (
>     <Row className={className}>
>       <Column>
> ```

Now this can't be used anymore in contexts where form layout is not column, so this is now less flexibe.

By the way, is Row actually used for anything?



_Code quote:_
```XML
    <Row className={className}>
      <Column>
```
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 37 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk3DRa13703fwSX127:-NNk3DRa13703fwSX128:b7s01u5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L37)):*
> ```XML
>  *  line id.
>  */
> const getVersionsUrlWithReturnToQueryString = (line: RouteLine) => {
> ```

By the way, this removes possibly existing query from url - is this intented? 
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 8 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk6UPL14070-a-55cq:-NNk6UPM9T31WTT2oxM6:b-nvcl2x) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L8)):*
> ```XML
> export const TimetableVersionTable = ({
>   data,
>   className,
> ```

Please default to `= ''` to avoid having 'undefined' strings as classNames in DOM if this is not given.

By the way, usually we have defined `interface Props` for these before the actual component, but as we don't have linter rules or anything for this I guess both approaches are OK.
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 15 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk6r185gWhrYSjhtRI:-NNk6r185gWhrYSjhtRJ:bezkeib) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L15)):*
> ```XML
>   return (
>     <table className={`${className} w-full`}>
> ```

className prop should be last in this list as it is intented for overriding default styles.
Also one could argue w-full should be passed from outside, but do as you wish regarding that.
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 26 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk7GSr8yC7XRbH99Ba:-NNk7GSr8yC7XRbH99Bb:b-l2hea2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L26)):*
Is span valid child for tbody element? Probably not:
>The only allowed child nodes of the HTML <tbody> tag are <tr> (table row) elements.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 29 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk8Is45Ex_VIjrmcUZ:-NNk8Is45Ex_VIjrmcU_:b-1dfecy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L29)):*
> ```XML
> };
> 
> const StatusTableDataCell = ({
> ```

Should be typed to JSX.Element.
But as this is used only once, I don't see reason why this is implemented as own component instead of inlining this.
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 81 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkARX-0ufLEuVOxmfX:-NNkARX-0ufLEuVOxmfY:bpr345y) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L81)):*
> ```XML
>       isDateInRange(
>         DateTime.now(),
>         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
> ```

Please comment why this is safe to do, or do we have bug here?
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 144 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkArnd0zMWdnQ3M9oC:-NNkArnd0zMWdnQ3M9oD:b-w1z5nu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L144)):*
> ```XML
>       timetableVersionsTableData.push({
>         routeLabelAndVariant: route.routeLabelAndVariant,
>         ...foo,
> ```

:notbad:, but please use e.g. `item` instead
___
*[`ui/src/locales/en-US/common.json` line 301 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk9guJ1RolxnRBc0YN:-NNk9guJ1RolxnRBc0YO:b-dl76cm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/locales/en-US/common.json#L301)):*
> ```JSON
>     "operatingCalendar": "Operating calendar",
>     "drafts": "Drafts",
>     "noDrafts": "No drafts",
> ```

Just note that "lines" has following translations too:
"draftsTitle": "Drafts",
    "noDrafts": "No drafts",

It is a matter of opinion if those should be combined, so do as you wish
___
*[`ui/src/locales/fi-FI/common.json` line 180 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkA0kPE7pD2nw2dY2Q:-NNkA0kPE7pD2nw2dY2R:b-ju5jn7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/locales/fi-FI/common.json#L180)):*
> ```JSON
>     "temporary": "Väliaikainen",
>     "draft": "Luonnos",
>     "special": "Special",
> ```

"Special" is not finnish. "Erityinen"? Or is this intented?

eleksi (changes requested):
> Reviewed 1 of 1 files at r1, 15 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 16 of 17 files reviewed, 23 unresolved discussions (waiting on @HenrikHartiala and @tk-vincit)

___
*[`ui/src/components/forms/common/ObservationDateInput.tsx` line 19 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkEgebArmBwGObG3cf:-NNkEgebArmBwGObG3cg:bfvyymj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/forms/common/ObservationDateInput.tsx#L19)):*
> ```XML
>   const dateInputId = 'observation-date-input';
>   return (
>     <Row className={className}>
> ```

What is the role if this `Row`
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 37 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkExw0-02p37E7EVXn:-NNkExw0-02p37E7EVXo:b70je7t) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L37)):*
> ```XML
>  *  line id.
>  */
> const getVersionsUrlWithReturnToQueryString = (line: RouteLine) => {
> ```

Could we just save the url from history when coming to this page or just do history.goBack() etc?
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 81 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkHCIY0XLnrUt1qDsE:-NNkHCIY0XLnrUt1qDsF:b-u0slg7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L81)):*
> ```XML
>             {line && (
>               <SimpleButton
>                 className="mt-auto"
> ```

What is this classname for
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 17 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkHJoR8SH1YNQ236HA:-NNkHJoR8SH1YNQ236HB:bk9v7vw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L17)):*
> ```XML
> 
>   const { timetableVersionsTableData } = useGetLineTimetableVersions({ label });
> ```

Is the observation date supposed to always be today, or can the "observation date line" in the visual view (in designs) be moved?
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 25 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkHkM-2XYBedzrIs2K:-NNkHkM-2XYBedzrIs2L:bm7f89r) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L25)):*
> ```XML
>           queryParams.returnTo as string,
>         )
>       : routeDetails[Path.timetables].getLink();
> ```

Again, I would rather try going back in history rather than hard coded url. If history has no back, then maybe go to that page.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 25 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkFxKE961rsbgslFKA:-NNkFxKE961rsbgslFKB:baw1h1p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L25)):*
> ```XML
>       return 'bg-gray-100';
>     default:
>       throw new Error(`Unkown priority: ${priority}`);
> ```

What if we have staging priority timetables (as I had when tried to test this)? The page crashes. If we do not support them and the app crashes instead, maybe we should filter them out completely so the view can be opened even if there were staging priority timetables.
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkJT0O31uC8w8LeDE7:-NNkJT0O31uC8w8LeDE8:bc0n9vr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L10)):*
> ```XML
> import { TimetablePriority } from '../../types/Priority';
> import { buildRouteLineLabelGqlFilter, mapToVariables } from '../../utils';
> import { getRouteLabelVariantText } from '../../utils/route';
> ```

Not part of this PR but if you want you could make this file contents to be exported from /utils/index :smile: 
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 12 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkMPWY3jk9RKI2bUST:-NNkMPWY3jk9RKI2bUSU:byqrekx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L12)):*
> ```XML
> import { getRouteLabelVariantText } from '../../utils/route';
> 
> export interface TimetablesVersionsRowData {
> ```

This looks funny with combination of camel case and snake case fields, but I guess that is ok :smile: 
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 13 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkJoe1-aD0cE7P3kxy:-NNkJoe1-aD0cE7P3kxz:br2alhf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L13)):*
> ```XML
> 
> export interface TimetablesVersionsRowData {
>   routeLabelAndVariant: string;
> ```

We have `unique_label` in route information, which includes route and variant. There is _ as a separator between the label and variant though, so it cant be used straight away. But just FYI there is pretty much similar field in the data model which is named unique label, if you want to utilize this info, it might affect this field naming and/or source.
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 140 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkNhxyDGOEnI3pPiMz:-NNkNhxyDGOEnI3pPiN-:bevsylh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L140)):*
> ```XML
>   const timetableVersionsTableData: TimetablesVersionsRowData[] = [];
> 
>   distinctMappedVersionsData.forEach((route) => {
> ```

This kind of looks like two `flatMap`s to me, instead of `forEach`.
___
*[`ui/src/locales/en-US/common.json` line 181 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkOQIL61_8lXpdwQXJ:-NNkOQIL61_8lXpdwQXK:b-twqj2c) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/locales/en-US/common.json#L181)):*
> ```JSON
>     "draft": "Draft",
>     "special": "Special",
>     "staging": "Staging"
> ```

This could be mapped to some different value as well. "Staging" might not tell the user much, this is more like "timetable being imported". I don't know if we want to be exact just tell the timetable is "staging" priority or actually tell them what the priority means.
___
*[`ui/src/locales/fi-FI/common.json` line 180 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkA0kPE7pD2nw2dY2Q:-NNkOI7e2e5W1pfoi0_5:bn4pxoi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/locales/fi-FI/common.json#L180)):*

"Special" is not finnish. "Erityinen"? Or is this intented?
</blockquote></details>

Erityispäivä?
___
*[`ui/src/locales/fi-FI/common.json` line 181 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkOcyI6qydGfA6InHX:-NNkOcyI6qydGfA6InHY:b7rliyx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/locales/fi-FI/common.json#L181)):*
> ```JSON
>     "draft": "Luonnos",
>     "special": "Special",
>     "staging": "Staging"
> ```

Same as for English version above... :thinking: 
___
*[`ui/src/locales/fi-FI/common.json` line 306 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkOnLCA7vgRWQWWo_r:-NNkOnLCA7vgRWQWWo_s:bh1uec9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/locales/fi-FI/common.json#L306)):*
> ```JSON
>   },
>   "timetableVersionsTableHeaders": {
>     "status": "Status",
> ```

I know it says this in designs, just wondering why this is not just "tila" or something like that.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 16 of 17 files reviewed, 23 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 25 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkHkM-2XYBedzrIs2K:-NNkjdt2AJGxVaGJSB73:b-1tnai) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L25)):*

Again, I would rather try going back in history rather than hard coded url. If history has no back, then maybe go to that page.
</blockquote></details>

Also it could be considered if this kind of logics should be extracted to a hook for a very least.
That would couple "both sides" of this hack into a single file.
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 36 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkjulkCwkRYNVllM37:-NNkjulkCwkRYNVllM38:bo23fu0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L36)):*
> ```XML
>   );
>   const onlyDraftTimetables = timetableVersionsTableData.filter(
>     (item) => item.vehicle_schedule_frame.priority === TimetablePriority.Draft,
> ```

Kind of matter of opinion, but as `useGetTimetableVerions` returns timetables already mapped to a data structure designed specially for UI's needs, this kind of splitting could be done in the hook itself.



_Code quote:_
```XML
  const timetablesExcludingDrafts = timetableVersionsTableData.filter(
    (item) => item.vehicle_schedule_frame.priority !== TimetablePriority.Draft,
  );
  const onlyDraftTimetables = timetableVersionsTableData.filter(
    (item) => item.vehicle_schedule_frame.priority === TimetablePriority.Draft,
  );
```
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 25 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkFxKE961rsbgslFKA:-NNkkQ0uC0HeyZwMkgp-:bm7cadp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L25)):*

What if we have staging priority timetables (as I had when tried to test this)? The page crashes. If we do not support them and the app crashes instead, maybe we should filter them out completely so the view can be opened even if there were staging priority timetables.
</blockquote></details>

Very good catch. Another alternative could be defining some kind of bg colour for those here.
Annoying that TS compiler doesn't complain here that we have unhandled cases. That is one reason why this kind of Record-based pattern is sometimes preferred:

```
export const mapPriorityToUiName = (key: Priority) => {
  const uiStrings: Record<Priority, string> = {
    [Priority.Standard]: i18n.t('priority.standard'),
    [Priority.Temporary]: i18n.t('priority.temporary'),
    [Priority.Draft]: i18n.t('priority.draft'),
  };
  return uiStrings[key];
};

```
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 12 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkMPWY3jk9RKI2bUST:-NNkkyzt3csMCU9vvX-f:b-k91odt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L12)):*

This looks funny with combination of camel case and snake case fields, but I guess that is ok :smile: 
</blockquote></details>

That is a good point. As this is custom datatype for UI's needs, there is no need to use snake_case :thinking:

tk-vincit (changes requested):
> Reviewed 3 of 25 files at r4.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 7 of 29 files reviewed, 23 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`package.json` line 26 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk1dja7XNHpai2Vd80:-NOJ1Ab5EAe3E3MBQCCI:b-8a4tgd) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/package.json#L26)):*

Yeah good point, that is the downside of this. But if I put it back like it was (having all the necessary commands in this script), then we need to update both always. And this should apply to `test:e2e:fast` as well. (Or is there another way of fixing this?)
So at least we have less text int he scripts :D

And about the other point, I did use the `yarn:qa` command every day but then the tests started to fail locally and everywhere so bad that I've just used CI. I do think that I'm going to continue using this command again later :+1: It is nice that it runs all the tests and eslints and prettier checks for me. Although I could just create a commithook for myself to run prettier.
</blockquote></details>

This could be solved by having more yarn scripts and composing this kind of commands from previously introduced scripts. But that also causes us to have more and more script definitions, which clutters things up and has thus also drawbacks.
But as discussed f2f, this is OK for now.  :+1: 
___
*[`ui/src/components/forms/common/ObservationDateInput.tsx` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk2W12Bbx8sTbUfYbn:-NOJ1jbA7vIeKBpjHjZ2:bp9i1nt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/forms/common/ObservationDateInput.tsx#L20)):*

The Row + Column combination is for the layout of this component. Now it will always be the same size look the same where ever you put it.
Im not sure what you mean by "can't be used in context where form layout is not column" I think you can now put this anywhere and it will look the same.

The Row is used for the layout of this component. If I remove the `Row`, then the outer component will be "Column" and this component will start to look different depending on what is the parent component. 
I can demonstrate this if needed, but you can also check what happens for example on line details page if you remove the Row component.

So with this Row + Colum combination, where ever you put it, it will always be so that the <label> (Tarkasteluhetki) will be on top of the input field and the size of the input will always be the same (minimum size so it won't take as much space as it is given). 
</blockquote></details>

As discussed f2f, defining size of component is arguably rather responsibility of the parent component.
But do as you wish here.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 37 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk3DRa13703fwSX127:-NOJ23_40yAzs2csWhoO:bx8sstd) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L37)):*

Good point, I am not sure. We might want them to persist there? Maybe this needs to be confirmed from Juha.
</blockquote></details>

Let's continue this discussion within the new place where this code is now moved, if there is still something to discuss.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 37 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkExw0-02p37E7EVXn:-NOJ2Jgl9D3k4MdrmTJX:be31d2m) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L37)):*

This is implemented like this to match the same behaviour and implementation as in LineDrafts. The reason is that if the URL is copypasted, the `Close` button in versions page would throw you out. (Copy paste it to blank page and this button would return you to blank page). 

We want this button to return you to the line page you came here from.
And our VehicleScheduleDetailsPage (and LineDetailsPage) is LineID based, but TimetableVersionsPage (and LineDraftsPage) is label based, we can't just remove the label and return to the corresponding linePage.
</blockquote></details>

Henrik has valid points about drawbacks of history.goBack, e.g. that the history might be empty (or it could even have some 3rd party page is user pastes url of this page e.g. after reading the news!)
___
*[`ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTable.tsx` line 34 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ3AHm1jf5JB_vw1qs:-NOJ3AHm1jf5JB_vw1qt:b2zce5o) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9e8120f4878ff6b499342eaa9680e3eed97369da/ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTable.tsx#L34)):*
> ```XML
> 
>   const getHeadingBgColor = (key: TimetablePriority) => {
> ```

If this is actually used only here for now, then it might make more sense to keep it here.
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 25 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkHkM-2XYBedzrIs2K:-NOJ3e1F1VXKDtUFqD2m:b-87u105) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L25)):*

Yeah so this is the same behaviour in which we ended up with when implementing LineDrafts. Currently this will make the "Close" buttons behaviour more sensible when this URL is copypasted (we dont want to "go back" we want to "close" this and return to the VehicleScheduleDetailsPage. I know that there could be arguments about other nonsensible behaviours in some use cases and I'm open for suggestions, but I do think that we should rather discuss this live, since it is very hard topic to discuss here.

All that said, the suggestion about hook extraction I can already do. Let's see if I can combine these two implementations :+1: 
</blockquote></details>

Let's continue this discussion (if needed) within the new hook implementation.
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 36 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkjulkCwkRYNVllM37:-NOJ4H-yDSAy2aIG7Lkm:b1lt0tw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L36)):*

Thanks, I thought of that but then I argumented myself that "what if someone does not want these separated"?  But you do have a point that it is not the RAW data anyway, and the user can just .concat these (or simply return all three variations from the hook) if doesn't want these separated.
So I'll change this :+1: 
</blockquote></details>

The hook itself could also return `allTimetables` if someone needs them :shrug: 
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 12 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ3rXuAP18RripbkSm:-NOJ3rXuAP18RripbkSn:bsyfrx5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L12)):*
> ```XML
> import { TimetableVersionTable } from './TimetableVersionTable';
> 
> export const LineTimetableVersionsPage = (): JSX.Element => {
> ```

Exported component name should be the same as file name, so please rename either one.
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 24 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ4WroD14u8Qh87fCo:-NOJ4WroD14u8Qh87fCp:b-hsap4y) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L24)):*
> ```XML
>       <FormRow mdColumns={2}>
>         <h2>{`${t('timetables.versionsTitle')} | ${t('lines.line', {
>           label,
> ```

I see potential misuse here, what if someone writes something stupid to the label part of url and then we just show it here :fearful: 
OK, maybe that one would be on user :grin: 
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 8 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk6UPL14070-a-55cq:-NOJ5k7a7dIhPPs0Gipl:b-2w9qwj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L8)):*

Check! 

Yeah, I think we have mixed use of this. Maybe the linter rule would be nice to add consistency to this. It would help readability if it is always defined the same :+1: 
</blockquote></details>

I once fixed all of these to default to `= ''`, but it is possible that someone has introduced classNames without default values since that :unamused: :smile: 
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 15 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk6r185gWhrYSjhtRI:-NOJ5zbz04puKQnHGbk9:b-z1fpju) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L15)):*

Didn't think of the ordering. And good point on the w-full. I moved it outside.
</blockquote></details>

Not a big thing, but string template is not needed anymore as we have only one className.
<table className={className}>`
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 21 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ6FrWEe_fAGkicb-Y:-NOJ6FrWEe_fAGkicb-Z:b8uyv54) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L21)):*
> ```XML
>       <TimetableVersionsTableHeader />
>       <tbody>
>         {data.length > 0 ? (
> ```

Btw, another way to write this without ternary. Not sure which one is pretier though.
```{(data || [].)map(...)}

<Visible show={!data.lenght} > no timetables </Visible}
```
___
*[`ui/src/components/timetables/versions/TimetableVersionTableHeader.tsx` line 3 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ6jIy-5aQd8IR3EWM:-NOJ6jIy-5aQd8IR3EWN:bc9z8lh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionTableHeader.tsx#L3)):*
> ```XML
> import { useTranslation } from 'react-i18next';
> 
> export const TimetableVersionsTableHeader = (): JSX.Element => {
> ```

Does it improve readability of anything to have this in own componet and in a separate file? :shrug: 
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 25 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkFxKE961rsbgslFKA:-NOJ7GOjA5oLZZaTPUrU:b-5qf03b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L25)):*

Yep, good catch. I think we do want to filter out the staging priorities, there has been no discussion that we should show them anywhere in the UI. 

For Special priority, I can now add a colour since the colours for those have been picked.
</blockquote></details>

There is still no color for special priority and this still throws error?
Maybe this would work better:
```
default:
  console.error(Unknown priority)
  return ''
```
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 54 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ8trx5szYAa-LPqMa:-NOJ8trx5szYAa-LPqMb:b-kg76sk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L54)):*
> ```XML
>     <tr className="text-center [&>td]:border [&>td]:border-light-grey [&>td]:p-4">
>       <td className={dayTypeClassName}>{data.dayType.nameI18n.fi_FI}</td>
> ```

What is the point of having i18n field and then hardcoding the ui to use fi_FI? :wink: 
Please see `parseI18nField` from `ui/src/i18n/utils.ts`.

(By the way, not sure if these UI's should rather show all different translations to the user :thinking: . But that is out of scope of this PR, so just use `parseI18nField` here and we are good to go.)
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 58 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ9EzJE1kCHPVxspbZ:-NOJ9EzJE1kCHPVxspb_:buo9nmt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L58)):*
> ```XML
>       <td>{mapToShortDate(data.vehicleScheduleFrame.validityEnd)}</td>
>       <td>{data.routeLabelAndVariant}</td>
>       <td>{data.vehicleScheduleFrame.nameI18n.fi_FI}</td>
> ```

Same here
___
*[`ui/src/hooks/line-timetable-versions/index.ts` line 1 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJTO4SDUyewTst6_a6:-NOJTO4SDUyewTst6_a7:b-tdzws5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/line-timetable-versions/index.ts#L1)):*
> ```XML
> export * from './useGetLineTimetableVersions';
> ```

Does it look like we are going to have more hooks here? If not, then own directory + index file for this is not very useful.
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 75 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJWzZaAdPT5_MOauc3:-NOJWzZaAdPT5_MOauc4:bwuz9e4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L75)):*
> ```XML
>  * vehicle service that is valid 'today') and change it's inEffect value to true
>  */
> const enrichDataWithInEffectAttribute = (
> ```

Optional comment, but function like this could definitely benefit from unit tests... Same for other chained mapping function in this file.
___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 10 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJVAfo79GMC0aKLfMv:-NOJVAfo79GMC0aKLfMw:b38zuev) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L10)):*
> ```XML
>    *  Attaches the 'returnTo' url to query string. This is used to return to the
>    *  same line timetable view from where we opened the versions page. This is
>    *  needed because dsaasdfafdsfadsadfsadfdsfaadfsadfadfdfadafswe need to use line label in the versions page rather than
> ```

dsaasdfafdsfadsadfsadfdsfaadfsadfadfdfadafswe :joy: 
___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 13 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJVSle8BR_cX_gKGjL:-NOJVSle8BR_cX_gKGjM:bp8bm37) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L13)):*
> ```XML
>    *  line id.
>    */
>   const getUrlWithReturnToQueryString = (url: string, returnToId?: UUID) => {
> ```

if returnToId is optional, it can be undefined. Does this then add `?returnTo=undefined` to url? Maybe that is not what we want...

tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 7 of 29 files reviewed, 24 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 4 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJYfHeALq799HjleUI:-NOJYfHeALq799HjleUJ:btmrd2m) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L4)):*
> ```XML
> import { useHistory } from 'react-router';
> 
> export const useVersionsOnCloseWithReturnToQueryParam = () => {
> ```

What does "version" mean in this context?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 3 of 30 files reviewed, 21 unresolved discussions (waiting on @eleksi and @tk-vincit)

___
*[`ui/src/components/forms/common/ObservationDateInput.tsx` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk2W12Bbx8sTbUfYbn:-NOJuQScCHv35wnfrWDE:b3oqey) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/forms/common/ObservationDateInput.tsx#L20)):*

As discussed f2f, defining size of component is arguably rather responsibility of the parent component.
But do as you wish here.
</blockquote></details>

:+1: Moved the row-component to parent to make the sizing more flexible.
___
*[`ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTable.tsx` line 34 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ3AHm1jf5JB_vw1qs:-NOK2GrB2DvS4NTGUJf5:buj2azj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9e8120f4878ff6b499342eaa9680e3eed97369da/ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTable.tsx#L34)):*

If this is actually used only here for now, then it might make more sense to keep it here.
</blockquote></details>

Check, I'll revert this.
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 36 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkjulkCwkRYNVllM37:-NOK2yGD-z9ACfWmCUgi:b-k08qa0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L36)):*

The hook itself could also return `allTimetables` if someone needs them :shrug: 
</blockquote></details>

yeah true.
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 12 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ3rXuAP18RripbkSm:-NOK2cgv2Q4l4w2yaoii:b1aytns) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L12)):*

Exported component name should be the same as file name, so please rename either one.
</blockquote></details>

Good one
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 24 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ4WroD14u8Qh87fCo:-NOK3pyX1MEzEtFu32GP:b2nh52p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L24)):*

I see potential misuse here, what if someone writes something stupid to the label part of url and then we just show it here :fearful: 
OK, maybe that one would be on user :grin: 
</blockquote></details>

:D 
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 8 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk6UPL14070-a-55cq:-NOK3yTZ5AczFED7knzE:b-3wzb56) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L8)):*

I once fixed all of these to default to `= ''`, but it is possible that someone has introduced classNames without default values since that :unamused: :smile: 
</blockquote></details>

Meant the interface Props. 
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 15 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNk6r185gWhrYSjhtRI:-NOK4ome4q5YjW5cTqMX:bm6ngfc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L15)):*

Not a big thing, but string template is not needed anymore as we have only one className.
<table className={className}>`
</blockquote></details>

:travolta:
___
*[`ui/src/components/timetables/versions/TimetableVersionTable.tsx` line 21 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ6FrWEe_fAGkicb-Y:-NOK6ygu7lnVrfpJFh4o:boov682) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionTable.tsx#L21)):*

Btw, another way to write this without ternary. Not sure which one is pretier though.
```{(data || [].)map(...)}

<Visible show={!data.lenght} > no timetables </Visible}
```
</blockquote></details>

as discussed f2f, minor change but no Visible component. 
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 54 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ8trx5szYAa-LPqMa:-NOKLf3m6iXJ7qqcLcHe:brmux6m) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L54)):*

What is the point of having i18n field and then hardcoding the ui to use fi_FI? :wink: 
Please see `parseI18nField` from `ui/src/i18n/utils.ts`.

(By the way, not sure if these UI's should rather show all different translations to the user :thinking: . But that is out of scope of this PR, so just use `parseI18nField` here and we are good to go.)
</blockquote></details>

Check!
___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 4 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJYfHeALq799HjleUI:-NOKOiKXAmIR3VAg4Dg2:b-byc3h8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L4)):*

What does "version" mean in this context?
</blockquote></details>

Yes, it is unnecessary strict naming here, should be generic so I'll remove it.
___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 10 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJVAfo79GMC0aKLfMv:-NOKOuv77sLbJv80fPJD:b4qkeyp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L10)):*

dsaasdfafdsfadsadfsadfdsfaadfsadfadfdfadafswe :joy: 
</blockquote></details>

:proud: :man_facepalming: This absolutely broke me :laughing: 
___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 13 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJVSle8BR_cX_gKGjL:-NOKPKXnDRTDCMgLCub3:b-dnqxp8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L13)):*

if returnToId is optional, it can be undefined. Does this then add `?returnTo=undefined` to url? Maybe that is not what we want...
</blockquote></details>

:+1: good one
___
*[`ui/src/components/timetables/versions/TimetableVersionTableHeader.tsx` line 3 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJ6jIy-5aQd8IR3EWM:-NOK7VXH5caPvOfJeawu:b-u2pg5i) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/components/timetables/versions/TimetableVersionTableHeader.tsx#L3)):*

Does it improve readability of anything to have this in own componet and in a separate file? :shrug: 
</blockquote></details>

Good one, It is actually more readable to have these in the file used.
___
*[`ui/src/hooks/line-timetable-versions/index.ts` line 1 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJTO4SDUyewTst6_a6:-NOKMvvu8AvcYgWtLLWC:b9n8iwv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/line-timetable-versions/index.ts#L1)):*

Does it look like we are going to have more hooks here? If not, then own directory + index file for this is not very useful.
</blockquote></details>

:+1: Removed
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 75 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJWzZaAdPT5_MOauc3:-NOKNDKn-coxAVtAmsBR:b44falc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L75)):*

Optional comment, but function like this could definitely benefit from unit tests... Same for other chained mapping function in this file.
</blockquote></details>

Yeah, that is true. But the next step is to try and move this logic in SQL function, and also this will change anyway change after we get this implementation complete (including reference days) so would not use time for tests yet. But noted

tk-vincit (approved):
> Reviewed 6 of 19 files at r6.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 9 of 30 files reviewed, 7 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOO5i-gD4wRJi9qRIb3:-NOO5i-gD4wRJi9qRIb4:bc3fl41) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/84c6f0750d721aa1eb2f881e4eceaec46bf5eef2/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*
> ```XML
>       <div className="mx-12 my-8">
>         <FormRow mdColumns={2} className="mb-8">
>           <Row>
> ```

Do we need row inside FormRow? :raised_eyebrow: 
I assume passing some className to ObservationDateControl would actually do what you want to do without introducing new dom node.
Same applies to other similar places in this PR.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 23 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOOEr1XBYfJklbG58LS:-NOOEr1XBYfJklbG58LT:b-cywa45) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/84c6f0750d721aa1eb2f881e4eceaec46bf5eef2/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L23)):*
> ```XML
>     [TimetablePriority.Special]: 'bg-hsl-light-purple',
>     [TimetablePriority.Draft]: 'bg-background',
>     [TimetablePriority.Staging]: '', // Staging priority should not be used anywhere in the UI
> ```

I guess those are actually shown within timetables import preview, so not sure if this is a bit misleading?
___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 4 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJYfHeALq799HjleUI:-NOOFg8Y4C8PVWS11CKT:bjxivhu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L4)):*

Yes, it is unnecessary strict naming here, should be generic so I'll remove it.
</blockquote></details>

Now that I think of this, also "onClose" might be too specific information.
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 13 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkJoe1-aD0cE7P3kxy:-NOOFNAg0f1o11OKW-gW:byqli2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L13)):*

We have `unique_label` in route information, which includes route and variant. There is _ as a separator between the label and variant though, so it cant be used straight away. But just FYI there is pretty much similar field in the data model which is named unique label, if you want to utilize this info, it might affect this field naming and/or source.
</blockquote></details>

Was anything done regarding this? Should anything be done regarding this?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 7 of 30 files reviewed, 6 unresolved discussions (waiting on @eleksi and @tk-vincit)

___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOO5i-gD4wRJi9qRIb3:-NOOIdu63MkDNRZty72b:bhmxrmp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/84c6f0750d721aa1eb2f881e4eceaec46bf5eef2/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*

Do we need row inside FormRow? :raised_eyebrow: 
I assume passing some className to ObservationDateControl would actually do what you want to do without introducing new dom node.
Same applies to other similar places in this PR.
</blockquote></details>

Good point. Now that it is not anymore inside the component to make the layout of the component, this is confusing here. Should be changed in both places now. Here and in ActionsRow.tsx
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 23 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOOEr1XBYfJklbG58LS:-NOOjk4l0NTX4u1AtK7C:bhjtfwt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/84c6f0750d721aa1eb2f881e4eceaec46bf5eef2/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L23)):*

I guess those are actually shown within timetables import preview, so not sure if this is a bit misleading?
</blockquote></details>

Hmm. Yeah that might be. I'll just add the same colour that was in the other function. 
___
*[`ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts` line 13 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NNkJoe1-aD0cE7P3kxy:-NOOm3MT3jMGQ40fb5kR:b-cznwqf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/77287eaff54c35f031136052f9f3597ed3af0791/ui/src/hooks/line-timetable-versions/useGetLineTimetableVersions.ts#L13)):*

Was anything done regarding this? Should anything be done regarding this?
</blockquote></details>

Forgot to write anything there, but we discussed f2f at the office at one point and yes this unique_label could be used, but it is in a fixed form of "<label>_<variant>".
Also if we need more unique attributes, then they will be added in uniqueLabel as well. So I think we decided to go with this approach where we define it separately and we can decide how it is presented in the UI.
___
*[`ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts` line 4 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501#-NOJYfHeALq799HjleUI:-NOOmwSL42G_6dd37CIU:bxq7v8b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc533e5f9366577de698a7bff3ecf458bf06e95a/ui/src/hooks/ui/useVersionsOnCloseWithReturnToQueryParam.ts#L4)):*

Now that I think of this, also "onClose" might be too specific information.
</blockquote></details>

Good point :D I'll remove the "onClose" from the other hook names as well. Unfortunately the more specific hooks names got a bit longer now. But it is more descriptive I think.

tk-vincit (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/501)* status: 7 of 30 files reviewed, 3 unresolved discussions (waiting on @eleksi)

PR #520, opened 2023-02-21
Add Foresight Telemetry step to e2e test workflow

sr-vala:
Resolves HSLDevcom/jore4#1178

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/520)

### DIFF
--- a/.github/workflows/run-cypress-tests.yml
+++ b/.github/workflows/run-cypress-tests.yml
@@ -13,6 +13,12 @@ jobs:
       CYPRESS_IMAGE_NAME: hsldevcom/jore4-cypress
 
     steps:
+      - name: Collect Workflow Telemetry
+        uses: runforesight/foresight-workflow-kit-action@v1
+        if: ${{ always() }}
+        with:
+          api_key: ${{ secrets.FORESIGHT_GITHUB_ACTIONS }}
+
       - name: Checkout code
         uses: actions/checkout@v3
 


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/520)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @sr-vala)

___
*[`.github/workflows/run-cypress-tests.yml` line 16 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/520#-NOnSabn6yQMzYLlSt5m:-NOnSabn6yQMzYLlSt5n:b610rhp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/01e891c8661830f97d3335b77f0b7bcd78c0d993/.github/workflows/run-cypress-tests.yml#L16)):*
> ```YAML
> 
>     steps:
>       - name: Collect Workflow Telemetry
> ```

the telemetry doesn't seem to be working like this, it has to be added after the cypress test runs to be able to report anything. More info on the usage on the website of the github action: https://github.com/runforesight/foresight-test-kit-action#usage

Also, the current suggestion is to use `if: success() || failure()` instead of `if: always()`. After you moved it to the end of this workflow job, it should be tested that if the cypress tests fail then the whole job fails (so that this action doesn't "fix" the failing job)

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/520)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @pallost)

___
*[`.github/workflows/run-cypress-tests.yml` line 16 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/520#-NOnSabn6yQMzYLlSt5m:-NOnUKTo9_SosTaSw7v1:b-21x5iw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/01e891c8661830f97d3335b77f0b7bcd78c0d993/.github/workflows/run-cypress-tests.yml#L16)):*

the telemetry doesn't seem to be working like this, it has to be added after the cypress test runs to be able to report anything. More info on the usage on the website of the github action: https://github.com/runforesight/foresight-test-kit-action#usage

Also, the current suggestion is to use `if: success() || failure()` instead of `if: always()`. After you moved it to the end of this workflow job, it should be tested that if the cypress tests fail then the whole job fails (so that this action doesn't "fix" the failing job)
</blockquote></details>

This is the general telemetry action for the workflow, not the test coverage step that you linked. This also needs to be active always. We can add the test kit step too if it looks interesting, but I'm not so sure in the case of e2e tests.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/520)* status: 0 of 1 files reviewed, all discussions resolved

PR #529, opened 2023-03-03
Dataset builder poc

pallost:


pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/529)* status: 0 of 6 files reviewed, 4 unresolved discussions

___
*[`test-db-manager/src/datasets/timetables/dataset.ts` line 17 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/529#-NPaOIF65k34fmXbp-uJ:-NPaOIF65k34fmXbp-uK:b9zvvh3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b07e75859f2ed54e46229b7cbe0bfd9f8f2697fe/test-db-manager/src/datasets/timetables/dataset.ts#L17)):*
> ```XML
> import { seedVehicleServiceBlocksByName } from './vehicleServiceBlocks';
> 
> const buildStopInJourneyPatternRef = ({
> ```

let's separate the builders from the dataset. There will be multiple datasets for multiple test setups, but the builders should be reusable
___
*[`test-db-manager/src/datasets/timetables/dataset.ts` line 23 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/529#-NPaOtTFDc4DPsFKJpNt:-NPaOtTFDc4DPsFKJpNu:b-t4k6kh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b07e75859f2ed54e46229b7cbe0bfd9f8f2697fe/test-db-manager/src/datasets/timetables/dataset.ts#L23)):*
> ```XML
>   sequenceNumber,
> }: {
>   id?: UUID;
> ```

instead of making custom types for each builder, you could reuse the original types. If you want the type fields to be camelCase, you could use these converters: https://gist.github.com/kuroski/9a7ae8e5e5c9e22985364d1ddbf3389d
___
*[`test-db-manager/src/datasets/timetables/dataset.ts` line 27 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/529#-NPaPGKgETiDiBSHiTjh:-NPaPGKgETiDiBSHiTji:b-9bkc0r) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b07e75859f2ed54e46229b7cbe0bfd9f8f2697fe/test-db-manager/src/datasets/timetables/dataset.ts#L27)):*
> ```XML
>   label: string;
>   sequenceNumber: number;
> }): StopInJourneyPatternRefInsertInput => ({
> ```

I like this return type! Makes it much more reusable...
___
*[`test-db-manager/src/datasets/timetables/dataset.ts` line 41 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/529#-NPaPppsB4m9jyIqvZFD:-NPaPppsB4m9jyIqvZFE:bgrwvts) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b07e75859f2ed54e46229b7cbe0bfd9f8f2697fe/test-db-manager/src/datasets/timetables/dataset.ts#L41)):*
> ```XML
> }: {
>   journeyPatternRefId: UUID;
>   labels?: string[]; // Either `labelPrefix` or explicit `labels` should be given.
> ```

I'd rather create two separate builder functions for these. If you have the label array, you don't need the prefix neither the stopsToCreate. It will be just a simple mapping function from labels to stopInJourneyPatterns. For the prefix setup though, you'll need both the labelPrefix and the stopsToCreate parameters.

But you could just split the builders like this:
```
buildStopSequenceFromLabels = (labels, journeyPatternRefId) => 
labels.map((label, sequence) => buildStopInJourneyPattern(label, journeyPatternRefId, sequence+1));

buildLabelArray = (labelPrefix, labelsToCreate) =>
range(0, labelsToCreate).map(no => `${.....}`)

buildStopSequenceWithLabelPrefix = (labelPrefix, stopsToCreate, journeyPatternRefId) =>
buildStopSequenceFromLabels(buildLabelArray(labelPrefix, labelsToCreate), journeyPatternRefId);
```

I mean... I think we should rather create nicely composable functions as builders and then compose them in ways that we want to use them

PR #531, opened 2023-03-07
Generate test data

pallost:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531)

Closes https://github.com/HSLdevcom/jore4/issues/1191


Leitsi (changes requested):
What is the general plan how does this relate to the current seed data? Since currently this does not replace it. And to me looks like it would be a bit difficult to replace the current seed (which is actually used as test data) with this kind of builder.

Would seed.ts and seed2.ts both live in our code base, and we would use seed2.ts as seed data for UI development / demoing, and seed.ts as test data? One problem I see with this is that our test/seed data builder code is a bit fragmented (UI & Hasura have their own tools), so this would just add to this problem (one more way to do this).

If we plan to replace the current seed builder code, then in my opinion that should already be in the scope of this pr. This would also demonstrate how easily this builder code is usable for test scenarios, which usually have a bit different requirements than generating lots of seed data.

> Reviewed 2 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-:-NQxrYZi10EPVvSSNjwl:b-92ymfw)* status: 20 of 24 files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala and @pallost)

___
*[`test-db-manager/src/seed2.ts` line 37 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxoc6zEmwyHs_Idqww:-NQxoc6zEmwyHs_Idqwx:b-hf69f6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L37)):*
> ```XML
>   const jp1 = 'a9136ad8-d185-4c7b-9969-057b65dc9b00';
>   const jpRef1 = buildJourneyPatternRefDeep(jp1, {
>     jpBase: {},
> ```

I would rather not use abbreviations in property names
___
*[`test-db-manager/src/seed2.ts` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjek4zEeEkLNKlZnK1:-NQxp32KAS2ym4s4fZjS:b-6ke0ru) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L50)):*

I think we could also use a bit more closer to real validity ranges here. Like half a year at a time. 

If the reason for this 8 year period is that we don't need to update this in 8 years, then I would suggest that we use "DateTime.now() + 6months" -kind of validity periods. 

Not sure if that is bad somehow, but at least it would always generate the same duration for the validity periods related to the seed date.
</blockquote></details>

If we want to dynamically adjust the ranges like this, instead of "now + x time" I would maybe use eg. "start of this year - end of this year" type of ranges, so at least the dates/months are always the same.

Leitsi (changes requested):
Discussed about this at the office. Ignore the part about seed being used in tests in my earlier comment, that is incorrect.
So, for now this would live as an alternative to the seed.ts. SGTM.

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-:-NQy-17J2RTP6-sGrqvc:bglvf8q)* status: 16 of 25 files reviewed, 12 unresolved discussions (waiting on @HenrikHartiala and @pallost)

___
*[`test-db-manager/src/builders/timetables/block.ts` line 51 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxueyB0_Yo5bTAY-B3:-NQxueyB0_Yo5bTAY-B4:b582kms) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L51)):*
> ```XML
> };
> 
> export const getBlockTimes = (block: BlockInsertInputDeep) => {
> ```

Should we only export functions that are intended to be used also from outside this file? I would maybe rather not export internal helpers like this one. Or is this an established practice in our code base?

Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531)* status: 16 of 25 files reviewed, 14 unresolved discussions (waiting on @HenrikHartiala and @pallost)

___
*[`test-db-manager/src/builders/timetables/block.ts` line 118 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy1_S8DTUgnHJ1FgKY:-NQy1_S9DTr9VeP3d66u:b-ysawtv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/32c1103ab50b24ed028a2ae2cee098cd975b740d/test-db-manager/src/builders/timetables/block.ts#L118)):*
> ```XML
> };
> 
> export const flattenBlock = (
> ```

Would it make sense to put all of these `flatten` methods to one file? Since they often have deal with many tables, call other flatten methods, and relate to the whole dataset rather than single tables.
___
*[`test-db-manager/src/types/types.d.ts` line 24 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy2mb-0EgeytfsN7yr:-NQy2mb-0EgeytfsN7ys:bhp8578) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/32c1103ab50b24ed028a2ae2cee098cd975b740d/test-db-manager/src/types/types.d.ts#L24)):*
> ```XML
>   ? `${T extends Capitalize<T> ? '_' : ''}${Lowercase<T>}${CamelToSnakeCase<U>}`
>   : S;
> type KeysToSnakeCase<T> = {
> ```

These snake conversion types are not used as far as I can see, so just remove? Also sounds like that kind of magic that I would rather avoid.

tk-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531)* status: 16 of 25 files reviewed, 22 unresolved discussions (waiting on @HenrikHartiala, @Leitsi, and @pallost)

___
*[`test-db-manager/src/seed2.ts` line 8 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxw4rp1o3inHXlYyht:-NQxw4rp1o3inHXlYyhu:b-hzb3n) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L8)):*
> ```XML
>   buildVehicleScheduleFrameDeep,
>   defaultBlockSeqParams,
>   defaultTptSeqParams,
> ```

Same as Leitsi mentioned before, I would avoid abbreviations.



_Code quote:_
```XML
  defaultTptSeqParams,
  defaultVjSeqParams,
  defaultVsSeqParams,
```
___
*[`test-db-manager/src/seed2.ts` line 37 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxoc6zEmwyHs_Idqww:-NQxvtWF3bxP8mULfVAF:b-methix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L37)):*

I would rather not use abbreviations in property names
</blockquote></details>

I agree. Since we have opted for transmodel approach and using verbose names all over our codebase, I don't see a reason why to break that with seed data.
___
*[`test-db-manager/src/seed2.ts` line 80 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy-TNx7z6ofOhbi6oG:-NQy-TNx7z6ofOhbi6oH:blo49kx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L80)):*
> ```XML
>       [SUN_DAY_TYPE]: {
>         ...defaultTptSeqParams,
>         ...defaultVjSeqParams,
> ```

These could also be extracted to a local variable to avoid code duplication, at least if you can figure out suitable name for these



_Code quote:_
```XML
        ...defaultTptSeqParams,
        ...defaultVjSeqParams,
        ...defaultBlockSeqParams,
        ...defaultVsSeqParams,
```
___
*[`test-db-manager/src/builders/common.ts` line 12 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy0DpK-qJphNNOlu5w:-NQy0DpK-qJphNNOlu5x:b-96ykyj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/common.ts#L12)):*
> ```XML
>   Duration.fromMillis(randomInt(min.toMillis(), max.toMillis()));
> 
> export type ConstantCount = number;
> ```

At least these two are used nowhere else, so no need to export those



_Code quote:_
```XML
export type ConstantCount = number;
export type RandomCount = { min: number; max: number };
```
___
*[`test-db-manager/src/builders/common.ts` line 31 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy12YWCw4fe6xWLWUC:-NQy12YWCw4fe6xWLWUD:bq074kb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/common.ts#L31)):*
> ```XML
>   index: number,
> ): T => {
>   if (method === 'modulo') {
> ```

We could also write this in type-safe way so that complier would already warn us before ending up in throw step:
https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript



_Code quote:_
```XML
  if (method === 'modulo') {
    return array[index % array.length];
  }
  if (method === 'random') {
    return array[randomInt(0, array.length)];
  }
  throw new Error(`Unknown array picker method: ${method}`);
```
___
*[`test-db-manager/src/builders/timetables/block.ts` line 51 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxueyB0_Yo5bTAY-B3:-NQy1bPqCBxqhS9z2dJE:b2s9ms1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L51)):*

Should we only export functions that are intended to be used also from outside this file? I would maybe rather not export internal helpers like this one. Or is this an established practice in our code base?
</blockquote></details>

It is not established practice to export things unless export is needed. So good catch & point.
___
*[`test-db-manager/src/builders/timetables/block.ts` line 54 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy2lpw8MaQNnzTAGTi:-NQy2lpw8MaQNnzTAGTj:bcfs0qk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L54)):*
> ```XML
>   const vehicleJourneys = block.vehicle_journeys.data;
> 
>   const { start } = getVehicleJourneyTimes(vehicleJourneys[0]);
> ```

Do we always have at least two items in this array?
___
*[`test-db-manager/src/builders/timetables/timeSequence.ts` line 58 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjjtafD_rCKv_LA8Q_:-NQy3x1H0wvKrpy-Xfr5:b-vq8oge) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/timeSequence.ts#L58)):*

Is there a possibility that this function is called with empty `timeSequence`? In that case this would need check length check to avoid exception.
</blockquote></details>

Yeah, I found a similar problem before (in this PR).
As this is "only" for generating seed data this kind of things are not critical, but still worth to be pointed out in PR.
___
*[`test-db-manager/src/builders/timetables/timetabledPassingTime.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy4aCEEQtvrsF_IKqJ:-NQy4aCF5q5qNEJCGHoh:b-ampx39) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/timetabledPassingTime.ts#L20)):*
> ```XML
>  * Builds a single timetables passing time
>  * @param vehicleJourneyId the vehicle journey this timetables passing time belongs to
>  * @param tptBase optional attributes for the timetables passing times
> ```

Second parameter is not documented. Even if it should be self-evident, having it here would make the documentation easier to read.
By the way, does intellisense go bananas if middle parameter is missing, or does it handle this well?
___
*[`test-db-manager/src/types/types.d.ts` line 29 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy7ytz1-X_ERpLnp_S:-NQy7ytz1-X_ERpLnp_T:bwj0igy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/types/types.d.ts#L29)):*
> ```XML
> 
> // makes the K keys within the T object required, leaves the rest as-is
> type RequiredKeys<T, K extends keyof T> = Required<Pick<T, K>> & Omit<T, K>;
> ```

Already defined in `test-db-manager/src/types/common.ts`.
Maybe these should be rather defined in that file, so that one would have to require these if needed, instead than defining these to global scope (as done now).
___
*[`test-db-manager/src/utils/random.ts` line 14 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy8ilS2YeNb0vSVg_-:-NQy8ilS2YeNb0vSVg_0:b-ys26rv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/utils/random.ts#L14)):*
> ```XML
> 
> /**
>  * Generates a pseudorandom integer within the given interval. Note that the max is never reached.
> ```

Not a big thing, but this would be more intuitive to the user if max could be reached. Could that be archieved just by replacing `(max` with `(max + 1`?

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531)* status: 16 of 25 files reviewed, 21 unresolved discussions (waiting on @HenrikHartiala, @Leitsi, and @tk-vincit)

___
*[`.github/workflows/check-seed-generators.yml` line 21 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjcWr4CCJ7G76QIXsh:-NQxlmbR5mKZR0nyFcov:bxch6z6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/.github/workflows/check-seed-generators.yml#L21)):*

Is there a way to give this as variable (or is it even an issue)? What I mean is that if we update node, do we need to remember to change this as well? Or does it even matter?
</blockquote></details>

well, this is also pinned in package.json. So if we update nodejs there but forget to increment it here, it'll complain (also vice-versa). So we don't need to parameterize this, luckily
___
*[`test-db-manager/src/seed2.ts` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjek4zEeEkLNKlZnK1:-NQxp4s933e6x1ryn2uQ:b-mxur0b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L50)):*

If we want to dynamically adjust the ranges like this, instead of "now + x time" I would maybe use eg. "start of this year - end of this year" type of ranges, so at least the dates/months are always the same.
</blockquote></details>

well, in case of UI seed data it's fine if it's not deterministic... I'll change it to dynamic. 
___
*[`test-db-manager/src/seed2.ts` line 91 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjeH1b4k47xM7bbDda:-NQxrgtq-5MgVyld-1S2:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L91)):*

I think this comment should be 2022-2023

(Also same comment here as above, maybe we should use some closer to realistic validity period here, so like 1-2months maybe?)
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/block.ts` line 51 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxueyB0_Yo5bTAY-B3:-NQy8Ibu-KvPb9KrRzh0:bm6a4qs) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L51)):*

Should we only export functions that are intended to be used also from outside this file? I would maybe rather not export internal helpers like this one. Or is this an established practice in our code base?
</blockquote></details>

this is a reusable piece of helper that could be used outside of this file in other builder, so I'd rather leave it exported. The other similar helper (getVehicleJourneyTimes) is already reused, here
___
*[`test-db-manager/src/builders/timetables/block.ts` line 118 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy1_S8DTUgnHJ1FgKY:-NQy8pnz0XVR2P_8ZHax:bjdgrix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/32c1103ab50b24ed028a2ae2cee098cd975b740d/test-db-manager/src/builders/timetables/block.ts#L118)):*

Would it make sense to put all of these `flatten` methods to one file? Since they often have deal with many tables, call other flatten methods, and relate to the whole dataset rather than single tables.
</blockquote></details>

well, I'm not sure... but I can do it if you wish
___
*[`test-db-manager/src/builders/timetables/timeSequence.ts` line 17 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjgoxNDS2SztV3rxZT:-NQxriNo1Nf1GQJbC6Tv:b-ax8r28) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/timeSequence.ts#L17)):*

Not very important, but we could use `i+=1` here to avoid the eslint-disable? (This eslint rule is new for me though)
</blockquote></details>

I'll actually rather exclude this stupid rule from eslint. The for loop increments are traditionally done with ++, I don't wanna change that :)
___
*[`test-db-manager/src/builders/timetables/timeSequence.ts` line 58 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjjtafD_rCKv_LA8Q_:-NQyRv-BBUbnl0RluOaN:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/timeSequence.ts#L58)):*

Yeah, I found a similar problem before (in this PR).
As this is "only" for generating seed data this kind of things are not critical, but still worth to be pointed out in PR.
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/timetabledPassingTime.ts` line 88 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjlGLH2GGCJ9I0P8Lu:-NQy-8Z_E3-LBt4eNdQs:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/timetabledPassingTime.ts#L88)):*

This seems little bit confusing. Could we use a named constant for this?
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/vehicleJourney.ts` line 77 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjmFRV5_LzSGWsttF3:-NQy4YSg7oTKtoDXrFwu:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/vehicleJourney.ts#L77)):*

Could benefit from a comment why non null assertion is safe.
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/vehicleJourney.ts` line 123 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjmdA574bKwzu5IvYD:-NQy57Yg90C2qQlAVmgo:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/vehicleJourney.ts#L123)):*

Are these supposed to be questions or explanations? 
If latter, then I would rephrase these questions in to something like "The amount of vehicle journeys to generate" and so on..

Same for all the question comments in these files
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/vehicleService.ts` line 103 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjnovg24K6msL7dO6U:-NQy6ZAe4oooHJBgB1Ka:btuwfe4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/vehicleService.ts#L103)):*

Is there another way to give these parameters than this `...rest`? When reviewing, I have no idea what this consists of.
</blockquote></details>

well, the structure goes such that there's a massive build context variable that contains all the information required to build all entities. Each function only uses only a fraction of that (these variables are named in the destructuring), the others can be passed on as-is. I'll rename the `...rest` variables to something better, maybe that helps...?
___
*[`test-db-manager/src/types/types.d.ts` line 24 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy2mb-0EgeytfsN7yr:-NQy73jg7ANXXvlX1wT7:b-lin212) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/32c1103ab50b24ed028a2ae2cee098cd975b740d/test-db-manager/src/types/types.d.ts#L24)):*

These snake conversion types are not used as far as I can see, so just remove? Also sounds like that kind of magic that I would rather avoid.
</blockquote></details>

you are right, I'll remove them.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531)* status: 12 of 28 files reviewed, 19 unresolved discussions (waiting on @HenrikHartiala, @Leitsi, and @tk-vincit)

___
*[`test-db-manager/src/seed2.ts` line 8 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxw4rp1o3inHXlYyht:-NR2k9whCbrECZwGkUVG:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L8)):*

Same as Leitsi mentioned before, I would avoid abbreviations.
</blockquote></details>

Done.
___
*[`test-db-manager/src/seed2.ts` line 37 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxoc6zEmwyHs_Idqww:-NR2GDCk52azjbb6g5qZ:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L37)):*

I agree. Since we have opted for transmodel approach and using verbose names all over our codebase, I don't see a reason why to break that with seed data.
</blockquote></details>

Done.
___
*[`test-db-manager/src/seed2.ts` line 80 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy-TNx7z6ofOhbi6oG:-NR2ZaMpCnEAHwieIHyw:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/seed2.ts#L80)):*

These could also be extracted to a local variable to avoid code duplication, at least if you can figure out suitable name for these
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/common.ts` line 12 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy0DpK-qJphNNOlu5w:-NR2_fvBBvyWusz0kx9G:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/common.ts#L12)):*

At least these two are used nowhere else, so no need to export those
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/common.ts` line 31 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy12YWCw4fe6xWLWUC:-NR2_gcx9FkHQZbt9eR_:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/common.ts#L31)):*

We could also write this in type-safe way so that complier would already warn us before ending up in throw step:
https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/block.ts` line 54 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy2lpw8MaQNnzTAGTi:-NR2mq_qBV8M_-Uh2TDZ:br3xty9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L54)):*

Do we always have at least two items in this array?
</blockquote></details>

good point. Handled the special cases where there are no stops in a vehicle journey... it looks pretty nasty now :grimacing: 
___
*[`test-db-manager/src/builders/timetables/timetabledPassingTime.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy4aCEEQtvrsF_IKqJ:-NQySHKq938lBfVu0xUQ:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/timetabledPassingTime.ts#L20)):*

Second parameter is not documented. Even if it should be self-evident, having it here would make the documentation easier to read.
By the way, does intellisense go bananas if middle parameter is missing, or does it handle this well?
</blockquote></details>

Done.
___
*[`test-db-manager/src/types/types.d.ts` line 29 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy7ytz1-X_ERpLnp_S:-NQySc6fB-69-2oWGP7Y:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/types/types.d.ts#L29)):*

Already defined in `test-db-manager/src/types/common.ts`.
Maybe these should be rather defined in that file, so that one would have to require these if needed, instead than defining these to global scope (as done now).
</blockquote></details>

Done.
___
*[`test-db-manager/src/utils/random.ts` line 14 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy8ilS2YeNb0vSVg_-:-NQySekV15eWES0N8Qbk:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/utils/random.ts#L14)):*

Not a big thing, but this would be more intuitive to the user if max could be reached. Could that be archieved just by replacing `(max` with `(max + 1`?
</blockquote></details>

Done.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531)* status: 14 of 28 files reviewed, 15 unresolved discussions (waiting on @HenrikHartiala, @Leitsi, and @tuokait)

___
*[`test-db-manager/src/builders/timetables/block.ts` line 51 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQxueyB0_Yo5bTAY-B3:-NRI7XOY4J76RUiuLgjM:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L51)):*

this is a reusable piece of helper that could be used outside of this file in other builder, so I'd rather leave it exported. The other similar helper (getVehicleJourneyTimes) is already reused, here
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/block.ts` line 54 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy2lpw8MaQNnzTAGTi:-NRI7a1U11ynCk2swwbQ:b-hbig3w) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L54)):*

Would we ever have vehicle journeys with no stops? Is that a developer error? 
Could this function check that the vehicleJourney have at least two stops and then if not, throw an error  (or if it is suitable case, then just use early return) etc.
Does our database even allow adding vehicleJourneys with less than 2 stops?
</blockquote></details>

If we ever wanted to use this setup to generate test data for integration tests, then yes, we might want to have an empty vehicle journey. Then in the integration test we could manually determine what kind of timetabled passing times we want to create under this vj (e.g. for testing the tpt ordering constraint)
___
*[`test-db-manager/src/builders/timetables/block.ts` line 70 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NR8NNFV-EVuNjZzbB56:-NRI88uY3qGdmFg_X1eI:bp35x4t) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/87d1d2930f29d2469f2f7f5a3f7db834ad4644cb/test-db-manager/src/builders/timetables/block.ts#L70)):*

Btw, us this supposed to be the last starting time and not the last ending time? If it is the last starting time, then maybe we should change the variable namings here to "firstStart" and "lastStart" instead of "start" and "end".
</blockquote></details>

very good catch! This is supposed to be "end". I will also rename the start and end variables for clarity.
___
*[`test-db-manager/src/builders/timetables/vehicleJourney.ts` line 156 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NR8Role4cobmG_29BJj:-NRI8a-K1hnUmMIbZOQ5:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/87d1d2930f29d2469f2f7f5a3f7db834ad4644cb/test-db-manager/src/builders/timetables/vehicleJourney.ts#L156)):*

This is also in question format
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/vehicleJourney.ts` line 165 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NR8RzmUDnnlDm5c3_kL:-NRI8fAB0HGqHm24bF4F:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/87d1d2930f29d2469f2f7f5a3f7db834ad4644cb/test-db-manager/src/builders/timetables/vehicleJourney.ts#L165)):*

This is also in question format
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/vehicleService.ts` line 103 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQjnovg24K6msL7dO6U:-NRI9BSp6-6QmybfcPyh:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/vehicleService.ts#L103)):*

Yes, thanks! This helped :) 
</blockquote></details>

Done.
___
*[`test-db-manager/src/builders/timetables/vehicleService.ts` line 83 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NR8TpEA9xwusBqt4dtU:-NRI93bz6zqjbKDWQZF8:b-ac41gu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/87d1d2930f29d2469f2f7f5a3f7db834ad4644cb/test-db-manager/src/builders/timetables/vehicleService.ts#L83)):*

Is this rename intentional or was it the traditional result of vsCode's rename?
</blockquote></details>

yeah, vsCode rename result... didn't notice it, now it's fixed, also in other files

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531)* status: 14 of 28 files reviewed, 9 unresolved discussions (waiting on @Leitsi and @tuokait)

___
*[`test-db-manager/src/builders/timetables/block.ts` line 54 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NQy2lpw8MaQNnzTAGTi:-NRIgezT6VA0C8Hr47R8:b-tw1qou) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c9d328ec7f11cc23fcd48e611b14299487daaf23/test-db-manager/src/builders/timetables/block.ts#L54)):*

If we ever wanted to use this setup to generate test data for integration tests, then yes, we might want to have an empty vehicle journey. Then in the integration test we could manually determine what kind of timetabled passing times we want to create under this vj (e.g. for testing the tpt ordering constraint)
</blockquote></details>

Ah, I see :+1: 
___
*[`test-db-manager/src/builders/timetables/vehicleService.ts` line 83 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/531#-NR8TpEA9xwusBqt4dtU:-NRIhWkT7biNmWbHj3uf:brm9vcg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/87d1d2930f29d2469f2f7f5a3f7db834ad4644cb/test-db-manager/src/builders/timetables/vehicleService.ts#L83)):*

yeah, vsCode rename result... didn't notice it, now it's fixed, also in other files
</blockquote></details>

Happens to me as well from time to time :D

PR #553, opened 2023-03-15
Change route export to use uniqueLabel

HenrikHartiala:
Also rename RouteInformationForMap and LineInformationForMap to RouteMapParams and LineMapParams

Resolves HSLdevcom/jore4#1150

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553)

### DIFF
--- a/codegen/schemaLoader.ts
+++ b/codegen/schemaLoader.ts
@@ -28,6 +28,10 @@ const hasuraOverrideTypeDefs = gql`
     destination_short_name_i18n: localized_string!
     route_shape: geography_linestring
     route_line: route_line!
+    # TODO: currently the unique_label is overridden here to be not nullable, but this should be
+    # done in hasura. unique_label field is computed field and it includes label field, which is
+    # not nullable, so this can't ever be null.
+    unique_label: String!
   }
   type route_route_set_input {
     name_i18n: localized_string
--- a/test-db-manager/src/generated/graphql.ts
+++ b/test-db-manager/src/generated/graphql.ts
@@ -6254,7 +6254,7 @@ export type RouteRoute = {
   /** A computed field, executes function "route.route_shape" */
   route_shape?: Maybe<Scalars['geography_linestring']>;
   /** Derived from label and variant. Routes are unique for each unique label for a certain direction, priority and validity period */
-  unique_label?: Maybe<Scalars['String']>;
+  unique_label: Scalars['String'];
   /** The point in time from which onwards the route is no longer valid. If NULL, the route is valid indefinitely after the start time of the validity period. */
   validity_end?: Maybe<Scalars['date']>;
   /** The point in time when the route becomes valid. If NULL, the route has been always valid before end time of validity period. */
--- a/ui/graphql.schema.json
+++ b/ui/graphql.schema.json
@@ -39012,9 +39012,13 @@
             "description": "Derived from label and variant. Routes are unique for each unique label for a certain direction, priority and validity period",
             "args": [],
             "type": {
-              "kind": "SCALAR",
-              "name": "String",
-              "ofType": null
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              }
             },
             "isDeprecated": false,
             "deprecationReason": null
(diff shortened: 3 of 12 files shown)


tk-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553#-NQdHZbbEW3V8jL8uLeo):*
Could have been in a separate commit. :wink: 
___
*[`codegen/schemaLoader.ts` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553#-NQdHiyZ4G4zDCNvpW4j:-NQdHiyZ4G4zDCNvpW4k:bas2upj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/12880225c90a44b08e56d95ab7f7fdac81c71298/codegen/schemaLoader.ts#L11)):*
> ```XML
>  */
> 
> // TODO: currently the unique_label is overridden here to be not nullable, but this should be
> ```

Ideally this comment would be in the same row than the override itself, and the second best alternative would be to move this right above that row.

Now this is very far away from the row which this concerns, so it is very likely that this gets forgotten if `unique_label: String!` row is e.g. removed at some point.
___
*[`ui/src/components/common/LineTableRow.tsx` line 61 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553#-NQdJQklBIXvxwC5Ml-9:-NQdJQklBIXvxwC5Ml-A:b-ez2qyo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/12880225c90a44b08e56d95ab7f7fdac81c71298/ui/src/components/common/LineTableRow.tsx#L61)):*
> ```XML
>   const { showRoutesOnMapByLineLabel } = useShowRoutesOnModal();
>   const dispatch = useAppDispatch();
>   const { selectedRouteUniqueLabels: selectedRouteLabels } =
> ```

Is this renamed here for a reason?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @tk-vincit)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553#-NQdHZbbEW3V8jL8uLeo):*

Could have been in a separate commit. :wink: 
</blockquote></details>

Yep, I realised this too late because both commits modify generated code, so I was lazy and asked Eetu at the office if it was okay to push these both to same commit instead of go through the generated code, or stash the modifications and re-generate the code. 
___
*[`codegen/schemaLoader.ts` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553#-NQdHiyZ4G4zDCNvpW4j:-NQdQZ_G4w8FsrQbGGtZ:b-nxpa2k) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/12880225c90a44b08e56d95ab7f7fdac81c71298/codegen/schemaLoader.ts#L11)):*

Ideally this comment would be in the same row than the override itself, and the second best alternative would be to move this right above that row.

Now this is very far away from the row which this concerns, so it is very likely that this gets forgotten if `unique_label: String!` row is e.g. removed at some point.
</blockquote></details>

The reason why it was on top was because I was thinking that you can't use `//` comments inside that gql template literal. But right there is hashtag for that.
___
*[`ui/src/components/common/LineTableRow.tsx` line 61 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/553#-NQdJQklBIXvxwC5Ml-9:-NQdQINI-17MTnWJBTYE:bxzgw16) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/12880225c90a44b08e56d95ab7f7fdac81c71298/ui/src/components/common/LineTableRow.tsx#L61)):*

Is this renamed here for a reason? 
</blockquote></details>

Good spot, when using the F2(rename) feature on visual studio, it sometimes does this renaming and it seems I did not spot all of them. Thanks! 

Found another like this in `RouteTableRow` and fixed it too.

PR #578, opened 2023-04-12
Show arrival times toggle

Perholehto:


eleksi (changes requested):
> Reviewed 6 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578)* status: 7 of 8 files reviewed, 6 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStxtXrE0gfqbde0mzq):*
We like to use present for of the verb in commit messages instead of past, as you used in the other "Add...". So this could also be "Fix..."
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 25 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStyIzk5UYNDG8jMxdu:-NStyIzl8NUBaucELczf:b-5nopmb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L25)):*
> ```XML
> export const VehicleScheduleDetailsPage = (): JSX.Element => {
>   const dispatch = useDispatch();
> ```

Could use `useAddDispatch()` which is typed better.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSu-plY6liqULWZEHij:-NSu-plY6liqULWZEHik:b5mwqv3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*
> ```XML
>       : line?.line_routes.filter((route) => route.label === routeLabel) || [];
> 
>   const colSpanAmount =
> ```

I cannot see straight away what is up with this dynamic span amount, but I wonder if this could be achieved with any less complexity. If you really want to keep using grid layout instead of flex, maybe @HenrikHartiala can check this and see if there is a better way.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStzw776fMRaUemLi09:-NStzw776fMRaUemLi0A:bdbx9hw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L72)):*
> ```XML
>         <FormRow mdColumns={6} className="mb-8 ">
>           <ObservationDateControl className="max-w-max" />
>           {activeView === TimetablesView.PASSING_TIMES_BY_STOP && (
> ```

We have `Visible` component that we use for showing and hiding stuff. It makes a code a little cleaner. I know this component does not use it in other places either, but if you want you can change all those as well.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 83 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStzS-y7m_bJO9XZvLl:-NStzS-y7m_bJO9XZvLm:b69096i) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L83)):*
> ```XML
>                     dispatch(setShowArrivalTimesAction(enabled))
>                   }
>                   testId="show-unused-stops-switch"
> ```

This is not "show-unused-stops-switch" ;).

We usually make a constant object out of testids and prefix it with parent component name. Example below:



_Code snippet:_
```XML
const testIds = {
  showArrivalTimesSwitch: 'VehicleScheduleDetailsPage::showArrivalTimesSwitch',
};
```
___
*[`ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTableRow.tsx` line 26 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStynaqEWEYe6_QMEkg:-NStynaqEWEYe6_QMEkh:b95deuq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTableRow.tsx#L26)):*
> ```XML
>         {minutes.map((item, key) => (
>           // eslint-disable-next-line react/no-array-index-key
>           <span key={`${item}-${key}`}>{padToTwoDigits(item)}</span>
> ```

Nice that you fixed this, it was annoying console warning.

HenrikHartiala (changes requested):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578)* status: all files reviewed, 7 unresolved discussions (waiting on @eleksi and @Perholehto)

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStxtXrE0gfqbde0mzq):*

We like to use present for of the verb in commit messages instead of past, as you used in the other "Add...". So this could also be "Fix..."
</blockquote></details>

Yeah, this way it says what the commit _will_ do if applied

Second thing, which I believe we haven't mentioned yet (nor do I think is documented anywhere) is that the Structure of the commit is 
```
Title of commit

Body of commit

Resolves HSLdevcom/jore4#<ticketnumber>
```
This way, we can see what is the "specification" of the commit and also it will automatically move the ticket on the board when this PR is merged to main. There are of course sometimes commits that do not resolve any specific tickets, but if there is a corresponding ticket, then it's nice to add here.

So for the `- ccc24d7: Add 'show arrival times'-toggle to timetable -commit`, you could add `Resolves HSLdevcom/jore4#1196` as I believe that is the corresponding ticket
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSuYOm8BO9FsBs18Nuv:-NSuYOm8BO9FsBs18Nuw:bkxehs4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx#L72)):*
> ```XML
>       >
>         <span className="flex flex-col">
>           <Visible visible={displayArrival}>
> ```

I know you did not do any changes here, 
**but** currently when toggling the arrivalTimes, the timetabled minutes and hours are  jumping around because now the toggle will dynamically change this condition, and the arrivalTimes are reserving space from the DOM and when they are not rendered at all, the layout changes.
So my suggestion is modify this as follows:
```
 <span
            className={`text-2xs leading-tight ${displayArrival ? '' : 'opacity-0'}`}
          >
            {displayedArrival}
</span>
```
So something like this. So with this cahnge we just modify the opacity to 0, so the reserved space won't change and everything will be in the same place regardless of the toggle state.

What do you think?
Also ping @eleksi , would like to hear your thought on this as well.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSu-plY6liqULWZEHij:-NSuFyB8-ryxI-gDV6dJ:b-jxw3j2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*

I cannot see straight away what is up with this dynamic span amount, but I wonder if this could be achieved with any less complexity. If you really want to keep using grid layout instead of flex, maybe @HenrikHartiala can check this and see if there is a better way.
</blockquote></details>

This will be a bit longer explanation with couple of different solutions, but ask if it is not clear :D.

Yeah, there is currently a minor problem with this, which I also stumbled on when I did my first changes to the UI a year ago :D. We have tailwind-css in use, which actually scans ALL the files and tries to find the tailwind classnames it can find. And then it generates the classtypes for those that it finds. It is an optimization thing. 

Now you create this col-span-4 or col-span-5 dynamically by changing only the number here. So when tailwind-css scans the files, it does not see "col-span-4" or "col-span-5", so it will not create the corresponding classnames for these, and hence this does not work. 

To make it work dynamically, you would need to do it this way:
```
  const colSpanClassName =
    activeView === TimetablesView.PASSING_TIMES_BY_STOP ? 'col-span-4' : 'col-span-5';
```
But as Eetu said, this might be a little bit unnecessarily complex implementation for the thing you are trying to achieve. So now you are using span 4 or span 5 depending on the view. But the mdColumn amount is a fixed 6, and we want to have the "Show versions" button in the right edge of the page. So maybe the best way is to either use `col-start-6` or `col-start-5 col-span-2`.

**One more solution** is to keep the mdColumn amount at 2, and just add this "showArrivalTimes" -part in the first column with the "ObservationDate"

FYI: We also have a bit of a mess with the "Column" and "FormColumn" -components, which we should fix at some point, but that part is maybe better to explain f2f.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 73 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSuQMDJ7hZh1Bz8ofUQ:-NSuQMDKDn5TWseYU6MS:bm8xx2b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L73)):*
> ```XML
>           <ObservationDateControl className="max-w-max" />
>           {activeView === TimetablesView.PASSING_TIMES_BY_STOP && (
>             <div className="justify-normal flex items-end">
> ```

I would maybe use "space-x-4" here rather then the "mr-4" in the SwitchLabel, but I think it is fine either way.
Also maybe use "col-span-2" for this as well, but this might benefit for a minor refactor in the layout anyway, so this is fine as well.

eleksi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578)* status: all files reviewed, 7 unresolved discussions (waiting on @HenrikHartiala and @Perholehto)

___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSuYOm8BO9FsBs18Nuv:-NSySERK8TMZHspgUmmM:brrsahx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx#L72)):*

Also the example above is just a POC  :D.
</blockquote></details>

I didn't test this branch so I didn't notice the jumping, good spot.

I'm not quite sure what we want to to with this, as hiding the arrival times also could be used to save space. If the same amount of space is used anyway, it really doesn't help with that. I think some jumping could be ok, as we clearly want to show different amount of information. But then again, if one switches arrival times on and off and tries to look at specific stop row, it could be hard if the stop row changes position.

I don't have strong opinion, I think this is also ok as is.

Btw. if you decide to go with Henkka's solution, use `invisible` instead of `opacity-0`
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSu-plY6liqULWZEHij:-NSyUM_wE8qS_K-87GYH:b6j2bjl) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*

This will be a bit longer explanation with couple of different solutions, but ask if it is not clear :D.

Yeah, there is currently a minor problem with this, which I also stumbled on when I did my first changes to the UI a year ago :D. We have tailwind-css in use, which actually scans ALL the files and tries to find the tailwind classnames it can find. And then it generates the classtypes for those that it finds. It is an optimization thing. 

Now you create this col-span-4 or col-span-5 dynamically by changing only the number here. So when tailwind-css scans the files, it does not see "col-span-4" or "col-span-5", so it will not create the corresponding classnames for these, and hence this does not work. 

To make it work dynamically, you would need to do it this way:
```
  const colSpanClassName =
    activeView === TimetablesView.PASSING_TIMES_BY_STOP ? 'col-span-4' : 'col-span-5';
```
But as Eetu said, this might be a little bit unnecessarily complex implementation for the thing you are trying to achieve. So now you are using span 4 or span 5 depending on the view. But the mdColumn amount is a fixed 6, and we want to have the "Show versions" button in the right edge of the page. So maybe the best way is to either use `col-start-6` or `col-start-5 col-span-2`.

**One more solution** is to keep the mdColumn amount at 2, and just add this "showArrivalTimes" -part in the first column with the "ObservationDate"

FYI: We also have a bit of a mess with the "Column" and "FormColumn" -components, which we should fix at some point, but that part is maybe better to explain f2f.
</blockquote></details>

Thanks. I suggest you use the last solution, IMO the simplest one.

And for the dynamic class problem, we could use tailwind safelist to fix this problem, but I don't think we want to use dynamic class names at least in this case.

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578)* status: all files reviewed, 7 unresolved discussions (waiting on @eleksi and @Perholehto)

___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSuYOm8BO9FsBs18Nuv:-NSyX2eR8pAebMpadCUF:bl8bxih) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx#L72)):*

I didn't test this branch so I didn't notice the jumping, good spot.

I'm not quite sure what we want to to with this, as hiding the arrival times also could be used to save space. If the same amount of space is used anyway, it really doesn't help with that. I think some jumping could be ok, as we clearly want to show different amount of information. But then again, if one switches arrival times on and off and tries to look at specific stop row, it could be hard if the stop row changes position.

I don't have strong opinion, I think this is also ok as is.

Btw. if you decide to go with Henkka's solution, use `invisible` instead of `opacity-0`
</blockquote></details>

I don't think that it matters much to "save space" by hiding the arrival times. But I personally hate jumping elements, because they in my opinion it reduces the usability and UX a lot and it just simply does not look good.
But Eetu great one with the `invisible`, that is actually the right one to use here. I was looking for something like that, but didn't find one.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSu-plY6liqULWZEHij:-NSyYLdiBeMIEpYg0ewv:beeknrh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*

Thanks. I suggest you use the last solution, IMO the simplest one.

And for the dynamic class problem, we could use tailwind safelist to fix this problem, but I don't think we want to use dynamic class names at least in this case.
</blockquote></details>

Oh right good one! There was the safelist, again forgot about that one :D. But yeah, no need at this point. For future reference the safelist can be found at `tailwind.config.js`.

And it seems that some patterns can be used for wider ranges as well. So no need to add 10 className:s where only the value changes.
https://tailwindcss.com/docs/content-configuration#using-regular-expressions

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578)* status: 6 of 8 files reviewed, 7 unresolved discussions (waiting on @eleksi and @HenrikHartiala)

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStxtXrE0gfqbde0mzq):*

Yeah, this way it says what the commit _will_ do if applied

Second thing, which I believe we haven't mentioned yet (nor do I think is documented anywhere) is that the Structure of the commit is 
```
Title of commit

Body of commit

Resolves HSLdevcom/jore4#<ticketnumber>
```
This way, we can see what is the "specification" of the commit and also it will automatically move the ticket on the board when this PR is merged to main. There are of course sometimes commits that do not resolve any specific tickets, but if there is a corresponding ticket, then it's nice to add here.

So for the `- ccc24d7: Add 'show arrival times'-toggle to timetable -commit`, you could add `Resolves HSLdevcom/jore4#1196` as I believe that is the corresponding ticket
</blockquote></details>

Nice that we care about a commit message format! I'm tired of reading "small fix" etc. 

I will gladly use that Resolves integration :) 

Maybe this should be documented, but it wouldn't have helped me cause I can't get to REAL wiki, yet.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 25 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStyIzk5UYNDG8jMxdu:-NTCpQ25D6L1lQSk2VIV:b-gvb7s2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L25)):*

Could use `useAddDispatch()` which is typed better.
</blockquote></details>

Changed this to use `useAppDispatch()` :+1: 
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSu-plY6liqULWZEHij:-NTCpVCL6-qq13pCcEX8:bljv8y5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*

Oh right good one! There was the safelist, again forgot about that one :D. But yeah, no need at this point. For future reference the safelist can be found at `tailwind.config.js`.

And it seems that some patterns can be used for wider ranges as well. So no need to add 10 className:s where only the value changes.
https://tailwindcss.com/docs/content-configuration#using-regular-expressions
</blockquote></details>

Yeah, that was kind of too complex. I tried to make it more simpler and used `col-start-5` strategy. If that is not simple/good enough, I will gladly change it  to other one.

Good to know that there is that kind of limiter in tailwind-css, not that it's a problem but needs to be given a though.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStzw776fMRaUemLi09:-NTCqkaW9GNhR3mHa2t8:b-v11t2a) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L72)):*

We have `Visible` component that we use for showing and hiding stuff. It makes a code a little cleaner. I know this component does not use it in other places either, but if you want you can change all those as well.
</blockquote></details>

Changed it to couple of places in this component, but left one because changing it to use `Visible` caused type errors and I didn't think it was worth to fix...
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 83 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStzS-y7m_bJO9XZvLl:-NTCrJcM05CINx4zy_Wj:b-hwfl87) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L83)):*

This is not "show-unused-stops-switch" ;).

We usually make a constant object out of testids and prefix it with parent component name. Example below:


</blockquote></details>

Just some basic copy/paste and forgot... Changed to use your snippet for testId!
___
*[`ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTableRow.tsx` line 26 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStynaqEWEYe6_QMEkg:-NTCrVbV5hBjC2XTywV8:b-8qfmfv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTableRow.tsx#L26)):*

Nice that you fixed this, it was annoying console warning.
</blockquote></details>

I want my console to be clear of RED!

HenrikHartiala (commented):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578)* status: 7 of 8 files reviewed, 5 unresolved discussions (waiting on @eleksi and @Perholehto)

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStxtXrE0gfqbde0mzq):*

Nice that we care about a commit message format! I'm tired of reading "small fix" etc. 

I will gladly use that Resolves integration :) 

Maybe this should be documented, but it wouldn't have helped me cause I can't get to REAL wiki, yet.
</blockquote></details>

Nice, yeah It _might_ be documented somewhere, but can't remember where. After we have time again to organize the new wiki, we could add a section about developing styles.
In UI repo readme, there is some guidelines (just one or two) but those could also be moved to the wiki.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSuYOm8BO9FsBs18Nuv:-NTD6AXH3DKeQCqWy2t_:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx#L72)):*

Now that I looked this more closely, it's really annoying that there is that small jump in elements. So I agree with Henrik and will change this to version with no jump-effect. 
</blockquote></details>

Thanks!
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSu-plY6liqULWZEHij:-NTD8sLP89ztnBV13VCw:b-dqod0z) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*

Yeah, that was kind of too complex. I tried to make it more simpler and used `col-start-5` strategy. If that is not simple/good enough, I will gladly change it  to other one.

Good to know that there is that kind of limiter in tailwind-css, not that it's a problem but needs to be given a though.
</blockquote></details>

Great, this is fine solution at least for me :)

eleksi (approved):
After changing the label font size and fixing alignment after that this is approved!

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-:-NTDZUzi2aVg5MkrUXvG:btzagl1)* status: all files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @Perholehto)

___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSuYOm8BO9FsBs18Nuv:-NTDSjOE26kQft43OA-U:beait0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTableRowPassingMinute.tsx#L72)):*

Thanks!
</blockquote></details>

Nice, thanks. In the future tailwind class merges could be done using tailwind merge (twMerge), but this is also ok!
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 25 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStyIzk5UYNDG8jMxdu:-NTDTKwFDOjyHfvTcXge:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L25)):*

Changed this to use `useAppDispatch()` :+1: 
</blockquote></details>

Thanks!
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NSu-plY6liqULWZEHij:-NTDWeEu9qWXcXWYFzks:b-79vvw3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L47)):*

Great, this is fine solution at least for me :)
</blockquote></details>

Much better now!
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 72 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStzw776fMRaUemLi09:-NTDTRwLEtz4vzfkglue:b-d988v5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L72)):*

Changed it to couple of places in this component, but left one because changing it to use `Visible` caused type errors and I didn't think it was worth to fix...
</blockquote></details>

Yeah, that is a known problem and I don't know how to fix this either. So no need to use Visible there.
___
*[`ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx` line 83 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStzS-y7m_bJO9XZvLl:-NTDWczk0j2X2YQ79D3A:b-83sjoh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/VehicleScheduleDetailsPage.tsx#L83)):*

Just some basic copy/paste and forgot... Changed to use your snippet for testId!
</blockquote></details>

Nice!
___
*[`ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTableRow.tsx` line 26 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578#-NStynaqEWEYe6_QMEkg:-NTDWjTPEfYdCCyxZrL1:b-t7uwr7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a9f1a4f2e897c55980b579966c65851643f7aabc/ui/src/components/timetables/vehicle-schedule-details/vehicle-service-table/VehicleServiceTableRow.tsx#L26)):*

I want my console to be clear of RED!
</blockquote></details>

Yep! :)

eleksi (approved):
> Reviewed 2 of 2 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/578)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @HenrikHartiala and @Perholehto)

PR #597, opened 2023-05-12
Make scripts runnable even if bash is not the default shell

Leitsi:
For example on PopOS sh uses dash by default,
and the scripts don't work on that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/597)


culka (approved):
:+1: 

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/597#-:-NVTAupj-VYV9ivStSe8:b-8fx2u7)* status: 0 of 4 files reviewed, all discussions resolved

PR #600, opened 2023-05-16
Add Hasura role header to websocket connections

Leitsi:
Used at least in auth hook, which gets spammed about every second. This was previously causing the auth backend to throw an error on each call because the required header was missing.

This does raise a question though, does the auth hook even work correctly, because even it failing all the time didn't seem to affect UI behavior at all.

Also tried adding "lazy: true" to the websocket parameters, as suggested by couple of separate issues:
  https://github.com/apollographql/subscriptions-transport-ws/issues/381#issuecomment-375905514
  https://github.com/apollographql/apollo-client/issues/3967#issuecomment-450255702
This stopped the hook being called completely (no more errors, right?), but also Apollo cache didn't work properly anymore.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/600)

### DIFF
--- a/ui/src/graphql/auth.ts
+++ b/ui/src/graphql/auth.ts
@@ -4,7 +4,7 @@ import { QueryHookOptions } from '@apollo/client/react/types/types';
 import { TypedDocumentNode } from '@graphql-typed-document-node/core';
 import { DocumentNode } from 'graphql';
 
-const REQUESTED_HASURA_ROLE_HEADER = 'x-hasura-role';
+export const REQUESTED_HASURA_ROLE_HEADER = 'x-hasura-role';
 
 export type Role = string;
 
--- a/ui/src/graphql/client.ts
+++ b/ui/src/graphql/client.ts
@@ -14,7 +14,7 @@ import { DateTime, Duration } from 'luxon';
 import introspectionResult from '../../graphql.schema.json';
 import { isDateLike, parseDate } from '../time';
 import { mapHttpToWs } from '../utils/url';
-import { authRoleMiddleware } from './auth';
+import { authRoleMiddleware, REQUESTED_HASURA_ROLE_HEADER } from './auth';
 
 const buildScalarMappingLink = () => {
   const typesMap = {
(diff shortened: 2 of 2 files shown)


culka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/600)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`ui/src/graphql/client.ts` line 88 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/600#-NVYbGH827P23cirTRRn:-NVYbGH827P23cirTRRo:b-zarmye) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9169d07a3ba9349c80f3ec2dfe582ad3fe141fb3/ui/src/graphql/client.ts#L88)):*
> ```XML
>       connectionParams: {
>         headers: {
>           'x-hasura-role': 'admin',
> ```

Since we do not have a proper auth role selection anyway, maybe use the same const `REQUESTED_HASURA_ROLE_HEADER` here as in `auth.ts`

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/600)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @culka)

___
*[`ui/src/graphql/client.ts` line 88 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/600#-NVYbGH827P23cirTRRn:-NWaLPFM6zHCSOB_7oRo:bevsic7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9169d07a3ba9349c80f3ec2dfe582ad3fe141fb3/ui/src/graphql/client.ts#L88)):*

Since we do not have a proper auth role selection anyway, maybe use the same const `REQUESTED_HASURA_ROLE_HEADER` here as in `auth.ts`
</blockquote></details>

:heavy_check_mark:

culka (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/600#-:-NWaR6Tu6R9kMrMhv-Le:bg7igkv)* status: 0 of 2 files reviewed, all discussions resolved

PR #606, opened 2023-05-22
Upgrade Cypress Docker image

sr-vala:
There seem to be no new images available with only Chrome installed anymore.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/606)


sr-vala:
Better to handle this via https://github.com/HSLdevcom/jore4-ui/pull/607

PR #610, opened 2023-05-29
Add default sorting for versions and daytype colours

HenrikHartiala:
These colours will help indicate what priority is in effect. Also add default sorting of the list. Later one can sort the list how one wants, but for now the default sorting is enough.

Resolves HSLdevcom/jore4#1152

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/610)

### DIFF
--- a/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx
+++ b/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx
@@ -21,12 +21,24 @@ const getStatusClassName = ({
   const statusClassNames: Record<TimetablePriority, string> = {
     [TimetablePriority.Standard]: 'bg-brand text-white',
     [TimetablePriority.Temporary]: 'bg-city-bicycle-yellow',
-    [TimetablePriority.Special]: 'bg-hsl-light-purple',
+    [TimetablePriority.Special]: 'bg-hsl-purple',
     [TimetablePriority.SubstituteByLineType]: 'bg-hsl-orange',
     [TimetablePriority.Draft]: 'bg-background',
     [TimetablePriority.Staging]: 'bg-hsl-red',
   };
   return statusClassNames[priority];
+};
+
+const getDayTypeClassName = (priority: TimetablePriority) => {
+  const dayTypeClassNames: Record<TimetablePriority, string> = {
+    [TimetablePriority.Standard]: 'bg-hsl-dark-green',
+    [TimetablePriority.Temporary]: 'bg-city-bicycle-yellow',
+    [TimetablePriority.Special]: 'bg-hsl-purple',
+    [TimetablePriority.SubstituteByLineType]: 'bg-hsl-orange',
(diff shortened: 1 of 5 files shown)


Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/610)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 58 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/610#-NWbEAwzA0dkmsxveU4d:-NWbEAwzA0dkmsxveU4e:bks0igi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6445f6b45e5b393a5b2c8705c8e8f568b5f4aa3/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L58)):*
> ```XML
>   const { onClose } = useTimetableVersionsReturnToQueryParam();
> 
>   const sortByInEffect = (
> ```

To be honest I would rather use lodash sortBy for these.
The JS array sort function syntax is a bit obscure, and also it mutates the original array.
Also, the names of these, `sortBy` sounds like the function itself does the sorting, though that is not the case: it just acts as the sorter function for Array.sort. 

So, something like:
```
const sortTimetables = (timetables: any /*TODO*/) => {
  return sortBy(timetables, [ 'inEffect', 'routeLabelAndVariant', 'validityStart']);
};
```
Probably have to wrap some of those in a function still, if types are uncertain.
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 69 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/610#-NWbDhWO5nQUo3ngWSAP:-NWbDhWO5nQUo3ngWSAQ:bxp4su7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6445f6b45e5b393a5b2c8705c8e8f568b5f4aa3/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L69)):*
> ```XML
>     // vehicleScheduleFrame always has validityStart
>     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
>     +b.vehicleScheduleFrame.validityStart! -
> ```

What's the purpose of `+` here? To cast falsy values to 0 (except undefined because it doesn't work)?
If that's so, I would maybe be more explicit, eg. `(b.vehicleScheduleFrame.validityStart || 0) - (a.vehicleScheduleFrame.validityStart || 0)`. This would also remove need for the eslint disable.

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/610#-:-NWh-Pm5-fn-8pfC6Rjb:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved

PR #616, opened 2023-06-06
E2e edit draft route

sr-vala:
Test that is described in [#1259](https://github.com/HSLdevcom/jore4/issues/1259)
Also noticed that origin and destination info was being filled twice, so changed the logic a bit in the other commit

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616)

### DIFF
--- a/cypress/e2e/editRoute.cy.ts
+++ b/cypress/e2e/editRoute.cy.ts
@@ -72,33 +72,32 @@ const lines: LineInsertInput[] = [
 const buildStopsOnInfrastrucureLinks = (
   infrastructureLinkIds: UUID[],
 ): StopInsertInput[] => [
-    {
-      ...buildStop({
-        label: stopLabels[0],
-        located_on_infrastructure_link_id: infrastructureLinkIds[0],
-      }),
-      validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
-      scheduled_stop_point_id: 'd4e6478a-adce-4c76-8579-c8ca2a6bb70f',
-      priority: Priority.Draft,
-      measured_location: {
-        type: 'Point',
-        coordinates: testInfraLinks[0].coordinates,
-      },
+  {
+    ...buildStop({
+      label: stopLabels[0],
+      located_on_infrastructure_link_id: infrastructureLinkIds[0],
+    }),
+    validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
+    scheduled_stop_point_id: 'd4e6478a-adce-4c76-8579-c8ca2a6bb70f',
+    priority: Priority.Draft,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[0].coordinates,
     },
-    {
-      ...buildStop({
-        label: stopLabels[1],
-        located_on_infrastructure_link_id: infrastructureLinkIds[1],
-      }),
-      validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
-      priority: Priority.Draft,
-      scheduled_stop_point_id: '3354eef5-0eaf-4b43-8b2f-14c867633342',
-      measured_location: {
-        type: 'Point',
-        coordinates: testInfraLinks[1].coordinates,
-      },
+  },
+  {
+    ...buildStop({
+      label: stopLabels[1],
+      located_on_infrastructure_link_id: infrastructureLinkIds[1],
+    }),
+    validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
+    scheduled_stop_point_id: '3354eef5-0eaf-4b43-8b2f-14c867633342',
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[1].coordinates,
     },
-  ];
+  },
+];
 
 const routes: RouteInsertInput[] = [
   {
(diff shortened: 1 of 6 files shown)


HenrikHartiala (changes requested):
> Reviewed 5 of 5 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616)* status: all files reviewed, 8 unresolved discussions (waiting on @sr-vala)

___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-NdeX2WA9zvU0wq5LzC7):*
This commit title is confusing. First it should say that we are modifying tests, because this will be amongst all the other commits in the repository as well. So something like "Remove extra form fill from editRoute test" or something that indicates what we are modifying here.

Secondly: A minor description to the commit body would help understanding what is being done. Now it took me a while to understand what we actually even want to do here.
___
*[`-- commits` line 10 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-NdebCVM5h7-10A9VCyf):*
Same comment here: the commit title should point towards tests. If you read this commit title without context, there is no way of even guessing what is done.

I believe the "setup" is indicating towards the testRouteLabels change and the teardown for the "deleteRoutesByLabel"? I understand the first but I don't understand the second, so again if there were a minor description why the teardown is unnecessary, it would help understanding why this change is done.
___
*[`cypress/e2e/editRoute.cy.ts` line 94 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-NdeWwUF4uRNoHqyX5Kn:-NdeWwUG3sOzdtM6Pc7T:b-as64ek) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/62f0cbae047a6747b768084178d472dfa75bc004/cypress/e2e/editRoute.cy.ts#L94)):*
> ```XML
>       }),
>       validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
>       priority: Priority.Draft,
> ```

I don't understand this removal, how is this related to the "fill origin and destination info only once"
___
*[`cypress/e2e/editRoute.cy.ts` line 136 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-NdeWnWODd7frLVpMIkW:-NdeWnWODd7frLVpMIkX:b-wrwv43) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/62f0cbae047a6747b768084178d472dfa75bc004/cypress/e2e/editRoute.cy.ts#L136)):*
> ```XML
>       is_traversal_forwards: true,
>     },
>     {
> ```

I don't understand this removal. How is this change related to the "fill origin and destination info only once"?
___
*[`cypress/e2e/editRoute.cy.ts` line 207 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-NdeerS19EWzlGrXfQS6:-NdeerS19EWzlGrXfQS7:b-ri63n5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6d81b8c71f0a86e76d7150df9d93a5cfe982e789/cypress/e2e/editRoute.cy.ts#L207)):*
> ```XML
> 
>   beforeEach(() => {
>     deleteRoutesByLabel(Object.values(testRouteLabels));
> ```

After writing all the comments I overviewed all the changes at once and realised that this third commit seems to be a fixup commit of the first commit. These things you remove here, are added in the first commit, so this should be squashed with it.
___
*[`cypress/e2e/editRoute.cy.ts` line 284 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-NdeVx5l8ufm9xH3Rwbm:-NdeVx5l8ufm9xH3Rwbn:b-p3d77o) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6d81b8c71f0a86e76d7150df9d93a5cfe982e789/cypress/e2e/editRoute.cy.ts#L284)):*
> ```XML
> 
>   it(
>     'Should show a warning when trying to change the priority of a draft route that has draft stops',
> ```

This change changes the same part as the earlier commit, seems to be just a prettier change so should be moved to the original commit.
___
*[`cypress/e2e/editRoute.cy.ts` line 98 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-Ndec8PT4kTIhaZrBozs:-Ndec8PT4kTIhaZrBozt:bgw94kw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b6ce0cf98f8370e895660c4f2226663890873667/cypress/e2e/editRoute.cy.ts#L98)):*
> ```XML
> const routes: RouteInsertInput[] = [
>   {
>     ...buildRoute({ label: 'T-reitti 1' }),
> ```

Nice, I like this masking variable is removed.
___
*[`cypress/pageObjects/RoutePropertiesForm.ts` line 96 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/616#-Nde_cr408uqc5Kx3K9j:-Nde_cr408uqc5Kx3K9k:brcpzwi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6d81b8c71f0a86e76d7150df9d93a5cfe982e789/cypress/pageObjects/RoutePropertiesForm.ts#L96)):*
> ```XML
>     }
> 
>     if (values.origin && values.destination) {
> ```

If we want to merge these, then I would rather extract the default test input values to variables and then use a `||` operator. So something like this
(I use this code block, since the "suggestion" diff looked very weird)
```
    const defaultTerminusOriginInput = {
          finnishName: 'Lähtöpaikka',
          swedishName: 'Ursprung',
          finnishShortName: 'LP',
          swedishShortName: 'UP',
        };
   const defaultTerminusDestinationInput = {
          finnishName: 'Määränpää',
          swedishName: 'Ändstation',
          finnishShortName: 'MP',
          swedishShortName: 'ÄS',
        },
        
      this.terminusNameInputs.fillTerminusNameInputsForm(
        values.origin || defaultTerminusOriginInput,
        values.destination || defaultTerminusDestinationInput,
      );
```

sr-vala:
Let's try this again in another PR.

PR #627, opened 2023-06-27
Add substitute operating period

Perholehto:
- Add LineTypeMultiSelectDropdown
- Add SubstituteDayOfWeeDropdown
- Add new page for creating, udating and removing substitute operating periods aka random reference days

This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)

### DIFF
--- a/docker/docker-compose.custom.yml
+++ b/docker/docker-compose.custom.yml
@@ -4,7 +4,8 @@ services:
     # Link to available jore4-hasura images in Docker Hub:
     # https://hub.docker.com/r/hsldevcom/jore4-hasura/tags?page=1&ordering=last_updated
     # The :hsl-tag contains the desired version of hsl specific hasura.
-    image: &hasura-image 'hsldevcom/jore4-hasura:hsl-main--20230711-f0391a96258129a48308711536ec733fa12fefd0'
+    # Waiting for merging feature-branch to main in hasura-repo
+    image: &hasura-image 'hsldevcom/jore4-hasura:hsl-main--20230810-f2cffa5fecb8bf3e6f2107a8c13fa73738bc85c9'
     # Waiting for database to be ready to avoid startup delay due to hasura crashing at startup if db is offline
     # Note: this should only be done in development setups as Kubernetes does not allow waiting for services to be ready
     depends_on:
--- a/test-db-manager/src/generated/graphql.ts
+++ b/test-db-manager/src/generated/graphql.ts
@@ -7243,6 +7243,20 @@ export type ServiceCalendarDayTypeActiveOnDayOfWeekAggregateBoolExpCount = {
   predicate: IntComparisonExp;
 };
 
+export type ServiceCalendarSubstituteOperatingDayByLineTypeAggregateBoolExp = {
+  count?: InputMaybe<ServiceCalendarSubstituteOperatingDayByLineTypeAggregateBoolExpCount>;
+};
+
+export type ServiceCalendarSubstituteOperatingDayByLineTypeAggregateBoolExpCount =
+  {
+    arguments?: InputMaybe<
+      Array<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeSelectColumn>
+    >;
+    distinct?: InputMaybe<Scalars['Boolean']>;
+    filter?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeBoolExp>;
+    predicate: IntComparisonExp;
+  };
+
 /** A dummy table that models the results of calculating the lengths of stop intervals from the given journey patterns. The table exists due to the limitations of Hasura and there is no intention to insert anything to it. */
 export type ServicePatternDistanceBetweenStopsCalculation = {
   __typename?: 'service_pattern_distance_between_stops_calculation';
@@ -10927,6 +10941,8 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineType = {
   /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
   substitute_day_of_week?: Maybe<Scalars['Int']>;
   substitute_operating_day_by_line_type_id: Scalars['uuid'];
+  /** The id of the substitute operating period */
+  substitute_operating_period_id: Scalars['uuid'];
   /** The date of operating day being superseded. */
   superseded_date: Scalars['date'];
   timezone: Scalars['String'];
@@ -10968,12 +10984,43 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeAggregateFi
     distinct?: InputMaybe<Scalars['Boolean']>;
   };
 
+/** order by aggregate values of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeAggregateOrderBy =
+  {
+    avg?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeAvgOrderBy>;
+    count?: InputMaybe<OrderBy>;
+    max?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeMaxOrderBy>;
+    min?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeMinOrderBy>;
+    stddev?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevOrderBy>;
+    stddev_pop?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevPopOrderBy>;
+    stddev_samp?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevSampOrderBy>;
+    sum?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeSumOrderBy>;
+    var_pop?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeVarPopOrderBy>;
+    var_samp?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeVarSampOrderBy>;
+    variance?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeVarianceOrderBy>;
+  };
+
+/** input type for inserting array relation for remote table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeArrRelInsertInput =
+  {
+    data: Array<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeInsertInput>;
+    /** upsert condition */
+    on_conflict?: InputMaybe<TimetablesServiceCalendarSubstituteOperatingDayByLineTypeOnConflict>;
+  };
+
 /** aggregate avg on columns */
 export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeAvgFields =
   {
     __typename?: 'timetables_service_calendar_substitute_operating_day_by_line_type_avg_fields';
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: Maybe<Scalars['Float']>;
+  };
+
+/** order by avg() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeAvgOrderBy =
+  {
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
   };
 
 /** Boolean expression to filter rows from the table "service_calendar.substitute_operating_day_by_line_type". All fields are combined with a logical 'AND'. */
@@ -10992,6 +11039,7 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeBoolExp = {
   end_time?: InputMaybe<IntervalComparisonExp>;
   substitute_day_of_week?: InputMaybe<IntComparisonExp>;
   substitute_operating_day_by_line_type_id?: InputMaybe<UuidComparisonExp>;
+  substitute_operating_period_id?: InputMaybe<UuidComparisonExp>;
   superseded_date?: InputMaybe<DateComparisonExp>;
   timezone?: InputMaybe<StringComparisonExp>;
   type_of_line?: InputMaybe<StringComparisonExp>;
@@ -11021,6 +11069,8 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeInsertInput
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: InputMaybe<Scalars['Int']>;
     substitute_operating_day_by_line_type_id?: InputMaybe<Scalars['uuid']>;
+    /** The id of the substitute operating period */
+    substitute_operating_period_id?: InputMaybe<Scalars['uuid']>;
     /** The date of operating day being superseded. */
     superseded_date?: InputMaybe<Scalars['date']>;
     timezone?: InputMaybe<Scalars['String']>;
@@ -11040,11 +11090,33 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeMaxFields =
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: Maybe<Scalars['Int']>;
     substitute_operating_day_by_line_type_id?: Maybe<Scalars['uuid']>;
+    /** The id of the substitute operating period */
+    substitute_operating_period_id?: Maybe<Scalars['uuid']>;
     /** The date of operating day being superseded. */
     superseded_date?: Maybe<Scalars['date']>;
     timezone?: Maybe<Scalars['String']>;
     /** The type of line this substitute operating day is bound to. */
     type_of_line?: Maybe<Scalars['String']>;
+  };
+
+/** order by max() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeMaxOrderBy =
+  {
+    /** Calculated timestamp for the instant from which the substituting public transit comes into effect. */
+    begin_datetime?: InputMaybe<OrderBy>;
+    created_at?: InputMaybe<OrderBy>;
+    /** Calculated timestamp for the instant (exclusive) until which the substituting public transit is in effect. */
+    end_datetime?: InputMaybe<OrderBy>;
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
+    substitute_operating_day_by_line_type_id?: InputMaybe<OrderBy>;
+    /** The id of the substitute operating period */
+    substitute_operating_period_id?: InputMaybe<OrderBy>;
+    /** The date of operating day being superseded. */
+    superseded_date?: InputMaybe<OrderBy>;
+    timezone?: InputMaybe<OrderBy>;
+    /** The type of line this substitute operating day is bound to. */
+    type_of_line?: InputMaybe<OrderBy>;
   };
 
 /** aggregate min on columns */
@@ -11059,11 +11131,33 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeMinFields =
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: Maybe<Scalars['Int']>;
     substitute_operating_day_by_line_type_id?: Maybe<Scalars['uuid']>;
+    /** The id of the substitute operating period */
+    substitute_operating_period_id?: Maybe<Scalars['uuid']>;
     /** The date of operating day being superseded. */
     superseded_date?: Maybe<Scalars['date']>;
     timezone?: Maybe<Scalars['String']>;
     /** The type of line this substitute operating day is bound to. */
     type_of_line?: Maybe<Scalars['String']>;
+  };
+
+/** order by min() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeMinOrderBy =
+  {
+    /** Calculated timestamp for the instant from which the substituting public transit comes into effect. */
+    begin_datetime?: InputMaybe<OrderBy>;
+    created_at?: InputMaybe<OrderBy>;
+    /** Calculated timestamp for the instant (exclusive) until which the substituting public transit is in effect. */
+    end_datetime?: InputMaybe<OrderBy>;
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
+    substitute_operating_day_by_line_type_id?: InputMaybe<OrderBy>;
+    /** The id of the substitute operating period */
+    substitute_operating_period_id?: InputMaybe<OrderBy>;
+    /** The date of operating day being superseded. */
+    superseded_date?: InputMaybe<OrderBy>;
+    timezone?: InputMaybe<OrderBy>;
+    /** The type of line this substitute operating day is bound to. */
+    type_of_line?: InputMaybe<OrderBy>;
   };
 
 /** response of any mutation on the table "service_calendar.substitute_operating_day_by_line_type" */
@@ -11101,6 +11195,7 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeOrderBy = {
   end_time?: InputMaybe<OrderBy>;
   substitute_day_of_week?: InputMaybe<OrderBy>;
   substitute_operating_day_by_line_type_id?: InputMaybe<OrderBy>;
+  substitute_operating_period_id?: InputMaybe<OrderBy>;
   superseded_date?: InputMaybe<OrderBy>;
   timezone?: InputMaybe<OrderBy>;
   type_of_line?: InputMaybe<OrderBy>;
@@ -11129,6 +11224,8 @@ export enum TimetablesServiceCalendarSubstituteOperatingDayByLineTypeSelectColum
   /** column name */
   SubstituteOperatingDayByLineTypeId = 'substitute_operating_day_by_line_type_id',
   /** column name */
+  SubstituteOperatingPeriodId = 'substitute_operating_period_id',
+  /** column name */
   SupersededDate = 'superseded_date',
   /** column name */
   Timezone = 'timezone',
@@ -11147,6 +11244,8 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeSetInput =
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: InputMaybe<Scalars['Int']>;
     substitute_operating_day_by_line_type_id?: InputMaybe<Scalars['uuid']>;
+    /** The id of the substitute operating period */
+    substitute_operating_period_id?: InputMaybe<Scalars['uuid']>;
     /** The date of operating day being superseded. */
     superseded_date?: InputMaybe<Scalars['date']>;
     timezone?: InputMaybe<Scalars['String']>;
@@ -11162,6 +11261,13 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevField
     substitute_day_of_week?: Maybe<Scalars['Float']>;
   };
 
+/** order by stddev() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevOrderBy =
+  {
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
+  };
+
 /** aggregate stddev_pop on columns */
 export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevPopFields =
   {
@@ -11170,12 +11276,26 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevPopFi
     substitute_day_of_week?: Maybe<Scalars['Float']>;
   };
 
+/** order by stddev_pop() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevPopOrderBy =
+  {
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
+  };
+
 /** aggregate stddev_samp on columns */
 export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevSampFields =
   {
     __typename?: 'timetables_service_calendar_substitute_operating_day_by_line_type_stddev_samp_fields';
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: Maybe<Scalars['Float']>;
+  };
+
+/** order by stddev_samp() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStddevSampOrderBy =
+  {
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
   };
 
 /** Streaming cursor of the table "service_calendar_substitute_operating_day_by_line_type" */
@@ -11202,6 +11322,8 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeStreamCurso
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: InputMaybe<Scalars['Int']>;
     substitute_operating_day_by_line_type_id?: InputMaybe<Scalars['uuid']>;
+    /** The id of the substitute operating period */
+    substitute_operating_period_id?: InputMaybe<Scalars['uuid']>;
     /** The date of operating day being superseded. */
     superseded_date?: InputMaybe<Scalars['date']>;
     timezone?: InputMaybe<Scalars['String']>;
@@ -11217,6 +11339,13 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeSumFields =
     substitute_day_of_week?: Maybe<Scalars['Int']>;
   };
 
+/** order by sum() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeSumOrderBy =
+  {
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
+  };
+
 /** update columns of table "service_calendar.substitute_operating_day_by_line_type" */
 export enum TimetablesServiceCalendarSubstituteOperatingDayByLineTypeUpdateColumn {
   /** column name */
@@ -11229,6 +11358,8 @@ export enum TimetablesServiceCalendarSubstituteOperatingDayByLineTypeUpdateColum
   SubstituteDayOfWeek = 'substitute_day_of_week',
   /** column name */
   SubstituteOperatingDayByLineTypeId = 'substitute_operating_day_by_line_type_id',
+  /** column name */
+  SubstituteOperatingPeriodId = 'substitute_operating_period_id',
   /** column name */
   SupersededDate = 'superseded_date',
   /** column name */
@@ -11254,12 +11385,26 @@ export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeVarPopField
     substitute_day_of_week?: Maybe<Scalars['Float']>;
   };
 
+/** order by var_pop() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeVarPopOrderBy =
+  {
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
+  };
+
 /** aggregate var_samp on columns */
 export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeVarSampFields =
   {
     __typename?: 'timetables_service_calendar_substitute_operating_day_by_line_type_var_samp_fields';
     /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
     substitute_day_of_week?: Maybe<Scalars['Float']>;
+  };
+
+/** order by var_samp() on columns of table "service_calendar.substitute_operating_day_by_line_type" */
+export type TimetablesServiceCalendarSubstituteOperatingDayByLineTypeVarSampOrderBy =
+  {
+    /** The ISO day of week (1=Monday, ... , 7=Sunday) of the day type used as the basis for operating day substitution. A NULL value indicates that there is no public transit at all, i.e. no vehicle journeys are operated within the given time period. */
+    substitute_day_of_week?: InputMaybe<OrderBy>;
   };
 
 /** aggregate variance on columns */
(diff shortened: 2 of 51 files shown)


HenrikHartiala (changes requested):
> Reviewed 1 of 2 files at r1, 48 of 48 files at r2.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: 48 of 49 files reviewed, 37 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 13 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBa-Gs6ULSmT2nRdbi:-NaBa-Gt3wBwVCGWUrxq:b9u8fp5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L13)):*
> ```XML
> };
> // eslint-disable-next-line @typescript-eslint/ban-types
> export const EnumMultiSelectDropdown = <TEnum extends Object>({
> ```

Out of curiosity, could we use `<TEnum extends object>` (lower case)? We could get rid of the eslint disable, but not sure if this achieves the same thing.
___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 49 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBbhKz3IsXsuAzSfT_:-NaBbhKz3IsXsuAzSfTa:b417z28) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L49)):*
> ```XML
>   });
> 
>   const getButtonContent = (inputValue: string | undefined): string => {
> ```

Can the inputValue be optional, or does it need to be set, just undefined is one of the options?



_Suggestion:_
```XML
  const getButtonContent = (inputValue?: string): string => {
```
___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 51 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBc3deENwJF9ukQXOD:-NaBc3deENwJF9ukQXOE:b-34xjug) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L51)):*
> ```XML
>   const getButtonContent = (inputValue: string | undefined): string => {
>     const enumValues = inputValue?.split(',');
>     if (isUndefined(enumValues) || first(enumValues) === '') return placeholder;
> ```

I personally don't like one liner if's because they make the flow of the function a bit hard to read. I suggest using explicit scopes.



_Suggestion:_
```XML
    if (isUndefined(enumValues) || first(enumValues) === ''){
      return placeholder;
    }
```
___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 51 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBfTOM4PQD01tyLz3O:-NaBfTOM4PQD01tyLz3P:b-ohywgk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L51)):*
> ```XML
>   const getButtonContent = (inputValue: string | undefined): string => {
>     const enumValues = inputValue?.split(',');
>     if (isUndefined(enumValues) || first(enumValues) === '') return placeholder;
> ```

Also, not sure if it's error prone if we just check the first value. One could be that we just. But I mean, that is not that readable, but it would also check that the there is no values at all. 
Well, this was really just thinking aloud and I think this is fine like this. If you get ideas from this then by all means.



_Suggestion:_
```XML
    if(!enumValues?.filter(Boolean).length) {
      return placeholder;
    }
```
___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 65 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBhOvU6zno9i1KdnqG:-NaBhOvU6zno9i1KdnqH:b-dfzt7z) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L65)):*
> ```XML
>     <MultiSelectListbox
>       id={id}
>       testId={testId || testIds.enumDropdown}
> ```

Kind of not part of this PR. since this is used in other places as well, but I was thinking that should we actually enforce giving the testId? Because is same page has two similar components, and neither of them have testId set, then we have two same testId's, and that may not be so nice.

What do you think?
___
*[`ui/src/components/forms/common/InputLabel.tsx` line 26 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBiOBZ303PvI1DL0yI:-NaBiOBZ303PvI1DL0yJ:b9vac2f) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/InputLabel.tsx#L26)):*
> ```XML
>   return (
>     <label className={className} htmlFor={`${translationPrefix}.${fieldPath}`}>
>       {t(`${translationPrefix}.${fieldPath.replace(/\.\d+/, '')}`)}
> ```

Why is this change done? It refactors existing code, does it deserve its own commit to be sure?
___
*[`ui/src/components/forms/timetables/LineTypeMultiSelectDropdown.tsx` line 9 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBjESL2wJk7URGrioo:-NaBjESL2wJk7URGriop:btg064g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/LineTypeMultiSelectDropdown.tsx#L9)):*
> ```XML
> interface Props extends FormInputProps {
>   id?: string;
>   testId?: string;
> ```

Same thing here, we could enforce the testId
___
*[`ui/src/components/forms/timetables/ObservationPeriodForm.tsx` line 9 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBppEZ-JqTR86LyMKH:-NaBppEZ-JqTR86LyMKI:b6x68no) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/ObservationPeriodForm.tsx#L9)):*
> ```XML
> 
> const testIds = {
>   startDate: 'TimetableFilterForm::startDate',
> ```

Ah, we should name the testId's matching the components name: https://github.com/HSLdevcom/jore4-ui#test-ids



_Suggestion:_
```XML
  startDate: 'ObservatoinPeriodForm::startDate',
```
___
*[`ui/src/components/forms/timetables/ObservationPeriodForm.tsx` line 31 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBmPjq0bNaJ9YSRHgH:-NaBmPjq0bNaJ9YSRHgI:b-x2bn7j) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/ObservationPeriodForm.tsx#L31)):*
> ```XML
>   return (
>     <>
>       <FormRow mdColumns={9}>
> ```

Hmm, I am not sure if this is the best way for the layout, because now we have this "ObservationPeriodForm" but it will always take the whole row. It would be better that the layout of the components inside of ObservationPeriodForm are defined in this component, but the "larger" layout is defined in the parent element.

And if you change this to be just <Row>, then maybe you also need to return a div, not an empty fragment, or the layout breaks again.

If this was confusing description, then let's take a telco :D 
___
*[`ui/src/components/forms/timetables/ObservationPeriodForm.tsx` line 57 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBrYvw8H-jYKF8_KXK:-NaBrYvw8H-jYKF8_KXL:bkf61dh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/ObservationPeriodForm.tsx#L57)):*
> ```XML
>         title={t('confirmReferenceDayFilterDialog.title')}
>         description={t('confirmReferenceDayFilterDialog.description')}
>         buttons={[
> ```

We have tried to avoid inline codes, this is somewhat confusing. But the confusion is minorly part of the component used here as well. Maybe it's the naming or something :D.
I guess this is kind of a warningPopUp which prevents changing the date? But if it always prevents it, could we rather just disable them when the form is dirty?

I do not have the silver bullet here, but it for me It feels like they (observation date range inputs) should be rather disabled and somehow indicated that "finish changes before changing the observation date range". 

If decided to change this, I also accept doing this iteratively, I believe it has been a long journey with this one for you already :D so I don't want to enforce everything to be fixed in this particular PR if you don't want to.
___
*[`ui/src/components/forms/timetables/SubstituteDayOfWeekDropdown.tsx` line 9 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBti-D6snIvum_ZPQI:-NaBti-D6snIvum_ZPQJ:budjbaz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/SubstituteDayOfWeekDropdown.tsx#L9)):*
> ```XML
> interface Props extends FormInputProps {
>   id?: string;
>   testId?: string;
> ```

enforce? Yes, No, Other: ?
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBu5gI1ToEhLMp8mK-:-NaBu5gI1ToEhLMp8mK0:b-8qth1h) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L12)):*
> ```XML
> 
> const testIds = {
>   closeButton: 'SettingsMainPage::closeButton',
> ```

Same comment about testId convention
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 15 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBuEXn74rdN64MF1aw:-NaBuEXoAFqxQtnVaSbV:b-2sitm2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L15)):*
> ```XML
> };
> 
> export const DaySettingsPage = (): JSX.Element => {
> ```

Hmm, should we use the Substitute-prefix for this? DaySettings sounds maybe a bit too generic. If decided to change, remember to change the folder names and filenames as well.
What do you think?
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 20 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBurGpA-nAHDFLNz8T:-NaBurGpA-nAHDFLNz8U:b-hz17h3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L20)):*
> ```XML
> 
>   const [isDirty, setIsDirty] = useState<boolean>(false);
> ```

Minor UX things, refreshing the page / using next or previous page is currently not interrupted even if there are changes in the form.
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 36 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBnpOkEvQReKsHKzrv:-NaBnpOkEvQReKsHKzrw:b-ipp1ud) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L36)):*
> ```XML
> 
>   return (
>     <Container className="">
> ```

empty className
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 42 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBwqCf3cWu4PsxaAbP:-NaBwqCf3cWu4PsxaAbQ:b-840y69) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L42)):*
> ```XML
>           testId={testIds.closeButton}
>           label={t('close')}
>           className="ml-auto text-base font-bold text-brand"
> ```

NIT, I'm not a fan of `ml-auto`, would rather use flex box keywords to get the button to the end.
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDay.tsx` line 24 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBvy_R67RClkrtzLgL:-NaBvy_R67RClkrtzLgM:bfuwpjg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDay.tsx#L24)):*
> ```XML
> }
> 
> export const RandomReferenceDay = ({ setIsDirty }: Props): JSX.Element => {
> ```

Would `OccasionalReferenceDay` sound better? Or `OccasionalSubstituteDay`? Random indicates to something that is out of our control :D 

Or actually, were this period's rather than day's? Should these be OccasionalSubstitutePeriods?
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDay.tsx` line 71 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBzspp3delksdkoh5f:-NaBzspp3delksdkoh5g:b-kakvje) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDay.tsx#L71)):*
> ```XML
>       showSuccessToast(t('timetables.settings.saveSuccess'));
>     } catch (err) {
>       showDangerToastWithError(t('errors.saveFailed'), err);
> ```

At some point, we might need more precise error messages (verbose versions), but it's okay to do that later as well.
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 39 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC0IFl8Ms4MKkmpckN:-NaC0IFl8Ms4MKkmpckO:b-mpqrfo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L39)):*
> ```XML
>   beginDate: mapToISODate(DateTime.now()) ?? '',
>   endDate: mapToISODate(DateTime.now()) ?? '',
>   beginTime: '04:30',
> ```

Should this actually be luxon's `Duration`? I'm not completely aware of how this should go with the zod and so on. This string just popped, and raised this thought
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 87 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC0oZa2Qx7qEvogbmE:-NaC0oZa2Qx7qEvogbmF:b412y9x) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L87)):*
> ```XML
> 
>   // Updates form dirty state to parent component. Parent component gives it to FilterForm
>   useEffect(() => setIsDirty(isDirty), [isDirty, setIsDirty]);
> ```

We maybe could also use redux for passing this information, what do you think?
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 92 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC6fKr0N7Pe1VqpbFP:-NaC6fKr0N7Pe1VqpbFQ:b-1ytoe7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L92)):*
> ```XML
>     const { id } = rowUnderDeleteOperation;
>     rowUnderDeleteOperation?.callback();
>     if (id) onRemove(id);
> ```

Same comment about one line ifs.
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 97 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC6wkO0HCQiII2Cxa3:-NaC6wkO0HCQiII2Cxa4:b-r45cm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L97)):*
> ```XML
>   };
> 
>   const handleRowRemove = (index: number, id: string | undefined) => {
> ```

is optional enough?



_Suggestion:_
```XML
  const handleRowRemove = (index: number, id?: string) => {
```
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 125 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC7huA7dHxDB8vehA5:-NaC7huA7dHxDB8vehA6:b-4lxh6p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L125)):*
> ```XML
>             </span>
>             <SimpleButton
>               className="h-full !px-3 text-xl"
> ```

is the "!important" required?
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx` line 27 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC8z5KEhkBq2itHIy6:-NaC8z5KEhkBq2itHIy7:undefined) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx#L27)):*
> ```XML
> 
> const testIds = {
>   periodnameInput: 'RandomReferenceDayForm::periodName',
> ```





_Suggestion:_
```XML
  periodnameInput: 'RandomReferenceDayFormRow::periodName',
```
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCCRPx0dC_uSGF8gjd:-NaCCRPx0dC_uSGF8gje:b-ub4gn0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx#L37)):*
> ```XML
> };
> 
> export const RandomReferenceDayFormRow = ({
> ```

I am little bit struggling to understand the relationship of these components, we have `RandomReferenceDayForm` and `RandomReferenceDayFormRow`. Then we also have RandomReferenceDay.tsx and DaySettingsPage.tsx. Maybe the naming could be polished a bit. 

So what we have in other terms:
* SubstituteOperatingDaySettingsPage (or Period etc.)
  * ObservationDatePeriod pickers
  * OccasionalSubstituteOperatingDaysContainer/Card/Section
    * SubstituteOperatingDayFormRow(s)
    * (the possibility to add new rows)
    * (the possibility to save and cancel editing occasional...)

We can also talk about this in a telco etc :+1: 
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx` line 111 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCFSw34Jjnm7jo9Tez:-NaCFSw34Jjnm7jo9Tf-:bmr7gyn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx#L111)):*
> ```XML
>       <Column className=" col-span-1  justify-end">
>         <SimpleButton
>           className="!px-3r h-full text-xl"
> ```

possibly an extra 'r' in 'px-3r'. Also is the !important required?
___
*[`ui/src/components/timetables/main/TimetablesMainPage.tsx` line 9 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCFwf_5wiZqkokMNQN:-NaCFwf_5wiZqkokMNQO:bwmqnpw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/main/TimetablesMainPage.tsx#L9)):*
> ```XML
> const testIds = {
>   importButton: 'TimetablesMainPage::importButton',
>   settingsButton: 'TimetablesMainPage::settingsButton',
> ```

Disclaimer: this is a bit nitpicking, but...

This is probably something that should be iterated with Juha, but In my opinion the `settings` is too generic for what it really is, which is SubstituteOperatingDaySettings.
So what comes to the button name, I would want to iterate this through Juha, if we want to use the "Asetukset" or rather something like "Päiväasetukset" or even "Korvauspäiväasetukset" / "Korvauspäivät".

I'm not demanding any changes in this PR regarding what the button text is, but maybe we should use the more descriptive naming within the code, 
___
*[`ui/src/components/timetables/main/TimetablesMainPage.tsx` line 16 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCGzqq1TNMVDj-deZv:-NaCGzqq1TNMVDj-deZw:bubyfwy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/main/TimetablesMainPage.tsx#L16)):*
> ```XML
> 
>   const importTimetablesRoute = routeDetails[Path.timetablesImport];
>   const settingsTimetablesRoute = routeDetails[Path.timetablesSettings];
> ```

Like said above, would rather use a more descriptive naming for this. So my suggestion is 
`const substituteOperatingPeriodSettings = routeDetails[Path.substituteOperatingPeriodSettings]
Still not sure if we should use the `Period` or `Day`.

If decided to not make changes to this, then maybe at least fix the inconsistency in the names :P



_Suggestion:_
```XML
  const timetableSettingsRoute = routeDetails[Path.timetableSettings];
```
___
*[`ui/src/hooks/reference-days/useCreateSubstituteOperatingPeriod.ts` line 41 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCHy6IETNsq9T2DS0R:-NaCHy6IETNsq9T2DS0S:b-vnhi7u) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/hooks/reference-days/useCreateSubstituteOperatingPeriod.ts#L41)):*
> ```XML
> };
> 
> const GQL_CREATE_SUBSTITUTE_OPERATING_PERIOD = gql`
> ```

Not sure if we have any convention about this, but we have usually positioned these GQL's on top part of the file rather than bottom.
___
*[`ui/src/hooks/reference-days/useDeleteOperatingPeriod.ts` line 15 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCK1qO1gCnQ-UxV2fd:-NaCK1qO1gCnQ-UxV2fe:b-7o56de) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/hooks/reference-days/useDeleteOperatingPeriod.ts#L15)):*
> ```XML
> };
> 
> const GQL_DELETE_SUBSTITUTE_OPERATING_PERIOD = gql`
> ```

Same comment about positioning.
___
*[`ui/src/hooks/reference-days/useEditSubstituteOperatingPeriod.ts` line 50 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCK7Ra0Ec-Nr4FpnWM:-NaCK7Ra0Ec-Nr4FpnWN:buo9nmt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/hooks/reference-days/useEditSubstituteOperatingPeriod.ts#L50)):*
> ```XML
> // Second mutation removes substitute operating days that are under "update"
> // Last mutation inserts new rows for each day and line type
> const GQL_UPDATE_SUBSTITUTE_PERIOD = gql`
> ```

Same here
___
*[`ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts` line 42 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCKjvw-E9t6uEdCD3N:-NaCKjvw-E9t6uEdCD3O:bolzzrp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts#L42)):*
> ```XML
> `;
> 
> export const useGetSubstituteOperatingPeriods = ({
> ```

NIT,The name of this hook indicates towards somehow getting the SubstituteOperatingPeriods as result, but then judging by the code, it returns the FormState, which is probably  different thing(?)
So maybe add the FormState there on the hook name etc? Maybe drop the "get" and it will be a more generic hook
What do you think?
___
*[`ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts` line 64 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCM80k2ni6z6mUm09W:-NaCM80k2ni6z6mUm09X:b-6fu3sk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts#L64)):*
> ```XML
> 
>   const mapDurationToString = (time: Maybe<Duration> | undefined) => {
>     if (time) return mapDurationToShortTime(time);
> ```

one lined if
___
*[`ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts` line 69 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCMOJh-_-xUC4XqMb2:-NaCMOJh-_-xUC4XqMb3:b-eaf5v9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts#L69)):*
> ```XML
> 
>   const mapSubstituteDayOfWeek = (
>     substituteDayOfWeek: Maybe<number> | undefined,
> ```

I guess in this case we really do want it not to be optional and rather have the possibility of it being undefined? Or actually should it even be undefined at any case? I think the Maybe<number> should be enough. But ah right, I think I've stumbled to this earlier as well. It should not be optional in the schema, but for some apparent reason it is.. So this is fine.
___
*[`ui/src/i18n/uiNameMappings.ts` line 69 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCOiM3DOYRI0x4uoTr:-NaCOiM3DOYRI0x4uoTs:b9ckf2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/i18n/uiNameMappings.ts#L69)):*
> ```XML
> ) => (direction === RouteDirectionEnum.Outbound ? '1' : '2');
> 
> export const mapTimetableDayOfWeekToUiName = (key: SubstituteDayOfWeek) => {
> ```

inconsistent naming



_Suggestion:_
```XML
export const mapSubstituteDayOfWeekToUiName = (key: SubstituteDayOfWeek) => {
```
___
*[`ui/src/uiComponents/Combobox.tsx` line 6 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCQBF94_yXS34sbzY_:-NaCQBF94_yXS34sbzYa:bz56tf5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/Combobox.tsx#L6)):*
> ```XML
> import { MdCheck, MdSearch } from 'react-icons/md';
> import { dropdownTransition } from './Listbox';
> import { OptionRenderPropArg } from './ListboxOptions';
> ```

Refactoring existing code (The part where ListBoxOptions is extracted to own file), maybe deserves it's own commit if possible?
___
*[`ui/src/uiComponents/index.ts` line 13 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCQSU75pv54J-BX04y:-NaCQSU75pv54J-BX04z:b-pgc8hf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/index.ts#L13)):*
> ```XML
> export * from './IconButton';
> export * from './IconToggle';
> export * from './Listbox';
> ```

should the newly created ListBoxOptions and ListBoxButton be here as well?
___
*[`ui/src/uiComponents/MultiSelectListbox.tsx` line 20 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCRDvPCKKr3e4yyNTr:-NaCRDvPCKKr3e4yyNTs:bbv0khb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/MultiSelectListbox.tsx#L20)):*
> ```XML
>   id?: string;
>   buttonContent: ReactNode;
>   testId?: string;
> ```

enforce testId?
___
*[`ui/src/uiComponents/MultiSelectListbox.tsx` line 46 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCTE0u3qigN_QtGeta:-NaCTE0u3qigN_QtGetb:b-cw3s9n) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/MultiSelectListbox.tsx#L46)):*
> ```XML
>     const selected = first(val.filter((x) => !value?.includes(x)));
> 
>     if (selected === AllOptionEnum.All) {
> ```

This is currently a bit hard to read and I think this could probably be simplified with some helper functions and we could get rid of the `let finalValue` in the process as well. 
___
*[`ui/src/uiComponents/MultiSelectListbox.spec.tsx` line 17 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCTdfIDW8Vy8RXcLKJ:-NaCTdfIDW8Vy8RXcLKK:b-721mg9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/MultiSelectListbox.spec.tsx#L17)):*
> ```XML
>   ];
> 
>   test('Opens dropdown when clicked', async () => {
> ```

Nice to have base for tests, but could maybe also have tests for the selecting functionality.
___
*[`ui/src/uiComponents/Toast.tsx` line 60 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCTt-x2XBnlMyOxldy:-NaCTt-x2XBnlMyOxldz:b-dds98v) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/Toast.tsx#L60)):*
> ```XML
>     <div className={`rounded-md bg-white ${className}`} data-testid={testId}>
>       <div className={`${bg} ${border} rounded-md border`}>
>         <Row className="my-6 ml-16 mr-16 items-center">
> ```

Hmm, is this change intentional? IF so, then maybe deserves its own commit with description

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: 48 of 49 files reviewed, 38 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 132 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCWt-c-YrXsTgiOO9U:-NaCWt-c-YrXsTgiOO9V:bycksqc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/59ae4fa11f12b6435a2b8a2e6754f827212c9459/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L132)):*
> ```XML
>             </SimpleButton>
>           </Row>
>           <Row className="my-8 justify-end space-x-4">
> ```

This Cancel and Save button are always enabled. I think they should be disabled as long as the sections form is not "dirty"

HenrikHartiala (changes requested):
> Reviewed 1 of 1 files at r3, 35 of 35 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: all files reviewed, 12 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 13 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBa-Gs6ULSmT2nRdbi:-Naf0weCAaRI8ONmHKJC:brmux6m) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L13)):*

It would be a good change but our EnumTypes don't work with `object`
</blockquote></details>

Check!
___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 51 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBc3deENwJF9ukQXOD:-Naf14iI7uH6gvWd8iNR:b8n9lmz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L51)):*

Is this only a personal opinion or is this a team-wide convention? One line if's were a thing in my old job... I don't personally have a strong feeling about this. Will fix these tho.
</blockquote></details>

It is my personal opinion, but I haven't seen them very often and also at least think I have a good reason why not to use them :D. The readability suffers and also harder to debug or make extensions to the if, because you would like to add a console.log for example, you would need to create the block first. Or even worse scenario is that you don't notice that it does not have a scope, and you just add a console log right below the if-clause thinking it is inside the if.

But I can also add an eslint rule about this, so then it would be a strict convention :D 

Anyway, thanks for the fix!
___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 51 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBfTOM4PQD01tyLz3O:-Naf2znI2g6_HYqM9T-y:bjebfwy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L51)):*

I didn't get your point, maybe eli5 in call? 

todo tarkista vielä enumValues:n pituus
</blockquote></details>

This might still be missing. I think we discussed this to be something like this.



_Suggestion:_
```XML
    if (isUndefined(enumValues) || (first(enumValues) === '' && !!enumValues.length))
```
___
*[`ui/src/components/forms/common/InputLabel.tsx` line 26 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBiOBZ303PvI1DL0yI:-Naf3lBtD_E6Gv7mdANX:b-wdj8zi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/InputLabel.tsx#L26)):*

Added to own commit
</blockquote></details>

Nice, thanks!
___
*[`ui/src/components/forms/timetables/LineTypeMultiSelectDropdown.tsx` line 9 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBjESL2wJk7URGrioo:-Naf4BEiCOLXF6HxSBLV:bi9estx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/LineTypeMultiSelectDropdown.tsx#L9)):*

done
</blockquote></details>

:pray: 
___
*[`ui/src/components/forms/timetables/ObservationPeriodForm.tsx` line 31 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBmPjq0bNaJ9YSRHgH:-NafA5Hn3Ni0WjoKxI98:b-n4ff02) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/ObservationPeriodForm.tsx#L31)):*

No `div`s but refactored, what do you think?
</blockquote></details>

Better, but I think there seems to still be some issues with the the outer layout. I can demo this in a telco :+1: 
___
*[`ui/src/components/forms/timetables/ObservationPeriodForm.tsx` line 57 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBrYvw8H-jYKF8_KXK:-NafAUTkEq7XeDYEZ4GV:b-3ymtqy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/ObservationPeriodForm.tsx#L57)):*

removed...
</blockquote></details>

Thanks, I think this warning text and disable results in better UX than to enforce the user to first click the date to notice that it can't be changed.
___
*[`ui/src/components/timetables/main/TimetablesMainPage.tsx` line 9 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCFwf_5wiZqkokMNQN:-NafBZPyCSUJ8cSy2Lm2:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/main/TimetablesMainPage.tsx#L9)):*

Let's see where this goes!
</blockquote></details>

:+1: 
___
*[`ui/src/components/timetables/main/TimetablesMainPage.tsx` line 16 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCGzqq1TNMVDj-deZv:-NafBfK87pTtuMDSFHpN:b-wdj8zi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/main/TimetablesMainPage.tsx#L16)):*

Went with the `period`
</blockquote></details>

Nice, thanks!
___
*[`ui/src/hooks/reference-days/useEditSubstituteOperatingPeriod.ts` line 49 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NafU4zSDiHdqzf6rH_L:-NafU4zSDiHdqzf6rH_M:b-7ne4oc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a4546c776f2326dcecf30c7173898eb540737134/ui/src/hooks/reference-days/useEditSubstituteOperatingPeriod.ts#L49)):*
> ```XML
> `;
> 
> // export const useEditSubstituteOperatingPeriod = () => {
> ```

commented code
___
*[`ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts` line 42 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCKjvw-E9t6uEdCD3N:-NafUphI90pX2xJGoz7H:b81yleo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/hooks/reference-days/useGetSubstituteOperatingPeriod.ts#L42)):*

This hook is refactored and should have functionality what the name suggests.
</blockquote></details>

Seems clear!
___
*[`ui/src/uiComponents/Combobox.tsx` line 6 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCQBF94_yXS34sbzY_:-NafV97Z05ssEuHR2dsC:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/Combobox.tsx#L6)):*

Added to own commit
</blockquote></details>

Thanks!
___
*[`ui/src/uiComponents/MultiSelectListbox.tsx` line 46 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCTE0u3qigN_QtGeta:-NafWOkf45GCwMQYmHDt:b-clwlb2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/MultiSelectListbox.tsx#L46)):*

Tried to refactor this. What you think?
</blockquote></details>

Yep, nice. I think it is more clear what happens here! :ok_hand: 
___
*[`ui/src/uiComponents/MultiSelectListbox.spec.tsx` line 17 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCTdfIDW8Vy8RXcLKJ:-NafWnrxAkWsbiLG33KT:b-gptzqp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/MultiSelectListbox.spec.tsx#L17)):*

Add todo
</blockquote></details>

I think the todo is missing 
___
*[`ui/src/utils/graphql.ts` line 10 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NafXCwhDe4GNnxkJG0P:-NafXCwhDe4GNnxkJG0Q:bulh5bz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a4546c776f2326dcecf30c7173898eb540737134/ui/src/utils/graphql.ts#L10)):*
> ```XML
>   return { variables };
> };
> export const mapToData = (data: ExplicitAny) => {
> ```

This is actually wrapping, not mapping. To be consistent with the other function names this is fine, but as a note we should change all these namings to something like "wrapInFoo" in a separate PR.

So "wrapInObject" "wrapInVariables" "wrapInData".
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDay.tsx` line 24 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBvy_R67RClkrtzLgL:-NafYu3p2lECs11qXnsl:bawdrd9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDay.tsx#L24)):*

Maybe there will be some clarification for this in the future. Decided to use `OccasionalSubstitutePeriod`
</blockquote></details>

Thats great!
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBu5gI1ToEhLMp8mK-:-NafC18TAwmQKzyEtr6-:b-u49nig) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L12)):*

done
</blockquote></details>

The new component name does not reflect to the testId anymore, pretty classic fix & refactor :D 
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 15 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBuEXn74rdN64MF1aw:-NafCVYB4i-r9iOuqj14:bllmts6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L15)):*

Naming changed. More info given in Telco.
</blockquote></details>

Check, seems good!
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 20 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBurGpA-nAHDFLNz8T:-NafCkfm6o3vHuySC7ao:bbohbal) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L20)):*

This seems to be something that isn't implemented on any existing form. Not so minor for me, but I think it shouldn't be this PR where it's implemented.
</blockquote></details>

Good point, yeah it could be implemented on higher level :+1:. Sounds good 
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 39 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC0IFl8Ms4MKkmpckN:-NafDFJ8ECaA3RYLMV8E:b-ag23ag) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L39)):*

Yes and no :D This `emptyRowObject` is inserted to form straight, no mappers etc. So Zod validations are run against those values and validation expects `string`. This is same for `DateTime`-objects. So inside Form `Duration`-objects are string.
</blockquote></details>

Okay, thanks for the explanation!
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 87 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC0oZa2Qx7qEvogbmE:-NafD_rVATt7ibyBpm0j:b2prout) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L87)):*

Yeah, this lives now inside Redux. Do we have some kind of guidelines to know when throw stuff to redux?
</blockquote></details>

Thanks!
I dont think we have a strict guideline for this, because we also have some state information in the query parameters, which makes it kind of situational. But maybe the cases where you need the information in multiple components, and it is something we do not want to put in query parameters. So the case where you would need to pass the information through components.
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 132 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCWt-c-YrXsTgiOO9U:-NafEP4pCVBeB8kn2U5M:b-oypg22) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/59ae4fa11f12b6435a2b8a2e6754f827212c9459/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L132)):*

These are now disabled
</blockquote></details>

This is now better UX in my opinion, but found one case where it did not work and here are the steps to reproduce:
* Add new period for today
* SAve period (here it works fine, the buttons go disabled after save)
* add another period for let's say tomorrow
* save
Result: SAving succesful, but the form seems to be dirty and the buttons won't go back to disabled.

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: 36 of 58 files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 51 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBfTOM4PQD01tyLz3O:-NbDrseAB603c_4mAx7N:bb7f2mf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L51)):*

This might still be missing. I think we discussed this to be something like this.
</blockquote></details>

done
___
*[`ui/src/components/forms/timetables/ObservationPeriodForm.tsx` line 31 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBmPjq0bNaJ9YSRHgH:-NbDugqM7kOBSj5FiAsC:begw0lt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/ObservationPeriodForm.tsx#L31)):*

Better, but I think there seems to still be some issues with the the outer layout. I can demo this in a telco :+1: 
</blockquote></details>

How about now?
___
*[`ui/src/components/forms/timetables/ObservationPeriodForm.tsx` line 57 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBrYvw8H-jYKF8_KXK:-NajitGw-QIUzNnEI3D7:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/forms/timetables/ObservationPeriodForm.tsx#L57)):*

Thanks, I think this warning text and disable results in better UX than to enforce the user to first click the date to notice that it can't be changed.
</blockquote></details>

Done.
___
*[`ui/src/uiComponents/index.ts` line 13 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCQSU75pv54J-BX04y:-Najjl9m01olmedqc3pY:b-htipa) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/index.ts#L13)):*

should the newly created ListBoxOptions and ListBoxButton be here as well?
</blockquote></details>

These are added
___
*[`ui/src/uiComponents/MultiSelectListbox.spec.tsx` line 17 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCTdfIDW8Vy8RXcLKJ:-NbDsZzTDrDwBtlX_1Ad:b-4ro67e) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/uiComponents/MultiSelectListbox.spec.tsx#L17)):*

I think the todo is missing 
</blockquote></details>

Now it's really added :) 
___
*[`ui/src/components/timetables/day-settings/DaySettingsPage.tsx` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaBu5gI1ToEhLMp8mK-:-Najl5qT6gZli4apTPWj:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/DaySettingsPage.tsx#L12)):*

The new component name does not reflect to the testId anymore, pretty classic fix & refactor :D 
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 87 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC0oZa2Qx7qEvogbmE:-NajjCf-8IEXCNkal7VK:bt3qw7c) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L87)):*

Thanks!
I dont think we have a strict guideline for this, because we also have some state information in the query parameters, which makes it kind of situational. But maybe the cases where you need the information in multiple components, and it is something we do not want to put in query parameters. So the case where you would need to pass the information through components.

</blockquote></details>

Maybe this is shady for me because we have those query parameters and I'm not yet used to those :D
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 132 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCWt-c-YrXsTgiOO9U:-NbDus8I3A3j4T_-NSQc:by4ypgr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/59ae4fa11f12b6435a2b8a2e6754f827212c9459/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L132)):*

This is now better UX in my opinion, but found one case where it did not work and here are the steps to reproduce:
* Add new period for today
* SAve period (here it works fine, the buttons go disabled after save)
* add another period for let's say tomorrow
* save
Result: SAving succesful, but the form seems to be dirty and the buttons won't go back to disabled.
</blockquote></details>

This should be now fixed. I made changed that we discussed in telco. I can demo this if needed.
___
*[`ui/src/hooks/reference-days/useEditSubstituteOperatingPeriod.ts` line 49 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NafU4zSDiHdqzf6rH_L:-NajkdGP23v62ancbg4K:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a4546c776f2326dcecf30c7173898eb540737134/ui/src/hooks/reference-days/useEditSubstituteOperatingPeriod.ts#L49)):*

commented code
</blockquote></details>

Done.

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: 36 of 58 files reviewed, 12 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/forms/common/EnumDropdown.tsx` line 65 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbUc2yMDkGOPjTTRBAX:-NbUc2yMDkGOPjTTRBAY:b-7ijr36) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ace284907a8c22ee92c782ad8d93ee425ffed0fd/ui/src/components/forms/common/EnumDropdown.tsx#L65)):*
> ```XML
>         value ? uiNameMapper(value as unknown as TEnum) : placeholder
>       }
>       buttonClassNames={disabled ? 'input-disabled' : ''}
> ```

This should not be necessary to add the buttonClassName here as well. It should be enough to give the `disabled` attribute.

I dove to HUI.ListboxButton, which takes the disabled as paramtere, but for some reason it does not make the button look like disabled which is unfortunate. (Or I did something wrong). There might be another way of making this also visually disabled. 

In any case, I'd at least bake the className logic inside the actual component getting disabledw, instead of putting it here and then passing it through the components.
___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 73 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbUgMFw2O-hVQSHNYfH:-NbUgMFw2O-hVQSHNYfI:bjturct) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ace284907a8c22ee92c782ad8d93ee425ffed0fd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L73)):*
> ```XML
>       testId={testId || testIds.enumDropdown}
>       buttonContent={getButtonContent(value)}
>       buttonClassNames={disabled ? 'input-disabled' : ''}
> ```

Same thing here, so you should not need to give the component both: the class and the `disabled` value.

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: 34 of 58 files reviewed, 12 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/forms/common/EnumMultiSelectDropdown.tsx` line 73 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbUgMFw2O-hVQSHNYfH:-NbZQXV_5fL4bT-Av9U_:bgzf9g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ace284907a8c22ee92c782ad8d93ee425ffed0fd/ui/src/components/forms/common/EnumMultiSelectDropdown.tsx#L73)):*

Same thing here, so you should not need to give the component both: the class and the `disabled` value.
</blockquote></details>

`disabled` comes from `formInputProps`
___
*[`ui/src/components/forms/common/EnumDropdown.tsx` line 65 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbUc2yMDkGOPjTTRBAX:-NbYPj_g93A5eTHx82Lh:bedhnrf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ace284907a8c22ee92c782ad8d93ee425ffed0fd/ui/src/components/forms/common/EnumDropdown.tsx#L65)):*

This should not be necessary to add the buttonClassName here as well. It should be enough to give the `disabled` attribute.

I dove to HUI.ListboxButton, which takes the disabled as paramtere, but for some reason it does not make the button look like disabled which is unfortunate. (Or I did something wrong). There might be another way of making this also visually disabled. 

In any case, I'd at least bake the className logic inside the actual component getting disabledw, instead of putting it here and then passing it through the components.
</blockquote></details>

I took a second look about `HUI.Listbox` that component takes `disabled` prop and should disable the input. I tested this with an example from their site  (https://headlessui.com/react/listbox#basic-example) by pasting that example to our code base. When I passed `disabled` as true, the input was disabled.

I think problem in our implementation is that we  use`buttonContent` and disabled prop will never get there.

HenrikHartiala (changes requested):
> Reviewed 16 of 22 files at r5, 3 of 8 files at r6, 1 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: 50 of 58 files reviewed, 5 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/hooks/urlQuery/useTimeRangeQueryParams.ts` line 30 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbnIg6DBbVs4WelpVBU:-NbnIg6DBbVs4WelpVBV:bxjnewn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d18ec2c591a4695fb532722e59aef020c149dc64/ui/src/hooks/urlQuery/useTimeRangeQueryParams.ts#L30)):*
> ```XML
>     start: DateTime,
>     end: DateTime,
>     showToast = true,
> ```

discussed f2f, probably not needed.
___
*[`ui/src/components/forms/common/EnumDropdown.tsx` line 65 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbUc2yMDkGOPjTTRBAX:-NbnPCYg2H8ebEZ5VFUd:b413qcg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ace284907a8c22ee92c782ad8d93ee425ffed0fd/ui/src/components/forms/common/EnumDropdown.tsx#L65)):*

I took a second look about `HUI.Listbox` that component takes `disabled` prop and should disable the input. I tested this with an example from their site  (https://headlessui.com/react/listbox#basic-example) by pasting that example to our code base. When I passed `disabled` as true, the input was disabled.

I think problem in our implementation is that we  use`buttonContent` and disabled prop will never get there.
</blockquote></details>

Yeah, okay. Maybe we need to refactor this at some point. Nice that you tested this :+1: 
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx` line 87 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaC0oZa2Qx7qEvogbmE:-NbnBvQt3yxR3H4ZdOlt:b-h980v8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayForm.tsx#L87)):*

Maybe this is shady for me because we have those query parameters and I'm not yet used to those :D
</blockquote></details>

Yeah, we have added those in some cases, where a user might want to a) refresh page or b) copypaste the url for a particular view to another user (this is probably just a guess), so it wouldn't be possible without the queryparams. Also it makes e2e testing a bit easier, e.g. map tests.
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCCRPx0dC_uSGF8gjd:-NbnCka14Z9_McVgMkv4:b-ewu11g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx#L37)):*

Naming is changed in multiple places. Does it make a sense now?
</blockquote></details>

Yes, the naming is now much clearer! Thanks
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodRow.tsx` line 43 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbnCqW806fMB-n3ggOG:-NbnCqW806fMB-n3ggOH:b-4n2w3e) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d18ec2c591a4695fb532722e59aef020c149dc64/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodRow.tsx#L43)):*
> ```XML
>   const { register, watch } = useFormContext<FormState>();
>   const deleted = watch(`periods.${index}.deleted`);
> ```

NIT, would `toBeDeleted`  be clearer? Since it is not yet actually deleted (or at least I assume so :D)?

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: 45 of 58 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/hooks/urlQuery/useTimeRangeQueryParams.ts` line 30 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbnIg6DBbVs4WelpVBU:-NbnSfOH2rou0uUqROua:b-hfpv0l) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d18ec2c591a4695fb532722e59aef020c149dc64/ui/src/hooks/urlQuery/useTimeRangeQueryParams.ts#L30)):*

discussed f2f, probably not needed.
</blockquote></details>

removed
___
*[`ui/src/components/forms/common/EnumDropdown.tsx` line 65 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbUc2yMDkGOPjTTRBAX:-NbnTy_X2ry1U0UX4quV:bulmy6s) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ace284907a8c22ee92c782ad8d93ee425ffed0fd/ui/src/components/forms/common/EnumDropdown.tsx#L65)):*

Yeah, okay. Maybe we need to refactor this at some point. Nice that you tested this :+1: 
</blockquote></details>

Maybe :) 
___
*[`ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NaCCRPx0dC_uSGF8gjd:-NbnStMHCih3dcA60v56:b-eapfem) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bbb94de8a8131ca36a11edfb2de71438170659cd/ui/src/components/timetables/day-settings/RandomReferenceDay/RandomReferenceDayFormRow.tsx#L37)):*

Yes, the naming is now much clearer! Thanks
</blockquote></details>

Resolved?
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodRow.tsx` line 43 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbnCqW806fMB-n3ggOG:-NbnSciO32rN83I0V0NV:bb7f2mf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d18ec2c591a4695fb532722e59aef020c149dc64/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodRow.tsx#L43)):*

NIT, would `toBeDeleted`  be clearer? Since it is not yet actually deleted (or at least I assume so :D)?
</blockquote></details>

done

HenrikHartiala (approved):
> Reviewed 5 of 8 files at r6, 2 of 3 files at r7, 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: all files reviewed, 1 unresolved discussion (waiting on @Perholehto)

___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodRow.tsx` line 43 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627#-NbnCqW806fMB-n3ggOG:-NbnqsGI3_yfg3FIRziZ:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d18ec2c591a4695fb532722e59aef020c149dc64/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodRow.tsx#L43)):*

done
</blockquote></details>

Thanks!

HenrikHartiala (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/627)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Perholehto)

PR #630, opened 2023-07-03
Add test for changing the validity time range of a timetable

sr-vala:
Resolves HSLdevcom/jore4#1292

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)


HenrikHartiala (commented):
Looks good. Some tests came in to my mind, which we could add at some point.
* Now we change the validity period in the passing times view, we could add a test about changing it in the original vehicle schedule view (check the result before changing the observation date)
* We could also test what happens if we change a validity period of a higher priority timetable outside of the observationDate range, so that we also have a lower priority timetable valid on the observation date.
  * For example: we have Saturday Standard valid, and Saturday temporary valid on observation date, we can only see the temporary since it is higher priority. Now we change the temporary priority outside of observation range, and as the result probably the standard priority timetable should be visible.
* some conflicting cases in which we try to extend the validity period on top of another same priority level timetable

These are probably the best to do when we populate the timetable straight to database and not by using the UI

> Reviewed 6 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-:-NapU0neEEJuSknepGtH:bu9gzhk)* status: all files reviewed, 5 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 198 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Nap-sR-991lyv79UFKn:-Nap-sR-991lyv79UFKo:b-3c21up) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L198)):*
> ```XML
>     cy.visit('/timetables');
> 
>     timetablesMainPage.getImportButton().click();
> ```

There could be some kind of TODO comment here, describing to change these kind of database initializations using the upcoming test data insert tool. In the end I think we do not want to use the UI to populate the test data for every test.

This could maybe even be in some nicely named helper/utils function if we need to do this in many test files. Not sure where to put this function though.. could it be in importTimetablesPage pageObject, etc. 
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 226 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapRxFZ9gb4HcVRfJ5r:-NapRxFZ9gb4HcVRfJ5s:b40h68d) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L226)):*
> ```XML
>       );
>       vehicleScheduleDetailsPage.routeTimetableList.routeTimetablesSection.vehicleServiceTable
>         .getTable()
> ```

I guess this just gets the first table it finds? Should we make this function more precise, since there can be multiple vehicleServiceTables visible at the same time?
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 227 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapSHGB6tGzBz_cAlZk:-NapSHGB6tGzBz_cAlZl:b2nbmjh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L227)):*
> ```XML
>       vehicleScheduleDetailsPage.routeTimetableList.routeTimetablesSection.vehicleServiceTable
>         .getTable()
>         .click();
> ```

We click this, so I think we are navigating to the passingTimes -view. We could add this information to the test title, and maybe think if we should add tests to test the other view as well.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 245 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapQsmt4pR37h_T-UEM:-NapQsmt4pR37h_T-UEN:bmrqn68) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L245)):*
> ```XML
>       vehicleScheduleDetailsPage.toast.checkSuccessToastHasMessage(
>         'Aikataulun voimassaolo tallennettu',
>       );
> ```

We could also add a check here that the timetable is not visible anymore here. (We might need multiple e2e tests for this view and functionality, I'll list some thoughts in the end)
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 255 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapWQt_7rW7m4xaE8A7:-NapWQt_7rW7m4xaE8A8:bx1s5of) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L255)):*
> ```XML
>       vehicleScheduleDetailsPage.vehicleJourneyGroupInfo
>         .getValidityTimeRange()
>         .should('contain', '3.3.2025 - 3.3.2026');
> ```

We should maybe validate the other two fields as well? The departure count and the departure timerange, that they haven't changed.

sr-vala (commented):
I added these to my checklist. I'll create tickets for these when the test data tool can be used.

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-:-Natj0yI7XlDkWGQmse_:bdh5m6m)* status: all files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 198 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Nap-sR-991lyv79UFKn:-NatVc5C9x5FkP5Mnb6a:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L198)):*

There could be some kind of TODO comment here, describing to change these kind of database initializations using the upcoming test data insert tool. In the end I think we do not want to use the UI to populate the test data for every test.

This could maybe even be in some nicely named helper/utils function if we need to do this in many test files. Not sure where to put this function though.. could it be in importTimetablesPage pageObject, etc. 
</blockquote></details>

Done.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 226 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapRxFZ9gb4HcVRfJ5r:-NatVeKy8PMRA9KgKzVZ:b-oqfzyl) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L226)):*

I guess this just gets the first table it finds? Should we make this function more precise, since there can be multiple vehicleServiceTables visible at the same time?
</blockquote></details>

I changed this to clicking the header button of the table instead.  Do you think this needs more parameters in addition to the day type?
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 227 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapSHGB6tGzBz_cAlZk:-Natu_I1AKV3l_O27C0-:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L227)):*

We click this, so I think we are navigating to the passingTimes -view. We could add this information to the test title, and maybe think if we should add tests to test the other view as well.
</blockquote></details>

Done.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 245 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapQsmt4pR37h_T-UEM:-NatubmTEKpUVWMXHdYk:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L245)):*

We could also add a check here that the timetable is not visible anymore here. (We might need multiple e2e tests for this view and functionality, I'll list some thoughts in the end)
</blockquote></details>

Done.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 255 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapWQt_7rW7m4xaE8A7:-NatuKgL5yE_iWs-F7G0:b-i02yfj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L255)):*

We should maybe validate the other two fields as well? The departure count and the departure timerange, that they haven't changed.
</blockquote></details>

A bit off topic, but I guess there's no harm doing these. Time range was a bit tricky to handle.

HenrikHartiala (changes requested):
> Reviewed 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)* status: all files reviewed, 3 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 198 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Nap-sR-991lyv79UFKn:-NavvB_48Il7a4khHBNY:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L198)):*

Done.
</blockquote></details>

Thanks!
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 226 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapRxFZ9gb4HcVRfJ5r:-NavvPxx82QIOX2Zc2Z6:b-5xbneg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L226)):*

I changed this to clicking the header button of the table instead.  Do you think this needs more parameters in addition to the day type?
</blockquote></details>

Yeah I think that is enough :ok_hand:  There shouldn't be more than one of same daytype timetables visible at the same time (only highest priority of each day type should be visible). Thanks! 
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 227 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapSHGB6tGzBz_cAlZk:-Navw4IDBm2PfcbXsmU9:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L227)):*

Done.
</blockquote></details>

:+1: 
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 245 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapQsmt4pR37h_T-UEM:-NavxPRL-_lp2QgEmSGP:b-dzrwuw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L245)):*

Done.
</blockquote></details>

Great, nice!
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 255 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapWQt_7rW7m4xaE8A7:-NavyPyr8iFdjZ0Pf25X:bswgslz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L255)):*

A bit off topic, but I guess there's no harm doing these. Time range was a bit tricky to handle.
</blockquote></details>

Well I mean to be sure nothing else changed there. The reason for this is that we can't match the result by any label or id. That is why I was thinking that we could check that the data is still the same. But yeah, might be unnecessary.

What comes to the implementation. I do think you would need to give some parameters for the vehicleJourneyGroupInfo as well, since again there can be multiple visible at the same time (even in passing times view, since there is both directions of route and possibly multiple routes?)

Or wait, actually this vehicleJourneyGroupInfo is under the vehicleServiceTable, which you already get by the heading name "lauantai" you could use that. So use the day name to get the right VehicleServiceTable, and then use that one to check the time range and trip count and triptimerange.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 306 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NavxRjL6tZHnJyngM1c:-NavxRjL6tZHnJyngM1d:b9en02h) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/cypress/e2e/timetableValidityPeriod.cy.ts#L306)):*
> ```XML
>         'Aikataulun voimassaolo tallennettu',
>       );
>       cy.contains('Ei vuoroja');
> ```

Heh, I already know a case where this would fail :D. I even had to test this that "does this really show 'Ei vuoroja', and yes it does if there are no other timetables left on observation date after the change :D But yeah we will face this in the upcoming tests. I'll make a note and probably transform it in to a ticket soon. It is not very critical though.

But now that I think about this a bit more, there are multiple routes visible at the same time, so I think it would be safer to make it more precise under which route this  text is, because there could be another route on the screen which has no scheduled departures. 
___
*[`cypress/pageObjects/VehicleJourneyGroupInfo.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Naw1sOv7daR0U-p6Ik2:-Naw1sOv7daR0U-p6Ik3:b-zhleow) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/cypress/pageObjects/VehicleJourneyGroupInfo.ts#L20)):*
> ```XML
>   verifyTripTimeRange(startTime: string, endTime: string) {
>     this.getTripTimeRange().should(($range) => {
>       // We need to strip whitespace because the span contains a line break character
> ```

ahh, pfft yeah :D. It is because it is in template literal and prettier and/or prettier has cut the too wide line in such manner that the line break has come in to action with the resulting text. There is a simple fix though, I'll post it where the code is, so you can make this simpler.
___
*[`ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx` line 83 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Naw5WvC6edcd7JGclH9:-Naw5WvC6edcd7JGclHA:b-ub197j) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx#L83)):*
> ```XML
>           <span>|</span>
>           <span data-testid={testIds.tripTimeRange}>
>             {`${mapDurationToShortTime(firstTrip.start_time)} ...
> ```

So here, the line break is created in the string part of this template literal, that is why it is also in the end result. But by adding the linebreak to the code part, the text continues without the line break. (prettier might still format a bit different than this example, but the main point is that the linebreak is done "within the code" not "within the string" :+1: 



_Suggestion:_
```XML
            {`${mapDurationToShortTime(firstTrip.start_time)
            } ... ${mapDurationToShortTime(lastTrip.start_time)}`}
```

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)* status: all files reviewed, 3 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/pageObjects/VehicleJourneyGroupInfo.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Naw1sOv7daR0U-p6Ik2:-Naw6yGP4cXic_lf-BYy:bjj7c01) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/cypress/pageObjects/VehicleJourneyGroupInfo.ts#L20)):*

ahh, pfft yeah :D. It is because it is in template literal and prettier and/or prettier has cut the too wide line in such manner that the line break has come in to action with the resulting text. There is a simple fix though, I'll post it where the code is, so you can make this simpler.
</blockquote></details>

Whoops, can't find how to edit the comments :D. `prettier and/or prettier` was supposed to be `prettier and/or the developer` :D

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)* status: all files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 255 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapWQt_7rW7m4xaE8A7:-Nb-7-sL01-dCjF9FBm9:b-3sueni) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L255)):*

Well I mean to be sure nothing else changed there. The reason for this is that we can't match the result by any label or id. That is why I was thinking that we could check that the data is still the same. But yeah, might be unnecessary.

What comes to the implementation. I do think you would need to give some parameters for the vehicleJourneyGroupInfo as well, since again there can be multiple visible at the same time (even in passing times view, since there is both directions of route and possibly multiple routes?)

Or wait, actually this vehicleJourneyGroupInfo is under the vehicleServiceTable, which you already get by the heading name "lauantai" you could use that. So use the day name to get the right VehicleServiceTable, and then use that one to check the time range and trip count and triptimerange.
</blockquote></details>

Tried to handle this with getRouteTimetableSectionByDayType, but maybe the passing times view needs its own version of this still.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 306 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NavxRjL6tZHnJyngM1c:-Naz-aSxB4KRzxvGiNuN:bmgvxkl) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/cypress/e2e/timetableValidityPeriod.cy.ts#L306)):*

Heh, I already know a case where this would fail :D. I even had to test this that "does this really show 'Ei vuoroja', and yes it does if there are no other timetables left on observation date after the change :D But yeah we will face this in the upcoming tests. I'll make a note and probably transform it in to a ticket soon. It is not very critical though.

But now that I think about this a bit more, there are multiple routes visible at the same time, so I think it would be safer to make it more precise under which route this  text is, because there could be another route on the screen which has no scheduled departures. 
</blockquote></details>

Added a more precise check for this.
___
*[`cypress/pageObjects/VehicleJourneyGroupInfo.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Naw1sOv7daR0U-p6Ik2:-Nb-6tLg-Nn863wLRFlj:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/cypress/pageObjects/VehicleJourneyGroupInfo.ts#L20)):*

Whoops, can't find how to edit the comments :D. `prettier and/or prettier` was supposed to be `prettier and/or the developer` :D 
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx` line 83 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Naw5WvC6edcd7JGclH9:-Nb-6uPr401taWwAe1JB:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx#L83)):*

So here, the line break is created in the string part of this template literal, that is why it is also in the end result. But by adding the linebreak to the code part, the text continues without the line break. (prettier might still format a bit different than this example, but the main point is that the linebreak is done "within the code" not "within the string" :+1: 
</blockquote></details>

Done.

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)* status: 8 of 14 files reviewed, 1 unresolved discussion (waiting on @sr-vala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 255 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapWQt_7rW7m4xaE8A7:-NbDto33BWAcAqBHUvA-:b-8uerux) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L255)):*

Tried to handle this with getRouteTimetableSectionByDayType, but maybe the passing times view needs its own version of this still.
</blockquote></details>

Okay, so there are couple of things here: 
First I think it would be better to use chaining get and find(?) commands. I believe find was the one that only tries to search the child elements. I think we have some function `findByTestId` etc.

So as pseudo, I would maybe do something like this ->
 ```
1. getVehicleSchedulePage
2. Inside vehicle schedulePage element -> findRouteTimetableSection(parameters: route label and direction)
3. Inside the found routeTimetableSection element -> finVehicleServiceTable(parameters:  day type)
Now you have the correct card, and you can perform the checks for elements inside it. So in this example you check that it has the correct tripCount and timerange and validity range.
```
Im not even sure if you need to get the vehicleSchedulePage at all here, but I put it in the example anyway.

The second thing I spotted is the responsibilities, which are a bit off here. For example 
```
vehicleScheduleDetailsPage.routeTimetablesSection.assertRouteHasNoSchedules(
        '99',
      );

```
Here you give the '99' parameter for a function that lies in routeTimetablesSection, but that section is already a single instance of one route. So the route label parameter should be passed to the parent component. So in this particular case an *rough example* would be something like
`routeTimetableList.getRouteTimetablesSection('99').assertRouteHasNoSchedules();`
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 306 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NavxRjL6tZHnJyngM1c:-NbEGuEs5SIuapUhMtd1:b-bir5s1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/cypress/e2e/timetableValidityPeriod.cy.ts#L306)):*

Added a more precise check for this.
</blockquote></details>

Added some thoughts to another comment below, the same principles apply.
___
*[`cypress/pageObjects/VehicleJourneyGroupInfo.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Naw1sOv7daR0U-p6Ik2:-NbDszxm1qGeSdsqr4EF:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/cypress/pageObjects/VehicleJourneyGroupInfo.ts#L20)):*

Done.
</blockquote></details>

Thanks!
___
*[`ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx` line 83 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Naw5WvC6edcd7JGclH9:-NbDt2HD1w9Kg_4FQ3v-:b4putsz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/88c37e704de3d1e6c0a192ed989f93fc7996b49e/ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx#L83)):*

Done.
</blockquote></details>

:ok_hand:

**Review by HenrikHartiala — CHANGES_REQUESTED** [review]
> Reviewed 2 of 6 files at r4, 3 of 9 files at r5, 14 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)* status: all files reviewed, 5 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 268 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcfX3mc-iG60symTyLo:-NcfX3mc-iG60symTyLp:b-mxh1c2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/e2e/timetableValidityPeriod.cy.ts#L268)):*
> ```XML
>       );
> 
>       route99InboundSaturdayVehicleService
> ```

FYI, this now does not have the same asserts as in the previous revision (validity assert missing). 



_Suggestion:_
```XML
      route99InboundSaturdayVehicleService
        .get()
        .should('contain', '3.3.2025 - 3.3.2026')
        .and('contain', '2 lähtöä')
        .and('contain', '07:10 ... 08:20');
```
___
*[`cypress/pageObjects/PassingTimesByStopSection.ts` line 14 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgDjNoA56PyrM7c0E1:-NcgDjNp5AVnv0htGPZj:b-dlaixv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/PassingTimesByStopSection.ts#L14)):*
> ```XML
>     parent: ParentPageObject,
>     dayType: string,
>     priority: TimetablePriority,
> ```

This (priority) is okay here for now, I just started started to think that we should not need the priority precision at all, because by my understanding this can only show one daytype / priority at once. The implementation on the code side of this element confuses me a bit, and maybe I'm missing something. 
This is just FYI, and I need to check at some point if I could refactor that component a little bit, would make this simpler as well.

Although it might even be better to have the priority here, and possibly the same thing with VehicleServiceTable IF in the future there would be a view that shows multiple of these in the same view.

Anyways, this was just FYI and thinking out loud, so no changes required anywhere on behalf of this.
___
*[`cypress/pageObjects/PassingTimesByStopSection.ts` line 22 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgDnK10qGc8h71A8R7:-NcgDnK10qGc8h71A8R8:b2kmhex) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/PassingTimesByStopSection.ts#L22)):*
> ```XML
> 
>   get() {
>     return cy.getByTestId(
> ```

You need to get the parent first, otherwise this searches the whole page and can find multiple.
So:



_Suggestion:_
```XML
    return this.getParent().findByTestId(
      `PassingTimesByStopSection::${this.dayType}::${this.priority}`,
    );
```
___
*[`cypress/pageObjects/PassingTimesByStopSection.ts` line 37 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgGN1v5kyVMpAauVLZ:-NcgGN1v5kyVMpAauVL_:b35q3c9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/PassingTimesByStopSection.ts#L37)):*
> ```XML
>   }
> 
>   getValidityTimeRange() {
> ```

if you want to simplify things, you could use the same way here as with the VehicleServiceTable in the first test. So instead of these three getters, just use the
```
route99InboundSaturdayPassingTimesSection
        .should('contain', '3.3.2025 - 3.3.2026');
        .and('contain', '2 lähtöä');
        .and('contain', '07:10 ... 08:20');
```

And then again those testIds could be possibly removed. Buuut of course this implemented way if more precise and will carry on even though something else changes.
So either is fine, would be nice if they are the same way in both tests for consistency though. You also mentioned something about iterating further the VehicleJourneyGroupInfo page object, so maybe that will also fix the consistency at once.
___
*[`cypress/pageObjects/RouteTimetablesSection.ts` line 18 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-Ncf_8s1C7jHhraD9ghV:-Ncf_8s1C7jHhraD9ghW:b-vi5ovk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/RouteTimetablesSection.ts#L18)):*
> ```XML
> 
>   assertRouteHasNoSchedules() {
>     this.get().findByTestId('RouteTimetablesSection::noSchedules');
> ```

Actually if you want to make this even simpler you, could just go for `this.get().should('contain', 'Ei vuoroja')` and then we would not even need the`noSchedules` testId. But this is also fine :+1: 
___
*[`cypress/pageObjects/VehicleScheduleDetailsPage.ts` line 11 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgKNbwDbtlseOieixT:-NcgKNbwDbtlseOieixU:bxdb47v) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/VehicleScheduleDetailsPage.ts#L11)):*
> ```XML
>   passingTimesByStopTable = new PassingTimesByStopTable();
> 
>   dayTypeDropDown = new DayTypeDropdown();
> ```

FYI, This should be removed in the future, when the possibly other tests are refactored to use the more safe way. No need to fix this in this particular PR.
___
*[`cypress/pageObjects/VehicleScheduleDetailsPage.ts` line 13 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgKCsa8tjSnK0dmWhs:-NcgKCsa8tjSnK0dmWht:b-kcipr9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/VehicleScheduleDetailsPage.ts#L13)):*
> ```XML
>   dayTypeDropDown = new DayTypeDropdown();
> 
>   vehicleJourneyGroupInfo = new VehicleJourneyGroupInfo();
> ```

I think this can also be removed, since it is ambiguous on this level and no more used anywhere (?)
___
*[`cypress/pageObjects/VehicleServiceTable.ts` line 29 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcfVqNs-cbc5B1dhHzz:-NcfVqNs-cbc5B1dhI--:braeq5y) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/VehicleServiceTable.ts#L29)):*
> ```XML
> 
>   getHeadingButton() {
>     return this.getParent().findByTestId('VehicleServiceTable::headingButton');
> ```

This needs to be `this.get()`. There shouldn't be any case where you should use the `this.getParent()`, other than inside `get()`

If you use this.getParent() and then find the `VehicleServiceTable::headingButton`, it searches the whole parent (instead of this particular VehicleServiceTable) which could have multiple of those. 
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopSection.tsx` line 64 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgM6gA87TvfO3N9Sob:-NcgM6gB9hB_GUCt-Sau:b-pbwr6e) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopSection.tsx#L64)):*
> ```XML
>           className="space-y-6"
>           key={`${dayType}${vehicleJourneyGroup.priority}`}
>           data-testid={testIds.passingTimesByStopSection(
> ```

Like said earlier, this component might need some refactor, because the structure is a bit confusing. This testId should be on the outer div, but because of the minor confusion currently, we seem to need the priority from the vehicleJourneyGroup. Just FYI, no changes required for now.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)* status: all files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 255 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NapWQt_7rW7m4xaE8A7:-NcucWO3EGDT04wES3BJ:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/5d36547d6801938a328e2cf189d54d42f006d545/cypress/e2e/timetableValidityPeriod.cy.ts#L255)):*

Okay, so there are couple of things here: 
First I think it would be better to use chaining get and find(?) commands. I believe find was the one that only tries to search the child elements. I think we have some function `findByTestId` etc.

So as pseudo, I would maybe do something like this ->
 ```
1. getVehicleSchedulePage
2. Inside vehicle schedulePage element -> findRouteTimetableSection(parameters: route label and direction)
3. Inside the found routeTimetableSection element -> finVehicleServiceTable(parameters:  day type)
Now you have the correct card, and you can perform the checks for elements inside it. So in this example you check that it has the correct tripCount and timerange and validity range.
```
Im not even sure if you need to get the vehicleSchedulePage at all here, but I put it in the example anyway.

The second thing I spotted is the responsibilities, which are a bit off here. For example 
```
vehicleScheduleDetailsPage.routeTimetablesSection.assertRouteHasNoSchedules(
        '99',
      );

```
Here you give the '99' parameter for a function that lies in routeTimetablesSection, but that section is already a single instance of one route. So the route label parameter should be passed to the parent component. So in this particular case an *rough example* would be something like
`routeTimetableList.getRouteTimetablesSection('99').assertRouteHasNoSchedules();`

</blockquote></details>

Done.
___
*[`cypress/pageObjects/PassingTimesByStopSection.ts` line 14 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgDjNoA56PyrM7c0E1:-NcuckqVAei3IAAXQ0C9:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/PassingTimesByStopSection.ts#L14)):*

This (priority) is okay here for now, I just started started to think that we should not need the priority precision at all, because by my understanding this can only show one daytype / priority at once. The implementation on the code side of this element confuses me a bit, and maybe I'm missing something. 
This is just FYI, and I need to check at some point if I could refactor that component a little bit, would make this simpler as well.

Although it might even be better to have the priority here, and possibly the same thing with VehicleServiceTable IF in the future there would be a view that shows multiple of these in the same view.

Anyways, this was just FYI and thinking out loud, so no changes required anywhere on behalf of this.
</blockquote></details>

Done.
___
*[`cypress/pageObjects/PassingTimesByStopSection.ts` line 22 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgDnK10qGc8h71A8R7:-NcukjoM809GuuVpth_D:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/PassingTimesByStopSection.ts#L22)):*

You need to get the parent first, otherwise this searches the whole page and can find multiple.
So:
</blockquote></details>

Done.
___
*[`cypress/pageObjects/VehicleScheduleDetailsPage.ts` line 13 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcgKCsa8tjSnK0dmWhs:-Ncuqm6y397b6DlTpZJT:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/VehicleScheduleDetailsPage.ts#L13)):*

I think this can also be removed, since it is ambiguous on this level and no more used anywhere (?)
</blockquote></details>

Done.
___
*[`cypress/pageObjects/VehicleServiceTable.ts` line 29 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcfVqNs-cbc5B1dhHzz:-NcurOGW0vc9bZI49VUU:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3f2f72cab27280c03c779a6b907589903de1791c/cypress/pageObjects/VehicleServiceTable.ts#L29)):*

This needs to be `this.get()`. There shouldn't be any case where you should use the `this.getParent()`, other than inside `get()`

If you use this.getParent() and then find the `VehicleServiceTable::headingButton`, it searches the whole parent (instead of this particular VehicleServiceTable) which could have multiple of those.
</blockquote></details>

Done.

HenrikHartiala (approved):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630)* status: all files reviewed, 3 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/timetableImport.cy.ts` line 212 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcvRUWN1TS6jtfITmwG:-NcvRUWN1TS6jtfITmwH:bwyljli) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/499431e88c426fab8d6f478bc9bc521231b21ade/cypress/e2e/timetableImport.cy.ts#L212)):*
> ```XML
>     { tags: [Tag.Smoke, Tag.Timetables, Tag.HastusImport] },
>     () => {
>       const route99InboundTimetableSection = new RouteTimetablesSection(
> ```

FYI, this is not related to the changeValidityPeriod tests, and would be nice to have them in their own commit. Also I would probably fix these in a separate PR, since there are other unsafe methods in these tests still that needs to be fixed as well. But to get forward and not expanding this PR out of its scope, no changes required here and you can do the rest of the fixes in another PR.
___
*[`cypress/pageObjects/PassingTimesByStopSection.ts` line 37 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcvO6yH2T6SX5F4Iavy:-NcvO6yH2T6SX5F4Iavz:b-n7it7a) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/499431e88c426fab8d6f478bc9bc521231b21ade/cypress/pageObjects/PassingTimesByStopSection.ts#L37)):*
> ```XML
>   }
> 
>   getValidityTimeRange() {
> ```

Are these used anymore? Or do you plan to use these in further iterations? If answer is now for both, then these can be removed and also the testIds from the code can be removed.
___
*[`cypress/pageObjects/PassingTimesByStopTable.ts` line 7 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcvOu0d8hu9YlB9VE-J:-NcvOu0d8hu9YlB9VE-K:bp0ffhu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/499431e88c426fab8d6f478bc9bc521231b21ade/cypress/pageObjects/PassingTimesByStopTable.ts#L7)):*
> ```XML
>     new PassingTimesByStopTableRowPassingTime();
> 
>   getTable() {
> ```

Is this used anywhere? If not then I would remove it for now. It also currently searches the whole page and can find multiple.
___
*[`ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx` line 11 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/630#-NcvPjLHC_tOKVNpCupF:-NcvPjLHC_tOKVNpCupG:brzkheb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/499431e88c426fab8d6f478bc9bc521231b21ade/ui/src/components/timetables/common/VehicleJourneyGroupInfo.tsx#L11)):*
> ```XML
> const testIds = {
>   changeValidityButton: 'VehicleJourneyGroupInfo::changeValidityButton',
>   validityTimeRange: 'VehicleJourneyGroupInfo::validityRangeSpan',
> ```

If these are not used or not planned to use in the future, we can remove them. But they can persist if planned to use in the future.

PR #632, opened 2023-07-11
Add new timetable page logic

HenrikHartiala:
Three commits:
* Add focus-visible to Switch component
* Update hasura image and generated resources (this function adds the SQL function which is used in the new fetching logic)
* Add new fetching logic to timetable page
  * This commit is pretty massive, but the roots of the refactoring and the new logic were deep. So this
commit replaces the old way of fetching vehicle schedules for vehicleScheduleDetails and also adds support
for substitute operating days by line type. Earlier we had this a bit awkward logic of 'fetch all data in two separate
requests and then combine the information to compile wanted end result and this resulted in only the in effect schedule even though this style fetched a lot of redundant data. So with this new logic we get all the necessary information from database function and then we play with it in different
views. Now we get all vehicle schedules for day types on the observation date, we combine them in to vehicleJourneyGroups and calculate which is the one in effect.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/632)

### DIFF
--- a/docker/docker-compose.custom.yml
+++ b/docker/docker-compose.custom.yml
@@ -4,7 +4,7 @@ services:
     # Link to available jore4-hasura images in Docker Hub:
     # https://hub.docker.com/r/hsldevcom/jore4-hasura/tags?page=1&ordering=last_updated
     # The :hsl-tag contains the desired version of hsl specific hasura.
-    image: &hasura-image 'hsldevcom/jore4-hasura:hsl-main--20230524-15a564f76a75f82bf5c789491387cdce4394932b'
+    image: &hasura-image 'hsldevcom/jore4-hasura:hsl-main--20230711-f0391a96258129a48308711536ec733fa12fefd0'
     # Waiting for database to be ready to avoid startup delay due to hasura crashing at startup if db is offline
     # Note: this should only be done in development setups as Kubernetes does not allow waiting for services to be ready
     depends_on:
--- a/test-db-manager/src/generated/graphql.ts
+++ b/test-db-manager/src/generated/graphql.ts
@@ -10068,6 +10068,249 @@ export type TimetablesReturnValueTimetableVersionVarianceFields = {
   priority?: Maybe<Scalars['Float']>;
 };
 
+/**
+ * This return value table is used in function vehicle_journey.get_vehicle_schedules_on_date. It consists of vehicle_journey_id, vehicle_schedule_frame_id or
+ * substitute_operating_day_by_line_type_id and also enriched with data, which are used in the UI side.
+ */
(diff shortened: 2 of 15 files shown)


Perholehto (approved):
Lot of logic in this PR :thought_balloon: Couldn't find anything to severe. I added one comment about readability, but it's not blocking and more to discuss readability  etc..

> Reviewed 2 of 2 files at r1, 4 of 4 files at r2, 9 of 10 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/632#-:-Na6HrhM4cKO0Ygxa1Yc:bjj1h3)* status: all files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`ui/src/hooks/vehicle-service/useGetRouteTimetables.ts` line 166 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/632#-Na6E7R0Bns43YJq4BAB:-Na6E7R0Bns43YJq4BAC:bmaijhs) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/66b61a1b6672cbac3535f8d0180c15d260141b95/ui/src/hooks/vehicle-service/useGetRouteTimetables.ts#L166)):*
> ```XML
>     if (
>       activeDayTypeIds?.includes(group.dayType.day_type_id) &&
>       group.priority ===
> ```

For me if-statements that have conditions that contains "complex" logic are hard to read. I think for better readability that logic should be extracted to its own function. Not sure if everyone agrees with this :smile: 



_Code snippet:_
```XML
    if (
      activeDayTypeIds?.includes(group.dayType.day_type_id) &&
      group.priority === findHighestPriority...(args...)) {
        ...
        
```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/632)* status: 2 of 15 files reviewed, 1 unresolved discussion (waiting on @niknii-vincit and @Perholehto)

___
*[`ui/src/hooks/vehicle-service/useGetRouteTimetables.ts` line 166 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/632#-Na6E7R0Bns43YJq4BAB:-NaGLv0JE8KmqLQXQRGN:b-og8yff) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/66b61a1b6672cbac3535f8d0180c15d260141b95/ui/src/hooks/vehicle-service/useGetRouteTimetables.ts#L166)):*

I agree with this :point_up:
</blockquote></details>

Ah, yeah this is totally true. I extracted to couple of variables to make it more readable.

PR #636, opened 2023-07-26
Add hastus check on export

HenrikHartiala:
Commit messages should have everything explained, but to sum up:
* Refactored route export selecting
* Reset selections on search
* Remove unintuitive behaviour when nothing was selected
* When exporting a line, only the routes that are valid on observation date are now exported
* Some other minor refactoring and clean up in code
* Add validation before export that all the exported routes have their hastus place set for first and last stop

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636)

### DIFF
--- a/ui/src/components/common/LineTableRow.tsx
+++ b/ui/src/components/common/LineTableRow.tsx
@@ -23,11 +23,14 @@ const GQL_LINE_TABLE_ROW = gql`
   fragment line_table_row on route_line {
     name_i18n
     short_name_i18n
+    validity_start
+    validity_end
     priority
     ...line_map_params
     line_routes {
       ...route_map_params
       unique_label
+      direction
       route_journey_patterns {
         journey_pattern_id
         journey_pattern_refs {
@@ -35,6 +38,11 @@ const GQL_LINE_TABLE_ROW = gql`
           vehicle_journeys {
             vehicle_journey_id
           }
+        }
+        scheduled_stop_point_in_journey_patterns {
+          journey_pattern_id
+          scheduled_stop_point_sequence
+          is_used_as_timing_point
         }
       }
     }
--- a/ui/src/components/common/LineTableRow.tsx
+++ b/ui/src/components/common/LineTableRow.tsx
@@ -5,11 +5,7 @@ import {
   useAppSelector,
   useShowRoutesOnModal,
 } from '../../hooks';
-import {
-  deselectRouteUniqueLabelsAction,
-  selectExport,
-  selectRouteUniqueLabelsAction,
-} from '../../redux';
+import { deselectRowAction, selectExport, selectRowAction } from '../../redux';
 import { routeHasTimetables } from '../../utils/route';
 import {
   RouteLineTableRow,
@@ -58,7 +54,7 @@ export const LineTableRow = ({
 }: Props): JSX.Element => {
   const { showRoutesOnMapByLineLabel } = useShowRoutesOnModal();
   const dispatch = useAppDispatch();
-  const { selectedRouteUniqueLabels } = useAppSelector(selectExport);
+  const { selectedRows } = useAppSelector(selectExport);
 
   const showLineRoutes = () => {
     showRoutesOnMapByLineLabel(line);
@@ -66,11 +62,10 @@ export const LineTableRow = ({
 
   const onSelectChanged = (event: React.ChangeEvent<HTMLInputElement>) => {
     const selected = event.target.checked;
-    const selectAction = selected
-      ? selectRouteUniqueLabelsAction
-      : deselectRouteUniqueLabelsAction;
 
-    dispatch(selectAction(line.line_routes.map((route) => route.label)));
+    const selectAction = selected ? selectRowAction : deselectRowAction;
+
+    dispatch(selectAction(line.label));
   };
 
   const hasRoutes = line.line_routes?.length > 0;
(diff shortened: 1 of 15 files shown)


niknii-vincit (changes requested):
Looks good :thumbsup:

Just a few nit-picky things, and a reminder/question whether a note should be made into a ticket. 

> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 8 of 8 files at r5, 1 of 1 files at r6, 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-:-NaHYrWM6JVzYqeuMA_o:b-23i5it)* status: all files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 57 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-NaHUKOe5MhX2HUuioey):*
Should this `NOTE` be added as a ticket to the backlog?
___
*[`ui/src/components/routes-and-lines/search/ExportToolbar.tsx` line 51 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-NaHPy_cAthoaMY3pOsj:-NaHPy_cAthoaMY3pOsk:b31n7cq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ddfc65cdb8165a70f405c54342bbbb81899db491/ui/src/components/routes-and-lines/search/ExportToolbar.tsx#L51)):*
> ```XML
>           // lines without routes are not selectable and need to be filtered out
>           searchResultLabels: lines
>             .filter((line) => !!line.line_routes.length)
> ```

If I've understood correctly, there might not be many lines in a route, so in that case this is a redundant comment :thinking:  

Just making sure: can there be a huge amount of lines in one route? If so, the twice called`.filter((line) => !!line.line_routes.length)` could be called only once and saved to a variable?
___
*[`ui/src/hooks/routes/useExportRoutes.ts` line 29 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-NaHXCjc3ZKQ8ZXTQ3PD:-NaHXCjc3ZKQ8ZXTQ3PE:btr2rih) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3533fc1edfa781ccce5b8a2af6b3171f2de945ff/ui/src/hooks/routes/useExportRoutes.ts#L29)):*
> ```XML
>             return currentObj;
>           }
>           return minObj;
> ```

I would prefer if these two were separate methods. It would improve readability, and perhaps figuring out the first and last stops in a journey could be useful elsewhere in the application (if not implemented elsewhere already..)?



_Suggestion:_
```XML
    const isEligibleJourneyPatterns = route.route_journey_patterns.map((jp) => {
      const firstStop = getFirstStop(...);
      const lastStop = getLastStop(...);
```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636)* status: all files reviewed, 2 unresolved discussions (waiting on @niknii-vincit)

___
*[`-- commits` line 57 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-NaHUKOe5MhX2HUuioey):*

Should this `NOTE` be added as a ticket to the backlog?
</blockquote></details>

That's a good point, but since this is very "maybe" thing and we do not have any specs or designs for it, so I would rather discuss briefly with design first. Or what do you think? 
___
*[`ui/src/components/routes-and-lines/search/ExportToolbar.tsx` line 51 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-NaHPy_cAthoaMY3pOsj:-NaLDgoE79CCk7j9eIv1:be7pzdw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/ddfc65cdb8165a70f405c54342bbbb81899db491/ui/src/components/routes-and-lines/search/ExportToolbar.tsx#L51)):*

If I've understood correctly, there might not be many lines in a route, so in that case this is a redundant comment :thinking:  

Just making sure: can there be a huge amount of lines in one route? If so, the twice called`.filter((line) => !!line.line_routes.length)` could be called only once and saved to a variable?
</blockquote></details>

I'm not completely sure what you mean. Briefly, the relationship is that line is the parent, and one line can have multiple routes. We also have two directions possible for each labeled route (this is how it goes for most cases, since the buses go forth and back). 
And when we search for lines in the UI, and then use the exporting selecting tool, all the lines that do not have any routes within have their checkbox disabled. That is why the filtering is done here.

But yeah, that is true that the "linesWithRoutes" etc could be in a variable, since it is done twice. 
___
*[`ui/src/hooks/routes/useExportRoutes.ts` line 29 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-NaHXCjc3ZKQ8ZXTQ3PD:-NaLGtNY0-jAwU9V8vNP:bdrn3g0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/3533fc1edfa781ccce5b8a2af6b3171f2de945ff/ui/src/hooks/routes/useExportRoutes.ts#L29)):*

I would prefer if these two were separate methods. It would improve readability, and perhaps figuring out the first and last stops in a journey could be useful elsewhere in the application (if not implemented elsewhere already..)?
</blockquote></details>

Great suggestion! I'll do this. I nearly already thought if I would make some generalised function to use in both cases, but then that left out.

niknii-vincit (approved):
Looking good! :lgtm_strong: 

> Reviewed 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-:-NaMFqxp-3VjJrx5rgjW:b-4xb56c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

___
*[`-- commits` line 57 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/636#-NaHUKOe5MhX2HUuioey):*

That's a good point, but since this is very "maybe" thing and we do not have any specs or designs for it, so I would rather discuss briefly with design first. Or what do you think? 
</blockquote></details>

Ah, maybe ticket to the backlog was too "big of a step". I meant to make sure it's on some tasklist, etc so it would not be forgotten :ok_hand:

PR #650, opened 2023-08-23
Add common substitute period

Perholehto:
Adds common substitute period form and container. Minor refactoring to occasional substitute periods functionality.  

Resolves HSLdevcom/jore4#1338

This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650)

### DIFF
--- a/ui/src/components/forms/timetables/ObservationPeriodForm.tsx
+++ b/ui/src/components/forms/timetables/ObservationPeriodForm.tsx
@@ -37,7 +37,10 @@ export const ObservationPeriodForm = (): JSX.Element => {
   const { t } = useTranslation();
   const { isInvalidDateRange } = useTimeRangeQueryParams();
   const {
-    settings: { isOccasionalSubstitutePeriodFormDirty },
+    settings: {
+      isOccasionalSubstitutePeriodFormDirty,
+      isCommonSubstitutePeriodFormDirty,
+    },
   } = useAppSelector(selectTimetable);
 
   return (
@@ -48,21 +51,32 @@ export const ObservationPeriodForm = (): JSX.Element => {
           dateInputId="startDate"
           testId={testIds.startDate}
           queryParamName={QueryParameterName.StartDate}
-          disabled={isOccasionalSubstitutePeriodFormDirty}
+          disabled={
+            isOccasionalSubstitutePeriodFormDirty ||
+            isCommonSubstitutePeriodFormDirty
+          }
         />
         <DateControl
           label={t('timetables.observationPeriodForm.endDate')}
           dateInputId="endDate"
           testId={testIds.endDate}
           queryParamName={QueryParameterName.EndDate}
-          disabled={isOccasionalSubstitutePeriodFormDirty}
+          disabled={
+            isOccasionalSubstitutePeriodFormDirty ||
+            isCommonSubstitutePeriodFormDirty
+          }
         />
       </div>
       <div className="h-12 pt-2">
         <Visible visible={isInvalidDateRange}>
           <ValidationError errorMessage={t('formValidation.timeRange')} />
         </Visible>
-        <Visible visible={isOccasionalSubstitutePeriodFormDirty}>
+        <Visible
+          visible={
+            isOccasionalSubstitutePeriodFormDirty ||
+            isCommonSubstitutePeriodFormDirty
+          }
+        >
           <WarningText
             warningMessage={t(
               'timetables.observationPeriodForm.warningMessage',
--- a/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx
+++ b/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx
@@ -1,0 +1,247 @@
+import { zodResolver } from '@hookform/resolvers/zod';
+import orderBy from 'lodash/orderBy';
+import range from 'lodash/range';
+import { DateTime } from 'luxon';
+import { useEffect, useMemo, useRef, useState } from 'react';
+import { FormProvider, useFieldArray, useForm } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
+import { TimetablesServiceCalendarSubstituteOperatingPeriod } from '../../../../generated/graphql';
+import { useAppDispatch, useTimeRangeQueryParams } from '../../../../hooks';
+import { Row } from '../../../../layoutComponents';
+import { setIsCommonSubstitutePeriodFormDirtyAction } from '../../../../redux/slices/timetable';
+import { mapToISODate, padToTwoDigits } from '../../../../time';
+import { ConfirmationDialog, SimpleButton } from '../../../../uiComponents';
+import {
+  mapDateTimeToFormState,
+  mapLineTypes,
+  parseSubstituteDayOfWeek,
+  submitFormByRef,
+} from '../../../../utils';
+import { commonSubstituteDayData } from '../common_substitute_day_data';
+import {
+  CommonDayType,
+  FormState,
+  schema,
+} from './CommonSubstitutePeriodForm.types';
+import { CommonSubstitutePeriodItem } from './CommonSubstitutePeriodItem';
+
+const testIds = {
+  cancelButton: 'CommonSubstitutePeriodForm::cancelButton',
+  saveButton: 'CommonSubstitutePeriodForm::saveButton',
+};
+
+type GeneratedDate = { name: string; date: DateTime };
+
+const generatePresetDatesForDateRange = (
+  startDate: DateTime,
+  endDate: DateTime,
+) => {
+  const years = range(startDate.year, endDate.year + 1);
+  const generatedDates: GeneratedDate[] = [];
+
+  // Generate individual days for each year
+  // return generated dates within date range
+  years.forEach((year) => {
+    commonSubstituteDayData.forEach((day) => {
+      const date = DateTime.fromISO(
+        `${year}-${padToTwoDigits(day.month)}-${padToTwoDigits(day.day)}`,
+      );
+      if (date >= startDate && date <= endDate) {
+        generatedDates.push({
+          name: `${day.name} ${year}`,
+          date,
+        });
+      }
+    });
+  });
+
+  return generatedDates;
+};
+
+export const mapCommonSubstituteOperatingPeriodsToCommonDays = (
+  commonSubstituteOperatingPeriods: TimetablesServiceCalendarSubstituteOperatingPeriod[],
+): CommonDayType[] => {
+  return commonSubstituteOperatingPeriods?.map((period) => {
+    const lineTypes: Set<string> = new Set();
+    period.substitute_operating_day_by_line_types.forEach(
+      (operatingDayByLineType) =>
+        lineTypes.add(operatingDayByLineType.type_of_line),
+    );
+
+    return {
+      periodId: period.substitute_operating_period_id,
+      periodName: period.period_name,
+      supersededDate: mapDateTimeToFormState(
+        period.substitute_operating_day_by_line_types[0].superseded_date,
+      ),
+      substituteDayOfWeek: parseSubstituteDayOfWeek(
+        period.substitute_operating_day_by_line_types[0].substitute_day_of_week,
+      ),
+      lineTypes: mapLineTypes(lineTypes),
+      fromDatabase: true,
+      created: false,
+      isPreset: period.is_preset,
+    };
+  });
+};
+
+const combineCommonDaysWithPresetDates = (
+  commonDays: CommonDayType[],
+  presetDays: GeneratedDate[],
+): CommonDayType[] => {
+  presetDays.forEach((presetDay) => {
+    const presetDayFoundInCommonDays = commonDays.some(
+      (commonDay) =>
+        presetDay.name === commonDay.periodName &&
+        mapToISODate(presetDay.date) === commonDay.supersededDate,
+    );
+
+    if (!presetDayFoundInCommonDays) {
+      commonDays.push({
+        periodName: presetDay.name,
+        supersededDate: mapDateTimeToFormState(presetDay.date),
+        lineTypes: '',
+        substituteDayOfWeek: '',
+        fromDatabase: false,
+        created: false,
+        isPreset: true,
+      });
+    }
+  });
+
+  return commonDays;
+};
+
+interface Props {
+  className?: string;
+  commonDays?: CommonDayType[];
+  onSubmit: (state: FormState) => void;
+}
+
+export const CommonSubstitutePeriodForm = ({
+  className,
+  commonDays,
+  onSubmit,
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+  const formRef = useRef<ExplicitAny>(null);
+  const dispatch = useAppDispatch();
+
+  const { startDate, endDate } = useTimeRangeQueryParams();
+
+  const [isResetting, setIsResetting] = useState<boolean>(false);
+
+  const generatedPresetDates = useMemo(
+    () => generatePresetDatesForDateRange(startDate, endDate),
+    [startDate, endDate],
+  );
+
+  const displayedCommonDays = combineCommonDaysWithPresetDates(
+    commonDays ?? [],
+    generatedPresetDates,
+  );
+
+  const orderedDisplayedCommonDays = orderBy(
+    displayedCommonDays,
+    ['supersededDate'],
+    ['asc'],
+  );
+
+  const methods = useForm<FormState>({
+    values: { commonDays: orderedDisplayedCommonDays },
+    resolver: zodResolver(schema),
+  });
+
+  const {
+    clearErrors,
+    control,
+    formState: { isDirty },
+    handleSubmit,
+    reset,
+    resetField,
+    setValue,
+  } = methods;
+
+  const { fields } = useFieldArray({
+    control,
+    name: 'commonDays',
+  });
+
+  const onReset = () => {
+    setIsResetting(false);
+    reset();
+  };
+
+  const onSave = () => {
+    submitFormByRef(formRef);
+  };
+
+  useEffect(() => {
+    dispatch(setIsCommonSubstitutePeriodFormDirtyAction(isDirty));
+  }, [dispatch, isDirty]);
+
+  const update = (index: number, created: boolean) => {
+    // Clear errors from specific item only when "removing" it
+    if (created) {
+      clearErrors(`commonDays.${index}`);
+    }
+
+    if (!created) {
+      setValue(`commonDays.${index}.created`, true, {
+        shouldDirty: true,
+        shouldTouch: true,
+      });
+    } else {
+      resetField(`commonDays.${index}`);
+    }
+  };
+
+  return (
+    <div className={className}>
+      {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+      <FormProvider {...methods}>
+        <form ref={formRef} onSubmit={handleSubmit(onSubmit)}>
+          <div className="flex flex-wrap justify-between gap-8">
+            {fields.map((field, index) => (
+              <CommonSubstitutePeriodItem
+                field={field}
+                index={index}
+                key={field.id}
+                update={update}
+              />
+            ))}
+          </div>
+          <Row className="my-8 justify-end space-x-4">
+            <SimpleButton
+              onClick={() => setIsResetting(true)}
+              disabled={!isDirty}
+              id="cancel-button"
+              testId={testIds.cancelButton}
+              inverted
+            >
+              {t('cancel')}
+            </SimpleButton>
+
+            <SimpleButton
+              onClick={onSave}
+              disabled={!isDirty}
+              id="save-button"
+              testId={testIds.saveButton}
+            >
+              {t('save')}
+            </SimpleButton>
+          </Row>
+        </form>
+      </FormProvider>
+      <ConfirmationDialog
+        isOpen={isResetting}
+        onCancel={() => setIsResetting(false)}
+        onConfirm={onReset}
+        title={t('confirmResetCommonSubstitutePeriodDialog.title')}
+        description={t('confirmResetCommonSubstitutePeriodDialog.description')}
+        confirmText={t('confirmResetCommonSubstitutePeriodDialog.confirmText')}
+        cancelText={t('confirmResetCommonSubstitutePeriodDialog.cancelText')}
+      />
+    </div>
+  );
+};
--- a/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.types.ts
+++ b/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.types.ts
@@ -1,0 +1,22 @@
+import { z } from 'zod';
+import { SubstituteDayOfWeek } from '../../../../types/enums';
+import { requiredDate } from '../../../forms/common/customZodSchemas';
+
+const commonDay = z.object({
+  periodId: z.string().optional(),
+  periodName: z.string(),
+  supersededDate: requiredDate,
+  substituteDayOfWeek: z.nativeEnum(SubstituteDayOfWeek),
+  lineTypes: z.string(),
+  fromDatabase: z.boolean(),
+  created: z.boolean(),
+  isPreset: z.boolean(),
+});
+
+export const schema = z.object({
+  commonDays: commonDay.array(),
+});
+
+export type CommonDayType = z.infer<typeof commonDay>;
+
+export type FormState = z.infer<typeof schema>;
--- a/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.types.ts
+++ b/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.types.ts
@@ -1,17 +1,50 @@
+import isEmpty from 'lodash/isEmpty';
 import { z } from 'zod';
 import { SubstituteDayOfWeek } from '../../../../types/enums';
 import { requiredDate } from '../../../forms/common/customZodSchemas';
 
-const commonDay = z.object({
-  periodId: z.string().optional(),
-  periodName: z.string(),
-  supersededDate: requiredDate,
-  substituteDayOfWeek: z.nativeEnum(SubstituteDayOfWeek),
-  lineTypes: z.string(),
-  fromDatabase: z.boolean(),
-  created: z.boolean(),
-  isPreset: z.boolean(),
-});
+interface CommonDay {
+  created: boolean;
+  fromDatabase: boolean;
+  substituteDayOfWeek: string | SubstituteDayOfWeek;
+  lineTypes: string;
+  periodId?: string;
+  periodName: string;
+  supersededDate: string;
+  isPreset: boolean;
+}
+
+type CommonDayProperty = 'substituteDayOfWeek' | 'lineTypes';
+
+const commonDayValidation = (
+  data: CommonDay,
+  property: CommonDayProperty,
+): boolean => {
+  if (data.created && !data.fromDatabase) {
+    return !isEmpty(data[property]);
+  }
+  return true;
+};
+
+const commonDay = z
+  .object({
+    periodId: z.string().optional(),
+    periodName: z.string(),
+    supersededDate: requiredDate,
+    substituteDayOfWeek: z.nativeEnum(SubstituteDayOfWeek).or(z.string()),
+    lineTypes: z.string(),
+    fromDatabase: z.boolean(),
+    created: z.boolean(),
+    isPreset: z.boolean(),
+  })
+  .refine((data) => commonDayValidation(data, 'substituteDayOfWeek'), {
+    message: 'substituteDayOfWeek',
+    path: ['substituteDayOfWeek'],
+  })
+  .refine((data) => commonDayValidation(data, 'lineTypes'), {
+    message: 'lineTypes',
+    path: ['lineTypes'],
+  });
 
 export const schema = z.object({
   commonDays: commonDay.array(),
--- a/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx
+++ b/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx
@@ -1,0 +1,111 @@
+import { FieldArrayWithId, useFormContext } from 'react-hook-form';
+import { Visible } from '../../../../layoutComponents';
+import { mapToShortDate } from '../../../../time';
+import { EditCloseButton } from '../../../../uiComponents/EditCloseButton';
+import { InputField } from '../../../forms/common';
+import { LineTypeMultiSelectDropdown } from '../../../forms/timetables/LineTypeMultiSelectDropdown';
+import { SubstituteDayOfWeekDropdown } from '../../../forms/timetables/SubstituteDayOfWeekDropdown';
+import { CommonDayType, FormState } from './CommonSubstitutePeriodForm.types';
+
+const testIds = {
+  periodName: 'CommonSubstitutePeriodItem::periodName',
+  supersededDate: 'CommonSubstitutePeriodItem::supersededDate',
+  substituteDayOfWeek: 'CommonSubstitutePeriodItem::substituteDayOfWeek',
+  lineTypes: 'CommonSubstitutePeriodItem::lineTypes',
+  editCloseButton: 'CommonSubstitutePeriodItem::editCloseButton',
+};
+
+interface Props {
+  index: number;
+  field: FieldArrayWithId<
+    {
+      commonDays: CommonDayType[];
+    },
+    'commonDays',
+    'id'
+  >;
+  update: (index: number, flag: boolean) => void;
+}
+
+const PeriodNameAndDateLabel = ({
+  periodName,
+  supersededDate,
+}: {
+  periodName: string;
+  supersededDate: string;
+}) => {
+  return (
+    <div className="flex-shrink-1 flex flex-grow-0 basis-11/12 flex-col">
+      <span className="font-semibold" data-testid={testIds.periodName}>
+        {periodName}
+      </span>
+      <span data-testid={testIds.supersededDate}>
+        {mapToShortDate(supersededDate)}
+      </span>
+    </div>
+  );
+};
+
+export const CommonSubstitutePeriodItem = ({
+  index,
+  field,
+  update,
+}: Props): JSX.Element => {
+  const { register, watch } = useFormContext<FormState>();
+  const created = watch(`commonDays.${index}.created`);
+
+  return (
+    <div className="flex basis-[28%] flex-wrap" key={field.id}>
+      <PeriodNameAndDateLabel
+        periodName={field.periodName}
+        supersededDate={field.supersededDate}
+      />
+      <div className="flex basis-1/12 justify-end">
+        <Visible visible={!field.fromDatabase}>
+          <EditCloseButton
+            showEdit={!created}
+            onEdit={() => update(index, created)}
+            onClose={() => update(index, created)}
+            testId={testIds.editCloseButton}
+          />
+        </Visible>
+      </div>
+      <div className="flex basis-full gap-2">
+        <InputField<FormState>
+          className="basis-1/2"
+          translationPrefix="timetables.settings"
+          testId={testIds.substituteDayOfWeek}
+          fieldPath={`commonDays.${index}.substituteDayOfWeek`}
+          // eslint-disable-next-line react/no-unstable-nested-components
+          inputElementRenderer={(props) => (
+            <SubstituteDayOfWeekDropdown
+              disabled={!field.fromDatabase && !created}
+              // eslint-disable-next-line react/jsx-props-no-spreading
+              {...props}
+            />
+          )}
+        />
+        <InputField<FormState>
+          className="basis-1/2"
+          translationPrefix="timetables.settings"
+          testId={testIds.lineTypes}
+          fieldPath={`commonDays.${index}.lineTypes`}
+          // eslint-disable-next-line react/no-unstable-nested-components
+          inputElementRenderer={(props) => (
+            <LineTypeMultiSelectDropdown
+              disabled={!field.fromDatabase && !created}
+              // eslint-disable-next-line react/jsx-props-no-spreading
+              {...props}
+            />
+          )}
+        />
+      </div>
+
+      <input
+        type="checkbox"
+        hidden
+        {...register(`commonDays.${index}.created`)}
+      />
+    </div>
+  );
+};
--- a/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx
+++ b/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx
@@ -23,9 +23,10 @@ export const CommonSubstitutePeriodSection = ({
   const { startDate, endDate } = useTimeRangeQueryParams();
 
   const {
-    presetSubstituteOperatingPeriods,
-    refetchSubstituteOperatingPeriods,
+    commonSubstituteOperatingPeriods,
+    refetchCommonSubstituteOperatingPeriods,
   } = useGetSubstituteOperatingPeriods({ startDate, endDate });
+
   const { prepareAndExecute: prepareAndExecuteCreate } =
     useCreateSubstituteOperatingPeriod();
 
@@ -47,7 +48,7 @@ export const CommonSubstitutePeriodSection = ({
     try {
       await prepareAndExecuteEdit({ form: { periods } });
       await prepareAndExecuteCreate({ form: { periods } });
-      refetchSubstituteOperatingPeriods();
+      refetchCommonSubstituteOperatingPeriods();
 
       showSuccessToast(t('timetables.settings.saveSuccess'));
     } catch (err) {
--- a/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx
+++ b/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx
@@ -1,0 +1,70 @@
+import { useTranslation } from 'react-i18next';
+import {
+  useCreateSubstituteOperatingPeriod,
+  useEditSubstituteOperatingPeriod,
+  useGetSubstituteOperatingPeriods,
+} from '../../../../hooks/substitute-operating-periods';
+import { useTimeRangeQueryParams } from '../../../../hooks/urlQuery';
+import { showDangerToastWithError, showSuccessToast } from '../../../../utils';
+import {
+  CommonSubstitutePeriodForm,
+  mapCommonSubstituteOperatingPeriodsToCommonDays,
+} from './CommonSubstitutePeriodForm';
+import { FormState } from './CommonSubstitutePeriodForm.types';
+
+interface Props {
+  className?: string;
+}
+
+export const CommonSubstitutePeriodSection = ({
+  className = '',
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+  const { startDate, endDate } = useTimeRangeQueryParams();
+
+  const {
+    presetSubstituteOperatingPeriods,
+    refetchSubstituteOperatingPeriods,
+  } = useGetSubstituteOperatingPeriods({ startDate, endDate });
+  const { prepareAndExecute: prepareAndExecuteCreate } =
+    useCreateSubstituteOperatingPeriod();
+
+  const { prepareAndExecute: prepareAndExecuteEdit } =
+    useEditSubstituteOperatingPeriod();
+
+  const onSubmit = async (form: FormState) => {
+    // Filter out days that are not already in database or are not added in UI
+    // Add to begin/end date for each day to match
+    // occasional substitute operating periods form type
+    const periods = form.commonDays
+      .filter((day) => day.created || day.fromDatabase)
+      .map((d) => ({
+        ...d,
+        beginDate: d.supersededDate,
+        endDate: d.supersededDate,
+      }));
+
+    try {
+      await prepareAndExecuteEdit({ form: { periods } });
+      await prepareAndExecuteCreate({ form: { periods } });
+      refetchSubstituteOperatingPeriods();
+
+      showSuccessToast(t('timetables.settings.saveSuccess'));
+    } catch (err) {
+      showDangerToastWithError(t('errors.saveFailed'), err);
+    }
+  };
+
+  return (
+    <div className={className}>
+      <h2>{t('timetables.settings.commonSubstituteDays')}</h2>
+      <CommonSubstitutePeriodForm
+        className="my-8"
+        commonDays={mapCommonSubstituteOperatingPeriodsToCommonDays(
+          commonSubstituteOperatingPeriods,
+        )}
+        onSubmit={onSubmit}
+      />
+    </div>
+  );
+};
--- a/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/index.ts
+++ b/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/index.ts
@@ -1,0 +1,1 @@
+export * from './CommonSubstitutePeriodSection';
--- a/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx
+++ b/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx
@@ -46,6 +46,7 @@ const emptyRowObject: PeriodType = {
   beginTime: '04:30',
   endTime: '28:30',
   toBeDeleted: false,
+  isPreset: false,
 };
 
 const mapDateTimeToFormState = (date: Maybe<DateTime> | undefined): string => {
@@ -112,6 +113,7 @@ const convertToPeriodSchema = (
       ),
       lineTypes: mapLineTypes(lineTypes),
       toBeDeleted: false,
+      isPreset: false,
     };
   });
   return { periods };
--- a/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx
+++ b/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx
@@ -97,10 +97,10 @@ const convertToPeriodSchema = (
   return { periods };
 };
 
-export const mapSubstituteOperatingPeriodsToFormState = (
-  data: TimetablesServiceCalendarSubstituteOperatingPeriod[],
+export const mapOccasionalSubstituteOperatingPeriodsToFormState = (
+  occasionalSubstituteOperatingPeriods: TimetablesServiceCalendarSubstituteOperatingPeriod[],
 ) => {
-  return convertToPeriodSchema(data);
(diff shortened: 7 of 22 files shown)


HenrikHartiala (changes requested):
> Reviewed 1 of 1 files at r1, 5 of 5 files at r2, 1 of 1 files at r3, 1 of 2 files at r4, 7 of 11 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650)* status: all files reviewed, 17 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcaklzX-byWlZMu8QX9):*
NIT,  layout? Not sure though if it is a change to layout either, if its just a minor padding change :D. But formatting also sounds too strong.
Maybe something like this?



_Suggestion:_
```
- f0e7bd1: Remove OccasionalSubstitutePeriodSection bottom padding
```
___
*[`-- commits` line 13 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcanvSCE43yEU3tB_6o):*
NIT, I think the hyphen is extra



_Suggestion:_
```
- 6e493de: Add common substitute days data
```
___
*[`-- commits` line 16 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-Ncat5Ev8PMRpPgFwQXo):*
NIT, `better place` is a bit non-informative :D. How about `Move substitute day helper functions under utils`or something similar a bit more descriptive?
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 34 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbGrib2YFF_OUvpUVo:-NcbGrib2YFF_OUvpUVp:b-xycwjm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L34)):*
> ```XML
> 
> const generatePresetDates = (startDate: DateTime, endDate: DateTime) => {
>   const years = range(startDate.year, endDate.year + 1);
> ```

Hmm, so we generate dates always at least for 1 year, and then filter the result by the startDate and endDate. Is there some particular benefit with this method, or would it be possible to just generate the dates that are found between the startDate and endDate?

I mean I do understand that it is easier to just generate all commonSubstituteDays at once, then filter, but would it be more complex to just generate by the dates?
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 52 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbNM6K9tg2qgbeMg7V:-NcbNM6K9tg2qgbeMg7W:bi8btk0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L52)):*
> ```XML
> };
> 
> export const mapSubstituteOperatingPeriodsToFormState = (
> ```

It feels like this function does more stuff than just maps periods to form state, it seems like it also generates the dates and does some replacing logic there as well. So could this be refactored to something bit simpler? Maybe the "generate & replace" stuff could be somewhere else, and this function could just map the results. 
(FYI, wrote this comment after the two comments below, because the more I watched this the more I started to understand)
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 57 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbHqc991IESIGIwKGe:-NcbHqc991IESIGIwKGf:bq0s8jk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L57)):*
> ```XML
>   endDate: DateTime,
> ) => {
>   const commonDays: CommonDayType[] = input?.map((item) => {
> ```

extra `?`, since input is not optional parameter. Also wondering if there is a more descriptive name for `input`, because now we are mapping `input`, and the items inside are named `item`, so the content gets a bit lost there.



_Suggestion:_
```XML
  const commonDays: CommonDayType[] = input.map((item) => {
```
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 83 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbJK_K7gg-gc3kNwK4:-NcbJK_K7gg-gc3kNwK5:b-5q6pmh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L83)):*
> ```XML
> 
>   // Remove dates that overlap with days from database
>   const filteredDays = presetDates.filter((d) => {
> ```

Hmm, we already filter the generated preset dates in the generatePresetDates once. What is this filter? I'm cant catch what are these "days from database"? 

Ahhh, wait so this is the combination of dates that are set and dates that are not set?
I would probably rethink the naming of these a little bit to make it more clear.
Now we have presetDates, commonDays, filteredDays and so on, it's a bit confusing.

Don't have the perfect solution straight away, but maybe something to iterate.
How about we generate the presetDates, then we enrich/replace the values with already set substitute days and in the end we have displayedSubstituteDays etc.
This enrich/replace could be its own helper function also.

(After this comment I wrote the comment related above, so maybe we could discuss this further in telco if needed.)
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 114 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbPd37Diuo1iWs-0VU:-NcbPd37Diuo1iWs-0VV:b-xte6sq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L114)):*
> ```XML
>   onSubmit: (state: FormState) => void;
> }): JSX.Element => {
>   const [isResetting, setIsResetting] = useState<boolean>(false);
> ```

NIT, these consts could be maybe rearranged and have some line breaks here and there. Maybe the most valuable thing here would be to at least move this useState declaration to bottom of the list, since it is the least "boiler" here :D But this is really nitpicking.
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 32 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbQjkN867jjhATvtYA:-NcbQjkN867jjhATvtYB:b-qe5r61) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L32)):*
> ```XML
> }
> 
> const EditAndRemoveButtons = ({
> ```

This naming implies that it returns multiple buttons, but in the end it actually returns only one button. 
So how about `EditRemoveButton` etc?
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 45 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbRy0k7R_L2NMfWMgZ:-NcbRy0k7R_L2NMfWMg_:blx102h) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L45)):*
> ```XML
>   return (
>     <div className="flex justify-end">
>       <Visible visible={showAnyButtons}>
> ```

This outer visible logic with `showAnyButtons` would be better on the parent. So we either render the component, or we don't render the component. Now it is baked inside this component, which reduces the readability by a tiny bit.

Another thing which came to mind is that this is pretty specific button which has a specific logic. You could make this more generic UI element, which takes in just onEdit(), onClose() and some flag determining which button to show, if you want, but this is not necessary, since don't know if this would be used anywhere else.
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 75 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbU2wEAosWKPJKgW1m:-NcbU2wF3m4dOVYrTugN:b-44r0fe) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L75)):*
> ```XML
> }) => {
>   return (
>     <div className="mb-0 flex flex-col">
> ```

`mb-0` seems extra
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 95 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbZMc_DqzkubpPCK35:-NcbZMc_DqzkubpPCK36:b35rk2u) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L95)):*
> ```XML
>   return (
>     <div
>       className="mr-8 mt-8 grid h-[20%] w-[30%] grid-cols-2 gap-0 gap-y-4"
> ```

Lots of style definitions here, also precise margins. You could check if this suits better:
Keep these here `grid h-[20%] w-[30%] grid-cols-2`
Change the parent element div like this:
```
<div className\="flex flex-wrap justify-between gap-8">
  {fields.map((field, index) => (
    <CommonSubstitutePeriodItem
      field={field}
      index={index}
      key={field.id}
      update={update}
    />
  ))}
</div>
```
(This might not completely match the layout, but is ofcourse iterable)

Generally it is better to have the kind of "layout" definitions given from outside the component (or with flexbox: on the parent), so that we won't enforce some particular style to a component which could be used elsewhere as well. (another comment below about styles coming from outside the component) 
And the second iteration to that is to use the nice flexbox definitions, which usually function a bit better than precise margins.

What do you think?
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx` line 53 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbaUCGECDV_WtuRYPq:-NcbaUCGECDV_WtuRYPr:bum9nda) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx#L53)):*
> ```XML
> 
>   return (
>     <div className="py-8">
> ```

As mentioned above, the minor layout things are generally better to be given from outside the component to avoid enforcing styles. So here I would take the className as parameter which defaults to empty string, and then give the `py-8` from outside this component.
```
interface Props {
  className?: string;
}

export const CommonSubstitutePeriodSection \= ({className \= ''}: Props): JSX.Element \=> {
...
```

and then give the `my-8` as parameter. Aaaand one last thing, doesn't really matter but margin might be the more desirable attribute here :D



_Suggestion:_
```XML
    <div className={className}>
```
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx` line 49 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcamHp29CcfyqpRmHil:-NcamHp29CcfyqpRmHim:bv9gvr5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6628085babe49a326a23db5157871c302bf02978/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx#L49)):*
> ```XML
>   endTime: '28:30',
>   toBeDeleted: false,
>   isPreset: false,
> ```

Is there presets in OccasionalSubstitutePeriod? Aren't the presets only on the General substitute days? (I am possibly misunderstanding something)

If the form is used for both (Occasional and general) then I would probably rename this to `SubstitutePeriodForm.tsx` (still not sure if this is the case). Maybe I'll find my answer later in this PR :D
___
*[`ui/src/data/common_substitute_days.ts` line 1 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcapmA76Hy4epTCXDgj:-NcapmA76Hy4epTCXDgk:b-oz3gos) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L1)):*
> ```XML
> type commonSubstituteDay = {
> ```

What comes to the folder choice, I'm not sure what the files in this `src/data` folder are, seems like some really old mock data, but this file could probably be better in the 'substitute-day-settings' folder where the rest of the components are. 

What do you think?
___
*[`ui/src/data/common_substitute_days.ts` line 2 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-Ncap1vHBxy0Uum9NXgU:-Ncap1vHBxy0Uum9NXgV:b8j884q) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L2)):*
> ```XML
> type commonSubstituteDay = {
>   name: string;
> ```

We probably need the translations here, so we could have this as `LocalizedString`. (Even though this is not at least yet required in the occasional day's)



_Suggestion:_
```XML
  name: LocalizedString;
```
___
*[`ui/src/data/common_substitute_days.ts` line 7 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcartfR0ic7W669M-aY:-NcartfR0ic7W669M-aZ:b-d296u5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/data/common_substitute_days.ts#L7)):*
> ```XML
> };
> 
> export const commonSubstituteDayData: commonSubstituteDay[] = [
> ```

One might also say that this commit adds dead code, so it could probably be squashed with the commit that uses this data. But I'm honestly not sure what is my mind on the matter. It is dead code yes, but also it usually reduces the cognitive load on review.
Although these are not used in the next commit either, so I would _at least_ move this commit just before the commit that uses this data (if not squashed).

What do you think?
___
*[`ui/src/hooks/substitute-operating-periods/useGetSubstituteOperatingPeriod.ts` line 63 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcanAYL-rmCh-EeUFW5:-NcanAYL-rmCh-EeUFW6:b-7mniuk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6628085babe49a326a23db5157871c302bf02978/ui/src/hooks/substitute-operating-periods/useGetSubstituteOperatingPeriod.ts#L63)):*
> ```XML
>   const substituteOperatingPeriods = mapSubstituteOperatingPeriodsResult(data);
> 
>   const [preset, notPreset] = partition(
> ```

This 'partition' seems like a nice util, I need to remember this :+1:

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650)* status: 10 of 21 files reviewed, 17 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcaklzX-byWlZMu8QX9):*

NIT,  layout? Not sure though if it is a change to layout either, if its just a minor padding change :D. But formatting also sounds too strong.
Maybe something like this?
</blockquote></details>

ok
___
*[`-- commits` line 13 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcanvSCE43yEU3tB_6o):*

NIT, I think the hyphen is extra
</blockquote></details>

I accidentally resolved this, but this commit is now squashed
___
*[`-- commits` line 16 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-Ncat5Ev8PMRpPgFwQXo):*

NIT, `better place` is a bit non-informative :D. How about `Move substitute day helper functions under utils`or something similar a bit more descriptive?
</blockquote></details>

under utils from now on
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 34 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbGrib2YFF_OUvpUVo:-Nd9O3-dDkhG72BH2Y7w:b521jil) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L34)):*

Hmm, so we generate dates always at least for 1 year, and then filter the result by the startDate and endDate. Is there some particular benefit with this method, or would it be possible to just generate the dates that are found between the startDate and endDate?

I mean I do understand that it is easier to just generate all commonSubstituteDays at once, then filter, but would it be more complex to just generate by the dates?
</blockquote></details>

This was the first solution and seems like it was also the latest one :laughing: 

After changes it still generates dates all dates but if generated date is not in time range it doesn't push it to return array.

Do you feel that it should not even create date object for comparison?
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 52 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbNM6K9tg2qgbeMg7V:-NdAVVSj5lE9fSsN9XIP:b-r53tpe) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L52)):*

It feels like this function does more stuff than just maps periods to form state, it seems like it also generates the dates and does some replacing logic there as well. So could this be refactored to something bit simpler? Maybe the "generate & replace" stuff could be somewhere else, and this function could just map the results. 
(FYI, wrote this comment after the two comments below, because the more I watched this the more I started to understand)
</blockquote></details>

Check reply below
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 57 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbHqc991IESIGIwKGe:-NdAVcGGDYfJyxM6OskH:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L57)):*

extra `?`, since input is not optional parameter. Also wondering if there is a more descriptive name for `input`, because now we are mapping `input`, and the items inside are named `item`, so the content gets a bit lost there.
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 83 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbJK_K7gg-gc3kNwK4:-NdAVo-a8-8k2ETPB6JP:bqfndjo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L83)):*

Hmm, we already filter the generated preset dates in the generatePresetDates once. What is this filter? I'm cant catch what are these "days from database"? 

Ahhh, wait so this is the combination of dates that are set and dates that are not set?
I would probably rethink the naming of these a little bit to make it more clear.
Now we have presetDates, commonDays, filteredDays and so on, it's a bit confusing.

Don't have the perfect solution straight away, but maybe something to iterate.
How about we generate the presetDates, then we enrich/replace the values with already set substitute days and in the end we have displayedSubstituteDays etc.
This enrich/replace could be its own helper function also.

(After this comment I wrote the comment related above, so maybe we could discuss this further in telco if needed.)
</blockquote></details>

I agree it was a mess. There was so much stuff in the `mapSubstitute...` function... Now that function only maps substitute operating periods to common days. 

Here is a lot of changes, I hope that code will explain it but let's have telco if it's hard to understand or messy or something
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 114 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbPd37Diuo1iWs-0VU:-NcfNvID0oBlgY5xnrZH:bupybby) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L114)):*

NIT, these consts could be maybe rearranged and have some line breaks here and there. Maybe the most valuable thing here would be to at least move this useState declaration to bottom of the list, since it is the least "boiler" here :D But this is really nitpicking.
</blockquote></details>

Added some spacing and moved useState to "bottom"
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 32 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbQjkN867jjhATvtYA:-NdAoJnf0wHnhTJXBuYl:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L32)):*

This naming implies that it returns multiple buttons, but in the end it actually returns only one button. 
So how about `EditRemoveButton` etc?
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 45 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbRy0k7R_L2NMfWMgZ:-NdAqV1Z4MVVqfefIPqB:b-hwpfi2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L45)):*

This outer visible logic with `showAnyButtons` would be better on the parent. So we either render the component, or we don't render the component. Now it is baked inside this component, which reduces the readability by a tiny bit.

Another thing which came to mind is that this is pretty specific button which has a specific logic. You could make this more generic UI element, which takes in just onEdit(), onClose() and some flag determining which button to show, if you want, but this is not necessary, since don't know if this would be used anywhere else.
</blockquote></details>

Moved the outer logic to parent.

I made this more generic not because I think it will be used at some place else. It is just cleaner now
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 75 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbU2wEAosWKPJKgW1m:-NcfNggd5LSU7XmeIgaV:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L75)):*

`mb-0` seems extra
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 95 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbZMc_DqzkubpPCK35:-NdE97uh0UvW8IUDpNKX:bvsozpp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L95)):*

Lots of style definitions here, also precise margins. You could check if this suits better:
Keep these here `grid h-[20%] w-[30%] grid-cols-2`
Change the parent element div like this:
```
<div className\="flex flex-wrap justify-between gap-8">
  {fields.map((field, index) => (
    <CommonSubstitutePeriodItem
      field={field}
      index={index}
      key={field.id}
      update={update}
    />
  ))}
</div>
```
(This might not completely match the layout, but is ofcourse iterable)

Generally it is better to have the kind of "layout" definitions given from outside the component (or with flexbox: on the parent), so that we won't enforce some particular style to a component which could be used elsewhere as well. (another comment below about styles coming from outside the component) 
And the second iteration to that is to use the nice flexbox definitions, which usually function a bit better than precise margins.

What do you think?
</blockquote></details>

Really good points! Made some changes here, removed margins and changed parent div. 

I also refactored child element to use flexbox instead of grid.

I understand what you mean by layout definitions given from outside, but seems like I have hard time to follow it when developing :/
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx` line 53 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbaUCGECDV_WtuRYPq:-NdEAJPR4drZmqpRNoW3:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodSection.tsx#L53)):*

As mentioned above, the minor layout things are generally better to be given from outside the component to avoid enforcing styles. So here I would take the className as parameter which defaults to empty string, and then give the `py-8` from outside this component.
```
interface Props {
  className?: string;
}

export const CommonSubstitutePeriodSection \= ({className \= ''}: Props): JSX.Element \=> {
...
```

and then give the `my-8` as parameter. Aaaand one last thing, doesn't really matter but margin might be the more desirable attribute here :D
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx` line 49 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcamHp29CcfyqpRmHil:-NdEGnNG-jZMNtUIs8tx:b-8jq1hc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6628085babe49a326a23db5157871c302bf02978/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx#L49)):*

Is there presets in OccasionalSubstitutePeriod? Aren't the presets only on the General substitute days? (I am possibly misunderstanding something)

If the form is used for both (Occasional and general) then I would probably rename this to `SubstitutePeriodForm.tsx` (still not sure if this is the case). Maybe I'll find my answer later in this PR :D
</blockquote></details>

There is not presets in OccasionalSubstitutePeriod and that's why it's set to `false`. Occasional and Common substitute periods have their own forms but they use same database table and same Graphql api. Because of that when fetching Substitute periods from database they will become occasional if have isPreset set to `false` and common periods have isPreset set to `true`. That `PeriodType` has `isPreset` property because that way it matches data from database and there is no need to add `isPreset` property when inserting/upserting data using Graphql.
___
*[`ui/src/data/common_substitute_days.ts` line 1 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcapmA76Hy4epTCXDgj:-NdEBE1l7BcHDsPtC53O:bzg4c4w) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L1)):*

What comes to the folder choice, I'm not sure what the files in this `src/data` folder are, seems like some really old mock data, but this file could probably be better in the 'substitute-day-settings' folder where the rest of the components are. 

What do you think?
</blockquote></details>

I couldn't find anything else than `src/data` folder where to put it, but if that is for mock data then maybe it can be under `substitute-day-settings`
___
*[`ui/src/data/common_substitute_days.ts` line 2 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-Ncap1vHBxy0Uum9NXgU:-NdED8ke67P9bgCBqpp4:bvlbbw1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L2)):*

We probably need the translations here, so we could have this as `LocalizedString`. (Even though this is not at least yet required in the occasional day's)
</blockquote></details>

Good point but this name will be used to insert new row to database and used in column `period_name` that column is `text` type. I think changing this to `LocalizedString` will bring more complexity than benefits. 

We can discuss this in telco if needed
___
*[`ui/src/data/common_substitute_days.ts` line 7 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcartfR0ic7W669M-aY:-NdEM5qt-SgM4Qw6Yg2z:b-cr4gx4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/data/common_substitute_days.ts#L7)):*

One might also say that this commit adds dead code, so it could probably be squashed with the commit that uses this data. But I'm honestly not sure what is my mind on the matter. It is dead code yes, but also it usually reduces the cognitive load on review.
Although these are not used in the next commit either, so I would _at least_ move this commit just before the commit that uses this data (if not squashed).

What do you think?
</blockquote></details>

This was in it's own commit because it was easy to separate and I thought it would be cleaner not to have everything in one commit. Now this commit is squashed to that one mega commit.
I think it make no difference if this is in it's own commit or squashed. It's dead code but it's also hard coded data so it's not code and one could argue that hard coded stuff is also bad etc..

HenrikHartiala (changes requested):
> Reviewed 11 of 11 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650)* status: all files reviewed, 4 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 13 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcanvSCE43yEU3tB_6o):*

I accidentally resolved this, but this commit is now squashed
</blockquote></details>

Check :+1: 
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 34 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbGrib2YFF_OUvpUVo:-NdUD8Ma7UQhZ3E-i3Og:b2nbbqc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L34)):*

This was the first solution and seems like it was also the latest one :laughing: 

After changes it still generates dates all dates but if generated date is not in time range it doesn't push it to return array.

Do you feel that it should not even create date object for comparison?
</blockquote></details>

Well, I think it doesn't matter so much that it is worth the effort to refactor this further now, because this change you made is now more clear what happens and more straight forward, so I most definitely think this is good now! Nice :ok_hand: 

Although after writing the comment I gave the code another look, can you just move the const date = ... inside the if? then it would not even create it.



_Suggestion:_
```XML
    commonSubstituteDayData.forEach((day) => {
      if (date >= startDate && date <= endDate) {
        const date = DateTime.fromISO(
        `${year}-${padToTwoDigits(day.month)}-${padToTwoDigits(day.day)}`,
      );
        generatedDates.push({
          name: `${day.name} ${year}`,
          date,
        });
      }
```
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 83 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbJK_K7gg-gc3kNwK4:-NdUFaak5gn6-1CxTf2f:b-qd96cv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L83)):*

I agree it was a mess. There was so much stuff in the `mapSubstitute...` function... Now that function only maps substitute operating periods to common days. 

Here is a lot of changes, I hope that code will explain it but let's have telco if it's hard to understand or messy or something
</blockquote></details>

Ahh, reeally really great work with the refactor! :muscle: This is so so much more clear and neat!
Awesome :star_struck: 
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 114 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbPd37Diuo1iWs-0VU:-NdUG7sm1z6nT5lUJoFs:b-un3dxg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L114)):*

Added some spacing and moved useState to "bottom"
</blockquote></details>

Nice, thanks! :pray: 
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 45 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbRy0k7R_L2NMfWMgZ:-NdUGd8B8D2xsRMG_W3u:bp763ow) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L45)):*

Moved the outer logic to parent.

I made this more generic not because I think it will be used at some place else. It is just cleaner now
</blockquote></details>

Great work with this one as well! And the `EditRemove` -> `EditClose` was a nice enhancement as well, since the "remove" part is more like a side effect to the "close"/"cancel". Now IF we would want to give the user a confirm dialog, it would be easy to implement. 
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 95 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbZMc_DqzkubpPCK35:-NdUK3C49UGGR_X2KRvD:b-6uhnml) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L95)):*

Really good points! Made some changes here, removed margins and changed parent div. 

I also refactored child element to use flexbox instead of grid.

I understand what you mean by layout definitions given from outside, but seems like I have hard time to follow it when developing :/
</blockquote></details>

Yeah it is easy to make the mistake even when knowing the reason behind it. 

Like said in the telco, the `basis-` attributes are new to me and I can't say if there are any downsides to them, but they most definitely are better than precision margins with each element. So nice one! 
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 33 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUHMqZDIamQP1O5HcJ:-NdUHMqZDIamQP1O5HcK:bypry8t) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L33)):*
> ```XML
> 
> const EditCloseButton = ({
>   flag,
> ```

This `flag` is kind of hard to understand what it does without actually reading the code. But for this generic EditCloseButton it is kind of hard to think of one, since something like `showEdit` or `isClosable` or whatever are also maybe a bit awkward. Maybe if you add a minor JSDoc comment on this component which explains that that when the flag is set to true, it shows the close option and on false it shows the edit option. Or something like that.

Minor edit/addition: maybe the "showEdit" could be better? If you check these two choices when using the component:

```
<EditCloseButton
            flag={created}
            onEdit={() => update(index, created)}
            onClose={() => update(index, created)}
          />
vs.
<EditCloseButton
            showEdit={created}
            onEdit={() => update(index, created)}
            onClose={() => update(index, created)}
          />

```
Even with the JSDoc comment, I think the showEdit or something that hinges the user towards what it is might be better, even though the parameter name is still a bit awkward. What do you think?

Btw, this component probably could now live in `uiComponents` in it's own file, since it is nicely already made generic :+1: 
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx` line 49 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcamHp29CcfyqpRmHil:-NdUOkP--t6nUFr5JX-n:bnutve6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6628085babe49a326a23db5157871c302bf02978/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodForm.tsx#L49)):*

There is not presets in OccasionalSubstitutePeriod and that's why it's set to `false`. Occasional and Common substitute periods have their own forms but they use same database table and same Graphql api. Because of that when fetching Substitute periods from database they will become occasional if have isPreset set to `false` and common periods have isPreset set to `true`. That `PeriodType` has `isPreset` property because that way it matches data from database and there is no need to add `isPreset` property when inserting/upserting data using Graphql.
</blockquote></details>

Okay then I think this is good :+1: 
So this is kind of the mapping or default values for the save. The thing I saw at first was that this is "occasional substitute day" and then it has "isPreset" flag, which it doesn't actually need since there can't be an occasional day that has "isPreset" set to true. So my point would be that this `OccasionalSubstitutePeriodForm.tsx` doesn't need to know about the preset stuff. But I can't wrap my head around any solution to suggest and it might not be worth it anyways, so all good.
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodSection.tsx` line 17 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUQDg62UWiNi8pN7K0:-NdUQDg62UWiNi8pN7K1:b-zd7jaj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodSection.tsx#L17)):*
> ```XML
>   findEarliestDate,
>   findLatestDate,
>   mapSubstituteOperatingPeriodsToFormState as mapOccasionalSubstituteOperatingPeriodsToFormState,
> ```

Seems like a VSCode rename leftover (?)
___
*[`ui/src/hooks/substitute-operating-periods/useGetSubstituteOperatingPeriod.ts` line 67 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUQw7q4SuJZUt8VtdB:-NdUQw7q4SuJZUt8VtdC:bcs72zt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/hooks/substitute-operating-periods/useGetSubstituteOperatingPeriod.ts#L67)):*
> ```XML
> 
>   const {
>     data: occasionalSubstituteOperatingPeriodData,
> ```

I really like that these are now separated :+1: :ok_hand: 
___
*[`ui/src/data/common_substitute_days.ts` line 1 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcapmA76Hy4epTCXDgj:-NdTzOjY6IUwHwL6u5UJ:b-i5n9rp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L1)):*

I couldn't find anything else than `src/data` folder where to put it, but if that is for mock data then maybe it can be under `substitute-day-settings`
</blockquote></details>

Thanks! 
Yeah, the reason why I would keep it under the substitute-day-settings folder, is that they are part of this feature only and also only used there. If we remove this feature, we have no more use for this data either.
___
*[`ui/src/data/common_substitute_days.ts` line 2 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-Ncap1vHBxy0Uum9NXgU:-NdU-aiRCRkZyZkm3LXR:b-n2oh4d) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L2)):*

Good point but this name will be used to insert new row to database and used in column `period_name` that column is `text` type. I think changing this to `LocalizedString` will bring more complexity than benefits. 

We can discuss this in telco if needed
</blockquote></details>

Ahh, right, these are saved to database. I just thought that these were used only for the hardcodes on page but yeah of course when the user sets a common substitute day, then it is saved to database.

Yeah no changes needed here. Maybe this comes somewhere in the future if we want to have translations for these. It kind of is the same for occasional subsitute days, but those can't have hardcoded translations so they would need user to input the translation and that doesn't sound fun.
___
*[`ui/src/data/common_substitute_days.ts` line 7 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcartfR0ic7W669M-aY:-NdU21tpE71Rr57rHnyD:b33zwvi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/data/common_substitute_days.ts#L7)):*

This was in it's own commit because it was easy to separate and I thought it would be cleaner not to have everything in one commit. Now this commit is squashed to that one mega commit.
I think it make no difference if this is in it's own commit or squashed. It's dead code but it's also hard coded data so it's not code and one could argue that hard coded stuff is also bad etc..
</blockquote></details>

Yeah I do understand why it was in it's own commit and I agree about that. Although do not agree on this on the hard coded argument, I think it's more like placeholder data than hard coded in the end. But that is different discussion :D *galloping horse*
But now that it is squashed, it's not so bad either. It is very small clear part in it's own file, so I think its fine there as well :)

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650)* status: 19 of 21 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 34 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbGrib2YFF_OUvpUVo:-NdUXH9Y2Ue-E_oErhu8:bu6qthu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L34)):*

Well, I think it doesn't matter so much that it is worth the effort to refactor this further now, because this change you made is now more clear what happens and more straight forward, so I most definitely think this is good now! Nice :ok_hand: 

Although after writing the comment I gave the code another look, can you just move the const date = ... inside the if? then it would not even create it.
</blockquote></details>

Hmm that could be nice, but if we move declaration of `date` inside if statement we can't use it in comparison because it is not declared before usage
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodSection.tsx` line 17 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUQDg62UWiNi8pN7K0:-NdUZr_B6XT0mX4gCN_b:bi1m5ce) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodSection.tsx#L17)):*

Seems like a VSCode rename leftover (?)
</blockquote></details>

Yeah, not first time and not last one... 
___
*[`ui/src/hooks/substitute-operating-periods/useGetSubstituteOperatingPeriod.ts` line 67 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUQw7q4SuJZUt8VtdB:-NdUYIcHAeLnA379fQYf:bqm0lwr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/hooks/substitute-operating-periods/useGetSubstituteOperatingPeriod.ts#L67)):*

I really like that these are now separated :+1: :ok_hand:
</blockquote></details>

You liked them so much that this needs to be blocked? :DD
___
*[`ui/src/data/common_substitute_days.ts` line 2 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-Ncap1vHBxy0Uum9NXgU:-NdUWQSDAo2nb21Kr9m5:b61r6t5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L2)):*

Ahh, right, these are saved to database. I just thought that these were used only for the hardcodes on page but yeah of course when the user sets a common substitute day, then it is saved to database.

Yeah no changes needed here. Maybe this comes somewhere in the future if we want to have translations for these. It kind of is the same for occasional subsitute days, but those can't have hardcoded translations so they would need user to input the translation and that doesn't sound fun.
</blockquote></details>

Yep, if there is need for translations, one possibility is to change `period_name` in database to `i18n` but that's different story

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650)* status: 18 of 22 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 33 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUHMqZDIamQP1O5HcJ:-NdUnEEX8Ac3rfyMKcoP:by4d9t6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L33)):*

This `flag` is kind of hard to understand what it does without actually reading the code. But for this generic EditCloseButton it is kind of hard to think of one, since something like `showEdit` or `isClosable` or whatever are also maybe a bit awkward. Maybe if you add a minor JSDoc comment on this component which explains that that when the flag is set to true, it shows the close option and on false it shows the edit option. Or something like that.

Minor edit/addition: maybe the "showEdit" could be better? If you check these two choices when using the component:

```
<EditCloseButton
            flag={created}
            onEdit={() => update(index, created)}
            onClose={() => update(index, created)}
          />
vs.
<EditCloseButton
            showEdit={created}
            onEdit={() => update(index, created)}
            onClose={() => update(index, created)}
          />

```
Even with the JSDoc comment, I think the showEdit or something that hinges the user towards what it is might be better, even though the parameter name is still a bit awkward. What do you think?

Btw, this component probably could now live in `uiComponents` in it's own file, since it is nicely already made generic :+1:
</blockquote></details>

Moved this to `uiComponents` that's good place for it now.
Changed `flag` to `showEdit`
Flipped `!` inside component to make more sense. Now if `showEdit` is `true` it shows edit button
Also added comment to clarify it more. I think it's ready to :ship:

HenrikHartiala (approved):
Great work! :muscle: 

> Reviewed 2 of 2 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-:-NdUrb_u7UBqQdA9_vEe:bibggeh)* status: all files reviewed, 3 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx` line 34 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NcbGrib2YFF_OUvpUVo:-NdUqWhpB5JBNLlqBgg6:b-mksmup) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d6f06634607c2c48398c3398d9c47ac403bffd7f/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodForm.tsx#L34)):*

Hmm that could be nice, but if we move declaration of `date` inside if statement we can't use it in comparison because it is not declared before usage
</blockquote></details>

haha, wth. I was only half a cup of my morning coffee when I checked that. Oh well.... :D 
___
*[`ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx` line 33 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUHMqZDIamQP1O5HcJ:-NdUrLaY96AG48ehLJYk:b-pagv6f) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/components/timetables/substitute-day-settings/CommonSubstitutePeriod/CommonSubstitutePeriodItem.tsx#L33)):*

Moved this to `uiComponents` that's good place for it now.
Changed `flag` to `showEdit`
Flipped `!` inside component to make more sense. Now if `showEdit` is `true` it shows edit button
Also added comment to clarify it more. I think it's ready to :ship:
</blockquote></details>

Nice, thanks! :shipit: 
___
*[`ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodSection.tsx` line 17 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-NdUQDg62UWiNi8pN7K0:-NdUrWS42B20Qbujb3MY:bun9gfx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bc99f29ff4ca43c43c3d5e8ee5f64cd5220d60c7/ui/src/components/timetables/substitute-day-settings/OccasionalSubstitutePeriod/OccasionalSubstitutePeriodSection.tsx#L17)):*

Yeah, not first time and not last one...
</blockquote></details>

Exactly :D 
___
*[`ui/src/data/common_substitute_days.ts` line 2 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/650#-Ncap1vHBxy0Uum9NXgU:-NdUqQRU2c4Nga0goNjp:bc1fb3y) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/6e493de13f89d18180b3b531fddf7c74c10880a2/ui/src/data/common_substitute_days.ts#L2)):*

Yep, if there is need for translations, one possibility is to change `period_name` in database to `i18n` but that's different story
</blockquote></details>

Yeah :+1:

PR #657, opened 2023-09-06
Add component test for PassingTimesByStopTable

Perholehto:
There is a few functions here that are used only to ease out creating test data. I'm not sure if that is the best way to do it, but without them initializing test data would clutter up the "whole" file.

Data used in tests shows that there is lot of unused data. That should be refactored to only take in as props what it really needs.

Resolves https://github.com/HSLdevcom/jore4/issues/1265

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)

### DIFF
--- a/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx
+++ b/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx
@@ -1,0 +1,303 @@
+import uniqueId from 'lodash/uniqueId';
+import { DateTime, Duration } from 'luxon';
+import {
+  InfrastructureNetworkDirectionEnum,
+  PassingTimeByStopFragment,
+  RouteDirectionEnum,
+  RouteWithJourneyPatternStopsFragment,
+  VehicleJourneyByStopFragment,
+} from '../../../generated/graphql';
+import { useAppDispatch } from '../../../hooks/redux';
+import { setShowArrivalTimesAction } from '../../../redux/slices/timetable';
+import {
+  fireEvent,
+  getAssertedElement,
+  render,
+} from '../../../utils/test-utils';
+import { PassingTimesByStopTable } from './PassingTimesByStopTable';
+
+const vehicleJourney = (vehicleTypeName: string) => ({
+  block: {
+    block_id: '',
+    vehicle_type: {
+      description_i18n: {
+        fi_FI: vehicleTypeName,
+      },
+      vehicle_type_id: '',
+    },
+  },
+  vehicle_journey_id: '',
+});
+
+const createScheduleStopPointInstance = (timingPlaceLabel: string) => ({
+  direction: InfrastructureNetworkDirectionEnum.Forward,
+  label: '',
+  located_on_infrastructure_link_id: '',
+  priority: 0,
+  scheduled_stop_point_id: '',
+  timing_place: {
+    label: timingPlaceLabel,
+    timing_place_id: '',
+  },
+  validity_end: DateTime.fromISO('2051-01-01T00:00:00.000+02:00'),
+  validity_start: DateTime.fromISO('1990-01-01T00:00:00.000+02:00'),
+});
+
+const createTimetabledPassingTime = ({
+  label,
+  passingTime,
+  arrivalTime,
+  departureTime,
+}: {
+  label: string;
+  passingTime: Duration;
+  arrivalTime?: Duration;
+  departureTime?: Duration;
+}): PassingTimeByStopFragment => {
+  return {
+    arrival_time: arrivalTime,
+    departure_time: departureTime,
+    passing_time: passingTime,
+    scheduled_stop_point_in_journey_pattern_ref: {
+      journey_pattern_ref: {
+        journey_pattern_ref_id: '',
+        observation_timestamp: DateTime.fromISO('2022-12-31T22:00:00+00:00'),
+      },
+      scheduled_stop_point_in_journey_pattern_ref_id: '',
+      scheduled_stop_point_instances: [],
+      scheduled_stop_point_label: label,
+    },
+    scheduled_stop_point_in_journey_pattern_ref_id: '',
+    timetabled_passing_time_id: uniqueId(),
+    vehicle_journey: vehicleJourney('Matala telibussi'),
+    vehicle_journey_id: uniqueId(),
+  };
+};
+
+const createTimetabledPassingTimeWithStopPoint = ({
+  label,
+  timingPlaceLabel,
+  passingTime,
+  arrivalTime,
+  departureTime,
+}: {
+  label: string;
+  timingPlaceLabel: string;
+  passingTime: Duration;
+  arrivalTime?: Duration;
+  departureTime?: Duration;
+}) => {
+  const timetabledPassingTime = createTimetabledPassingTime({
+    label,
+    passingTime,
+    arrivalTime,
+    departureTime,
+  });
+
+  const scheduledStopPointInstance =
+    createScheduleStopPointInstance(timingPlaceLabel);
+
+  return {
+    ...timetabledPassingTime,
+    scheduled_stop_point_in_journey_pattern_ref: {
+      ...timetabledPassingTime.scheduled_stop_point_in_journey_pattern_ref,
+      scheduled_stop_point_instances: [scheduledStopPointInstance],
+    },
+  };
+};
+
+describe(`<${PassingTimesByStopTable.name} />`, () => {
+  const vehicleJourneys: VehicleJourneyByStopFragment[] = [
+    {
+      journey_pattern_ref_id: '',
+      timetabled_passing_times: [
+        createTimetabledPassingTimeWithStopPoint({
+          label: 'H2210',
+          timingPlaceLabel: '1ELIEL',
+          passingTime: Duration.fromISO('PT9H37M'),
+          arrivalTime: Duration.fromISO('PT9H37M'),
+        }),
+        createTimetabledPassingTimeWithStopPoint({
+          label: 'H2210',
+          timingPlaceLabel: '1ELIEL',
+          passingTime: Duration.fromISO('PT9H42M'),
+          arrivalTime: Duration.fromISO('PT9H42M'),
+        }),
+        createTimetabledPassingTime({
+          label: 'H221',
+          passingTime: Duration.fromISO('PT13H19M'),
+          departureTime: Duration.fromISO('PT13H19M'),
+        }),
+        createTimetabledPassingTime({
+          label: 'H221',
+          passingTime: Duration.fromISO('PT13H25M'),
+          departureTime: Duration.fromISO('PT13H25M'),
+        }),
+        createTimetabledPassingTime({
+          label: 'H221',
+          passingTime: Duration.fromISO('PT14H55M'),
+          arrivalTime: Duration.fromISO('PT14H58M'),
+          departureTime: Duration.fromISO('PT14H59M'),
+        }),
+        createTimetabledPassingTime({
+          label: 'H221',
+          passingTime: Duration.fromISO('PT14H25M'),
+          departureTime: Duration.fromISO('PT14H25M'),
+        }),
+      ],
+      vehicle_journey_id: '',
+    },
+  ];
+
+  const route: RouteWithJourneyPatternStopsFragment = {
+    name_i18n: {},
+    origin_name_i18n: {},
+    origin_short_name_i18n: {},
+    destination_name_i18n: {},
+    destination_short_name_i18n: {},
+    on_line_id: '',
+    label: '',
+    direction: RouteDirectionEnum.Outbound,
+    route_id: '',
+    priority: 0,
+    route_journey_patterns: [
+      {
+        journey_pattern_id: '',
+        on_route_id: '',
+        scheduled_stop_point_in_journey_patterns: [
+          {
+            is_loading_time_allowed: false,
+            is_regulated_timing_point: false,
+            is_used_as_timing_point: true,
+            is_via_point: false,
+            journey_pattern: {
+              journey_pattern_id: '',
+              on_route_id: '',
+            },
+            journey_pattern_id: '',
+            scheduled_stop_point_label: 'H2210',
+            scheduled_stop_point_sequence: 1,
+            scheduled_stop_points: [],
+            via_point_name_i18n: null,
+            via_point_short_name_i18n: null,
+          },
+        ],
+      },
+    ],
+  };
+
+  const DispatchSetShowArrivalTimesActionWrapper = ({
+    children,
+  }: {
+    children: JSX.Element;
+  }) => {
+    const dispatch = useAppDispatch();
+    dispatch(setShowArrivalTimesAction(true));
+    return children;
+  };
+
+  const selectors = {
+    timeContainer:
+      '[data-testid="PassingTimesByStopTableRowPassingTime::timeContainer"]',
+    hour: '[data-testid="PassingTimesByStopTableRowPassingTime::hour"]',
+    departureTime:
+      '[data-testid="PassingTimesByStopTableRowPassingMinute::departureTime"]',
+    arrivalTime:
+      '[data-testid="PassingTimesByStopTableRowPassingMinute::arrivalTime"]',
+    button: '[data-testid="PassingTimesByStopTableRowPassingMinute::button"]',
+  };
+
+  const shouldHaveCorrectLabelTexts = (
+    row: HTMLElement,
+    textContentToHave: string[],
+  ) => {
+    const column = row.querySelector('td');
+    textContentToHave.forEach((text) => expect(column).toHaveTextContent(text));
+  };
+
+  const shouldHaveCorrectHourText = (timeContainer: Element, hour: string) => {
+    expect(timeContainer.querySelector(selectors.hour)).toHaveTextContent(hour);
+  };
+
+  const shouldHaveCorrectDepartureTimes = (
+    timeContainer: Element,
+    minutes: string[],
+  ) => {
+    const departureTimes = timeContainer.querySelectorAll(
+      selectors.departureTime,
+    );
+    expect(departureTimes).toHaveLength(minutes.length);
+    departureTimes.forEach((departureTime, index) => {
+      expect(departureTime).toHaveTextContent(minutes[index]);
+    });
+  };
+
+  test('should render labels and times correctly', () => {
+    const { getByTestId } = render(
+      <DispatchSetShowArrivalTimesActionWrapper>
+        <PassingTimesByStopTable
+          vehicleJourneys={vehicleJourneys}
+          route={route}
+        />
+      </DispatchSetShowArrivalTimesActionWrapper>,
+    );
+
+    const rowH2210 = getByTestId('PassingTimesByStopTableRow::H2210');
+    shouldHaveCorrectLabelTexts(rowH2210, ['H2210', '1ELIEL']);
+
+    const timeContainerH2210 = getAssertedElement(
+      rowH2210.querySelector(selectors.timeContainer),
+    );
+
+    shouldHaveCorrectHourText(timeContainerH2210, '9');
+    shouldHaveCorrectDepartureTimes(timeContainerH2210, ['37', '42']);
+
+    const rowH221 = getByTestId('PassingTimesByStopTableRow::H221');
+
+    shouldHaveCorrectLabelTexts(rowH221, ['H221', 'Interpoloitu']);
+
+    const timeContainersH221 = rowH221.querySelectorAll(
+      selectors.timeContainer,
+    );
+
+    shouldHaveCorrectHourText(timeContainersH221[0], '13');
+    shouldHaveCorrectDepartureTimes(timeContainersH221[0], ['19', '25']);
+
+    shouldHaveCorrectHourText(timeContainersH221[1], '14');
+    shouldHaveCorrectDepartureTimes(timeContainersH221[1], ['25', '55']);
+
+    const arrivalTimes = timeContainersH221[1].querySelectorAll(
+      selectors.arrivalTime,
+    );
+    expect(arrivalTimes[0]).toHaveClass('invisible');
+
+    expect(arrivalTimes[1]).not.toHaveClass('invisible');
+    expect(arrivalTimes[1]).toHaveTextContent('58');
+  });
+
+  test('should open and close popup correctly', () => {
+    const { getByText, getByTestId, queryByText } = render(
+      <PassingTimesByStopTable
+        vehicleJourneys={vehicleJourneys}
+        route={route}
+      />,
+    );
+    expect(queryByText('Matala telibussi')).toBeNull();
+
+    const passingTimesByStopTableRow = getByTestId(
+      'PassingTimesByStopTableRow::H2210',
+    );
+
+    const button = getAssertedElement(
+      passingTimesByStopTableRow.querySelector(selectors.button),
+    );
+    fireEvent.click(button);
+
+    expect(getByText('Matala telibussi')).toBeVisible();
+
+    const closeButton = getByTestId('Popover::closeButton');
+    fireEvent.click(closeButton);
+
+    expect(queryByText('Matala telibussi')).toBeNull();
+  });
+});
--- a/ui/src/utils/test-utils/misc.ts
+++ b/ui/src/utils/test-utils/misc.ts
@@ -8,3 +8,8 @@ Date.now = jest.fn(() => 1487076708000);
 
 // mock the lodash debounce module to call the function immediately
 jest.mock('lodash/debounce', () => jest.fn((fn) => fn));
+
+export const getAssertedElement = <T>(obj: T | null | undefined): T => {
+  expect(obj).toBeDefined();
+  return obj as T;
+};


Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 159 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Nde9PkS4yLiAAlNVkTN:-Nde9PkS4yLiAAlNVkTO:bffpj1r) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L159)):*
> ```XML
>   };
> 
>   const DispatchSetShowArrivalTimesActionWrapper = ({
> ```

This wrapper is something I couldn't find out better way to do it. 

We have Redux already hooked up with our `customRender` function and it works nicely. But.. here component under test have no functionality to change Redux state. I still wanted to test that when `showArrivalTimes` is `true` arrivalTimes are shown on DOM. 

This was only way I figured out and worked here.

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: all files reviewed, 8 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdjAmV54Q0wBdDCbhON):*
Ah, is links the PR with the URL as well (?)
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 28 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj3W1Q6QyuBzEYO8-n:-Ndj3W1Q6QyuBzEYO8-o:b-dqxcej) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L28)):*
> ```XML
> });
> 
> const createScheduleStopPointInstance = (label: string) => ({
> ```

This was first confusing that we have label = '' but the parameter is not set here. But it seemed there is a minor misunderstanding (comment below). So this is actually the label for timing place, which is different than the stop label.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 32 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdiBre82Kdn3tXvXbp5:-NdiBre82Kdn3tXvXbp6:b-9lexm6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L32)):*
> ```XML
>   label: '',
>   located_on_infrastructure_link_id: '',
>   priority: 10,
> ```

If this should be the standard priority, then I think that it is best to use the enum.



_Suggestion:_
```XML
  priority: Priority.Standard,
```
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 42 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdiBNkf860pPXo7Cgo4:-NdiBNkf860pPXo7Cgo5:b-z0fwhc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L42)):*
> ```XML
> });
> 
> const createTimetabledPassingTime = (
> ```

Im not sure, since not used that much, but did we have some builders already on the test-db-manager? `timetabledPassingTime.ts`. Not even sure if they are usable here :D
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 109 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndi8ZhLAIC7qdzNCcR7:-Ndi8ZhLAIC7qdzNCcR8:b-qidcjk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L109)):*
> ```XML
>   const route: RouteWithJourneyPatternStopsFragment = {
>     name_i18n: {
>       fi_FI: undefined,
> ```

Out of curiosity: Is there a particular reason why these are precisely 'undefined' and not just empty strings like the other attributes are set? Not that it matters very much, since this component parameters or parameter types needs to be changed anyways :D 
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 159 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Nde9PkS4yLiAAlNVkTN:-NdjFb7cAQQY2U5_Zcfd:b-wmr14d) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L159)):*

This wrapper is something I couldn't find out better way to do it. 

We have Redux already hooked up with our `customRender` function and it works nicely. But.. here component under test have no functionality to change Redux state. I still wanted to test that when `showArrivalTimes` is `true` arrivalTimes are shown on DOM. 

This was only way I figured out and worked here.
</blockquote></details>

Did a really quick check, but did not get any better solutions. I think this is a fine solution, maybe a minor comment might be in place, but not necessary it is pretty easy to understand.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 181 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj2ndj0hBhuRz0tG8m:-Ndj2ndj0hBhuRz0tG8n:bm5jpr1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L181)):*
> ```XML
>     expect(getByText('H221')).toBeVisible();
>     // Should show twice because it is used as timing point
>     expect(getAllByText('H2210')).toHaveLength(2);
> ```

First this was confusing, but I think there is a minor misunderstanding here. The label for timing place is not the same as the label for the stop. Timing place labels are something like "1ELIEL" which is kind of a shortened version of "Elielinaukio" etc.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 184 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj0vDZ0hGzhmW0Jawy:-Ndj0vDZ0hGzhmW0Jawz:b-ckmj03) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L184)):*
> ```XML
> 
>     // Hours shown correctly
>     expect(getByText('9')).toBeVisible();
> ```

These currently just checks everywhere for the '9', so it kind of could be in the minutes or in the hours. There should be testIds already for the hours and minutes. Also maybe we could check that they are on the correct rows as well: so we could first get the correct stop row and then within that row check that the hours/minutes are correct.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 192 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdjB2gG6nWLLCEfCqt-:-NdjB2gG6nWLLCEfCqt0:b-w046id) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L192)):*
> ```XML
>     expect(getByText('55')).toBeVisible();
> 
>     expect(getByText('Interpoloitu')).toBeVisible();
> ```

This also just checks if "Interpoloitu" is on the screen, but it does not check if it is on correct stop. So I think you should first get the wanted stop element, and within that element check that it has "interpoloitu" text inside it.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 210 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj3s7z54GUOpNX5jmi:-Ndj3s7z54GUOpNX5jmj:b-iyqspy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L210)):*
> ```XML
> 
>     const button = passingTimesByStopTableRow.querySelectorAll(
>       'td:nth-child(2) > span > span:nth-child(2) > button',
> ```

There should be a testId for this as well.

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: all files reviewed, 8 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdjAmV54Q0wBdDCbhON):*

Ah, is links the PR with the URL as well (?)
</blockquote></details>

I have no clue what you mean :D

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: all files reviewed, 8 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdjAmV54Q0wBdDCbhON):*

I have no clue what you mean :D
</blockquote></details>

Me grubby language. You no understand. Couple of letters were missing :D 

What I meant was that I did not know that the URL to the ticket also works with the linking of PR and ticket. We have used the `HSLdevcom/jore4#<ticketnumber>`. But this might be a better way since it would work also with stuff that doesn't know how to transform the ticketID to an actual URL (github surely does do the linking, but source tree does not and even VSCode adds an extra ascii character to the URL and breaks it).

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdjAmV54Q0wBdDCbhON):*

Me grubby language. You no understand. Couple of letters were missing :D 

What I meant was that I did not know that the URL to the ticket also works with the linking of PR and ticket. We have used the `HSLdevcom/jore4#<ticketnumber>`. But this might be a better way since it would work also with stuff that doesn't know how to transform the ticketID to an actual URL (github surely does do the linking, but source tree does not and even VSCode adds an extra ascii character to the URL and breaks it).
</blockquote></details>

Not sure if there is a downside to use full URL. I think it's simpler just to copypaste URL rather than format `HSLdevcom/jore4#<ticketnumber>`.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 28 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj3W1Q6QyuBzEYO8-n:-Ndo_72R3hEozgralgNk:b-gz41da) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L28)):*

This was first confusing that we have label = '' but the parameter is not set here. But it seemed there is a minor misunderstanding (comment below). So this is actually the label for timing place, which is different than the stop label.
</blockquote></details>

renamed this to `timingPlaceLabel` 
I was just me who was confused by this :D
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 32 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdiBre82Kdn3tXvXbp5:-NdjKhMGDj4lHHMfQ5fn:b-4ihzs2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L32)):*

If this should be the standard priority, then I think that it is best to use the enum.
</blockquote></details>

This was discussed in telco and changed it to `0` to indicate that it's not used.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 42 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdiBNkf860pPXo7Cgo4:-NdjKJBBDlXxFr1QWqrJ:b-noez92) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L42)):*

Im not sure, since not used that much, but did we have some builders already on the test-db-manager? `timetabledPassingTime.ts`. Not even sure if they are usable here :D
</blockquote></details>

How I see them they are used to create input data and here we need output from GraphQL
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 109 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndi8ZhLAIC7qdzNCcR7:-Ndo__Aw3fB3wBhrIFvK:b8mxito) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L109)):*

Out of curiosity: Is there a particular reason why these are precisely 'undefined' and not just empty strings like the other attributes are set? Not that it matters very much, since this component parameters or parameter types needs to be changed anyways :D
</blockquote></details>

Discussed in telco if you can recall, but no real reason and now there is only empty objects left.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 181 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj2ndj0hBhuRz0tG8m:-NdjM2uz7akI3HvYLFzW:bijump7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L181)):*

First this was confusing, but I think there is a minor misunderstanding here. The label for timing place is not the same as the label for the stop. Timing place labels are something like "1ELIEL" which is kind of a shortened version of "Elielinaukio" etc.
</blockquote></details>

Oh, I thought it was kind of a confusing, but so is this whole domain :D I'll use "1ELIEL" for tests now
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 184 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj0vDZ0hGzhmW0Jawy:-Ndo_oCG564Y_Ozx0SDL:b-am1a5q) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L184)):*

These currently just checks everywhere for the '9', so it kind of could be in the minutes or in the hours. There should be testIds already for the hours and minutes. Also maybe we could check that they are on the correct rows as well: so we could first get the correct stop row and then within that row check that the hours/minutes are correct.
</blockquote></details>

I think we also went through this in telco and here is changes that were agreed
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 192 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdjB2gG6nWLLCEfCqt-:-Ndoa2XW4ZhTc6bRRy09:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L192)):*

This also just checks if "Interpoloitu" is on the screen, but it does not check if it is on correct stop. So I think you should first get the wanted stop element, and within that element check that it has "interpoloitu" text inside it.
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 210 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ndj3s7z54GUOpNX5jmi:-Ndoa4bHDee3SFf99kQd:b-upn236) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c01d14d71a44a2e7387f23b26db2fabe4ab414f9/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L210)):*

There should be a testId for this as well.
</blockquote></details>

telco for this one also
___
*[`ui/src/utils/test-utils/misc.ts` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdoaQiq9yeJEAKtab-v:-NdoaQiq9yeJEAKtab-w:beurj1b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/utils/test-utils/misc.ts#L12)):*
> ```XML
> jest.mock('lodash/debounce', () => jest.fn((fn) => fn));
> 
> export const assertDefined = <T>(obj: T | null | undefined): T => {
> ```

This helper is from Internet :incoming_envelope:

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/utils/test-utils/misc.ts` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdoaQiq9yeJEAKtab-v:-Ndoapqc3hG0i1grdnKv:beudr3u) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/utils/test-utils/misc.ts#L12)):*

This helper is from Internet :incoming_envelope:
</blockquote></details>

And why... This is needed when we have optional object.  `querySelector()` returns ` Element | null` 

```ts
const timeContainerH2210 = rowH2210.querySelector(selectors.timeContainer);
expect(timeContainerH2210).toBeDefined();
const timeContainerH2210Element = timeContainerH2210 as Element;
// or use ! operator
```

with helper 
```ts
const timeContainerH2210 = assertDefined(
rowH2210.querySelector(selectors.timeContainer),
);
```

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdjAmV54Q0wBdDCbhON):*

Not sure if there is a downside to use full URL. I think it's simpler just to copypaste URL rather than format `HSLdevcom/jore4#<ticketnumber>`.
</blockquote></details>

Don't know that either. Currently it feels like there is only upsides to it :D and you can still see all the same information as with the different format.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 71 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ne26E1u1CLK6i4XiYwv:-Ne26E1u1CLK6i4XiYww:bhx346l) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L71)):*
> ```XML
>   label: string,
>   timingPlaceLabel: string,
>   passingTime: string,
> ```

passingTime, arrivalTime and departureTime should probably be type of Duration. Also passingTime is actually a computed field. You probably should not need to require it as parameter.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 99 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ne25M8X7ftnBwwB2sx_:-Ne25M8X7ftnBwwB2sxa:b-xfbeao) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L99)):*
> ```XML
>       timetabled_passing_times: [
>         createTimetabledPassingTimeWithStopPoint(
>           'H2210',
> ```

Would make this a lot more clear and readable if you take the parameters as one object. So that the names would be visible here as well. 

Addition. Oh, and even more because you have two optionals in the function as well.

By using the object, you don't need to explicitly say that something is undefined either.



_Suggestion:_
```XML
        createTimetabledPassingTimeWithStopPoint({
          stopPointLabel: 'H2210',
          timingPointLabel: '1ELIEL',
          arrivalTime: 'PT9H37M',
          departureTime: 'PT9H37M',
          
        }),
```
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 110 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ne28MbjBFRN1Ra7LFtr:-Ne28MbjBFRN1Ra7LFts:bnfti1a) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L110)):*
> ```XML
>           'PT9H42M',
>         ),
>         createTimetabledPassingTime('H221', 'PT13H19M', undefined, 'PT13H19M'),
> ```

Same thing here, if the function takes in the parameters as object, would be a neater :+1: 
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 178 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NeCC04A4Ia0l659MVnx:-NeCC04A4Ia0l659MVny:b-3t0vup) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L178)):*
> ```XML
> 
>   test('should render labels and times correctly', () => {
>     const shouldhaveCorrectLabelTexts = (
> ```

Would take these helper functions out of this test for test readability.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 192 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NeCDJo0-X_RWox6lWmR:-NeCDJo0-X_RWox6lWmS:b-ymzsf8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L192)):*
> ```XML
>       hour: string,
>     ) => {
>       expect(timeContainer.querySelector(selectors.hour)).toHaveTextContent(
> ```

Thought about these selectors and was thinking that this might not be the most readable / long lasting solution if we want to continue creating these component tests. Was thinking if there were a way to just kind of make a utilfunction that "queryByTestId" etc, which has the `[data-testId=` -parts there and takes the testId as parameter. But I don't actually have a solution here, since have not done this kind of jest tests in a while. Was more so reflecting this with the cypress commands.
We can also iterate these further later if we find out some new pattern for these, so no changes required here.
___
*[`ui/src/utils/test-utils/misc.ts` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdoaQiq9yeJEAKtab-v:-NeCB9RjBU-JUuAvlaMf:b-tgfvfq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/utils/test-utils/misc.ts#L12)):*

And why... This is needed when we have optional object.  `querySelector()` returns ` Element | null` 

```ts
const timeContainerH2210 = rowH2210.querySelector(selectors.timeContainer);
expect(timeContainerH2210).toBeDefined();
const timeContainerH2210Element = timeContainerH2210 as Element;
// or use ! operator
```

with helper 
```ts
const timeContainerH2210 = assertDefined(
rowH2210.querySelector(selectors.timeContainer),
);
```
</blockquote></details>

I am not completely sure what is the function of this `assertDefined` other than change the "possible null" to not null or fail the test with this expect, but I think this is unnecessary. You could just leave this function out (and make the elements optional in the functions) and then the test will fail in the next expect-clauses if it does not exist.

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 71 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ne26E1u1CLK6i4XiYwv:-NeCIb7YDIXX2v4O5xg7:bb6x5nn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L71)):*

passingTime, arrivalTime and departureTime should probably be type of Duration. Also passingTime is actually a computed field. You probably should not need to require it as parameter.
</blockquote></details>

Hmm, not sure about that `passingTime` is not required here.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 99 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ne25M8X7ftnBwwB2sx_:-NeDArGp0tKHJW_TsqS0:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L99)):*

Would make this a lot more clear and readable if you take the parameters as one object. So that the names would be visible here as well. 

Addition. Oh, and even more because you have two optionals in the function as well.

By using the object, you don't need to explicitly say that something is undefined either.
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 110 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-Ne28MbjBFRN1Ra7LFtr:-NeDArt84QPXaQWZBJj7:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L110)):*

Same thing here, if the function takes in the parameters as object, would be a neater :+1:
</blockquote></details>

Done.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 178 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NeCC04A4Ia0l659MVnx:-NeDAsaBDrEPpBAF4pWS:bnmcv4j) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L178)):*

Would take these helper functions out of this test for test readability.
</blockquote></details>

I removed the helper functions, oh wait! Nope, moved them out from inside the test.
___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 192 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NeCDJo0-X_RWox6lWmR:-NeDB05N5bG_GaeyPVnL:b-2h6ke8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L192)):*

Thought about these selectors and was thinking that this might not be the most readable / long lasting solution if we want to continue creating these component tests. Was thinking if there were a way to just kind of make a utilfunction that "queryByTestId" etc, which has the `[data-testId=` -parts there and takes the testId as parameter. But I don't actually have a solution here, since have not done this kind of jest tests in a while. Was more so reflecting this with the cypress commands.
We can also iterate these further later if we find out some new pattern for these, so no changes required here.
</blockquote></details>

No changes here, discussion should continue some other time
___
*[`ui/src/utils/test-utils/misc.ts` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdoaQiq9yeJEAKtab-v:-NeDBHoY5S0w-wQo1I-c:bsdvfdc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/utils/test-utils/misc.ts#L12)):*

I am not completely sure what is the function of this `assertDefined` other than change the "possible null" to not null or fail the test with this expect, but I think this is unnecessary. You could just leave this function out (and make the elements optional in the functions) and then the test will fail in the next expect-clauses if it does not exist.
</blockquote></details>

And now it's `getAssertedElement`

HenrikHartiala (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657)* status: all files reviewed, 1 unresolved discussion (waiting on @Perholehto)

___
*[`ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx` line 178 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NeCC04A4Ia0l659MVnx:-NeDGjtk91DzdIsp8XXi:b-14zo32) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/components/timetables/passing-times-by-stop/PassingTimesByStopTable.spec.tsx#L178)):*

I removed the helper functions, oh wait! Nope, moved them out from inside the test.
</blockquote></details>

:D :ok_man: Thanks!
___
*[`ui/src/utils/test-utils/misc.ts` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/657#-NdoaQiq9yeJEAKtab-v:-NeDH-5YEUSoyd5VOR6E:b4putsz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b64670714706289de5c2089da5c87529a76b85d1/ui/src/utils/test-utils/misc.ts#L12)):*

And now it's `getAssertedElement`
</blockquote></details>

:ok_hand:

PR #662, opened 2023-09-12
Try experimentalModifyObstructiveThirdPartyCode in Cypress config

sr-vala:
Test if this setting affects test stability. Details for the setting: https://docs.cypress.io/guides/guides/web-security#Modifying-Obstructive-Third-Party-Code

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/662)


sr-vala:
This is false by default and probably should stay that way.

PR #667, opened 2023-09-18
Add tests for occasional substitute operating periods

sr-vala:
Resolves HSLdevcom/jore4#1434

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)

### DIFF
--- a/cypress/e2e/occasionalSubstituteDays.cy.ts
+++ b/cypress/e2e/occasionalSubstituteDays.cy.ts
@@ -1,0 +1,626 @@
+import {
+  GetInfrastructureLinksByExternalIdsResult,
+  InfraLinkAlongRouteInsertInput,
+  JourneyPatternInsertInput,
+  LineInsertInput,
+  RouteDirectionEnum,
+  RouteInsertInput,
+  RouteTypeOfLineEnum,
+  StopInJourneyPatternInsertInput,
+  StopInsertInput,
+  buildLine,
+  buildRoute,
+  buildStop,
+  buildStopInJourneyPattern,
+  buildTimingPlace,
+  extractInfrastructureLinkIdsFromResponse,
+  mapToGetInfrastructureLinksByExternalIdsQuery,
+} from '@hsl/jore4-test-db-manager';
+import { defaultDayTypeIds } from '@hsl/timetables-data-inserter';
+import { DateTime, Duration } from 'luxon';
+import { Tag } from '../enums';
+import {
+  RouteTimetablesSection,
+  SubstituteDaySettingsPage,
+  Toast,
+  VehicleScheduleDetailsPage,
+  VehicleServiceTable,
+} from '../pageObjects';
+import { UUID } from '../types';
+import {
+  SupportedResources,
+  insertToDbHelper,
+  removeFromDbHelper,
+} from '../utils';
+
+// These infralink IDs exist in the 'infraLinks.sql' test data file.
+// These form a straight line on Eerikinkatu in Helsinki.
+// Coordinates are partial since they are needed only for the stop creation.
+
+const testInfraLinks = [
+  {
+    externalId: '445156',
+    coordinates: [24.925682785, 60.163824160000004, 7.3515],
+  },
+  {
+    externalId: '442331',
+    coordinates: [24.927565858, 60.1644843305, 9.778500000000001],
+  },
+  {
+    externalId: '442424',
+    coordinates: [24.929825718, 60.165285984, 9.957],
+  },
+  {
+    externalId: '442325',
+    coordinates: [24.93312261043133, 60.16645636069328, 13.390046659939703],
+  },
+];
+
+const stopLabels = ['H1231', 'H1232', 'H1233', 'H1234'];
+
+const lines: LineInsertInput[] = [
+  {
+    ...buildLine({ label: '99' }),
+    line_id: '08d1fa6b-440c-421e-ad4d-0778d65afe60',
+    type_of_line: RouteTypeOfLineEnum.StoppingBusService,
+  },
+];
+
+const timingPlaces = [
+  buildTimingPlace('ad07b4bd-1d38-4ed5-a0b6-451887f3deb2', '1AACKT'),
+  buildTimingPlace('38e9a1d5-d92f-4379-90c6-0c748f3db9a7', '1ELIMK'),
+  buildTimingPlace('19f429b5-ea6c-4747-a048-2a064d2b6d6b', '1AURLA'),
+];
+
+const buildStopsOnInfrastrucureLinks = (
+  infrastructureLinkIds: UUID[],
+): StopInsertInput[] => [
+  {
+    ...buildStop({
+      label: stopLabels[0],
+      located_on_infrastructure_link_id: infrastructureLinkIds[0],
+    }),
+    scheduled_stop_point_id: '52dfb94f-d37f-47d2-ae28-a06f2208fe13',
+    timing_place_id: timingPlaces[0].timing_place_id,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[0].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: stopLabels[1],
+      located_on_infrastructure_link_id: infrastructureLinkIds[1],
+    }),
+    scheduled_stop_point_id: 'c6bac741-4c95-4d78-afa4-f43928ae3ff4',
+    timing_place_id: null,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[1].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: stopLabels[2],
+      located_on_infrastructure_link_id: infrastructureLinkIds[2],
+    }),
+    scheduled_stop_point_id: '7f0cee71-d847-42e3-9bff-f68207634d59',
+    timing_place_id: timingPlaces[1].timing_place_id,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[2].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: stopLabels[3],
+      located_on_infrastructure_link_id: infrastructureLinkIds[3],
+    }),
+    scheduled_stop_point_id: 'a4580ffd-6f58-49fc-a4ae-9d6f7cdfedc4',
+    timing_place_id: timingPlaces[2].timing_place_id,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[3].coordinates,
+    },
+  },
+];
+
+const routes: RouteInsertInput[] = [
+  {
+    ...buildRoute({ label: '99' }),
+    route_id: 'a87d68ff-4e72-4a25-b8db-67606a72a962',
+    on_line_id: lines[0].line_id,
+    validity_start: DateTime.fromISO('2023-01-01T13:08:43.315+03:00'),
+    validity_end: DateTime.fromISO('2043-06-30T13:08:43.315+03:00'),
+    direction: RouteDirectionEnum.Inbound,
+  },
+];
+
+const buildInfraLinksAlongRoute = (
+  infrastructureLinkIds: UUID[],
+): InfraLinkAlongRouteInsertInput[] => [
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[0],
+    infrastructure_link_sequence: 0,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[1],
+    infrastructure_link_sequence: 1,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[2],
+    infrastructure_link_sequence: 2,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[3],
+    infrastructure_link_sequence: 3,
+    is_traversal_forwards: true,
+  },
+];
+
+const journeyPatterns: JourneyPatternInsertInput[] = [
+  {
+    journey_pattern_id: 'c9167102-8815-419e-92e3-2d249000bae5',
+    on_route_id: routes[0].route_id,
+  },
+];
+
+const stopsInJourneyPattern: StopInJourneyPatternInsertInput[] = [
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[0],
+    scheduledStopPointSequence: 0,
+    isUsedAsTimingPoint: true,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[1],
+    scheduledStopPointSequence: 1,
+    isUsedAsTimingPoint: false,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[2],
+    scheduledStopPointSequence: 2,
+    isUsedAsTimingPoint: true,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[3],
+    scheduledStopPointSequence: 3,
+    isUsedAsTimingPoint: true,
+  }),
+];
+
+const timetableDataInput = {
+  _journey_pattern_refs: {
+    route99inbound: {
+      route_label: '99',
+      route_direction: 'inbound',
+      journey_pattern_id: journeyPatterns[0].journey_pattern_id,
+      _stop_points: [
+        {
+          scheduled_stop_point_sequence: 1,
+          scheduled_stop_point_label: 'H1231',
+        },
+        {
+          scheduled_stop_point_sequence: 2,
+          scheduled_stop_point_label: 'H1232',
+        },
+        {
+          scheduled_stop_point_sequence: 3,
+          scheduled_stop_point_label: 'H1233',
+        },
+        {
+          scheduled_stop_point_sequence: 4,
+          scheduled_stop_point_label: 'H1234',
+        },
+      ],
+    },
+  },
+  _vehicle_schedule_frames: {
+    spring2023: {
+      validity_start: DateTime.fromISO('2025-01-01'),
+      validity_end: DateTime.fromISO('2025-12-31'),
+      name: 'Kevät 2025',
+      booking_label: 'Spring booking label',
+      _vehicle_services: {
+        saturday: {
+          day_type_id: defaultDayTypeIds.SATURDAY,
+          _blocks: {
+            block: {
+              _vehicle_journeys: {
+                route99Inbound1: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT7H05M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT7H12M'),
+                      departure_time: Duration.fromISO('PT7H13M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT7H19M'),
+                      departure_time: Duration.fromISO('PT7H23M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT7H27M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+                route99Inbound2: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT8H20M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT8H22M'),
+                      departure_time: Duration.fromISO('PT8H23M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT8H26M'),
+                      departure_time: Duration.fromISO('PT8H27M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT8H29M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+              },
+            },
+          },
+        },
+        sunday: {
+          day_type_id: defaultDayTypeIds.SUNDAY,
+          _blocks: {
+            block: {
+              _vehicle_journeys: {
+                route99Inbound1: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT11H05M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT11H12M'),
+                      departure_time: Duration.fromISO('PT11H13M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT11H19M'),
+                      departure_time: Duration.fromISO('PT11H23M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT11H27M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+                route99Inbound2: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT12H20M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT12H22M'),
+                      departure_time: Duration.fromISO('PT12H23M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT12H26M'),
+                      departure_time: Duration.fromISO('PT12H27M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT12H29M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+                route99Inbound3: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT13H05M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT13H12M'),
+                      departure_time: Duration.fromISO('PT13H13M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT13H19M'),
+                      departure_time: Duration.fromISO('PT13H23M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT13H27M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+  },
+};
+
+const route99InboundTimetableSection = new RouteTimetablesSection(
+  '99',
+  'inbound',
+);
+
+const route99InboundSaturdayVehicleService = new VehicleServiceTable(
+  route99InboundTimetableSection,
+  'LA',
+);
+
+const route99InboundSundayVehicleService = new VehicleServiceTable(
+  route99InboundTimetableSection,
+  'SU',
+);
+
+describe('Occasional substitute operating periods', () => {
+  let substituteDaySettingsPage: SubstituteDaySettingsPage;
+  let toast: Toast;
+  let vehicleScheduleDetailsPage: VehicleScheduleDetailsPage;
+
+  const baseDbResources = {
+    lines,
+    routes,
+    journeyPatterns,
+    stopsInJourneyPattern,
+  };
+  let dbResources: SupportedResources;
+
+  before(() => {
+    cy.task<GetInfrastructureLinksByExternalIdsResult>(
+      'hasuraAPI',
+      mapToGetInfrastructureLinksByExternalIdsQuery(
+        testInfraLinks.map((infralink) => infralink.externalId),
+      ),
+    ).then((res) => {
+      const infraLinkIds = extractInfrastructureLinkIdsFromResponse(res);
+      const stops = buildStopsOnInfrastrucureLinks(infraLinkIds);
+      const infraLinksAlongRoute = buildInfraLinksAlongRoute(infraLinkIds);
+      dbResources = {
+        ...baseDbResources,
+        timingPlaces,
+        stops,
+        infraLinksAlongRoute,
+      };
+    });
+  });
+
+  beforeEach(() => {
+    removeFromDbHelper(dbResources);
+    insertToDbHelper(dbResources);
+    cy.task('truncateTimetablesDatabase');
+
+    substituteDaySettingsPage = new SubstituteDaySettingsPage();
+    toast = new Toast();
+    vehicleScheduleDetailsPage = new VehicleScheduleDetailsPage();
+
+    cy.setupMapTiles();
+    cy.mockLogin();
+
+    cy.task('insertHslTimetablesDatasetToDb', timetableDataInput);
+
+    cy.visit('/timetables/settings');
+  });
+
+  afterEach(() => {
+    removeFromDbHelper(dbResources);
+    cy.task('truncateTimetablesDatabase');
+  });
+
+  it(
+    'Should create and delete an occasional substitute operating period successfully',
+    { tags: [Tag.Timetables, Tag.Smoke] },
+    () => {
+      // Add an occasional substitute day
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm
+        .getAddOccasionalSubstitutePeriodButton()
+        .click();
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm.fillNthOccasionalSubstitutePeriodForm(
+        {
+          nth: 0,
+          formInfo: {
+            name: 'Poikkeusjakson nimi',
+            beginDate: '2025-03-08',
+            beginTime: '04:30',
+            endDate: '2025-03-08',
+            endTime: '28:30',
+            substituteDay: 'Sunnuntai',
+            lineTypes: ['Peruslinja'],
+          },
+        },
+      );
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm
+        .getSaveButton()
+        .click();
+      cy.wait('@gqlCreateSubstituteOperatingPeriod')
+        .its('response.statusCode')
+        .should('equal', 200);
+      toast.checkWarningToastHasMessage('Suodatusjaksoa muutettu');
+
+      // Check that the Saturday schedule is the same as the Sunday schedule
+      cy.visit(
+        `timetables/lines/${lines[0].line_id}?observationDate=2025-03-08&routeLabels=${routes[0].label}`,
+      );
+      vehicleScheduleDetailsPage.getShowAllValidSwitch().click();
+      route99InboundSaturdayVehicleService
+        .get()
+        .should('contain', '8.3.2025 - 8.3.2025')
+        .and('contain', '3 lähtöä')
+        .and('contain', '11:05 ... 13:05');
+      route99InboundSundayVehicleService
+        .get()
+        .should('contain', '1.1.2025 - 31.12.2025')
+        .and('contain', '3 lähtöä')
+        .and('contain', '11:05 ... 13:05');
+
+      // Remove the substitute day
+      cy.visit('/timetables/settings');
+      substituteDaySettingsPage.observationPeriodForm.setStartDate(
+        '2025-03-08',
+      );
+      substituteDaySettingsPage.observationPeriodForm.setEndDate('2025-03-08');
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm.deleteNthOccasionalSubstituteDay(
+        0,
+      );
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm
+        .getSaveButton()
+        .click();
+
+      // Check that the Saturday schedule is not the same as the Sunday schedule
+      cy.visit(
+        `timetables/lines/${lines[0].line_id}?observationDate=2025-03-08&routeLabels=${routes[0].label}`,
+      );
+      vehicleScheduleDetailsPage.getShowAllValidSwitch().click();
+      route99InboundSaturdayVehicleService
+        .get()
+        .should('contain', '1.1.2025 - 31.12.2025')
+        .and('contain', '2 lähtöä')
+        .and('contain', '07:05 ... 08:20');
+      route99InboundSundayVehicleService
+        .get()
+        .should('contain', '1.1.2025 - 31.12.2025')
+        .and('contain', '3 lähtöä')
+        .and('contain', '11:05 ... 13:05');
+    },
+  );
+
+  it(
+    "Should create a 'No traffic' day successfully",
+    { tags: [Tag.Timetables] },
+    () => {
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm
+        .getAddOccasionalSubstitutePeriodButton()
+        .click();
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm.fillNthOccasionalSubstitutePeriodForm(
+        {
+          nth: 0,
+          formInfo: {
+            name: 'Ei liikennöintiä -päivä',
+            beginDate: '2025-01-18',
+            beginTime: '04:30',
+            endDate: '2025-01-18',
+            endTime: '28:30',
+            substituteDay: 'Ei liikennöintiä',
+            lineTypes: ['Peruslinja'],
+          },
+        },
+      );
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm
+        .getSaveButton()
+        .click();
+      cy.wait('@gqlCreateSubstituteOperatingPeriod')
+        .its('response.statusCode')
+        .should('equal', 200);
+
+      // Check the timetable on the date when there should be no operation
+      cy.visit(
+        `timetables/lines/${lines[0].line_id}?observationDate=2025-01-18&routeLabels=${routes[0].label}`,
+      );
+      route99InboundSaturdayVehicleService
+        .get()
+        .should('contain', 'Ei liikennöintiä');
+      // Check that next Saturday's timetable remains unaffected
+      cy.visit(
+        `timetables/lines/${lines[0].line_id}?observationDate=2025-01-25&routeLabels=${routes[0].label}`,
+      );
+      route99InboundSaturdayVehicleService
+        .get()
+        .should('contain', '1.1.2025 - 31.12.2025')
+        .and('contain', '2 lähtöä')
+        .and('contain', '07:05 ... 08:20');
+    },
+  );
+
+  it(
+    'Should create a substitute period based on a partial day successfully',
+    { tags: [Tag.Timetables] },
+    () => {
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm
+        .getAddOccasionalSubstitutePeriodButton()
+        .click();
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm.fillNthOccasionalSubstitutePeriodForm(
+        {
+          nth: 0,
+          formInfo: {
+            name: 'nimi',
+            beginDate: '2025-03-15',
+            beginTime: '12:00',
+            endDate: '2025-03-15',
+            endTime: '13:00',
+            substituteDay: 'Sunnuntai',
+            lineTypes: ['Peruslinja'],
+          },
+        },
+      );
+      substituteDaySettingsPage.occasionalSubstitutePeriodForm
+        .getSaveButton()
+        .click();
+      cy.wait('@gqlCreateSubstituteOperatingPeriod')
+        .its('response.statusCode')
+        .should('equal', 200);
+
+      // Check that the setting was applied to the selected date
+      cy.visit(
+        `timetables/lines/${lines[0].line_id}?observationDate=2025-03-15&routeLabels=${routes[0].label}`,
+      );
+      route99InboundSaturdayVehicleService
+        .get()
+        .should('contain', '15.3.2025 - 15.3.2025')
+        .and('contain', '1 lähtöä')
+        .and('contain', '12:20 ... 12:20');
+      // Check that next Saturday's timetable remains unaffected
+      cy.visit(
+        `timetables/lines/${lines[0].line_id}?observationDate=2025-03-22&routeLabels=${routes[0].label}`,
+      );
+      route99InboundSaturdayVehicleService
+        .get()
+        .should('contain', '1.1.2025 - 31.12.2025')
+        .and('contain', '2 lähtöä')
+        .and('contain', '07:05 ... 08:20');
+    },
+  );
+
+  // TODO: 'Should not let the user create an occasional reference day that overlaps a static reference day'
+});
--- a/cypress/e2e/timetablePassingTimes.cy.ts
+++ b/cypress/e2e/timetablePassingTimes.cy.ts
@@ -270,6 +270,7 @@ describe('Timetable import and export', () => {
     // Import a timetable for the exported route
     navbar.getTimetablesLink().click();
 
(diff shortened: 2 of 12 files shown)


HenrikHartiala (changes requested):
> Reviewed 6 of 6 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: all files reviewed, 10 unresolved discussions (waiting on @sr-vala)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewBz4ZDfoaLnjP7Yqh):*
This is a great separate commit! :+1: 

Would rephrase this commit title a bit. It now implies to changes to hastus import. I read this and thought that "wait, that should already be possible?" But then realised reading the code that it's about page objects.
Maybe something like "Add support to e2e tests for multiple hastus import files" or something like that.
___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewWGSAAnD7IOnv2dvx):*
This body does not provide any additional information to the commit title. But you could add things like:
Add and rename testIds to OccasionalSubstitutePeriodRow.tsx
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 247 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewJf2Y3GNZ8w4NAqm_:-NewJf2Y3GNZ8w4NAqma:bs60krj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L247)):*
> ```XML
>   });
> 
>   it(
> ```

You could add a couple more empty lines between some of the commands to make it a bit more readable. Now there is none after 
`      cy.visit('/timetables');`
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 268 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewX8GqE0P7u9j8bdpf:-NewX8GqE0P7u9j8bdpg:b-prdxfv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L268)):*
> ```XML
>       cy.visit('/timetables');
>       timetablesMainPage.getSettingsButton().click();
>       // Clicking the '+' button does not always work without some waiting
> ```

This (and the similar waits) should be investigated further rather than implement "some waiting"
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 413 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewZ5-u2loHCI4tMvZv:-NewZ5-u2loHCI4tMvZw:b-x0oqog) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L413)):*
> ```XML
>           beginTime: '10:00',
>           endDate: '2025-03-15',
>           endTime: '12:00',
> ```

This currently strips only from the beginning, we could probably use the same test for stripping also from the end. But I can also see that the imported file only has 3 departures, so that can't be tested with this base data. Will be easier to implement with the timetables-data-inserter.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 18 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewMkYlBcEArTvRb6m2:-NewMkYlBcEArTvRb6m3:bp3a44d) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L18)):*
> ```XML
>   openOccasionalSubstitutePeriodForm() {
>     this.getAddOccasionalSubstitutePeriodButton().click();
>     // Form sometimes closes/disappears so we have to check if it is visible
> ```

This really needs to be investigated further. This kind of hack for "some random reason behaviour" is not very long lasting fix.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 41 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewO7l92TgJ5sg6abHU:-NewO7l92TgJ5sg6abHV:bx5cy39) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L41)):*
> ```XML
> 
>   getOccasionalSubstitutePeriodFormPeriodName() {
>     return cy.getByTestId('OccasionalSubstitutePeriodRow::periodName');
> ```

There can be multiple of these fields visible at the same time on screen (If there is already an occasional substitute period present in the time range, or if you press the add-icon multiple times to add multiple at the same time.
Currently not sure what would be the best way to handle this. You probably need to get all the rows, use them by index and then search the fields inside those elements. 
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 68 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewPOCGE5QZU4omivi0:-NewPOCGE5QZU4omivi1:b-dhu68g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L68)):*
> ```XML
>   selectDayType(dayType: string) {
>     this.getOccasionalSubstitutePeriodFormDayTypeDropdown().click();
>     cy.get('li').contains(dayType).click();
> ```

cy.get(..) searches the whole screen, you should narrow this down to search only inside the dropdown.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 77 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewQOt85W0jjlY3cqJs:-NewQOt85W0jjlY3cqJt:b-3ay2ie) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L77)):*
> ```XML
>   selectLineType(lineType: string) {
>     this.getOccasionalSubstitutePeriodFormLineTypesDropdown().click();
>     // Sometimes checkboxes are all checked and sometimes none are checked.
> ```

This random behaviour should be investigated, and definitely not implemented to tests. Probably @Perholehto might know which should be the default (all checked or none checked by default)
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 84 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewTrQz28TK3qkP234O:-NewTrQz28TK3qkP234P:bi9y4kj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L84)):*
> ```XML
>       }
>     });
>     cy.get('li').contains(lineType).click();
> ```

This searches for the whole screen, you should narrow this down as well.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 93 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewMEE01-QVoNpZmXLy:-NewMEE01-QVoNpZmXLz:bivt1s0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L93)):*
> ```XML
>     // // The form sometimes closes for some reason unless wait is used before filling it
>     // eslint-disable-next-line cypress/no-unnecessary-waiting
>     cy.wait(500);
> ```

This really should be investigated further: what is the reason behind this. The "cy.wait(500)" is not a stable or good fix, because if it is random, then this 500ms wait might not be enough either.
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-NewUtQl6ZSYsjX7PUK1:b-9p0e2e) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*
> ```XML
>   }
> 
>   getWarningToast() {
> ```

FYI, if there are multiple toasts visible on screen, all these will just get the first. This is something that might not be a problem, since we are probably not spamming toasts in these tests, but as an example: if you make a test that results in a toast. Then you pretty soon(within the 5 seconds or whatever the timeout for toasts is) you perform another action that results also in a similar toast, and you assert that the toast is visible, you might get a green light from the first toast that has not yet disappeared from the screen.

So just to keep this in mind :+1:

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 6 of 17 files reviewed, 10 unresolved discussions (waiting on @HenrikHartiala and @Perholehto)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewBz4ZDfoaLnjP7Yqh):*

This is a great separate commit! :+1: 

Would rephrase this commit title a bit. It now implies to changes to hastus import. I read this and thought that "wait, that should already be possible?" But then realised reading the code that it's about page objects.
Maybe something like "Add support to e2e tests for multiple hastus import files" or something like that.
</blockquote></details>

Done.
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 247 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewJf2Y3GNZ8w4NAqm_:-Nf9nN3a-DpXD62TEqWj:bumhy9i) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L247)):*

You could add a couple more empty lines between some of the commands to make it a bit more readable. Now there is none after 
`      cy.visit('/timetables');`
</blockquote></details>

Added some comments and line breaks to split up the tests into sections.
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 268 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewX8GqE0P7u9j8bdpf:-Nf9nWrb3AiapMNbWmrW:bfnojoi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L268)):*

This (and the similar waits) should be investigated further rather than implement "some waiting"
</blockquote></details>

This behavior cannot be replicated manually. Maybe a component test should be created?
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 413 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewZ5-u2loHCI4tMvZv:-Nf9ojLI98CLbx5Q-1DG:b-429iil) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L413)):*

This currently strips only from the beginning, we could probably use the same test for stripping also from the end. But I can also see that the imported file only has 3 departures, so that can't be tested with this base data. Will be easier to implement with the timetables-data-inserter.
</blockquote></details>

Added a TODO for this.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 18 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewMkYlBcEArTvRb6m2:-Nf9ooJbALiPheX6GLyO:bfnojoi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L18)):*

This really needs to be investigated further. This kind of hack for "some random reason behaviour" is not very long lasting fix.
</blockquote></details>

This behavior cannot be replicated manually. Maybe a component test should be created?
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 41 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewO7l92TgJ5sg6abHU:-Nf9oqeb5LwqCVxe4hTO:bxrsqtw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L41)):*

There can be multiple of these fields visible at the same time on screen (If there is already an occasional substitute period present in the time range, or if you press the add-icon multiple times to add multiple at the same time.
Currently not sure what would be the best way to handle this. You probably need to get all the rows, use them by index and then search the fields inside those elements.
</blockquote></details>

We create these fields in the tests so at all times we know and control how many of these fields are visible, so using indexes should not be an issue.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 68 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewPOCGE5QZU4omivi0:-Nf9mF1C8BRHAnVagZmg:b-b7qcek) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L68)):*

cy.get(..) searches the whole screen, you should narrow this down to search only inside the dropdown.
</blockquote></details>

The dropdown list does not have a test-id and I did not find a way to add it.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 77 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewQOt85W0jjlY3cqJs:-Nf9mTmr0fEcdx0WIhCg:bfnojoi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L77)):*

This random behaviour should be investigated, and definitely not implemented to tests. Probably @Perholehto might know which should be the default (all checked or none checked by default)
</blockquote></details>

This behavior cannot be replicated manually. Maybe a component test should be created?
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 84 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewTrQz28TK3qkP234O:-Nf9miUF3TIFi97tUQc2:bfmefmk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L84)):*

This searches for the whole screen, you should narrow this down as well.
</blockquote></details>

A test-id for the list would help with this.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 93 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewMEE01-QVoNpZmXLy:-Nf9mpvJ65hVI6ebq0uk:bsbckyq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L93)):*

This really should be investigated further: what is the reason behind this. The "cy.wait(500)" is not a stable or good fix, because if it is random, then this 500ms wait might not be enough either.
</blockquote></details>

This behavior cannot be replicated manually, so the issue is only when using the site via Cypress. Maybe a component test should be created.
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-Nf9pCsI4YwbabOukQrS:bqzqvkr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

FYI, if there are multiple toasts visible on screen, all these will just get the first. This is something that might not be a problem, since we are probably not spamming toasts in these tests, but as an example: if you make a test that results in a toast. Then you pretty soon(within the 5 seconds or whatever the timeout for toasts is) you perform another action that results also in a similar toast, and you assert that the toast is visible, you might get a green light from the first toast that has not yet disappeared from the screen.

So just to keep this in mind :+1:
</blockquote></details>

Testing the toast visibility has been very unstable in general for some reason. I have not found a way to check for multiple toasts in a row.

HenrikHartiala (changes requested):
> Reviewed 1 of 15 files at r7, 15 of 15 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: all files reviewed, 7 unresolved discussions (waiting on @Perholehto and @sr-vala)

___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 247 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewJf2Y3GNZ8w4NAqm_:-Nhzd-aWE9HZZGNxUBdD:b-979das) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L247)):*

Added some comments and line breaks to split up the tests into sections.
</blockquote></details>

This thread jumped out from the code for some reason, So I'll open a new one.
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 413 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewZ5-u2loHCI4tMvZv:-Nhzh_lk6MEG6zbKtb01:b89z570) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L413)):*

Added a TODO for this.
</blockquote></details>

I didn't see a TODO for this, but now I also realised that this is probably already implemented (I think?). And thirdly my initial comment about "can't be tested with this base data" was wrong. I think I confused departures and just a route's stops :D. So yeah 3 departures should be enough to strip from start and end, which results in 1 departure, and if I checked the code correctly, this was already done. (thread jumped out of code).
So all good :+1: 
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 456 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NhzdBFbEGotzB-Z403e:-NhzdBFcDFB4Mfz9drWU:bozeuce) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/39d80c66fdc71dffdd995b5e91efafd2ed39f7c8/cypress/e2e/occasionalSubstituteDays.cy.ts#L456)):*
> ```XML
>     { tags: [Tag.Timetables, Tag.Smoke] },
>     () => {
>       // Clicking the '+' button does not always work without some waiting
> ```

Now that I had another look on this, my theory is that if the clicking happens just at the same time these are loaded, then cypress might miss with the click when the button changes place as a result of simultaneous rendering of the occasional days. At least I think it's something like this. So this probably is _again_ something that can be properly fixed with a loading indicator. 

These waits are OK until we get the loading indicators, but I think the comment should be clarified a bit, because now it feels like a random behaviour and random fix. So something like "We need to wait the queries, because sometimes the UI jumps after the queries are done and cypress tries to click at the same time missing the button" or something like this if it makes sense. 
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 41 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewO7l92TgJ5sg6abHU:-Ni-8B1G5WARswo-J-ns:b1x2k19) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L41)):*

We create these fields in the tests so at all times we know and control how many of these fields are visible, so using indexes should not be an issue.
</blockquote></details>

Yeah, the "getAll" and index should be enough :+1: 
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 93 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewMEE01-QVoNpZmXLy:-Ni-BLLb1QQWIXBZ5pQ1:b-pcxdfb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L93)):*

This behavior cannot be replicated manually, so the issue is only when using the site via Cypress. Maybe a component test should be created.
</blockquote></details>

Hmm, I don't see how a component test would fix this issue if this is only cypress related.
I'll try to investigate why this is so flaky.
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-Ni-6bYOAOI6M87JOtSu:b7y8ew) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

Testing the toast visibility has been very unstable in general for some reason. I have not found a way to check for multiple toasts in a row.
</blockquote></details>

Hmm, yeah. We need to think some solution. First that comes in to mind is that in the tests we get all toasts found from screen and check if there is more than one visible and in that case use array index.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: all files reviewed, 7 unresolved discussions (waiting on @HenrikHartiala and @Perholehto)

___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 413 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewZ5-u2loHCI4tMvZv:-Ni31zHW4UK-kzlw3a_Y:b-drvt3w) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/e2e/occasionalSubstituteDays.cy.ts#L413)):*

I didn't see a TODO for this, but now I also realised that this is probably already implemented (I think?). And thirdly my initial comment about "can't be tested with this base data" was wrong. I think I confused departures and just a route's stops :D. So yeah 3 departures should be enough to strip from start and end, which results in 1 departure, and if I checked the code correctly, this was already done. (thread jumped out of code).
So all good :+1:
</blockquote></details>

Yeah I changed this after the timetables data inserter was available in the UI repo.
___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 456 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NhzdBFbEGotzB-Z403e:-Ni32_b7C2B6efiGtfSp:b-n31qmw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/39d80c66fdc71dffdd995b5e91efafd2ed39f7c8/cypress/e2e/occasionalSubstituteDays.cy.ts#L456)):*

Now that I had another look on this, my theory is that if the clicking happens just at the same time these are loaded, then cypress might miss with the click when the button changes place as a result of simultaneous rendering of the occasional days. At least I think it's something like this. So this probably is _again_ something that can be properly fixed with a loading indicator. 

These waits are OK until we get the loading indicators, but I think the comment should be clarified a bit, because now it feels like a random behaviour and random fix. So something like "We need to wait the queries, because sometimes the UI jumps after the queries are done and cypress tries to click at the same time missing the button" or something like this if it makes sense.
</blockquote></details>

Cypress has a lot of built-in checks for element "actionability", so I would be surprised if this was caused by something like what you described.
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-Ni3DdSm13jpLwFeZhlh:by4d7se) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

Hmm, yeah. We need to think some solution. First that comes in to mind is that in the tests we get all toasts found from screen and check if there is more than one visible and in that case use array index.
</blockquote></details>

Now that I think about this, I'm not really sure what the actual issue would be in your first message. The method is just a getter method for the danger toast element, and in the tests we are checking that a danger toast with a specific message is shown with checkDangerToastHasMessage. There is an issue with checking multiple toasts on the screen simultaneously, and by that I mean that Cypress does not even find/see some toasts, so indexes won't help in that case.

HenrikHartiala (commented):
> Reviewed 1 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 17 of 18 files reviewed, 6 unresolved discussions (waiting on @Perholehto and @sr-vala)

___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-Nihq43I9RAF2x3x_KOy:b-283mnb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

Now that I think about this, I'm not really sure what the actual issue would be in your first message. The method is just a getter method for the danger toast element, and in the tests we are checking that a danger toast with a specific message is shown with checkDangerToastHasMessage. There is an issue with checking multiple toasts on the screen simultaneously, and by that I mean that Cypress does not even find/see some toasts, so indexes won't help in that case.
</blockquote></details>

The first message issue is that if you do let's say two actions consecutively and both are supposed to be successful actions and we check for successful toast after each action. So after the second action we have two toasts on screen. Now let's say that the second action fails resulting in a failed toast. Now if in the test we check after both action if we can find successful toast using this "cy.get", they BOTH will find a successful toast, even though the second one failed and has a fail toast. So basically the issue is that these work reliably only when we have one toast on screen at a time.

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 2 of 18 files reviewed, 7 unresolved discussions (waiting on @Perholehto and @sr-vala)

___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 59 at r11](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NinhG6e-5nmuf6cE71s:-NinhG6e-5nmuf6cE71t:b9f5k29) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/eab35c153c9c91776231cd46a7a7f18c89cbcb03/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L59)):*
> ```XML
>   }
> 
>   selectLineType(lineType: string) {
> ```

Now that I see this function without the workaround, I also noticed that this function is named "selectLineType", but it actually opens the dropdown, clicks "kaikki" and then clicks the linetype. 

The component is a multiselect component, and you can have multiple selected at the same time. Also the "kaikki" toggle will have different effect depending on what the initial state is. So now if you use this function to "select Peruslinja" but the selcetor already has something else selected, this will first click "kaikki" and then the "lineType" given as a parameter, which probably will have a different result than what the user of this function expects.

For the `selectDayType` above it works fine, because it is a singleSelect, and the result does not depend on what the initial state is. But it is different with this component, so you need to tweak this function a bit, or remove it and have the specific logic in each test separately.

Also, this function does not work if using with english locales.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 2 of 18 files reviewed, 7 unresolved discussions (waiting on @HenrikHartiala and @Perholehto)

___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 41 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewO7l92TgJ5sg6abHU:-NirjPwJCNZS7a6--OdY:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L41)):*

Yeah, the "getAll" and index should be enough :+1:
</blockquote></details>

Done.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 93 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewMEE01-QVoNpZmXLy:-NirjOLtDIRy4YFHTP_2:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L93)):*

Hmm, I don't see how a component test would fix this issue if this is only cypress related.
I'll try to investigate why this is so flaky.
</blockquote></details>

Done.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 59 at r11](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NinhG6e-5nmuf6cE71s:-NirjTDrDvl8-kfXmD4u:bxvlv78) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/eab35c153c9c91776231cd46a7a7f18c89cbcb03/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L59)):*

Now that I see this function without the workaround, I also noticed that this function is named "selectLineType", but it actually opens the dropdown, clicks "kaikki" and then clicks the linetype. 

The component is a multiselect component, and you can have multiple selected at the same time. Also the "kaikki" toggle will have different effect depending on what the initial state is. So now if you use this function to "select Peruslinja" but the selcetor already has something else selected, this will first click "kaikki" and then the "lineType" given as a parameter, which probably will have a different result than what the user of this function expects.

For the `selectDayType` above it works fine, because it is a singleSelect, and the result does not depend on what the initial state is. But it is different with this component, so you need to tweak this function a bit, or remove it and have the specific logic in each test separately.

Also, this function does not work if using with english locales.
</blockquote></details>

Changed the behavior so that all checked checkboxes are unchecked before the list item is clicked.
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-Nirkhqy6UCmtkx0Wktw:b-5dcyzr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

The first message issue is that if you do let's say two actions consecutively and both are supposed to be successful actions and we check for successful toast after each action. So after the second action we have two toasts on screen. Now let's say that the second action fails resulting in a failed toast. Now if in the test we check after both action if we can find successful toast using this "cy.get", they BOTH will find a successful toast, even though the second one failed and has a fail toast. So basically the issue is that these work reliably only when we have one toast on screen at a time.
</blockquote></details>

Well, we could check in all such cases that a danger toast does not exist. But I guess then we should also check that a warning toast does not exist.

HenrikHartiala (changes requested):
> Reviewed 4 of 16 files at r11, 2 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 8 of 19 files reviewed, 2 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 460 at r13](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-Nit7zOl81CE6DurJJgK:-Nit7zOl81CE6DurJJgL:b-ys9mav) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/01613c78c034d5f0f80b15981e9dc48aa0d890e7/cypress/e2e/occasionalSubstituteDays.cy.ts#L460)):*
> ```XML
>         .getAddOccasionalSubstitutePeriodButton()
>         .click();
>       substituteDaySettingsPage.occasionalSubstitutePeriodForm.fillOccasionalSubstitutePeriodForm(
> ```

I think it would be better if you here get the nth value and within that you fill it. Or does it work like that?
Can you just:
```
substituteDaySettingsPage.occasionalSubstitutePeriodForm.getOccasionalSubstitutePeriodRow()
      .eq(0)
      .within(() => {
      substituteDaySettingsPage.occasionalSubstitutePeriodForm.fillOccasionalSubstitutePeriodForm(()
}
```
Does the scope narrow down for the fill-function call?

If this is not possible, then I would at least give the parameters as object, so that it is clear what the parameters here are.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 59 at r11](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NinhG6e-5nmuf6cE71s:-NitJPKqDXI1uLhrYbek:bfc4mfo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/eab35c153c9c91776231cd46a7a7f18c89cbcb03/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L59)):*

Changed the behavior so that all checked checkboxes are unchecked before the list item is clicked.
</blockquote></details>

If you really want to keep this within this automated "fillOccasionalSubstitutePeriodForm and not have the logic of simply clicking the boxes that needs to be clicked in the actual test, then this is fine. 

I still strongly recommend that we would handle those separately in each test, because this feels clumsy: "I want to have everything else checked but 'U-linja'". This will first uncheck everything, then check everything back but the U-linja, instead of just clicking U-linja once.

But like I said, if this has to be in the automated "fillOccasionalSubstitutePeriodForm" as it is, then so be it. At least the result of this should always be what the user wants.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 120 at r13](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NitAK5fDV-emZokX5M2:-NitAK5fDV-emZokX5M3:befyjj1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/01613c78c034d5f0f80b15981e9dc48aa0d890e7/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L120)):*
> ```XML
>   }
> 
>   deleteOccasionalSubstituteDay(nth: number) {
> ```

I'd include the `nth` in the function name.



_Suggestion:_
```XML
  deleteNthOccasionalSubstituteDay(nth: number) {
```
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-Nit1JElBtBjqo5zuq--:bqp42f3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

Well, we could check in all such cases that a danger toast does not exist. But I guess then we should also check that a warning toast does not exist.
</blockquote></details>

But then we would not know if we got a second successful toast visible. What if the second action is completely different? The test fails, even though the toast message is visible.
Again example:
We do an action that results in success toast which has the message "Foobar".
-> We check that we have success toast visible with the message Foobar - FOUND
We do another (different) action that results in success toast which has the message "Punz".
-> We check that we have success toast visible with the message "Punz" - Now here we have a possibility that it does not find this, even though it is visible on the screen, because we use cy.get() which returns the first one it finds. So if this finds the "Foobar" message, it will do the assertion for message: "Punz" and fail.

If we need to check that an action results in success, it's very clumsy to make the assertion so that "Well, at least nothing resulted in failed toast, so we probably can assume everything went fine". To make sure that everything works correctly, we need to check that all actions and behaviours related to that action happens. We can't make assumptions based on things that should not at least happen in tests.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 8 of 19 files reviewed, 2 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/occasionalSubstituteDays.cy.ts` line 460 at r13](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-Nit7zOl81CE6DurJJgK:-NitN11jCAgbkSPx43c-:b-hgyk3b) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/01613c78c034d5f0f80b15981e9dc48aa0d890e7/cypress/e2e/occasionalSubstituteDays.cy.ts#L460)):*

I think it would be better if you here get the nth value and within that you fill it. Or does it work like that?
Can you just:
```
substituteDaySettingsPage.occasionalSubstitutePeriodForm.getOccasionalSubstitutePeriodRow()
      .eq(0)
      .within(() => {
      substituteDaySettingsPage.occasionalSubstitutePeriodForm.fillOccasionalSubstitutePeriodForm(()
}
```
Does the scope narrow down for the fill-function call?

If this is not possible, then I would at least give the parameters as object, so that it is clear what the parameters here are.
</blockquote></details>

Ah pff.t Maybe not move the nth choosing logic here, because then it would mean that the user of the fill-function would need to know this particular thing. So maybe keep em like they are and maybe just make the parameters as object and include the "nth" in the function name or something. Idk.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 7 of 19 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 59 at r11](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NinhG6e-5nmuf6cE71s:-Nj6Rr8wEcXmkzTEMa0W:b-gnkxbb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/eab35c153c9c91776231cd46a7a7f18c89cbcb03/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L59)):*

If you really want to keep this within this automated "fillOccasionalSubstitutePeriodForm and not have the logic of simply clicking the boxes that needs to be clicked in the actual test, then this is fine. 

I still strongly recommend that we would handle those separately in each test, because this feels clumsy: "I want to have everything else checked but 'U-linja'". This will first uncheck everything, then check everything back but the U-linja, instead of just clicking U-linja once.

But like I said, if this has to be in the automated "fillOccasionalSubstitutePeriodForm" as it is, then so be it. At least the result of this should always be what the user wants.
</blockquote></details>

I don't really get what you're trying to say, and it's getting a bit painful trying to read someone else's mind and trying shift the implementation accordingly. Feel free to iterate on the method later if you have ideas.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 120 at r13](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NitAK5fDV-emZokX5M2:-Nj6R2vz3w7V_H-vGZXY:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/01613c78c034d5f0f80b15981e9dc48aa0d890e7/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L120)):*

I'd include the `nth` in the function name.
</blockquote></details>

Done.
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-Nj6QmBUCqq7umCntIGj:b-7i7ju6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

But then we would not know if we got a second successful toast visible. What if the second action is completely different? The test fails, even though the toast message is visible.
Again example:
We do an action that results in success toast which has the message "Foobar".
-> We check that we have success toast visible with the message Foobar - FOUND
We do another (different) action that results in success toast which has the message "Punz".
-> We check that we have success toast visible with the message "Punz" - Now here we have a possibility that it does not find this, even though it is visible on the screen, because we use cy.get() which returns the first one it finds. So if this finds the "Foobar" message, it will do the assertion for message: "Punz" and fail.

If we need to check that an action results in success, it's very clumsy to make the assertion so that "Well, at least nothing resulted in failed toast, so we probably can assume everything went fine". To make sure that everything works correctly, we need to check that all actions and behaviours related to that action happens. We can't make assumptions based on things that should not at least happen in tests.
</blockquote></details>

I don't think toasts are the most important way to check for any resulting state of a test in any case. Better not to focus on them too much.

HenrikHartiala (changes requested):
> Reviewed 8 of 16 files at r11, 2 of 5 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: all files reviewed, 4 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/timetableImport.cy.ts` line 401 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCxR0q-ke_KsgXy6nr:-NjCxR0rE7nKKAj1N3rP:ba7dy2w) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableImport.cy.ts#L401)):*
> ```XML
>       const routesAndLinesPage = new RoutesAndLinesPage();
> 
>       const IMPORT_FILENAME = 'hastusImportSaturday.exp';
> ```

Please use different commits or AT LEAST explain these changes in the commit body if you are making changes that are unrelated to the actual commit.
___
*[`cypress/e2e/timetablePassingTimes.cy.ts` line 274 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzF_M4bUpQGLdBgxk:-NjCzF_M4bUpQGLdBgxl:ba7dy2w) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetablePassingTimes.cy.ts#L274)):*
> ```XML
> 
>     // TODO: Change timetable importing to proper test data generation when it is available
>     const IMPORT_FILENAME = 'hastusImportSaturday.exp';
> ```

Please use different commits or AT LEAST explain these changes in the commit body if you are making changes that are unrelated to the actual commit.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 273 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzXQLBvqt6ni3dDdj:-NjCzXQLBvqt6ni3dDdk:ba7dy2w) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableValidityPeriod.cy.ts#L273)):*
> ```XML
>     navbar.getTimetablesLink().click();
> 
>     const IMPORT_FILENAME = 'hastusImportSaturday.exp';
> ```

Please use different commits or AT LEAST explain these changes in the commit body if you are making changes that are unrelated to the actual commit.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 275 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzY0M24ykQA1E4DlA:-NjCzY0M24ykQA1E4DlB:bm2fnj8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableValidityPeriod.cy.ts#L275)):*
> ```XML
>     const IMPORT_FILENAME = 'hastusImportSaturday.exp';
> 
>     cy.setupMapTiles();
> ```

These seem to be leftover from something else, seems completely unrelated to any changes on this pull request.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 59 at r11](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NinhG6e-5nmuf6cE71s:-NjCtWXhByE7WjSXbaKh:beo2xks) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/eab35c153c9c91776231cd46a7a7f18c89cbcb03/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L59)):*

I don't really get what you're trying to say, and it's getting a bit painful trying to read someone else's mind and trying shift the implementation accordingly. Feel free to iterate on the method later if you have ideas.
</blockquote></details>

Well, I tried to ask you a bunch of questions related to exactly this in slack, so that I would understand what you are initially trying to do and then give suggestions based on that. But since I only got code changes but no follow-up questions, no explanations or no answers to any questions it is hard to continue this.
___
*[`cypress/pageObjects/Toast.ts` line 14 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NewUtQl6ZSYsjX7PUK0:-NjCqz2g3zg3o4X5xjWK:b-sk1nwm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/14e7301a45a78789b93a823fd9590242dd3727bd/cypress/pageObjects/Toast.ts#L14)):*

I don't think toasts are the most important way to check for any resulting state of a test in any case. Better not to focus on them too much.
</blockquote></details>

It seems that you are missing the point, because that is a completely different topic. The point is that if we want to test that everything works as they are intended (toasts included), then we need a way to actually check the toasts as well. What is the point of adding a toast check if it does not check the correct toast?

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: all files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/timetablePassingTimes.cy.ts` line 274 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzF_M4bUpQGLdBgxk:-NjGHM1iDo-CYv7bw-h3:b-hf5749) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetablePassingTimes.cy.ts#L274)):*

Please use different commits or AT LEAST explain these changes in the commit body if you are making changes that are unrelated to the actual commit.
</blockquote></details>

The test data was originally inserted with the 'hastusImportSunday.exp' file, so I probably renamed 'hastusImport.exp' to include the day type in that file's name also. That doesn't matter now that the timetable data is inserted with timetables-data-inserter.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 273 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzXQLBvqt6ni3dDdj:-NjGIBZk5fxqSmY0VVfL:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableValidityPeriod.cy.ts#L273)):*

Please use different commits or AT LEAST explain these changes in the commit body if you are making changes that are unrelated to the actual commit.
</blockquote></details>

Done.
___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 275 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzY0M24ykQA1E4DlA:-NjGIC0b3IHbSKzhkErE:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableValidityPeriod.cy.ts#L275)):*

These seem to be leftover from something else, seems completely unrelated to any changes on this pull request.
</blockquote></details>

Done.
___
*[`cypress/pageObjects/OccasionalSubstitutePeriodForm.ts` line 59 at r11](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NinhG6e-5nmuf6cE71s:-NjGIH6M-sEIb9QcNHZX:b-wf7ked) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/eab35c153c9c91776231cd46a7a7f18c89cbcb03/cypress/pageObjects/OccasionalSubstitutePeriodForm.ts#L59)):*

Well, I tried to ask you a bunch of questions related to exactly this in slack, so that I would understand what you are initially trying to do and then give suggestions based on that. But since I only got code changes but no follow-up questions, no explanations or no answers to any questions it is hard to continue this.
</blockquote></details>

'fillNthOccasionalSubstitutePeriodForm' should work when filling a new form and also when editing an existing form. I have not understood from your comments what exactly you think is wrong here.

HenrikHartiala (changes requested):
> Reviewed 2 of 6 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 15 of 19 files reviewed, 2 unresolved discussions (waiting on @sr-vala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 275 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzY0M24ykQA1E4DlA:-NjGkT231HzL7d9GA_B6:b-8l8bbi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableValidityPeriod.cy.ts#L275)):*

Done.
</blockquote></details>

These is still here.

sr-vala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 15 of 19 files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/timetableValidityPeriod.cy.ts` line 275 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCzY0M24ykQA1E4DlA:-NjGssKZ5Efw1o33NXtM:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableValidityPeriod.cy.ts#L275)):*

These is still here.
</blockquote></details>

Done.
___
*[`cypress/e2e/timetableImport.cy.ts` line 401 at r14](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667#-NjCxR0q-ke_KsgXy6nr:-NjGt0ilDg7WDL7OzCa_:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9eb3064e752e8166fa5a5c4b862e7749e928dad5/cypress/e2e/timetableImport.cy.ts#L401)):*

Please use different commits or AT LEAST explain these changes in the commit body if you are making changes that are unrelated to the actual commit.
</blockquote></details>

Done.

HenrikHartiala (approved):
> Reviewed 1 of 6 files at r15.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/667)* status: 16 of 19 files reviewed, 1 unresolved discussion (waiting on @sr-vala)

PR #668, opened 2023-09-19
Revert back to Cypress 12.17.3

sr-vala:
Revert back to Cypress 12 because of problems in 13

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/668)


sr-vala:
This is probably not needed.

PR #671, opened 2023-09-21
Update pinned hasura image

HenrikHartiala:
As explained in the committed comment, we removed the FK constraints from return_value.timetable_version because we do not want to make it part of the data model, since it is only a dummy table for function return value. But because of this, the graphql-codegen now types it as nullable, even though it can't ever be null in our case. To fix this, added an if-statement with an explanatory comment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/671)


Leitsi (approved):
:lgtm:

> Reviewed 3 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/671#-:-NerBMNk7AjQkzQW_weB:bnfp4nl)* status: 3 of 5 files reviewed, all discussions resolved

PR #695, opened 2023-10-31
Add info why timetable row is not clickable

Perholehto:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/695)

### DIFF
--- a/ui/src/components/common/RouteLineTableRow.tsx
+++ b/ui/src/components/common/RouteLineTableRow.tsx
@@ -8,7 +8,7 @@ import { isRoute } from '../../graphql';
 import { useAlertsAndHighLights } from '../../hooks';
 import { Column, Row, Visible } from '../../layoutComponents';
 import { Path, routeDetails } from '../../router/routeDetails';
-import { mapToShortDate, MAX_DATE, MIN_DATE } from '../../time';
+import { MAX_DATE, MIN_DATE, mapToShortDate } from '../../time';
 import { LocatorButton } from '../../uiComponents';
 import { LineDetailsButton } from './LineDetailsButton';
 import { LineTimetablesButton } from './LineTimetablesButton';
@@ -141,7 +141,7 @@ export const RouteLineTableRow = ({
         >
           <Row className="items-center">
             <Column className="w-1/2 font-bold">
-              <Row>
+              <Row className="items-center">
                 <h2>
                   {isRoute(rowItem) ? (
                     <RouteLabel
@@ -153,6 +153,9 @@ export const RouteLineTableRow = ({
                   )}
                 </h2>
                 {displayInformation.rowIcon}
+                {rowVariant === RouteLineTableRowVariant.Timetables &&
+                  !hasTimetables &&
+                  t('timetables.noTimetables')}
               </Row>
               <p>{rowItem.name_i18n.fi_FI}</p>
             </Column>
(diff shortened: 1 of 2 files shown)


HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/695)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/common/RouteLineTableRow.tsx` line 45 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/695#-Ni4Pcvy2u1dPfWahEEG:-Ni4Pcvy2u1dPfWahEEH:bbjehoo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/75424d17c96269a591c316b676e3ae94cf9c561e/ui/src/components/common/RouteLineTableRow.tsx#L45)):*
> ```XML
> const yBorderClassnames = 'border-y border-y-light-grey';
> 
> const getDisplayInformation = (
> ```

This function really calls for a refactor (probably initially made by me). But something doesn't feel right if you need to pass in a translation function as a parameter. I think that is a cue to componentize some stuff. I really am not sure what is the best solution here and  feels like this would need a bit more thorough change. 

We could take a telco to think of some ideas?
___
*[`ui/src/components/common/RouteLineTableRow.tsx` line 64 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/695#-Ni4QPeiA7_gLKAgVuJx:-Ni4QPeiA7_gLKAgVuJy:b-uu2rfc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/75424d17c96269a591c316b676e3ae94cf9c561e/ui/src/components/common/RouteLineTableRow.tsx#L64)):*
> ```XML
>               }`}
>             />
>             {!hasTimetables && t('timetables.noTimetables')}
> ```

This is now under "rowIcon" and I think the conditional text should not be part of "icon"

HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/695)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/common/RouteLineTableRow.tsx` line 45 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/695#-Ni4Pcvy2u1dPfWahEEG:-Ni4TUbKBvdHg9i1V-5c:b-g66oyc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/75424d17c96269a591c316b676e3ae94cf9c561e/ui/src/components/common/RouteLineTableRow.tsx#L45)):*

This function really calls for a refactor (probably initially made by me). But something doesn't feel right if you need to pass in a translation function as a parameter. I think that is a cue to componentize some stuff. I really am not sure what is the best solution here and  feels like this would need a bit more thorough change. 

We could take a telco to think of some ideas?
</blockquote></details>

As said on slack, you shouldn't run this through getDiusplayInformation() -function and just add it in the parent

HenrikHartiala (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/695#-:-Ni4a0WZ5oE58DCUXWiJ:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @Perholehto)

PR #703, opened 2023-11-06
Revert and refactor rewrites

HenrikHartiala:
- Revert removing the rewrites function as it actually handles the initial routing. The warning seems to be irrelevant for us. Not sure how to suppress it though.
- Refactor to make the structure of index.tsx and _app.tsx towards how they are intended to be.
- Relocate and rename Main for consistency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/703)

### DIFF
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -1,5 +1,14 @@
 module.exports = {
   compress: false, // We have server rendering disabled so can't use compression.
+  async rewrites() {
+    return [
+      {
+        // Rewrite everything to `pages/index`
+        source: '/:any*',
+        destination: '/',
+      },
+    ];
+  },
   webpack: (config) => {
     // eslint-disable-next-line no-param-reassign
     config.resolve.alias = {
--- a/ui/src/components/main/MainPage.tsx
+++ b/ui/src/components/main/MainPage.tsx
@@ -1,15 +1,15 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { LOGIN_URL } from '../api/user';
-import { useAppSelector } from '../hooks';
-import { selectUser } from '../redux';
-import { SimpleButton } from '../uiComponents';
+import { LOGIN_URL } from '../../api/user';
+import { useAppSelector } from '../../hooks';
+import { selectUser } from '../../redux';
+import { SimpleButton } from '../../uiComponents';
 
 const testIds = {
   main: 'Main',
 };
 
-export const Main: React.FC = () => {
+export const MainPage: React.FC = () => {
   const { userInfo } = useAppSelector(selectUser);
   const { t } = useTranslation();
 
@@ -49,4 +49,4 @@ export const Main: React.FC = () => {
 };
 
 // eslint-disable-next-line import/no-default-export
-export default Main;
+export default MainPage;
--- a/ui/src/pages/_app.tsx
+++ b/ui/src/pages/_app.tsx
@@ -1,14 +1,9 @@
 import 'maplibre-gl/dist/maplibre-gl.css';
 import type { AppProps } from 'next/app';
 import { useEffect, useState } from 'react';
-import { UserProvider } from '../auth/UserProvider';
 import '../generated/fontello/css/hsl-icons.css';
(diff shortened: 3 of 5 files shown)


Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/703)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`ui/src/pages/index.tsx` line 17 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/703#-Ni_7IIRC6QGGdSVE0M5:-Ni_7IIRC6QGGdSVE0M6:b-mrix68) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d44cb48d11f93d4f4883d38392d9d6c4471fc817/ui/src/pages/index.tsx#L17)):*
> ```XML
>         <link rel="icon" href="/favicon.ico" />
>       </Head>
>       <ApolloProvider>
> ```

I'm not too sure about this change overall. What's the motivation here, will this make adding some new functionality easier?

Some thoughts (take with a grain of salt since I'm not familiar with "architectures" in React apps):
- the previous version made more sense to me in some way: `_app` had all the boilerplate and this `index` had only the actual HTML. Now they're mixed.
- if left like this, do we really need to split the`_app` and `_index`? The `_app` feels a bit empty (which I guess is fine as long as it has a clear responsibility, but: see previous point)
- overall, this `pages` folder feels a bit funny when all the actual Pages are stored inside `components` folder. Makes me think if it would make sense to have all the page components here instead. Though that's not this PRs issue necessarily, has existed before.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/703)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`ui/src/pages/index.tsx` line 17 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/703#-Ni_7IIRC6QGGdSVE0M5:-NidEm392SUzpXo6FSjY:b-ju32jg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d44cb48d11f93d4f4883d38392d9d6c4471fc817/ui/src/pages/index.tsx#L17)):*

I'm not too sure about this change overall. What's the motivation here, will this make adding some new functionality easier?

Some thoughts (take with a grain of salt since I'm not familiar with "architectures" in React apps):
- the previous version made more sense to me in some way: `_app` had all the boilerplate and this `index` had only the actual HTML. Now they're mixed.
- if left like this, do we really need to split the`_app` and `_index`? The `_app` feels a bit empty (which I guess is fine as long as it has a clear responsibility, but: see previous point)
- overall, this `pages` folder feels a bit funny when all the actual Pages are stored inside `components` folder. Makes me think if it would make sense to have all the page components here instead. Though that's not this PRs issue necessarily, has existed before.
</blockquote></details>

Like explained in slack, some of this is result of having next, but not using it's routing or the pages folder further than the `index.tsx`, which probably has to be there, or it would cause more configuring / workarounds. 

And in my opinion the previous version had displayed stuff in two different files, because the title and footer were on `index.tsx`, but everything else were in `_app.tsx`. So when I removed the rewrites -function, we lost stuff that were on `index.tsx`, but nothing else, so I did not spot that we had footer and title missing from the pages (after refresh).
Now the _app.tsx just has it's necessary codes for starting up the site, but `index.tsx` has the actual structure there in one place.

So if we would use the `pages` folder like it's meant to be used, it also do the routing for us. for examples `pages/foo/bar.tsx` would result in a route `/foo/bar` (https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts)

When I saw this pages - components split when I joined the project, I asked that why are we not using the pages, but can't remember what the answer was. 

But yeah, I actually wanted to leave this refactoring change out from this PR, because it is not relevant with the revert, but Jarno asked that I would include these changes also in the PR, so I did.

And lastly, we do have some other strange structures still remaining in the UI regarding components, but thats for another time.

PR #724, opened 2023-11-17
Add functionality to delete timetable

Perholehto:
Closes hsldevcom/jore4#1578

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)

### DIFF
--- a/ui/src/components/timetables/common/AffectedRouteLabels.tsx
+++ b/ui/src/components/timetables/common/AffectedRouteLabels.tsx
@@ -1,0 +1,19 @@
+interface Props {
+  affectedRouteLabels: string[];
+  text: string;
+}
+
+export const AffectedRouteLabels = ({
+  affectedRouteLabels,
+  text,
+}: Props): JSX.Element => {
+  return (
+    <div className="my-6 flex flex-row items-center space-x-6">
+      <i className="icon-alert text-hsl-red" />
+      <div className="space-y-1 text-sm">
+        <p className="font-bold">{text}</p>
+        <p>{affectedRouteLabels.join(', ')}</p>
+      </div>
+    </div>
+  );
+};
--- a/ui/src/components/timetables/common/ChangeTimetablesValidityForm.tsx
+++ b/ui/src/components/timetables/common/ChangeTimetablesValidityForm.tsx
@@ -12,6 +12,7 @@ import {
   InputField,
   requiredString,
 } from '../../forms/common';
+import { AffectedRouteLabels } from './AffectedRouteLabels';
 
 export const schema = z.object({
   validityStart: requiredString.regex(/[0-9]{4}-[0-9]{2}-[0-9]{2}/),
@@ -82,15 +83,10 @@ export const ChangeTimetablesValidityForm = ({
           </FormRow>
         </FormColumn>
         <Visible visible={affectedRouteLabels.length > 1}>
-          <div className="my-6 flex flex-row items-center space-x-6">
-            <i className="icon-alert text-hsl-red" />
-            <div className="space-y-1 text-sm">
-              <p className="font-bold">
-                {t('changeTimetablesValidityModal.noticeChangesInRoutes')}
-              </p>
-              <p>{affectedRouteLabels.join(', ')}</p>
-            </div>
-          </div>
+          <AffectedRouteLabels
+            affectedRouteLabels={affectedRouteLabels}
+            text={t('changeTimetablesValidityModal.noticeChangesInRoutes')}
+          />
         </Visible>
         <Row className="mt-6 justify-end space-x-4">
           <SimpleButton onClick={onCancel} inverted>
--- a/ui/src/components/timetables/versions/DeleteTimetableModal.tsx
+++ b/ui/src/components/timetables/versions/DeleteTimetableModal.tsx
@@ -1,0 +1,76 @@
+import { Dialog } from '@headlessui/react';
+import { useTranslation } from 'react-i18next';
+import {
+  useAppDispatch,
+  useAppSelector,
+  useVehicleScheduleFrameValidity,
+} from '../../../hooks';
+import { useDeleteVehicleScheduleFrame } from '../../../hooks/vehicle-schedule-frame/useDeleteVehicleScheduleFrame';
+import {
+  closeDeleteTimetableModalAction,
+  selectDeleteTimetableModal,
+} from '../../../redux';
+import { ConfirmModal } from '../../../uiComponents/ConfirmModal';
+import { AffectedRouteLabels } from '../common/AffectedRouteLabels';
+
+interface Props {
+  fetchTimetableVersions: () => void;
+}
+
+const testIds = {
+  closeButton: 'DeleteTimetableModal::closeButton',
+  confirmButton: 'DeleteTimetableModal::confirmButton',
+  description: 'DeleteTimetableModal::description',
+};
+
+export const DeleteTimetableModal = ({
+  fetchTimetableVersions,
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+  const dispatch = useAppDispatch();
+
+  const { isOpen, vehicleScheduleFrameId } = useAppSelector(
+    selectDeleteTimetableModal,
+  );
+
+  const { affectedRouteLabels } = useVehicleScheduleFrameValidity(
+    vehicleScheduleFrameId,
+  );
+
+  const { deleteVehicleScheduleFrameMutation } = useDeleteVehicleScheduleFrame({
+    onCompleted: fetchTimetableVersions,
+  });
+
+  const onClose = () => {
+    dispatch(closeDeleteTimetableModalAction());
+  };
+
+  const onConfirm = async (id: string) => {
+    await deleteVehicleScheduleFrameMutation({
+      vehicle_schedule_frame_id: id,
+    });
+
+    dispatch(closeDeleteTimetableModalAction());
+  };
+
+  return (
+    <ConfirmModal
+      heading={t('deleteTimetableModal.title')}
+      isOpen={isOpen}
+      onCancel={onClose}
+      onConfirm={() =>
+        vehicleScheduleFrameId && onConfirm(vehicleScheduleFrameId)
+      }
+      confirmButtonText={t('deleteTimetableModal.confirmText')}
+      cancelButtonText={t('deleteTimetableModal.cancelText')}
+    >
+      <Dialog.Description data-testid={testIds.description}>
+        {t('deleteTimetableModal.description')}
+      </Dialog.Description>
+      <AffectedRouteLabels
+        affectedRouteLabels={affectedRouteLabels}
+        text={t('deleteTimetableModal.affectedRoutesText')}
+      />
+    </ConfirmModal>
+  );
+};
--- a/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx
+++ b/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx
@@ -1,15 +1,21 @@
 import { useTranslation } from 'react-i18next';
+import { useAppDispatch } from '../../../hooks';
 import { TimetableVersionRowData } from '../../../hooks/useGetTimetableVersions';
 import {
   mapDayOfWeekToUiName,
   mapTimetablePriorityToUiName,
 } from '../../../i18n/uiNameMappings';
 import { parseI18nField } from '../../../i18n/utils';
+import { Visible } from '../../../layoutComponents';
+import { openDeleteTimetableModalAction } from '../../../redux';
 import { mapToShortDate } from '../../../time';
 import { TimetablePriority } from '../../../types/enums';
+import { AlignDirection, SimpleDropdownMenu } from '../../../uiComponents';
+import { SimpleDropdownMenuItem } from '../../routes-and-lines/line-details/SimpleDropdownMenuItem';
 
 const testIds = {
   timetableVersionTableRow: 'TimetableVersionTableRow::row',
+  deleteTimetableMenuItem: 'TimetableVersionTableRow::deleteTimetableMenuItem',
 };
 
 const getStatusClassName = ({
@@ -51,6 +57,14 @@ interface Props {
 
 export const TimetableVersionTableRow = ({ data }: Props): JSX.Element => {
   const { t } = useTranslation();
+  const dispatch = useAppDispatch();
+
+  const onClick = () => {
+    if (data.vehicleScheduleFrame.id) {
+      dispatch(openDeleteTimetableModalAction(data.vehicleScheduleFrame.id));
+    }
+  };
+
   const statusClassName = ` ${getStatusClassName({
     priority: data.vehicleScheduleFrame.priority,
     inEffect: data.inEffect,
@@ -65,6 +79,7 @@ export const TimetableVersionTableRow = ({ data }: Props): JSX.Element => {
         dayOfWeek: mapDayOfWeekToUiName(data.substituteDay.substituteDayOfWeek),
       })
     : t('timetables.noService');
+
   const dayTypeClassName = `${getDayTypeClassName(
     data.vehicleScheduleFrame.priority,
   )} bg-opacity-25`;
@@ -89,7 +104,17 @@ export const TimetableVersionTableRow = ({ data }: Props): JSX.Element => {
       <td>{parseI18nField(data.vehicleScheduleFrame.nameI18n)}</td>
       <td>!Muokkaaja</td>
       <td>!Muokattu</td>
-      <td>...</td>
+      <td>
+        <Visible visible={!!data.vehicleScheduleFrame.id}>
+          <SimpleDropdownMenu alignItems={AlignDirection.Right} testId="menu">
+            <SimpleDropdownMenuItem
+              onClick={onClick}
+              testId={testIds.deleteTimetableMenuItem}
+              text={t('timetables.versions.deleteTimetable')}
+            />
+          </SimpleDropdownMenu>
+        </Visible>
+      </td>
     </tr>
   );
 };
--- a/ui/src/components/timetables/versions/TimetableVersionsPage.tsx
+++ b/ui/src/components/timetables/versions/TimetableVersionsPage.tsx
@@ -14,6 +14,7 @@ import { TimetablePriority } from '../../../types/enums';
 import { CloseIconButton } from '../../../uiComponents';
 import { TimeRangeControl } from '../../common';
 import { FormColumn, FormRow } from '../../forms/common';
+import { DeleteTimetableModal } from './DeleteTimetableModal';
 import { TimetableVersionTable } from './TimetableVersionTable';
 
 const testIds = {
@@ -33,7 +34,7 @@ export const TimetableVersionsPage = (): JSX.Element => {
       endDate,
     });
   // Then we can fetch the timetable versions using SQL functions
-  const { versions } = useGetTimetableVersions({
+  const { versions, fetchTimetableVersions } = useGetTimetableVersions({
     // We need to use memoized value which only changes when the loading status is changed. Otherwise we
     // end up in infinite loop
     journeyPatternIdsGroupedByRouteLabel: useMemo(
@@ -99,6 +100,7 @@ export const TimetableVersionsPage = (): JSX.Element => {
           data={onlyDraftTimetables}
         />
       </Container>
+      <DeleteTimetableModal fetchTimetableVersions={fetchTimetableVersions} />
     </Container>
   );
 };
--- a/ui/src/generated/graphql.tsx
+++ b/ui/src/generated/graphql.tsx
@@ -21908,6 +21908,21 @@ export type LineMapParamsFragment = {
   }>;
 };
 
+export type DeleteVehicleScheduleFrameMutationVariables = Exact<{
+  vehicle_schedule_frame_id: Scalars['uuid'];
+}>;
+
+export type DeleteVehicleScheduleFrameMutation = {
+  __typename?: 'mutation_root';
+  timetables?: {
+    __typename?: 'timetables_timetables_mutation_frontend';
+    timetables_delete_vehicle_schedule_vehicle_schedule_frame_by_pk?: {
+      __typename?: 'timetables_vehicle_schedule_vehicle_schedule_frame';
+      vehicle_schedule_frame_id: UUID;
+    } | null;
+  } | null;
+};
+
 export type VehicleScheduleFrameWithRoutesFragment = {
   __typename?: 'timetables_vehicle_schedule_vehicle_schedule_frame';
   vehicle_schedule_frame_id: UUID;
@@ -27500,6 +27515,61 @@ export type GetTimetableVersionsByJourneyPatternIdsQueryResult =
     GetTimetableVersionsByJourneyPatternIdsQuery,
     GetTimetableVersionsByJourneyPatternIdsQueryVariables
   >;
+export const DeleteVehicleScheduleFrameDocument = gql`
+  mutation DeleteVehicleScheduleFrame($vehicle_schedule_frame_id: uuid!) {
+    timetables {
+      timetables_delete_vehicle_schedule_vehicle_schedule_frame_by_pk(
+        vehicle_schedule_frame_id: $vehicle_schedule_frame_id
+      ) {
+        vehicle_schedule_frame_id
+      }
+    }
+  }
+`;
+export type DeleteVehicleScheduleFrameMutationFn = Apollo.MutationFunction<
+  DeleteVehicleScheduleFrameMutation,
+  DeleteVehicleScheduleFrameMutationVariables
+>;
+
+/**
+ * __useDeleteVehicleScheduleFrameMutation__
+ *
+ * To run a mutation, you first call `useDeleteVehicleScheduleFrameMutation` within a React component and pass it any options that fit your needs.
+ * When your component renders, `useDeleteVehicleScheduleFrameMutation` returns a tuple that includes:
+ * - A mutate function that you can call at any time to execute the mutation
+ * - An object with fields that represent the current status of the mutation's execution
+ *
+ * @param baseOptions options that will be passed into the mutation, supported options are listed on: https://www.apollographql.com/docs/react/api/react-hooks/#options-2;
+ *
+ * @example
+ * const [deleteVehicleScheduleFrameMutation, { data, loading, error }] = useDeleteVehicleScheduleFrameMutation({
+ *   variables: {
+ *      vehicle_schedule_frame_id: // value for 'vehicle_schedule_frame_id'
+ *   },
+ * });
+ */
+export function useDeleteVehicleScheduleFrameMutation(
+  baseOptions?: Apollo.MutationHookOptions<
+    DeleteVehicleScheduleFrameMutation,
+    DeleteVehicleScheduleFrameMutationVariables
+  >,
+) {
+  const options = { ...defaultOptions, ...baseOptions };
+  return Apollo.useMutation<
+    DeleteVehicleScheduleFrameMutation,
+    DeleteVehicleScheduleFrameMutationVariables
+  >(DeleteVehicleScheduleFrameDocument, options);
+}
+export type DeleteVehicleScheduleFrameMutationHookResult = ReturnType<
+  typeof useDeleteVehicleScheduleFrameMutation
+>;
+export type DeleteVehicleScheduleFrameMutationResult =
+  Apollo.MutationResult<DeleteVehicleScheduleFrameMutation>;
+export type DeleteVehicleScheduleFrameMutationOptions =
+  Apollo.BaseMutationOptions<
+    DeleteVehicleScheduleFrameMutation,
+    DeleteVehicleScheduleFrameMutationVariables
+  >;
 export const GetVehicleScheduleFrameWithRoutesDocument = gql`
   query GetVehicleScheduleFrameWithRoutes($vehicle_schedule_frame_id: uuid!) {
     timetables {
@@ -28139,6 +28209,7 @@ export function useGetTimetableVersionsByJourneyPatternIdsAsyncQuery() {
 }
 export type GetTimetableVersionsByJourneyPatternIdsAsyncQueryHookResult =
   ReturnType<typeof useGetTimetableVersionsByJourneyPatternIdsAsyncQuery>;
+
 export function useGetVehicleScheduleFrameWithRoutesAsyncQuery() {
   return useAsyncQuery<
     GetVehicleScheduleFrameWithRoutesQuery,
--- a/ui/src/graphql/client.ts
+++ b/ui/src/graphql/client.ts
@@ -229,6 +229,12 @@ const buildCacheDefinition = () => {
               return incoming;
             },
           },
+          timetables_vehicle_service_get_timetable_versions_by_journey_pattern_ids:
+            {
+              merge(_, incoming = []) {
+                return incoming;
+              },
+            },
         },
       },
       timetables_passing_times_timetabled_passing_time: {
--- a/ui/src/hooks/useGetTimetableVersions.ts
+++ b/ui/src/hooks/useGetTimetableVersions.ts
@@ -65,6 +65,7 @@ export interface TimetableVersionRowData {
     nameI18n: LocalizedString;
   };
   vehicleScheduleFrame: {
+    id: UUID | undefined;
     nameI18n: LocalizedString;
     priority: TimetablePriority;
     validityStart?: DateTime | null;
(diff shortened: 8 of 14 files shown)


HenrikHartiala (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)* status: 1 of 13 files reviewed, 2 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 12 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NkEU6yF83mn2TS9r7Xc):*




_Suggestion:_
```
  - Hook can take a callback function, which is executed when delete is successful
```
___
*[`-- commits` line 13 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NkEUi1hAPK2AI9UWPvF):*




_Suggestion:_
```
  - Added vsf id to TimetableVersionRowData. Vsf id is needed to delete timetable
```
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 28 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NkEgi4S0t4Ki-gSNLjD:-NkEgi4S0t4Ki-gSNLjE:brg6alv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/f03e068bf96541ca8ea2a64ae8db3852fabaaeab/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L28)):*
> ```XML
>   const { label } = useRequiredParams<{ label: string }>();
>   const { startDate, endDate } = useTimeRangeQueryParams();
>   const [isDeleteConfirmationOpen, setIsDeleteConfirmationOpen] =
> ```

I'd add the `Dialog`, `Modal` here for clarity



_Suggestion:_
```XML
  const [isDeleteConfirmationDialogOpen, setIsDeleteConfirmationDialogOpen] =
```
___
*[`ui/src/components/timetables/versions/TimetableVersionsPage.tsx` line 82 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NkEllIWEJtpmunAIyqX:-NkEllIWEJtpmunAIyqY:b-tll86p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/f03e068bf96541ca8ea2a64ae8db3852fabaaeab/ui/src/components/timetables/versions/TimetableVersionsPage.tsx#L82)):*
> ```XML
>   };
> 
>   const handleDialogOpen = (callback: () => void) => {
> ```

This seems a bit too complex, I think we could refactor this as discussed.

HenrikHartiala (changes requested):
> Reviewed 5 of 12 files at r7, 8 of 11 files at r8, 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)* status: all files reviewed, 9 unresolved discussions (waiting on @Perholehto)

___
*[`-- commits` line 2 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgOXOv0vH-weha2fsR):*
:muscle: 
___
*[`-- commits` line 12 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgQJx_9x7M5BnQPDE6):*
This commit is probably the one that could have the `Resolves HSLdevcom/jore4#1578` 
___
*[`ui/src/components/timetables/common/AffectedRouteLabels.tsx` line 6 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgQg5aDBX-IgNFN_93:-NlgQg5aDBX-IgNFN_94:bga9fki) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/common/AffectedRouteLabels.tsx#L6)):*
> ```XML
> }
> 
> export const AffectedRouteLabels = ({
> ```

Disclaimer: this is mostly thinking out loud and in two pats :D 
Hmm, can't figure out precisely what this component is, but this is now named as "AffectedRouteLabels" and it is located in `/common`. At first sight this seems pretty precise component, so should this actually either have different name or be located in some precise folder?
What do you think?

MORE: 
Hmm, yeah also kept reading the code and saw that this is used in ChangeTimetableValidityForm.tsx, which is located in the common folder and this is also used in the modal. So it is good that this is it's own component to have the same styles in both. This is probably fine as it is, but if these thoughts strike new ideas for the name or something to maybe try to tell more what this is, then would be nice. :D I mean the problem might not even be in the name or in the components, could also be our existing folder structure.
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 33 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgW8TL2OPikDrFCcU9:-NlgW8TL2OPikDrFCcUA:b-o73lkc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L33)):*
> ```XML
>   const dispatch = useAppDispatch();
> 
>   const deleteTimetableModalState = useAppSelector(selectDeleteTimetableModal);
> ```

You can also destructure these already here:



_Suggestion:_
```XML
  const { isOpen, vehicleScheduleFrameId } = useAppSelector(selectDeleteTimetableModal);
```
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 55 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgWp0mDb42y46rQufX:-NlgWp0mDb42y46rQufY:b-da0w11) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L55)):*
> ```XML
>       });
>       dispatch(closeDeleteTimetableModalAction());
>     }
> ```

Wondering if we should give some errortoast if it is missing? Or is it even possible to not have the id in place? 
Hmm, saw the same kind of logic further in the code. Could we disable the button or handle this some other way than having this if-condition here?
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 70 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgXCQfAr_J9TIJOeBh:-NlgXCQfAr_J9TIJOeBi:b-ihu62m) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L70)):*
> ```XML
>         {t('deleteTimetableModal.description')}
>       </Dialog.Description>
>       <Visible visible={affectedRouteLabels.length > 1}>
> ```

Do we not want to show this if there is only 1 affected?
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 62 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgZyZt0lcsmks0AmG2:-NlgZyZu-pVPA20P16Gi:b3q04f8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L62)):*
> ```XML
> 
>   const onClick = () => {
>     if (data.vehicleScheduleFrame.id) {
> ```

We could rather have the button disabled if some conditions are not met, and then here just always dispatch. And actually you already have the disabled, so we shouldn't need to check it here.
___
*[`ui/src/locales/en-US/common.json` line 470 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgbMLEEv2_8dXVGOfi:-NlgbMLEEv2_8dXVGOfj:b-yic6jp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/locales/en-US/common.json#L470)):*
> ```JSON
>   "deleteTimetableModal": {
>     "title": "Remove schedule",
>     "description": "The schedule is deleted and cannot be restored",
> ```

Would use future tense



_Suggestion:_
```JSON
    "description": "The schedule will be deleted and cannot be restored",
```
___
*[`ui/src/locales/en-US/common.json` line 473 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-Nlgc4jvBLpgYQGJR_re:-Nlgc4jvBLpgYQGJR_rf:bm9ooq5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/locales/en-US/common.json#L473)):*
> ```JSON
>     "confirmText": "Delete",
>     "cancelText": "Do not delete",
>     "affectedRoutes": "The schedules are also removed from the following routes",
> ```

Ahh, no I see why the modal had the "length > 1" and not "length > 0". We could also just explicitly say that "The schedules will be removed from the following routes" and then list all the affected routes, as the keyword "affectedRoutes" implies.
What do you think?
___
*[`ui/src/redux/slices/loader.ts` line 35 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgcnXh-XQ10vqzkzwH:-NlgcnXh-XQ10vqzkzwI:b-bkwyrq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/redux/slices/loader.ts#L35)):*
> ```XML
> ];
> 
> export const joreOperations = [...importOperations, Operation.DeleteTimetable];
> ```

initially thought that these would always be in some arrays, for example this could be in some `const timetableOperations`, but since there is only one currently, this is also fine. Just need to keep in mind that when we add more operations, we would have them categorised in their own arrays and then spread here. But thats just my initial thought.
___
*[`ui/src/redux/slices/modals.ts` line 142 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgdS410eBa98xQHQ4_:-NlgdS410eBa98xQHQ4a:b-hgvpu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/redux/slices/modals.ts#L142)):*
> ```XML
>     },
>     closeDeleteTimetableModal: (state) => {
>       state.deleteTimetableModal = initialState.deleteTimetableModal;
> ```

NIT, I know that this style is used in some of the redux actions, but for me this feels a bit awkward, because it is harder to read (after reading the closeDeleteTimetableModal you need to check what is actually the initial state) and also the `isOpen` should explicitly be set to false, instead of the initialState. But I guess this is a preference thing, so not blocking comment :D

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)* status: 4 of 17 files reviewed, 9 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 12 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgQJx_9x7M5BnQPDE6):*

This commit is probably the one that could have the `Resolves HSLdevcom/jore4#1578`
</blockquote></details>

added to commit message
___
*[`ui/src/components/timetables/common/AffectedRouteLabels.tsx` line 6 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgQg5aDBX-IgNFN_93:-Nlh30Wu-NtNpBW5bB91:b-6n5agh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/common/AffectedRouteLabels.tsx#L6)):*

Disclaimer: this is mostly thinking out loud and in two pats :D 
Hmm, can't figure out precisely what this component is, but this is now named as "AffectedRouteLabels" and it is located in `/common`. At first sight this seems pretty precise component, so should this actually either have different name or be located in some precise folder?
What do you think?

MORE: 
Hmm, yeah also kept reading the code and saw that this is used in ChangeTimetableValidityForm.tsx, which is located in the common folder and this is also used in the modal. So it is good that this is it's own component to have the same styles in both. This is probably fine as it is, but if these thoughts strike new ideas for the name or something to maybe try to tell more what this is, then would be nice. :D I mean the problem might not even be in the name or in the components, could also be our existing folder structure.
</blockquote></details>

So this component's logic is extracted to own component to avoid duplication in `DeleteTimetableModal`. I couldn't figure out better place to put this new component and it's now close to it's origin.
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 33 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgW8TL2OPikDrFCcU9:-Nlh3sJn8o54KF-EcoAo:b-sxkoxv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L33)):*

You can also destructure these already here:
</blockquote></details>

Yeah, good catch!
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 55 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgWp0mDb42y46rQufX:-NlhV6L04O8OA7ugrqu8:bxtbxl5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L55)):*

Wondering if we should give some errortoast if it is missing? Or is it even possible to not have the id in place? 
Hmm, saw the same kind of logic further in the code. Could we disable the button or handle this some other way than having this if-condition here?
</blockquote></details>

Here `id` is typed to `string | undefined` because we need to have initial value in Redux and I think it's better to have `string | undefined` than use `string` and then initial value is `''`.
It's not possible to open this modal with `undefined` Action takes only `string` so this check is only to keep compiler happy.
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 70 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgXCQfAr_J9TIJOeBh:-Nlhbd7j60EOGAnqqrQv:b5jie3f) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L70)):*

Do we not want to show this if there is only 1 affected?
</blockquote></details>

Replied to other comment about this
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 62 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgZyZt0lcsmks0AmG2:-NlhWCLS169elEJ33TsJ:bjdwon7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L62)):*

We could rather have the button disabled if some conditions are not met, and then here just always dispatch. And actually you already have the disabled, so we shouldn't need to check it here.
</blockquote></details>

I don't think there is a way to skip checking if `data.vehicleScheduleFrame.id` is a string. Here it should always be `string`, could be handled with non-null assertion but I don't think it's better.
Ofc we could move that check to inline in `OnClick` as you suggested in Slack but I don't think that makes any difference?
___
*[`ui/src/locales/en-US/common.json` line 470 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgbMLEEv2_8dXVGOfi:-NlhXMWrAqgFrbPmw_my:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/locales/en-US/common.json#L470)):*

Would use future tense
</blockquote></details>

:+1: 
___
*[`ui/src/locales/en-US/common.json` line 473 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-Nlgc4jvBLpgYQGJR_re:-NlhYNL-AS6veBX5HheH:b-o92yit) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/locales/en-US/common.json#L473)):*

Ahh, no I see why the modal had the "length > 1" and not "length > 0". We could also just explicitly say that "The schedules will be removed from the following routes" and then list all the affected routes, as the keyword "affectedRoutes" implies.
What do you think?
</blockquote></details>

The idea with `length >1` is that when you have only one label it's always label of the line which version you were watching and clicked to open the modal. So this is additional information that is showed when delete has impact to somewhere else that is not already on the screen
___
*[`ui/src/redux/slices/loader.ts` line 35 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgcnXh-XQ10vqzkzwH:-Nlh_1cSEU2yF3BTSPv3:byq5nx2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/redux/slices/loader.ts#L35)):*

initially thought that these would always be in some arrays, for example this could be in some `const timetableOperations`, but since there is only one currently, this is also fine. Just need to keep in mind that when we add more operations, we would have them categorised in their own arrays and then spread here. But thats just my initial thought.
</blockquote></details>

Good to know. I think these will get grouped up when there is operations that should be grouped.

Maybe Readme should I have this thought ? Now it has this line: 
`"just add the chosen Operation to joreOperations in loader.ts."`
___
*[`ui/src/redux/slices/modals.ts` line 142 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgdS410eBa98xQHQ4_:-NlhbCZ69zT8-RigMszY:b-rt11d8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/redux/slices/modals.ts#L142)):*

NIT, I know that this style is used in some of the redux actions, but for me this feels a bit awkward, because it is harder to read (after reading the closeDeleteTimetableModal you need to check what is actually the initial state) and also the `isOpen` should explicitly be set to false, instead of the initialState. But I guess this is a preference thing, so not blocking comment :D
</blockquote></details>

Not sure how I feel about this or  both ways are okay to me. Changed this and will try to prefer explicitly set these values. I think it's better if we have common way to do these and not just do what feels good

HenrikHartiala (changes requested):
> Reviewed 2 of 2 files at r11, 9 of 11 files at r12, 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)* status: all files reviewed, 5 unresolved discussions (waiting on @Perholehto)

___
*[`ui/src/components/timetables/common/AffectedRouteLabels.tsx` line 6 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgQg5aDBX-IgNFN_93:-Nlw11oI5B-7jZnJY2aC:b-qb5qpn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/common/AffectedRouteLabels.tsx#L6)):*

So this component's logic is extracted to own component to avoid duplication in `DeleteTimetableModal`. I couldn't figure out better place to put this new component and it's now close to it's origin.
</blockquote></details>

Yes I mean the extraction of this is completely OK and very good, but there is something in this that makes me wonder. I mean yes it is used for "affected" routelabels, but do we need to make that precision in this component name? Could the nature of this component name be more generic somehow? Don't have a good suggestion for this, but the point is that there is kind of nothing that makes this component actually dependant on affected route labels, it is just icon, styles, text and then joined strings, so practically just a generic visual component.
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 55 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgWp0mDb42y46rQufX:-Nlw3Egz3GvOnEigK8sR:b-2ehvd4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L55)):*

Here `id` is typed to `string | undefined` because we need to have initial value in Redux and I think it's better to have `string | undefined` than use `string` and then initial value is `''`.
It's not possible to open this modal with `undefined` Action takes only `string` so this check is only to keep compiler happy.
</blockquote></details>

For me, if we need to make if-clauses just to keep compiler happy, but no need to handle the block of code after the if-block, then it feels like a hint that some responsibilities are off.
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 70 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgXCQfAr_J9TIJOeBh:-NlwHYQc7KG9zzfPmMmG:b-h7jm06) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L70)):*

Replied to other comment about this
</blockquote></details>

Couple of general things:
I think we should not hide the information from the user even though if it is the only affected route. Now if the user clicks "poista aikataulu", and there is only one affected route, the modal does not tell the user what route is being deleted.
Also I would explicitly show the "affected routes" even if there is only one. Because it can be confusing to the user that there is one case where we do not show the affected routes, and it might make the user wonder "Hmm, what routes are actually affected?" IF the user has already seen this same confirmation modal with the affected routes listing.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 62 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgZyZt0lcsmks0AmG2:-NlvslvQ-6H0bR4vLjhd:bquwt4d) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L62)):*

I don't think there is a way to skip checking if `data.vehicleScheduleFrame.id` is a string. Here it should always be `string`, could be handled with non-null assertion but I don't think it's better.
Ofc we could move that check to inline in `OnClick` as you suggested in Slack but I don't think that makes any difference?
</blockquote></details>

The only thing it makes, is that it removes the checking responsibility out from the actual clickHandler which's responsibility is to act when correct things are in place and clicked, but now we check if the `id` is in place implying that there is a case where it is not in place.
As an example I did have to ask in the `onConfirm` case that "Wondering if we should give some errortoast if it is missing?" because we had to check the id there. 

I don't have a silver bullet for these either, so these are fine as they are, up to you if you want to make some changes still :+1: 
___
*[`ui/src/locales/en-US/common.json` line 473 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-Nlgc4jvBLpgYQGJR_re:-Nlvx7LF6Sj5sQQmx1FI:bffscg1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/locales/en-US/common.json#L473)):*

The idea with `length >1` is that when you have only one label it's always label of the line which version you were watching and clicked to open the modal. So this is additional information that is showed when delete has impact to somewhere else that is not already on the screen
</blockquote></details>

I can understand that yes, but then we should probably also bring that information in`affectedRoutes`. I would assume that `affectedRoutes` are the routes that are affected, but in this case they are more like "affected routes other than the one picked", if you can catch my breath :D. So something like "otherAffectedRoutes" or something. 

And btw other thing is that should we also have some postfix in this translation key like `text`?
___
*[`ui/src/redux/slices/loader.ts` line 35 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgcnXh-XQ10vqzkzwH:-Nlvz56w5UVMsjoOagPc:bior6q1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/redux/slices/loader.ts#L35)):*

Good to know. I think these will get grouped up when there is operations that should be grouped.

Maybe Readme should I have this thought ? Now it has this line: 
`"just add the chosen Operation to joreOperations in loader.ts."`
</blockquote></details>

Probably could be yeah, but after we have more operations here, I think it is obvious after that. Let's check how this will go forward and then add the readme if it's necessary. Maybe we do want to put all of the operations straight in to this operations, or partly.
___
*[`ui/src/redux/slices/modals.ts` line 142 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgdS410eBa98xQHQ4_:-NlvziVWCQZWpilm6zFv:brla4f2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/redux/slices/modals.ts#L142)):*

Not sure how I feel about this or  both ways are okay to me. Changed this and will try to prefer explicitly set these values. I think it's better if we have common way to do these and not just do what feels good
</blockquote></details>

Thanks,
If for any reason at all, we had some modal open as default, we would have to explicitly use `isOpen:false` to close it in that case anyway. Or just the thought that changing initial state has side effects to closeAction makes me want to explicitly always set these. 
For some `reset` action I think it's correct to use the initial state.

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)* status: 4 of 17 files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/timetables/common/AffectedRouteLabels.tsx` line 6 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgQg5aDBX-IgNFN_93:-Nm5CYI9-g_AFgg-0zD0:b-wmrwwn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/common/AffectedRouteLabels.tsx#L6)):*

Yes I mean the extraction of this is completely OK and very good, but there is something in this that makes me wonder. I mean yes it is used for "affected" routelabels, but do we need to make that precision in this component name? Could the nature of this component name be more generic somehow? Don't have a good suggestion for this, but the point is that there is kind of nothing that makes this component actually dependant on affected route labels, it is just icon, styles, text and then joined strings, so practically just a generic visual component.
</blockquote></details>

I don't think this should be more generic here. It's now used in two places, both which have context where `affectedRouteLabels` makes a sense. I understand that this is simple component and it shows only a icon, etc... But if I now make this component somehow more generic and it's used, let's say in front page. Then we have a need to change how we show affected route labels, now we have coupled this component to other places and we need to extend it to handle two different cases or make a new component.
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 55 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgWp0mDb42y46rQufX:-Nm5G20O5bC5e_KH-_sm:brao9md) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L55)):*

For me, if we need to make if-clauses just to keep compiler happy, but no need to handle the block of code after the if-block, then it feels like a hint that some responsibilities are off.
</blockquote></details>

Maybe this is Telco stuff?
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 70 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgXCQfAr_J9TIJOeBh:-Nm5FyQi30fY8f8gV92s:bpt1bah) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L70)):*

Couple of general things:
I think we should not hide the information from the user even though if it is the only affected route. Now if the user clicks "poista aikataulu", and there is only one affected route, the modal does not tell the user what route is being deleted.
Also I would explicitly show the "affected routes" even if there is only one. Because it can be confusing to the user that there is one case where we do not show the affected routes, and it might make the user wonder "Hmm, what routes are actually affected?" IF the user has already seen this same confirmation modal with the affected routes listing.
</blockquote></details>

I think this is something we can discuss in Telco ?
___
*[`ui/src/locales/en-US/common.json` line 473 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-Nlgc4jvBLpgYQGJR_re:-Nm5FiVz2ScNKlLFg8aj:b-34ithr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/locales/en-US/common.json#L473)):*

I can understand that yes, but then we should probably also bring that information in`affectedRoutes`. I would assume that `affectedRoutes` are the routes that are affected, but in this case they are more like "affected routes other than the one picked", if you can catch my breath :D. So something like "otherAffectedRoutes" or something. 

And btw other thing is that should we also have some postfix in this translation key like `text`?
</blockquote></details>

Added postfix! 

But not sure if I catch your breath in first one.
___
*[`ui/src/redux/slices/modals.ts` line 142 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgdS410eBa98xQHQ4_:-Nm5FryN3skS3Wq1NToI:b-7m2y43) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/redux/slices/modals.ts#L142)):*

Thanks,
If for any reason at all, we had some modal open as default, we would have to explicitly use `isOpen:false` to close it in that case anyway. Or just the thought that changing initial state has side effects to closeAction makes me want to explicitly always set these. 
For some `reset` action I think it's correct to use the initial state.
</blockquote></details>

noted

Perholehto (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)* status: 4 of 17 files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/timetables/common/AffectedRouteLabels.tsx` line 6 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgQg5aDBX-IgNFN_93:-NmA4bOF6ldORp1cxsfn:b-7vfgsl) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/common/AffectedRouteLabels.tsx#L6)):*

I don't think this should be more generic here. It's now used in two places, both which have context where `affectedRouteLabels` makes a sense. I understand that this is simple component and it shows only a icon, etc... But if I now make this component somehow more generic and it's used, let's say in front page. Then we have a need to change how we show affected route labels, now we have coupled this component to other places and we need to extend it to handle two different cases or make a new component.
</blockquote></details>

Resolved in telco
___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 70 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgXCQfAr_J9TIJOeBh:-NmA4fwN8ikNwmtZ1WS5:be9y6zm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L70)):*

I think this is something we can discuss in Telco ?
</blockquote></details>

Removed conditional rendering, moved id checking to inline and `text` postfix
___
*[`ui/src/locales/en-US/common.json` line 473 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-Nlgc4jvBLpgYQGJR_re:-NmA4sqxADATb2_PsGs9:bx4suz3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/locales/en-US/common.json#L473)):*

Added postfix! 

But not sure if I catch your breath in first one.
</blockquote></details>

resolved in telco

HenrikHartiala (approved):
> Reviewed 2 of 2 files at r15, 8 of 11 files at r16, 3 of 3 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Perholehto)

___
*[`ui/src/components/timetables/versions/DeleteTimetableModal.tsx` line 70 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/724#-NlgXCQfAr_J9TIJOeBh:-NmA9vAA06Bue00jZ_Jq:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/b4f9c8cfb013e497ad1066c97475f400a2a5e61e/ui/src/components/timetables/versions/DeleteTimetableModal.tsx#L70)):*

Removed conditional rendering, moved id checking to inline and `text` postfix
</blockquote></details>

Thanks!

PR #742, opened 2023-12-07
Fix run-integration-tests actions order

HenrikHartiala:
setup-node action is run after we install dependencies, which can cause the version restriction to fail to incompatible versions if the cache has another version at that time. So we need to first setup the correct node version, and then run yarn install. This is also the order in all other workflows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/742)

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -14,6 +14,10 @@ jobs:
         with:
           submodules: recursive
 
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18.18.2'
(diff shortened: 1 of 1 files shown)


culka (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/742#-:-Nl3OId456oYwyYJeXJr:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #747, opened 2023-12-14
Fix stop order rendering

HenrikHartiala:
This rendering relied that passingTimesByStop was populated in the correct order. It seemed that there was a possibility that this was populated partly wrong, or it used partly some cache(?) or it was not immutable or something, because we had cases where the first stop was actually rendered last and the only possibiity is that the last stop was added last to passingTimesByStop object, which will put it last in the array when we use Object.entries(passingTimesByStop).

The mapping logic is now changed so that we always map stopsInJourneyPattern, which always should have the stops in correct order, and changes to timetables won't change that. It also is a bit more readable to have the mapping this way.

Resolves HSLdevcom/jore4#1547

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/747)


Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/747)* status: 0 of 2 files reviewed, all discussions resolved

___
*[`-- commits` line 12 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/747#-NlbxP-g5L9HUBQSykym):*
> which always should have the stops in correct order

After some digging this does seem to be the case, but I had to go through quite a few components and graphql fragments to "verify" this.
Seems to work because the `route` here is fetched with `GetRouteWithJourneyPattern` which uses fragments `route\_with\_journey\_pattern\_stops` -> `journey\_pattern\_with\_stops` -> `scheduled\_stop\_point\_in\_journey\_patterns( order\_by: { scheduled\_stop\_point\_sequence: asc } )`

Sorting these here would be trivial, but maybe a bit useless since this probably isn't the only place where we rely on the query sorting the stops correctly.

A "proper" fix I think would be to eg. rename the property in query (using an alias that is, I think),  to `ordered\_scheduled\_stop\_point\_in\_journey\_patterns`.
There's quite a few other queries that fetch `scheduled\_stop\_point\_in\_journey\_patterns`, and only this one query here has them sorted.
Worth it?

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/747#-:-Nlc0jTyEQAIb21pb19W:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

Leitsi (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/747#-:-Nlcj0AyBlBdrjdkBfFQ:bnfp4nl)* status: 0 of 8 files reviewed, all discussions resolved

PR #769, opened 2024-02-20
Timetables version details sidepanel

HenrikHartiala:
Commits: 
* Refactor VehicleJourneyGroupInfo parameters to require only necessary
This component had unnecessary requirements for parameters. It required `VehicleJourneyGroup`
as parameter, even though it did not need all attributes within that type.
Also the VehicleJourneyGroup made the requirements to be so that vehicleJourney's had to be
type `VehicleJourneyWithService` which reduced the reusability greatly or forced to use that fragment.
Only the start_time's are required for this component which is now the only required attribute. (<Pick..>)

* Add timetable version details panel functionality
This version details will open the timetable schedules of the selected version, which corresponds to the selected
versions vehicleScheduleFrame's schedules. This view is now currently the only way you can view all the timetables
within one vehileScheduleFrame, which is crucial if the user wants to change the validity of a timetable, because
it affects the validity period of the whole vehicle schedule frame.
This is pretty big commit and it has some overlapping components with existing code, and the existing code can be also refactored
to use some of the new components. For example `TimetableHeading` can be relocated and used with the existing
  * Added a optional `onChange` callback function to `ChangeTimetablesValidityModal`, so that we add custom
functionality in case of change e.g. re-fetching some view data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)

### DIFF
--- a/cypress/e2e/timetableImport.cy.ts
+++ b/cypress/e2e/timetableImport.cy.ts
@@ -727,10 +727,10 @@ describe('Timetable import', () => {
           '2023-06-30',
         );
         timetableVersionsPage.timetableVersionTableRow
-          .getRow()
+          .getRows()
           .should('have.length', 2);
         timetableVersionsPage.timetableVersionTableRow
-          .getRow()
+          .getRows()
           .eq(0)
           .should('contain', 'Voimassa')
           .and('contain', 'Sunnuntai')
@@ -738,7 +738,7 @@ describe('Timetable import', () => {
           .and('contain', '31.12.2049')
           .and('contain', '99');
         timetableVersionsPage.timetableVersionTableRow
-          .getRow()
+          .getRows()
           .eq(1)
           .should('contain', 'Luonnos')
           .and('contain', 'Sunnuntai')
--- a/cypress/e2e/timetableVersionDetailsPanel.cy.ts
+++ b/cypress/e2e/timetableVersionDetailsPanel.cy.ts
@@ -1,0 +1,612 @@
+import {
+  GetInfrastructureLinksByExternalIdsResult,
+  InfraLinkAlongRouteInsertInput,
+  JourneyPatternInsertInput,
+  LineInsertInput,
+  RouteDirectionEnum,
+  RouteInsertInput,
+  RouteTypeOfLineEnum,
+  StopInJourneyPatternInsertInput,
+  StopInsertInput,
+  buildLine,
+  buildRoute,
+  buildStop,
+  buildStopInJourneyPattern,
+  buildTimingPlace,
+  extractInfrastructureLinkIdsFromResponse,
+  mapToGetInfrastructureLinksByExternalIdsQuery,
+} from '@hsl/jore4-test-db-manager';
+import { defaultDayTypeIds } from '@hsl/timetables-data-inserter';
+import { DateTime, Duration } from 'luxon';
+import {
+  ChangeTimetablesValidityForm,
+  TimetableVersionDetailsPanel,
+  TimetableVersionsPage,
+  Toast,
+  VehicleServiceRow,
+} from '../pageObjects';
+import { UUID } from '../types';
+import {
+  SupportedResources,
+  insertToDbHelper,
+  removeFromDbHelper,
+} from '../utils';
+
+// TODO: Extract the basedata somewhere, so that it can be used in multiple tests. No need
+// to copy paste 400+ lines in front of every test.
+
+// TODO: Also after extracting base data, expand the data so that we have different labeled routes
+// in the dataset and add asserts to the tests for those details
+
+// These infralink IDs exist in the 'infraLinks.sql' test data file.
+// These form a straight line on Eerikinkatu in Helsinki.
+// Coordinates are partial since they are needed only for the stop creation.
+
+const testInfraLinks = [
+  {
+    externalId: '445156',
+    coordinates: [24.925682785, 60.163824160000004, 7.3515],
+  },
+  {
+    externalId: '442331',
+    coordinates: [24.927565858, 60.1644843305, 9.778500000000001],
+  },
+  {
+    externalId: '442424',
+    coordinates: [24.929825718, 60.165285984, 9.957],
+  },
+  {
+    externalId: '442325',
+    coordinates: [24.93312261043133, 60.16645636069328, 13.390046659939703],
+  },
+];
+
+const stopLabels = ['H1231', 'H1232', 'H1233', 'H1234'];
+
+const lines: LineInsertInput[] = [
+  {
+    ...buildLine({ label: '99' }),
+    line_id: 'f148d51b-36ff-4321-8cf1-049946f75f73',
+    type_of_line: RouteTypeOfLineEnum.StoppingBusService,
+  },
+];
+
+const timingPlaces = [
+  buildTimingPlace('50623e7b-abd0-4c9a-85fa-f88ff7f65a06', '1AACKT'),
+  buildTimingPlace('f8a93c6f-5ef7-4b09-ae5e-0a04ea8597e9', '1ELIMK'),
+  buildTimingPlace('31a2592e-0af1-48b7-8f2f-373dcca39ddd', '1AURLA'),
+];
+
+const buildStopsOnInfrastrucureLinks = (
+  infrastructureLinkIds: UUID[],
+): StopInsertInput[] => [
+  {
+    ...buildStop({
+      label: stopLabels[0],
+      located_on_infrastructure_link_id: infrastructureLinkIds[0],
+    }),
+    scheduled_stop_point_id: 'd9f0bc78-45f2-4d44-9cac-4674f856a400',
+    timing_place_id: timingPlaces[0].timing_place_id,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[0].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: stopLabels[1],
+      located_on_infrastructure_link_id: infrastructureLinkIds[1],
+    }),
+    scheduled_stop_point_id: '63bd05f9-de46-4fa3-bdd9-10e9a81702e3',
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[1].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: stopLabels[2],
+      located_on_infrastructure_link_id: infrastructureLinkIds[2],
+    }),
+    scheduled_stop_point_id: 'f732ceb2-fc41-4843-8164-ead6ec7dd33b',
+    timing_place_id: timingPlaces[1].timing_place_id,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[2].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: stopLabels[3],
+      located_on_infrastructure_link_id: infrastructureLinkIds[3],
+    }),
+    scheduled_stop_point_id: '322a32cc-7a50-402b-9c01-5dc6a6b39af6',
+    timing_place_id: timingPlaces[2].timing_place_id,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[3].coordinates,
+    },
+  },
+];
+
+const routes: RouteInsertInput[] = [
+  {
+    ...buildRoute({ label: '99' }),
+    route_id: '829e9d55-aa25-4ab9-858b-f2a5aa81d931',
+    on_line_id: lines[0].line_id,
+    validity_start: DateTime.fromISO('2023-01-01T13:08:43.315+03:00'),
+    validity_end: DateTime.fromISO('2043-06-30T13:08:43.315+03:00'),
+    direction: RouteDirectionEnum.Inbound,
+  },
+  {
+    ...buildRoute({ label: '99' }),
+    route_id: '06577560-4535-46ec-8240-0b989b0fed87',
+    on_line_id: lines[0].line_id,
+    validity_start: DateTime.fromISO('2023-01-01T13:08:43.315+03:00'),
+    validity_end: DateTime.fromISO('2043-06-30T13:08:43.315+03:00'),
+    direction: RouteDirectionEnum.Outbound,
+  },
+];
+
+const buildInfraLinksAlongRoute = (
+  infrastructureLinkIds: UUID[],
+): InfraLinkAlongRouteInsertInput[] => [
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[0],
+    infrastructure_link_sequence: 0,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[1],
+    infrastructure_link_sequence: 1,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[2],
+    infrastructure_link_sequence: 2,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[0].route_id,
+    infrastructure_link_id: infrastructureLinkIds[3],
+    infrastructure_link_sequence: 3,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[1].route_id,
+    infrastructure_link_id: infrastructureLinkIds[3],
+    infrastructure_link_sequence: 0,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[1].route_id,
+    infrastructure_link_id: infrastructureLinkIds[2],
+    infrastructure_link_sequence: 1,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[1].route_id,
+    infrastructure_link_id: infrastructureLinkIds[1],
+    infrastructure_link_sequence: 2,
+    is_traversal_forwards: true,
+  },
+  {
+    route_id: routes[1].route_id,
+    infrastructure_link_id: infrastructureLinkIds[0],
+    infrastructure_link_sequence: 3,
+    is_traversal_forwards: true,
+  },
+];
+
+const journeyPatterns: JourneyPatternInsertInput[] = [
+  {
+    journey_pattern_id: '72ff8c33-33aa-4169-803e-53f2426a4508',
+    on_route_id: routes[0].route_id,
+  },
+  {
+    journey_pattern_id: '45217fb9-f95f-4193-8597-fb285e859d04',
+    on_route_id: routes[1].route_id,
+  },
+];
+
+const stopsInJourneyPattern: StopInJourneyPatternInsertInput[] = [
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[0],
+    scheduledStopPointSequence: 0,
+    isUsedAsTimingPoint: true,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[1],
+    scheduledStopPointSequence: 1,
+    isUsedAsTimingPoint: false,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[2],
+    scheduledStopPointSequence: 2,
+    isUsedAsTimingPoint: true,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[0].journey_pattern_id,
+    stopLabel: stopLabels[3],
+    scheduledStopPointSequence: 3,
+    isUsedAsTimingPoint: true,
+  }),
+];
+
+const stopInAnotherJourneyPattern: StopInJourneyPatternInsertInput[] = [
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[1].journey_pattern_id,
+    stopLabel: stopLabels[3],
+    scheduledStopPointSequence: 0,
+    isUsedAsTimingPoint: true,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[1].journey_pattern_id,
+    stopLabel: stopLabels[2],
+    scheduledStopPointSequence: 1,
+    isUsedAsTimingPoint: true,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[1].journey_pattern_id,
+    stopLabel: stopLabels[1],
+    scheduledStopPointSequence: 2,
+    isUsedAsTimingPoint: false,
+  }),
+  buildStopInJourneyPattern({
+    journeyPatternId: journeyPatterns[1].journey_pattern_id,
+    stopLabel: stopLabels[0],
+    scheduledStopPointSequence: 3,
+    isUsedAsTimingPoint: true,
+  }),
+];
+
+const timetableDataInput = {
+  _journey_pattern_refs: {
+    route99inbound: {
+      route_label: '99',
+      route_direction: 'inbound',
+      journey_pattern_id: journeyPatterns[0].journey_pattern_id,
+      _stop_points: [
+        {
+          scheduled_stop_point_sequence: 1,
+          scheduled_stop_point_label: 'H1231',
+          timing_place_label: '1AACKT',
+        },
+        {
+          scheduled_stop_point_sequence: 2,
+          scheduled_stop_point_label: 'H1232',
+        },
+        {
+          scheduled_stop_point_sequence: 3,
+          scheduled_stop_point_label: 'H1233',
+          timing_place_label: '1ELIMK',
+        },
+        {
+          scheduled_stop_point_sequence: 4,
+          scheduled_stop_point_label: 'H1234',
+          timing_place_label: '1AURLA',
+        },
+      ],
+    },
+    route99outbound: {
+      route_label: '99',
+      route_direction: 'outbound',
+      journey_pattern_id: journeyPatterns[1].journey_pattern_id,
+      _stop_points: [
+        {
+          scheduled_stop_point_sequence: 1,
+          scheduled_stop_point_label: 'H1234',
+          timing_place_label: '1AURLA',
+        },
+        {
+          scheduled_stop_point_sequence: 2,
+          scheduled_stop_point_label: 'H1233',
+          timing_place_label: '1ELIMK',
+        },
+        {
+          scheduled_stop_point_sequence: 3,
+          scheduled_stop_point_label: 'H1232',
+        },
+        {
+          scheduled_stop_point_sequence: 4,
+          scheduled_stop_point_label: 'H1231',
+          timing_place_label: '1AACKT',
+        },
+      ],
+    },
+  },
+  _vehicle_schedule_frames: {
+    year2023: {
+      validity_start: DateTime.fromISO('2023-01-01'),
+      validity_end: DateTime.fromISO('2023-12-31'),
+      name: '2023',
+      booking_label: '2023 booking label',
+      _vehicle_services: {
+        saturday: {
+          day_type_id: defaultDayTypeIds.SATURDAY,
+          _blocks: {
+            block: {
+              _vehicle_journeys: {
+                route99Inbound1: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT7H05M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT7H12M'),
+                      departure_time: Duration.fromISO('PT7H12M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT7H19M'),
+                      departure_time: Duration.fromISO('PT7H23M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT7H27M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+                route99Inbound2: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT8H20M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT8H22M'),
+                      departure_time: Duration.fromISO('PT8H22M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT8H26M'),
+                      departure_time: Duration.fromISO('PT8H27M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT8H29M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+                route99Inbound3: {
+                  _journey_pattern_ref_name: 'route99inbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT8H30M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT8H39M'),
+                      departure_time: Duration.fromISO('PT8H39M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT8H50M'),
+                      departure_time: Duration.fromISO('PT8H50M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: Duration.fromISO('PT8H59M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+                route99Outbound1: {
+                  _journey_pattern_ref_name: 'route99outbound',
+                  _passing_times: [
+                    {
+                      _scheduled_stop_point_label: 'H1234',
+                      arrival_time: null,
+                      departure_time: Duration.fromISO('PT7H28M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1233',
+                      arrival_time: Duration.fromISO('PT7H38M'),
+                      departure_time: Duration.fromISO('PT7H38M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1232',
+                      arrival_time: Duration.fromISO('PT8H00M'),
+                      departure_time: Duration.fromISO('PT8H00M'),
+                    },
+                    {
+                      _scheduled_stop_point_label: 'H1231',
+                      arrival_time: Duration.fromISO('PT8H15M'),
+                      departure_time: null,
+                    },
+                  ],
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+  },
+};
+
+describe('Timetable version details panel', () => {
+  const baseDbResources = {
+    lines,
+    routes,
+    journeyPatterns,
+    stopsInJourneyPattern,
+    stopInAnotherJourneyPattern,
+  };
+  let dbResources: SupportedResources;
+  let timetableVersionsPage: TimetableVersionsPage;
+  let timetableVersionDetailsPanel: TimetableVersionDetailsPanel;
+  let changeTimetablesValidityForm: ChangeTimetablesValidityForm;
+  let vehicleServiceRow: VehicleServiceRow;
+  let toast: Toast;
+
+  before(() => {
+    cy.task<GetInfrastructureLinksByExternalIdsResult>(
+      'hasuraAPI',
+      mapToGetInfrastructureLinksByExternalIdsQuery(
+        testInfraLinks.map((infralink) => infralink.externalId),
+      ),
+    ).then((res) => {
+      const infraLinkIds = extractInfrastructureLinkIdsFromResponse(res);
+      const stops = buildStopsOnInfrastrucureLinks(infraLinkIds);
+      const infraLinksAlongRoute = buildInfraLinksAlongRoute(infraLinkIds);
+      dbResources = {
+        ...baseDbResources,
+        timingPlaces,
+        stops,
+        infraLinksAlongRoute,
+      };
+    });
+  });
+
+  beforeEach(() => {
+    cy.task('truncateTimetablesDatabase');
+    removeFromDbHelper(dbResources);
+    insertToDbHelper(dbResources);
+    cy.task('insertHslTimetablesDatasetToDb', timetableDataInput);
+
+    timetableVersionsPage = new TimetableVersionsPage();
+    timetableVersionDetailsPanel = new TimetableVersionDetailsPanel();
+    changeTimetablesValidityForm = new ChangeTimetablesValidityForm();
+    vehicleServiceRow = new VehicleServiceRow();
+    toast = new Toast();
+
+    cy.setupTests();
+    cy.mockLogin();
+    cy.visit(
+      `/timetables/lines/99/versions?startDate=2023-03-04&endDate=2024-03-04`,
+    );
+  });
+
+  afterEach(() => {
+    removeFromDbHelper(dbResources);
+    cy.task('truncateTimetablesDatabase');
+  });
+
+  it('Should open, have correct details, and close', () => {
+    timetableVersionsPage.timetableVersionTableRow
+      .getRow()
+      .should('have.length', 1);
+
+    timetableVersionsPage.timetableVersionTableRow.openNthRowVersionDetailsPanel(
+      0,
+    );
+
+    timetableVersionDetailsPanel
+      .getHeading()
+      .should('contain', 'Aikataulu voimassa')
+      .and('contain', '1.1.2023 - 31.12.2023');
+
+    timetableVersionDetailsPanel.getRows().should('have.length', 2);
+
+    timetableVersionDetailsPanel
+      .getRows()
+      .eq(0)
+      .findByTestId('DirectionBadge::outbound')
+      .should('be.visible');
+
+    timetableVersionDetailsPanel.getRows().eq(0).should('contain', '99');
+
+    timetableVersionDetailsPanel.toggleExpandNthRow(0);
+    timetableVersionDetailsPanel.getRows()
+      .eq(0)
+      .findByTestId('VehicleServiceRow::row')
+      .findByTestId('VehicleServiceRow::hour')
+      .eq(0)
+      .should('contain', '07');
+
+    timetableVersionDetailsPanel
+      .getRows()
+      .eq(1)
+      .findByTestId('DirectionBadge::inbound')
+      .should('be.visible');
+
+    timetableVersionDetailsPanel.toggleExpandNthRow(1);
+    timetableVersionDetailsPanel.getRows()
+      .eq(1)
+      .findByTestId('VehicleServiceRow::row')
+      .then((rows) => {
+        // hour 07 departures
+        cy.wrap(rows)
+          .eq(0)
+          .within(() => {
+            vehicleServiceRow.getHour().should('contain', '07');
+            vehicleServiceRow.getMinute().eq(0).contains('05');
+          });
+
+        // hour 08 departures
+        cy.wrap(rows)
+          .eq(1)
+          .within(() => {
+            vehicleServiceRow.getHour().should('contain', '08');
+            vehicleServiceRow.getMinute().eq(0).contains('20');
+            vehicleServiceRow.getMinute().eq(1).contains('30');
+          });
+      });
+
+    timetableVersionDetailsPanel.close();
+
+    timetableVersionDetailsPanel.getHeading().should('not.exist');
+  });
+
+  it('Should change the validity period and update all correct validities', () => {
+    timetableVersionsPage.timetableVersionTableRow.openNthRowVersionDetailsPanel(
+      0,
+    );
+
+    timetableVersionDetailsPanel
+      .getHeading()
+      .should('contain', 'Aikataulu voimassa')
+      .and('contain', '1.1.2023 - 31.12.2023');
+
+    timetableVersionDetailsPanel.toggleExpandNthRow(0);
+    timetableVersionDetailsPanel.vehicleJourneyGroupInfo
+      .getChangeValidityButton()
+      .eq(0)
+      .click();
+
+    changeTimetablesValidityForm.setValidityEndDate('2024-03-31');
+    changeTimetablesValidityForm.getSaveButton().click();
+
+    toast.getSuccessToast().should('be.visible');
+
+    // Check that the panel heading's validity changed
+    timetableVersionDetailsPanel.getHeading().contains('1.1.2023 - 31.3.2024');
+
+    timetableVersionDetailsPanel.toggleExpandNthRow(1);
+
+    // Check that both timetable card validity periods changed
+    timetableVersionDetailsPanel.vehicleJourneyGroupInfo
+      .getValidityTimeRange()
+      .eq(0)
+      .contains('1.1.2023 - 31.3.2024');
+    timetableVersionDetailsPanel.vehicleJourneyGroupInfo
+      .getValidityTimeRange()
+      .eq(1)
+      .contains('1.1.2023 - 31.3.2024');
+
+    // Check that the version row's validity changed
+    timetableVersionsPage.timetableVersionTableRow
+      .getRow()
+      .eq(0)
+      .findByTestId('TimetableVersionTableRow::validityEnd')
+      .contains('31.3.2024');
+  });
+});
--- a/cypress/e2e/timetableVersionDetailsPanel.cy.ts
+++ b/cypress/e2e/timetableVersionDetailsPanel.cy.ts
@@ -502,7 +502,7 @@ describe('Timetable version details panel', () => {
 
   it('Should open, have correct details, and close', () => {
     timetableVersionsPage.timetableVersionTableRow
-      .getRow()
+      .getRows()
       .should('have.length', 1);
 
     timetableVersionsPage.timetableVersionTableRow.openNthRowVersionDetailsPanel(
@@ -525,7 +525,8 @@ describe('Timetable version details panel', () => {
     timetableVersionDetailsPanel.getRows().eq(0).should('contain', '99');
 
     timetableVersionDetailsPanel.toggleExpandNthRow(0);
-    timetableVersionDetailsPanel.getRows()
+    timetableVersionDetailsPanel
+      .getRows()
       .eq(0)
       .findByTestId('VehicleServiceRow::row')
       .findByTestId('VehicleServiceRow::hour')
@@ -539,7 +540,8 @@ describe('Timetable version details panel', () => {
       .should('be.visible');
 
     timetableVersionDetailsPanel.toggleExpandNthRow(1);
-    timetableVersionDetailsPanel.getRows()
+    timetableVersionDetailsPanel
+      .getRows()
       .eq(1)
       .findByTestId('VehicleServiceRow::row')
       .then((rows) => {
@@ -604,7 +606,7 @@ describe('Timetable version details panel', () => {
 
     // Check that the version row's validity changed
     timetableVersionsPage.timetableVersionTableRow
-      .getRow()
+      .getRows()
       .eq(0)
       .findByTestId('TimetableVersionTableRow::validityEnd')
       .contains('31.3.2024');
--- a/cypress/pageObjects/TimetableVersionTableRow.ts
+++ b/cypress/pageObjects/TimetableVersionTableRow.ts
@@ -1,10 +1,10 @@
 export class TimetableVersionTableRow {
-  getRow() {
+  getRows() {
     return cy.getByTestId('TimetableVersionTableRow::row');
   }
 
   openNthRowVersionDetailsPanel(rowNumber: number) {
-    this.getRow()
+    this.getRows()
       .eq(rowNumber)
       .findByTestId('TimetableVersionTableRow::actions')
       .click();
--- a/cypress/pageObjects/TimetableVersionTableRow.ts
+++ b/cypress/pageObjects/TimetableVersionTableRow.ts
@@ -2,4 +2,13 @@ export class TimetableVersionTableRow {
   getRow() {
     return cy.getByTestId('TimetableVersionTableRow::row');
   }
+
+  openNthRowVersionDetailsPanel(rowNumber: number) {
+    this.getRow()
+      .eq(rowNumber)
+      .findByTestId('TimetableVersionTableRow::actions')
+      .click();
+
+    cy.getByTestId('TimetableVersionTableRow::versionPanelMenuItem').click();
+  }
 }
--- a/cypress/pageObjects/index.ts
+++ b/cypress/pageObjects/index.ts
@@ -1,3 +1,4 @@
+export * from './ChangeTimetablesValidityForm';
 export * from './ChangeValidityForm';
 export * from './ConfirmationDialog';
 export * from './EditRoutePage';
@@ -20,7 +21,6 @@ export * from './RouteTimetablesSection';
 export * from './RoutesAndLinesPage';
 export * from './SearchResultsPage';
 export * from './StopForm';
-export * from './stop-registry';
 export * from './SubstituteDaySettingsPage';
 export * from './TemplateRouteSelector';
 export * from './TerminusNameInputs';
@@ -29,3 +29,5 @@ export * from './TimetablesMainPage';
 export * from './Toast';
 export * from './VehicleScheduleDetailsPage';
 export * from './VehicleServiceTable';
+export * from './stop-registry';
+export * from './timetables';
--- a/cypress/pageObjects/timetables/index.ts
+++ b/cypress/pageObjects/timetables/index.ts
@@ -1,0 +1,1 @@
+export * from './versions';
--- a/cypress/pageObjects/timetables/versions/VehicleServiceRow.ts
+++ b/cypress/pageObjects/timetables/versions/VehicleServiceRow.ts
@@ -1,0 +1,9 @@
+export class VehicleServiceRow {
+  getHour = () => {
+    return cy.getByTestId('VehicleServiceRow::hour');
+  };
+
+  getMinute = () => {
+    return cy.getByTestId('VehicleServiceRow::minute');
+  };
+}
--- a/cypress/pageObjects/timetables/versions/index.ts
+++ b/cypress/pageObjects/timetables/versions/index.ts
@@ -1,0 +1,2 @@
+export * from './VehicleServiceRow';
+export * from './timetable-version-details-panel';
--- a/cypress/pageObjects/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.ts
+++ b/cypress/pageObjects/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.ts
@@ -1,0 +1,24 @@
+import { VehicleJourneyGroupInfo } from '../../../VehicleJourneyGroupInfo';
+
+export class TimetableVersionDetailsPanel {
+  vehicleJourneyGroupInfo = new VehicleJourneyGroupInfo();
+
+  getHeading = () => {
+    return cy.getByTestId('TimetableVersionPanelHeading::heading');
+  };
+
+  getRows = () => {
+    return cy.getByTestId('ExpandableRouteTimetableRow::row');
+  };
+
+  toggleExpandNthRow = (rowNumber: number) => {
+    this.getRows()
+      .eq(rowNumber)
+      .findByTestId('ExpandableRouteTimetableRow::AccordionButton')
+      .click();
+  };
+
+  close = () => {
+    cy.getByTestId('TimetableVersionPanelHeading::closeButton').click();
+  };
+}
--- a/cypress/pageObjects/timetables/versions/timetable-version-details-panel/index.ts
+++ b/cypress/pageObjects/timetables/versions/timetable-version-details-panel/index.ts
@@ -1,0 +1,1 @@
+export * from './TimetableVersionDetailsPanel';
--- a/ui/src/components/map/modal/Modal.tsx
+++ b/ui/src/components/map/modal/Modal.tsx
@@ -1,5 +1,6 @@
-import { FunctionComponent, useEffect } from 'react';
+import { FunctionComponent } from 'react';
 import { useTranslation } from 'react-i18next';
+import { useCallbackOnKeyEscape } from '../../../hooks';
 import { Row } from '../../../layoutComponents';
 import { SimpleButton } from '../../../uiComponents';
 import { ModalBody } from './ModalBody';
@@ -54,17 +55,7 @@ export const Modal: FunctionComponent<Props> = ({
   onSave,
   children,
 }) => {
-  const closeOnEscape = (e: KeyboardEvent) => {
-    if (e.key === 'Escape') {
-      e.preventDefault();
-      onClose();
-    }
-  };
-
-  useEffect(() => {
-    document.addEventListener('keydown', closeOnEscape, true);
-    return () => document.removeEventListener('keydown', closeOnEscape, true);
-  });
+  useCallbackOnKeyEscape(onClose);
 
   return (
     <div data-testid={testId} className="overflow-auto bg-white">
--- a/ui/src/components/map/stops/CreateStopMarker.tsx
+++ b/ui/src/components/map/stops/CreateStopMarker.tsx
@@ -1,6 +1,7 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { useDispatch } from 'react-redux';
 import { theme } from '../../../generated/theme';
+import { useCallbackOnKeyEscape } from '../../../hooks';
 import { resetEnabledModesAction } from '../../../redux';
 import { Coords } from '../../../types';
 import { Circle } from '../markers';
@@ -34,17 +35,8 @@ export const CreateStopMarker = ({ onCursorMove }: Props): JSX.Element => {
     setMouseCoords(undefined);
   };
 
-  const detectKeyDown = (e: KeyboardEvent) => {
-    if (e.key === 'Escape') {
-      e.preventDefault();
-      dispatch(resetEnabledModesAction());
-    }
-  };
-
-  useEffect(() => {
-    document.addEventListener('keydown', detectKeyDown, true);
-    return () => document.removeEventListener('keydown', detectKeyDown, true);
-  });
+  const resetModes = () => dispatch(resetEnabledModesAction());
+  useCallbackOnKeyEscape(resetModes);
 
   return (
     <div
--- a/ui/src/components/routes-and-lines/line-details/DirectionBadge.tsx
+++ b/ui/src/components/routes-and-lines/line-details/DirectionBadge.tsx
@@ -1,3 +1,4 @@
+import { useTranslation } from 'react-i18next';
 import { RouteDirectionEnum } from '../../../generated/graphql';
 import { mapDirectionToShortUiName } from '../../../i18n/uiNameMappings';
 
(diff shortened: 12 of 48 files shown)


Leitsi (changes requested):
> Reviewed 4 of 4 files at r1, 20 of 20 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: all files reviewed, 15 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/timetables/versions/index.ts` line 4 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA55AyDWinbAUXtHGg:-NrA55AyDWinbAUXtHGh:b-nluoem) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/index.ts#L4)):*
> ```XML
> export * from './TimetableVersionTableRow';
> export * from './TimetableVersionsPage';
> export * from './timetable-version-details-panel/TimetableVersionDetailsPanel';
> ```

Shouldn't this just export the whole folder, `export * from './timetable-version-details-panel;`
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 116 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAEhGA9slc0Hhuxl2B:-NrAEhGA9slc0Hhuxl2C:b-n99v8p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L116)):*
> ```XML
>       <td className={dayTypeClassName}>
>         {parseI18nField(data.dayType.nameI18n)}
> ```

Could just use the `dayType` variable here since it is already parsed above.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 117 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA7F-k1cIiQ9rhqCbx:-NrA7F-l0iPVitvG50B9:blf1u7m) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L117)):*
> ```XML
>       <td className={dayTypeClassName}>
>         {parseI18nField(data.dayType.nameI18n)}
>         {/* TODO: <Visible> component was changed so can't be used here as it is, since it currently
> ```

Think I saw this exact error on some other table. Am in favor of reverting, I don't see why the id couldn't just be added to some separate div wrapper. FYI @niknii-vincit 
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 122 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA6ehfDK7hAQiHvy2Z:-NrA6ehgED4auM2INeD2:b-a5w3q) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L122)):*
> ```XML
>           <IconButton
>             onClick={openVersionPanel}
>             icon={<i className="icon-calendar" />}
> ```

This icon could use some tooltip text.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 145 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAEYHa2g4Yz_rmutQr:-NrAEYHa2g4Yz_rmutQs:b7ppq67) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L145)):*
> ```XML
>               status: statusText,
>               schedule: vehicleScheduleFrameName,
>               dayType: parseI18nField(dayType),
> ```

dayType is already parsed above, `const dayType = parseI18nField(data.dayType.nameI18n);`.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrANN0jDLXaIc-XvX0T:-NrANN0jDLXaIc-XvX0U:b-2vxkns) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L37)):*
> ```XML
>             direction={direction}
>             className="my-5 ml-12"
>             titleName={t(`directionEnum.${direction}`)}
> ```

In every instance where this `DirectionBadge` is used, there's always `titleName={t(`directionEnum.${direction}`)}` Since `direction? is passed as parameter as well, this could/should probably be handled inside the `DirectionBadge` component.
Not an issue of this PR, but could maybe refactor here all the same, or just kick the can down the road.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-NrAOtpm3vQoMxBF-4mo:bhrndg2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*
> ```XML
>           />
>           <div className="ml-3.5 mt-1">
>             <span className="block font-bold">{routeLabel}</span>
> ```

Could be a `h3`?
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/index.ts` line 2 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA5Mbk5v2Ni6SIKkG9:-NrA5Mbk5v2Ni6SIKkGA:b-g9f0qn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/index.ts#L2)):*
> ```XML
> export * from './ExpandableRouteTimetableRow';
> export * from './TimetableHeading';
> ```

`RouteTimetableCard` missing here.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx` line 16 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA5dOhEbR2igeJN0dA:-NrA5dOhEbR2igeJN0dB:bk2ow06) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx#L16)):*
> ```XML
> }
> 
> const getOddRowColor = (key: TimetablePriority) => {
> ```

This seems to be copypasted from `VehicleServiceTable`. Also both of these are only passed as parameter to `VehicleServiceRow`.
I think I would maybe remove this whole getter from this file. It doesn't make much sense since we use a hardcoded priority here. Maybe add `bg-hsl-neutral-blue` as default for the `oddRowColor` parameter in `VehicleServiceRow`.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx` line 43 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAJ2Ix1SeZFk9VKiEQ:-NrAJ2Ix1SeZFk9VKiER:b-c21nu7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx#L43)):*
> ```XML
>       sectionIdentifier={sectionIdentifier}
>     >
>       <div className="space-y-2" id={sectionIdentifier}>
> ```

Top margin missing here, looks about `mt-4`
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableHeading.tsx` line 10 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAKkDuBrsQbF4ZEp8a:-NrAKkDuBrsQbF4ZEp8b:bapcv8h) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableHeading.tsx#L10)):*
> ```XML
> import { getTimetableHeadingBgColor } from '../../vehicle-schedule-details';
> 
> export const TimetableHeading: React.FC<{
> ```

This looks quite like the `timetableHeadingButton` in `VehicleServiceTable` (where it is clickable, though). Could be an opportunity for refactoring.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAGmnpAsZND4iX8WSq:-NrAGmnpAsZND4iX8WSr:bny5kr9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L37)):*
> ```XML
>   }, [dispatch, location.pathname]);
> 
>   // Sort reverse alphabetically, so that 'outbound' (1) gets sorted as first.
> ```

Hard to see from just the code that it actually works that way. A unit test would be nice, _if_ one could be implemented without too much overhead. But doesn't look simple, maybe I'll just take your word for it.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 61 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAG4jOBFIf4EQ6fCCr:-NrAG4jOBFIf4EQ6fCCs:bm1zg67) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L61)):*
> ```XML
>             />
>           );
>         })}
> ```

Can drop the return if you want.



_Suggestion:_
```XML
        {sortedByDirectionRouteTimetableRowInfo?.map((rowInfo) => (
          <RouteTimetableCard
            key={`${rowInfo.label}.${rowInfo.direction}`}
            routeTimetableRowInfo={rowInfo}
            dayTypeNameI18n={dayType?.name_i18n}
            createdAt={createdAt}
          />
        ))}
```
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 18 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMQEyC3G2ramvsRy6:-NrAMQEyC3G2ramvsRy7:b-cdrt2h) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L18)):*
> ```XML
>   <Row className="mb-4 ml-2 justify-between">
>     <div>
>       <span className="block font-bold">{t('timetables.scheduleValid')}</span>
> ```

Isn't span + .block just a div?
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 19 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMH1w-baU0kyAN0ls:-NrAMH1w-baU0kyAN0lt:bpmbg34) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L19)):*
> ```XML
>     <div>
>       <span className="block font-bold">{t('timetables.scheduleValid')}</span>
>       <span className="text-xl font-bold">{`${mapToShortDate(
> ```

Could use some header elements here, for these two?
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 25 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMd8VA_uGa6QXFQ6Y:-NrAMd8VA_uGa6QXFQ6Z:bbsvbiv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L25)):*
> ```XML
>     <CloseIconButton
>       label={t('close')}
>       className=" text-base font-bold text-brand"
> ```

Extra whitespace here, yikes.



_Suggestion:_
```XML
      className="text-base font-bold text-brand"
```
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 169 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAV7Px0NfQpL_1BN74:-NrAV7Px0NfQpL_1BN75:b-iugwsh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L169)):*
> ```XML
> };
> 
> export const useVehicleScheduleFrameSchedules = (
> ```

Lots of stuff here, but no tests that I can see. At least some E2E test that checks this view would be nice?
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 176 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrARNeR4P5sen4IRXEq:-NrARNeS0lBXaeLjm5k_:bdon7wp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L176)):*
> ```XML
>   });
> 
>   const dayType =
> ```

Lots of optional chaining here in this hook. I think it would simplify things a lot if you add an early return for the case when there is no result, so something like:

```
  const frame = vsfResult.data?.timetables?.timetables_vehicle_schedule_vehicle_schedule_frame_by_pk;
  if (!frame) {
    return {
      timetableRowInfo: [],
      dayType: undefined,
      createdAt: undefined,
      validityStart: undefined,
      validityEnd: undefined
    };
  }

// these can't be undefined anynmore :tada: 
const dayType = frame.vehicle_services[0].day_type;
const createdAt = frame.created_at;
const priority = frame.priority;
const validityStart = frame?.validity_start;
const validityEnd = frame.validity_end;
```

This simplification should help with the helper functions here as well.
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 202 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrATASj4SSm-frZFn5C:-NrATASj4SSm-frZFn5D:b7pu320) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L202)):*
> ```XML
>     );
> 
>   const journeysGroupedByLabelAndDirection =
> ```

I'm thinking if this call should be inside `createTimetableRowInfo`, and just pass `allJourneys`as parameter from here. To limit the scope where you need to be aware of `JourneysGroupedByLabelAndDirection`. No biggie though.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 4 of 28 files reviewed, 15 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`ui/src/components/timetables/versions/index.ts` line 4 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA55AyDWinbAUXtHGg:-NrZGf_D8c9hDQfHeIdr:b-oidhef) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/index.ts#L4)):*

Shouldn't this just export the whole folder, `export * from './timetable-version-details-panel;`
</blockquote></details>

yes, good one. I might have moved this in to a folder afterwards and then VS code has autochanged this path and I did not notice it.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 116 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAEhGA9slc0Hhuxl2B:-NrZHDeM8_92b4E8bdX5:b9u7bpw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L116)):*

Could just use the `dayType` variable here since it is already parsed above.
</blockquote></details>

Yes. Also noticed these when we went through some of the existing code with Niki on Thursday. For me these were broken for a while and I probably "fixed" things after the rebase. But yeah, everything were fine in the existing code so I'll just revert tthese
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 117 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA7F-k1cIiQ9rhqCbx:-NrZHcRM7kg9vZd8j0D5:by9oetq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L117)):*

Think I saw this exact error on some other table. Am in favor of reverting, I don't see why the id couldn't just be added to some separate div wrapper. FYI @niknii-vincit
</blockquote></details>

Yep, discussed this with Niki on Thursday.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 122 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA6ehfDK7hAQiHvy2Z:-NrZIU0b6SIsMSM9yFtE:b12vxgf) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L122)):*

This icon could use some tooltip text.
</blockquote></details>

:+1: Added
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 145 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAEYHa2g4Yz_rmutQr:-NrZL5qyCxYjaxHa54bu:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L145)):*

dayType is already parsed above, `const dayType = parseI18nField(data.dayType.nameI18n);`.
</blockquote></details>

:+1: 
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrANN0jDLXaIc-XvX0T:-NrZMkQv2eX__PJ3X3qA:b-cx14ei) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L37)):*

In every instance where this `DirectionBadge` is used, there's always `titleName={t(`directionEnum.${direction}`)}` Since `direction? is passed as parameter as well, this could/should probably be handled inside the `DirectionBadge` component.
Not an issue of this PR, but could maybe refactor here all the same, or just kick the can down the road.
</blockquote></details>

Good one, I'll keep this in mind and will try to remember to add a refactor commit after this.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-NrZMS-qBIbcSndZO037:bw4wt6d) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*

Could be a `h3`?
</blockquote></details>

Hmm, not sure if that is the correct one from e.g. accessibility point of view. This is a list element, and the label is just bolded. I think we should not use heading here. What does @niknii-vincit think?
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/index.ts` line 2 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA5Mbk5v2Ni6SIKkG9:-NrZNLW04mrtL5_xJWWe:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/index.ts#L2)):*

`RouteTimetableCard` missing here.
</blockquote></details>

:+1: 
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx` line 16 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA5dOhEbR2igeJN0dA:-NrZOM_l1dndCt26Lek0:b7sm1f0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx#L16)):*

This seems to be copypasted from `VehicleServiceTable`. Also both of these are only passed as parameter to `VehicleServiceRow`.
I think I would maybe remove this whole getter from this file. It doesn't make much sense since we use a hardcoded priority here. Maybe add `bg-hsl-neutral-blue` as default for the `oddRowColor` parameter in `VehicleServiceRow`.
</blockquote></details>

Good one! Like you said, it's just copy pasted. Now that I look at it, it really doesn't make much sense. I think the VehicleServiceRow code could be implemented with a bit different way, but that is for the future. I'll now put it as default parameter as you said :ok_hand: 
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx` line 43 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAJ2Ix1SeZFk9VKiEQ:-NrZSO0mCJWzDRZLjVgJ:b4putsz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx#L43)):*

Top margin missing here, looks about `mt-4`
</blockquote></details>

:ok_hand: 
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableHeading.tsx` line 10 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAKkDuBrsQbF4ZEp8a:-NrZSRvC5fWlEt4XdXBv:bqkbzno) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableHeading.tsx#L10)):*

This looks quite like the `timetableHeadingButton` in `VehicleServiceTable` (where it is clickable, though). Could be an opportunity for refactoring.
</blockquote></details>

Yep, like mentioned these can probably be combined with the clickability as parameterized. Will look in to this after.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAGmnpAsZND4iX8WSq:-NrZTHkM3VnCXfhijMde:b-ewai10) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L37)):*

Hard to see from just the code that it actually works that way. A unit test would be nice, _if_ one could be implemented without too much overhead. But doesn't look simple, maybe I'll just take your word for it.
</blockquote></details>

Yeah good one, the localeCompare is the alphabet sorting, but actually this comment is not completely correct. **We do** two separate sorting here. 
Added sorting functions to utils. Also the code now speaks for itself so removed the comment completely.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 61 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAG4jOBFIf4EQ6fCCr:-NrZsPtK6sagSUbcqJvM:bx96mgv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L61)):*

Can drop the return if you want.
</blockquote></details>

:+1: removed, more readable also
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 18 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMQEyC3G2ramvsRy6:-NrZtWhY9X4sqkEmyCeT:bd3dya0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L18)):*

Isn't span + .block just a div?
</blockquote></details>

Well I think yes and now. Semantically they are different and probably accessibility-wise we should use span here. The "block" classname is just for the layout, but we still want it to be a span. At least thats how I understand it. What do you think @niknii-vincit 
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 19 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMH1w-baU0kyAN0ls:-NrZtuxW6lBy0A_3O-og:b-kqspln) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L19)):*

Could use some header elements here, for these two?
</blockquote></details>

Yeah, this is the heading of the panel, so these could probably be headings then :D 
What does @niknii-vincit think?
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 25 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMd8VA_uGa6QXFQ6Y:-NrZug_w2LZh507Ey6-Y:bjjukyx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L25)):*

Extra whitespace here, yikes.
</blockquote></details>

noice :+1: fixed
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 169 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAV7Px0NfQpL_1BN74:-NrZulzVDtZftRQC7loF:b5973ga) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L169)):*

Lots of stuff here, but no tests that I can see. At least some E2E test that checks this view would be nice?
</blockquote></details>

Yep, should definitely be tested somehow. I'll check this after other fixes.
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 176 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrARNeR4P5sen4IRXEq:-NrZv9KV0_iAp0v-2MJn:b-jflkao) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L176)):*

Lots of optional chaining here in this hook. I think it would simplify things a lot if you add an early return for the case when there is no result, so something like:

```
  const frame = vsfResult.data?.timetables?.timetables_vehicle_schedule_vehicle_schedule_frame_by_pk;
  if (!frame) {
    return {
      timetableRowInfo: [],
      dayType: undefined,
      createdAt: undefined,
      validityStart: undefined,
      validityEnd: undefined
    };
  }

// these can't be undefined anynmore :tada: 
const dayType = frame.vehicle_services[0].day_type;
const createdAt = frame.created_at;
const priority = frame.priority;
const validityStart = frame?.validity_start;
const validityEnd = frame.validity_end;
```

This simplification should help with the helper functions here as well.
</blockquote></details>

I hated the optionals and knew that something is off, but my brains were already a bit empty to think this through again. 
Big thanks! :tada: This helped a lot and simplified things.
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 202 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrATASj4SSm-frZFn5C:-Nr_-IDv9lhSVQvVA6nE:b-emmrr5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L202)):*

I'm thinking if this call should be inside `createTimetableRowInfo`, and just pass `allJourneys`as parameter from here. To limit the scope where you need to be aware of `JourneysGroupedByLabelAndDirection`. No biggie though.
</blockquote></details>

Good one! On this level it is just noise, as you said. Made the change

niknii-vincit:
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 117 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA7F-k1cIiQ9rhqCbx:-Nr_4zz31WQeR_EoGkF1:b-qdaepd) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L117)):*

Yep, discussed this with Niki on Thursday.
</blockquote></details>

Yup! I reverted the div wrapping and the id's have been changed to the childrens' pre-existing wrapping divs :ok_hand:

niknii-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 17 of 28 files reviewed, 15 unresolved discussions (waiting on @Leitsi)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-Nr_EbheBGoq5Znxot_i:b708ghv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*

Hmm, not sure if that is the correct one from e.g. accessibility point of view. This is a list element, and the label is just bolded. I think we should not use heading here. What does @niknii-vincit think?
</blockquote></details>

I agree that this should be a heading, but `h4`, because the entire section should have a `h3` above (as mentioned in another comment). They are list items sure, but they are the headings for the expandable section below them.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 18 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMQEyC3G2ramvsRy6:-Nr_DpF3-jZcm2MFmN6r:b-3b9qlo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L18)):*

Well I think yes and now. Semantically they are different and probably accessibility-wise we should use span here. The "block" classname is just for the layout, but we still want it to be a span. At least thats how I understand it. What do you think @niknii-vincit
</blockquote></details>

I think this should be a heading. The heading level is defined by the location we open the dialog in, so in this case (if the heading levels in the main page would correctly start from `h1`, instead of `h2`), it would be `h3`. I think the same style can be kept with spans within the `h3` element, but it being a header, makes it more scannable.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 19 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMH1w-baU0kyAN0ls:-Nr_ERfDCH0Xom9XqwAV:b-xuzoc6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L19)):*

Yeah, this is the heading of the panel, so these could probably be headings then :D 
What does @niknii-vincit think?
</blockquote></details>

Same here as above

niknii-vincit:
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-Nr_tpgpDMSeMNfmPyBw:bdzsldp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*

I agree that this should be a heading, but `h4`, because the entire section should have a `h3` above (as mentioned in another comment). They are list items sure, but they are the headings for the expandable section below them.
</blockquote></details>

Consulted Accessibility expert, and they said that the modal should start with a `h2`, because it's basically a new document for screen readers.

niknii-vincit:
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-Nr_ud993h078Fhmhls0:-Nr_ud993h078Fhmhls1:b-rullhj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a467d0457126de369d95ac48815c0959faa1bfd8/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L50)):*
> ```XML
>   return (
>     <Visible visible={isOpen}>
>       <div className="fixed top-0 right-0 mt-20 mr-4 h-[90%] w-1/3 overflow-scroll rounded-md bg-white p-4 shadow-[0_3px_10px_rgb(0,0,0,0.2)] shadow-slate-500">
> ```

This might need the `aria-modal=true` and `role=dialog` attributes. 

Here's something to read on this: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/, which is pretty hard to read for some reason (ironic), anyway there is the codepen example, which was way easier to read and understand from (at least for me)

Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 17 of 28 files reviewed, 7 unresolved discussions (waiting on @HenrikHartiala and @niknii-vincit)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-Nrd9juT0fc8jDEQKQRr:bao76de) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*

Consulted Accessibility expert, and they said that the modal should start with a `h2`, because it's basically a new document for screen readers.
</blockquote></details>

Roger. Could still be some header instead of a span though. Too bad tailwind doesn't have any classes for applying all text styles at once to make it look like a specific header element... Even if its just font size and weight.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx` line 16 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA5dOhEbR2igeJN0dA:-NrdBERR6o0Ew3vBzgq5:bhp2zfd) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx#L16)):*

Good one! Like you said, it's just copy pasted. Now that I look at it, it really doesn't make much sense. I think the VehicleServiceRow code could be implemented with a bit different way, but that is for the future. I'll now put it as default parameter as you said :ok_hand:
</blockquote></details>

Is not a default parameter now though, after all. But is fine like this as well. :heavy_check_mark: 
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 99 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrdD0SbABCBGJjHGrxF:-NrdD0SbABCBGJjHGrxG:b-mrxxsv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a467d0457126de369d95ac48815c0959faa1bfd8/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L99)):*
> ```XML
>     const updatedAcc = { ...acc };
> 
>     if (label && direction) {
> ```

I think this could be simplified as well with an early return. Since label and direction are both required for a route, you could just add an early return if `journey_pattern_route` is not set (can this even happen? oh well). So something like:

```
const updatedAcc = { ...acc };

const route = item.journey_pattern_ref.journey_pattern_instance?.journey_pattern_route
if (!route) { return updatedAcc; }

const label = route.unique_label;
const direction = route.direction;
const nameI18n = route.name_i18n;
```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 13 of 30 files reviewed, 7 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-Nro8CnF1poI1IA5YAlr:baip8o4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*

Roger. Could still be some header instead of a span though. Too bad tailwind doesn't have any classes for applying all text styles at once to make it look like a specific header element... Even if its just font size and weight.
</blockquote></details>

I'll put this as h3 (since the modal heading is h2 and these are childs of that), but still raises some questions, if the heading size for this list item is based on what are the parent headings. What if this component was to be used in somewhere else, where it is under h3? Wouldn't we need two different heading sizes for this at that point? For us this is probably enough, but should we at the given scenario somehow dynamically determine the heading?
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx` line 16 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrA5dOhEbR2igeJN0dA:-Nrnsw5rAXSu9vXVRTEH:b-9pck0q) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/RouteTimetableCard.tsx#L16)):*

Is not a default parameter now though, after all. But is fine like this as well. :heavy_check_mark:
</blockquote></details>

Ah pfft. true, should be fixed now as default parameter in VehicleServiceRow.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-Nr_ud993h078Fhmhls0:-NrntV3uBflXz1GmLk1j:b-tpv8yp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a467d0457126de369d95ac48815c0959faa1bfd8/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L50)):*

This might need the `aria-modal=true` and `role=dialog` attributes. 

Here's something to read on this: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/, which is pretty hard to read for some reason (ironic), anyway there is the codepen example, which was way easier to read and understand from (at least for me)
</blockquote></details>

Hmm. But this is **not** a modal. It's more like a dialog/panel. (Modal = block interaction with the rest of the page)
This is what the link also says: "Tells assistive technologies that the windows underneath the current dialog are not available for interaction (inert)."
But in this case they are available for interaction. 
And by my understanding the role=dialog does not change this fact. 

Correct me if I understood something wrong, but otherwise: 
should we still add "role=dialog" for this and not the aria-modal=true?
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 18 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMQEyC3G2ramvsRy6:-Nro77gI7HZ_3JNNoQ-k:bjzujj7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L18)):*

I think this should be a heading. The heading level is defined by the location we open the dialog in, so in this case (if the heading levels in the main page would correctly start from `h1`, instead of `h2`), it would be `h3`. I think the same style can be kept with spans within the `h3` element, but it being a header, makes it more scannable.
</blockquote></details>

Okay, so I now modified this that the wrapping div is the h2 (this was the suggested one from the other comment: ` Consulted Accessibility expert, and they said that the modal should start with a `h2`, because it's basically a new document for screen readers.`)
And also changed the classNames of the spans to better match figma.
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 19 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMH1w-baU0kyAN0ls:-Nro7abi0vEVUEeVlFVP:bz8d4eh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L19)):*

Same here as above
</blockquote></details>

So now the surrounding div is the h2, and these two remained as spans for different stylings. They both are part of the same heading.
___
*[`ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts` line 99 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrdD0SbABCBGJjHGrxF:-NrnyBW36ULHm5KiEckp:bwfpgvy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a467d0457126de369d95ac48815c0959faa1bfd8/ui/src/hooks/vehicle-schedule-frame/useVehicleScheduleFrameSchedules.ts#L99)):*

I think this could be simplified as well with an early return. Since label and direction are both required for a route, you could just add an early return if `journey_pattern_route` is not set (can this even happen? oh well). So something like:

```
const updatedAcc = { ...acc };

const route = item.journey_pattern_ref.journey_pattern_instance?.journey_pattern_route
if (!route) { return updatedAcc; }

const label = route.unique_label;
const direction = route.direction;
const nameI18n = route.name_i18n;
```
</blockquote></details>

Good one!
I think at some point I've learnt to avoid early returns, but for example here it simplifies the code and readability so much that it is definitely beneficial. Thanks

niknii-vincit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 13 of 30 files reviewed, 6 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-Nrov9z05vx2MNYGItw-:bix5hu8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*

I'll put this as h3 (since the modal heading is h2 and these are childs of that), but still raises some questions, if the heading size for this list item is based on what are the parent headings. What if this component was to be used in somewhere else, where it is under h3? Wouldn't we need two different heading sizes for this at that point? For us this is probably enough, but should we at the given scenario somehow dynamically determine the heading?
</blockquote></details>

Yes! Absolutely `h3` for these child elements, I should have written the `h2` part in the correct modal "title" spot, but alas :shrug:  

I think once this is used elsewhere and if it's heading doesn't match there, it could be parametrised as an attribute perhaps? I don't think we need to solve that issue now "for the future".
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-Nr_ud993h078Fhmhls0:-Nrowe814xORhnP5zu7_:boe6npc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a467d0457126de369d95ac48815c0959faa1bfd8/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L50)):*

Hmm. But this is **not** a modal. It's more like a dialog/panel. (Modal = block interaction with the rest of the page)
This is what the link also says: "Tells assistive technologies that the windows underneath the current dialog are not available for interaction (inert)."
But in this case they are available for interaction. 
And by my understanding the role=dialog does not change this fact. 

Correct me if I understood something wrong, but otherwise: 
should we still add "role=dialog" for this and not the aria-modal=true?
</blockquote></details>

Ah I misunderstood, I somehow thought it's a "lite" modal, because it covers some of the pre-existing UI. Yes the correct way would be to add `role=dialog` here. :ok_hand:  Good job staying vigilant, I'm a little scatterbrained at the moment.

niknii-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 13 of 30 files reviewed, 6 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 18 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMQEyC3G2ramvsRy6:-NroxAiM54tPEbJa_Ypd:b4u9lob) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L18)):*

Okay, so I now modified this that the wrapping div is the h2 (this was the suggested one from the other comment: ` Consulted Accessibility expert, and they said that the modal should start with a `h2`, because it's basically a new document for screen readers.`)
And also changed the classNames of the spans to better match figma.
</blockquote></details>

Looks great! 
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 19 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAMH1w-baU0kyAN0ls:-NroxubDCbYFFQkqAwnQ:beex6gu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L19)):*

So now the surrounding div is the h2, and these two remained as spans for different stylings. They both are part of the same heading.
</blockquote></details>

Imo it looks great :thumbsup:

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 13 of 30 files reviewed, 6 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx` line 40 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAOtpm3vQoMxBF-4mn:-NsDIxzr7RoeMGUUo0SK:b-owz02y) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/ExpandableRouteTimetableRow.tsx#L40)):*

Yes! Absolutely `h3` for these child elements, I should have written the `h2` part in the correct modal "title" spot, but alas :shrug:  

I think once this is used elsewhere and if it's heading doesn't match there, it could be parametrised as an attribute perhaps? I don't think we need to solve that issue now "for the future".
</blockquote></details>

Yep lets go with that :+1: 
___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-Nr_ud993h078Fhmhls0:-NsDJDIT63kJmEA5Ouz0:b-p768g3) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a467d0457126de369d95ac48815c0959faa1bfd8/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.tsx#L50)):*

Ah I misunderstood, I somehow thought it's a "lite" modal, because it covers some of the pre-existing UI. Yes the correct way would be to add `role=dialog` here. :ok_hand:  Good job staying vigilant, I'm a little scatterbrained at the moment.
</blockquote></details>

Added the `role=dialog`

Leitsi (changes requested):
> Reviewed 6 of 20 files at r6.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 18 of 44 files reviewed, 9 unresolved discussions (waiting on @HenrikHartiala and @niknii-vincit)

___
*[`cypress/e2e/timetableVersionDetailsPanel.cy.ts` line 36 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO4YKpDD66cNXn0xmh:-NsO4YKpDD66cNXn0xmi:b-j5wj0f) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/e2e/timetableVersionDetailsPanel.cy.ts#L36)):*
> ```XML
> 
> // TODO: Extract the basedata somewhere, so that it can be used in multiple tests. No need
> // to copy paste 400+ lines in front of every test.
> ```

Yep, :TLDR:.
___
*[`cypress/e2e/timetableVersionDetailsPanel.cy.ts` line 446 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO4gpL1qieAdr4vbhy:-NsO4gpL1qieAdr4vbhz:b7qf4mc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/e2e/timetableVersionDetailsPanel.cy.ts#L446)):*
> ```XML
> 
> describe('Timetable version details panel', () => {
>   // let vehicleScheduleDetailsPage: VehicleScheduleDetailsPage;
> ```

This was forgotten here.
___
*[`cypress/e2e/timetableVersionDetailsPanel.cy.ts` line 501 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO5Aqa2wCiXoM1WKHD:-NsO5Aqa2wCiXoM1WKHE:bwn7lwk) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/e2e/timetableVersionDetailsPanel.cy.ts#L501)):*
> ```XML
> 
>   afterEach(() => {
>     cy.task('emptyDownloadsFolder');
> ```

Are we actually downloading something here though?
___
*[`cypress/e2e/timetableVersionDetailsPanel.cy.ts` line 613 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO7_MV2vMuTJ6LV2xs:-NsO7_MV2vMuTJ6LV2xt:bisrvyx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/e2e/timetableVersionDetailsPanel.cy.ts#L613)):*
> ```XML
>       .getRows()
>       .eq(0)
>       .findByTestId('TimetableVersionTableRow::row::validityEnd')
> ```

Slightly bothers me that the test ids leak to tests. Should we add `findNthRowValidityEnd`?

Though, there's quite a few `findByTestId` calls earlier in this file and that solution wouldn't really work for them...
___
*[`cypress/pageObjects/TimetableVersionTableRow.ts` line 6 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO5MdYDGIoQHwHXn5b:-NsO5MdYDGIoQHwHXn5c:bhx4sfu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/pageObjects/TimetableVersionTableRow.ts#L6)):*
> ```XML
>   }
> 
>   openNthRowVersionDetailsPanel({ rowNumber }: { rowNumber: number }) {
> ```

This would be just fine as `(rowNumber: number)` as well, in my opinion. The `Nth` in function title already tells what that means.
___
*[`cypress/pageObjects/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.ts` line 14 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO4J4h79IT_Ocb279R:-NsO4J4i6QjJCYURAVRl:bwgjegg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/pageObjects/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.ts#L14)):*
> ```XML
>   };
> 
>   expandNthRow = ({ rowNumber }: { rowNumber: number }) => {
> ```

Nit: To be precise, this _toggles_, right? Just happens to expand since it's closed by default.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 28 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO2JcdDB75pjt9QGyP:-NsO2JcdDB75pjt9QGyQ:bk0qe63) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L28)):*
> ```XML
>     'TimetableVersionTableRow::row::validityStart',
>   timetableVersionTableRowValidityEnd:
>     'TimetableVersionTableRow::row::validityEnd',
> ```

Nit: don't think we need the `::row::` on these validity testids here.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 31 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO35R72K9IuiakaUi_:-NsO35R72K9IuiakaUia:b-4hwbwa) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L31)):*
> ```XML
>   timetableVersionTableRowActions: 'TimetableVersionTableRow::actions',
>   deleteTimetableMenuItem: 'TimetableVersionTableRow::deleteTimetableMenuItem',
>   versionPanelMenuItem: 'TimetableVersionTableRow:versionPanelMenuItem',
> ```

Some inconsistency, some test ids use `:` as separator and some `::`. Hardly matters though.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 17 of 44 files reviewed, 8 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`cypress/e2e/timetableVersionDetailsPanel.cy.ts` line 446 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO4gpL1qieAdr4vbhy:-NsOAy0I8x4jZ5-pH9Wz:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/e2e/timetableVersionDetailsPanel.cy.ts#L446)):*

This was forgotten here.
</blockquote></details>

:+1: 
___
*[`cypress/e2e/timetableVersionDetailsPanel.cy.ts` line 501 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO5Aqa2wCiXoM1WKHD:-NsRtXi97I4iY7Qeg-KP:b-d34222) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/e2e/timetableVersionDetailsPanel.cy.ts#L501)):*

Are we actually downloading something here though?
</blockquote></details>

Good catch and nope, probably copypaste leftover. 
___
*[`cypress/e2e/timetableVersionDetailsPanel.cy.ts` line 613 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO7_MV2vMuTJ6LV2xs:-NsS1jWHB_fASshkkbBC:b6hxwyj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/e2e/timetableVersionDetailsPanel.cy.ts#L613)):*

Slightly bothers me that the test ids leak to tests. Should we add `findNthRowValidityEnd`?

Though, there's quite a few `findByTestId` calls earlier in this file and that solution wouldn't really work for them...
</blockquote></details>

Yep, I do think that we need to think of some guidelines for how we create these tests and create/use pageObjects, and because of that I did not want to invest time to these yet. That said, I did make some changes to these. (e.g. in some places the .getRows wasn't used)
Also hmm.. probably should use wihtin here, instead of multiple findBy's.. Well, like said, I'd rather do these refactors in another PR, where I can go through these with more focus on just the pageObjects. I do have some ideas for the pageObjects.
___
*[`cypress/pageObjects/TimetableVersionTableRow.ts` line 6 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO5MdYDGIoQHwHXn5b:-NsRtn0PBINNG93_3Akz:bu21deh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/pageObjects/TimetableVersionTableRow.ts#L6)):*

This would be just fine as `(rowNumber: number)` as well, in my opinion. The `Nth` in function title already tells what that means.
</blockquote></details>

True :+1: 
___
*[`cypress/pageObjects/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.ts` line 14 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO4J4h79IT_Ocb279R:-NsRuobrDqd1BdAB2Di2:bk8akgr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/cypress/pageObjects/timetables/versions/timetable-version-details-panel/TimetableVersionDetailsPanel.ts#L14)):*

Nit: To be precise, this _toggles_, right? Just happens to expand since it's closed by default.
</blockquote></details>

Good one! I'll change this to toggleExpandNthRow probably could also be `toggleNthRow`, but maybe this is more clear.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 28 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO2JcdDB75pjt9QGyP:-NsRvGd-6rFtxDJ7WWyo:bylloie) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L28)):*

Nit: don't think we need the `::row::` on these validity testids here.
</blockquote></details>

Thanks, yeah now that I look at it,  it's clearly redundant.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 31 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsO35R72K9IuiakaUi_:-NsRvjwTCTb--spnmdf-:b-fyeaxv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L31)):*

Some inconsistency, some test ids use `:` as separator and some `::`. Hardly matters though.
</blockquote></details>

We do have a guideline in readme to use double colons, so you are correct. This was a pure negligent error. Quickly skimmed through the changes and I believe this was the only one with single colon.

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-:-NsSTjU10bohnizMof8m:bnfp4nl)* status: 17 of 44 files reviewed, 5 unresolved discussions (waiting on @niknii-vincit)

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 17 of 44 files reviewed, 4 unresolved discussions (waiting on @niknii-vincit)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableHeading.tsx` line 10 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NrAKkDuBrsQbF4ZEp8a:-NsSTzqL5kORDgBIP2sc:b-1h6pa1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90cd3711726f37c80988be0cb0dda8e3e30a6899/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableHeading.tsx#L10)):*

Yep, like mentioned these can probably be combined with the clickability as parameterized. Will look in to this after.
</blockquote></details>

:heavy_check_mark:  let's check this separately from this PR then.

niknii-vincit:
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 24 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsSw9Vw0i7T07zHGWhn:-NsSw9Vw0i7T07zHGWho:b-ypyz36) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L24)):*
> ```XML
>       )} - ${mapToShortDate(validityEnd)}`}</span>
>     </h2>
>     <CloseIconButton
> ```

Because there is a close button here, I think the dialog should be able to be closed with the press of "Esc", so maybe you could use this code from `Modal.tsx`(rows 57-67):



_Code snippet:_
```XML
  const closeOnEscape = (e: KeyboardEvent) => {
    if (e.key === 'Escape') {
      e.preventDefault();
      onClose();
    }
  };

  useEffect(() => {
    document.addEventListener('keydown', closeOnEscape, true);
    return () => document.removeEventListener('keydown', closeOnEscape, true);
  });
```

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 17 of 48 files reviewed, 2 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 24 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsSw9Vw0i7T07zHGWhn:-NsgV3AE9np8rr92CBAr:b-q63gtg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L24)):*

Because there is a close button here, I think the dialog should be able to be closed with the press of "Esc", so maybe you could use this code from `Modal.tsx`(rows 57-67):
</blockquote></details>

Created a hook for this :+1:

niknii-vincit:
*[`ui/src/hooks/useCallbackOnKeyEscape.ts` line 12 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsgYd5w7fwgjBvrso2Y:-NsgYd5w7fwgjBvrso2Z:b-s4ixx6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90d2e18d0102118e6d50551f1116f771c44c8b84/ui/src/hooks/useCallbackOnKeyEscape.ts#L12)):*
> ```XML
>     };
> 
>     document.addEventListener('keydown', onKeyEscape, true);
> ```

Should this be keyup instead? So it doesn't happen INSTANTLY if someone presses the wrong button? At least I have the tendency to keep pressing the button and going elsewhere to stop the button action from happening :sweat_smile:

niknii-vincit (commented):
> Reviewed 2 of 11 files at r3, 9 of 14 files at r4, 1 of 1 files at r5, 17 of 20 files at r6, 2 of 4 files at r7, 2 of 2 files at r8, 4 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 47 of 48 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala and @Leitsi)

___
*[`ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx` line 24 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsSw9Vw0i7T07zHGWhn:-NsgYKg26lhULT5eCUry:bccpe83) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9a8ef330db1953ff20fe34dce69cf86337621a95/ui/src/components/timetables/versions/timetable-version-details-panel/TimetableVersionPanelHeading.tsx#L24)):*

Created a hook for this :+1:
</blockquote></details>

Awesome!

niknii-vincit:
:shipit:

niknii-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769)* status: 47 of 48 files reviewed, all discussions resolved (waiting on @Leitsi)

___
*[`ui/src/hooks/useCallbackOnKeyEscape.ts` line 12 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-NsgYd5w7fwgjBvrso2Y:-Nsg_fh2Efgyio0-NKWy:b-gsb174) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/90d2e18d0102118e6d50551f1116f771c44c8b84/ui/src/hooks/useCallbackOnKeyEscape.ts#L12)):*

Should this be keyup instead? So it doesn't happen INSTANTLY if someone presses the wrong button? At least I have the tendency to keep pressing the button and going elsewhere to stop the button action from happening :sweat_smile:
</blockquote></details>

Let's fix it later on if it's a problem, so :shipit:

niknii-vincit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/769#-:-Nsg_rqh75IGePExuF6Q:bnfp4nl)* status: 47 of 48 files reviewed, all discussions resolved (waiting on @Leitsi)

PR #771, opened 2024-02-23
Add stop searching by label

Leitsi:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771)

### DIFF
--- a/cypress/e2e/stop-registry/stopSearch.cy.ts
+++ b/cypress/e2e/stop-registry/stopSearch.cy.ts
@@ -1,0 +1,216 @@
+import {
+  GetInfrastructureLinksByExternalIdsResult,
+  Priority,
+  StopInsertInput,
+  StopRegistryNameType,
+  StopRegistryStopPlace,
+  buildStop,
+  buildTimingPlace,
+  extractInfrastructureLinkIdsFromResponse,
+  mapToDeleteStopPlaceMutation,
+  mapToGetInfrastructureLinksByExternalIdsQuery,
+} from '@hsl/jore4-test-db-manager';
+import { DateTime } from 'luxon';
+import { Tag } from '../../enums';
+import { StopSearchBar } from '../../pageObjects/stop-registry/StopSearchBar';
+import { StopSearchResultsPage } from '../../pageObjects/stop-registry/StopSearchResultsPage';
+import { UUID } from '../../types';
+import {
+  SupportedResources,
+  insertToDbHelper,
+  removeFromDbHelper,
+} from '../../utils';
+
+// These infralink IDs exist in the 'infraLinks.sql' test data file.
+// These form a straight line on Eerikinkatu in Helsinki.
+// Coordinates are partial since they are needed only for the stop creation.
+
+const testInfraLinks = [
+  {
+    externalId: '445156',
+    coordinates: [24.926699622176628, 60.164181083308065, 10.0969999999943],
+  },
+  {
+    externalId: '442424',
+    coordinates: [24.92904198486008, 60.16490775039894, 0],
+  },
+  {
+    externalId: '442325',
+    coordinates: [24.932072417514647, 60.166003223527824, 0],
+  },
+];
+
+const timingPlaces = [
+  buildTimingPlace('057ebb3f-61bc-46f1-9018-f65257d11efb', '1AACKT'),
+  buildTimingPlace('75f8e23d-4bcf-455a-9a14-262f71b4ea11', '1AURLA'),
+];
+
+const stopPlaceData: Array<Partial<StopRegistryStopPlace>> = [
+  {
+    name: { lang: 'fi_FI', value: 'Puistokaari' },
+  },
+  {
+    name: { lang: 'fi_FI', value: 'Lapinrinne' },
+    alternativeNames: [
+      {
+        name: { lang: 'sv_FI', value: 'Lappbrinken' },
+        nameType: StopRegistryNameType.Translation,
+      },
+    ],
+  },
+  {
+    name: { lang: 'fi_FI', value: 'Tuusulanväylä' },
+  },
+];
+
+const buildScheduledStopPoints = (
+  infrastructureLinkIds: UUID[],
+): StopInsertInput[] => [
+  {
+    ...buildStop({
+      label: 'H1122',
+      located_on_infrastructure_link_id: infrastructureLinkIds[0],
+    }),
+    validity_start: DateTime.fromISO('2020-03-20'),
+    scheduled_stop_point_id: '772a788b-75ab-4bf1-94af-f76182886344',
+    timing_place_id: timingPlaces[0].timing_place_id,
+    priority: Priority.Draft,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[0].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: 'H1234',
+      located_on_infrastructure_link_id: infrastructureLinkIds[1],
+    }),
+    validity_start: DateTime.fromISO('2020-04-20'),
+    validity_end: DateTime.fromISO('2050-05-31'),
+    scheduled_stop_point_id: '92048575-488c-4811-8b8f-de40ee90d9b1',
+    timing_place_id: timingPlaces[1].timing_place_id,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[1].coordinates,
+    },
+  },
+  {
+    ...buildStop({
+      label: 'H2233',
+      located_on_infrastructure_link_id: infrastructureLinkIds[1],
+    }),
+    validity_start: DateTime.fromISO('2020-02-20'),
+    validity_end: DateTime.fromISO('2050-05-31'),
+    scheduled_stop_point_id: '8e7a7175-6bd5-42f1-bdc7-ad0e531beede',
+    timing_place_id: null,
+    measured_location: {
+      type: 'Point',
+      coordinates: testInfraLinks[2].coordinates,
+    },
+  },
+];
+
+describe('Stop search', () => {
+  let stopSearchBar: StopSearchBar;
+  let stopSearchResultsPage: StopSearchResultsPage;
+  const baseDbResources = {
+    timingPlaces,
+  };
+  let dbResources: SupportedResources &
+    Required<Pick<SupportedResources, 'stops'>>;
+  let stopPlaceIds: Array<string>;
+
+  before(() => {
+    cy.task<GetInfrastructureLinksByExternalIdsResult>(
+      'hasuraAPI',
+      mapToGetInfrastructureLinksByExternalIdsQuery(
+        testInfraLinks.map((infralink) => infralink.externalId),
+      ),
+    ).then((res) => {
+      const infraLinkIds = extractInfrastructureLinkIdsFromResponse(res);
+
+      const stops = buildScheduledStopPoints(infraLinkIds);
+      dbResources = {
+        ...baseDbResources,
+        stops,
+      };
+    });
+  });
+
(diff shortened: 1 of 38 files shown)


HenrikHartiala (changes requested):
> Reviewed 2 of 2 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 4 of 4 files at r5, 2 of 2 files at r6, 5 of 5 files at r7, 2 of 2 files at r8, 2 of 2 files at r9, 17 of 17 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771)* status: all files reviewed, 4 unresolved discussions (waiting on @Leitsi)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nr_F4hn7Mo3R9-C8fRT):*
:ok_hand: 
___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nr_FRDd1kapG_K0nT6F):*
also good one :+1: 
___
*[`-- commits` line 15 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nr_GXj62GT7ysHE772E):*
Could have been avoided if the name was something like "navTranslationKey" etc. But yeah it's okay to remove redundant translationKeys.
___
*[`-- commits` line 41 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre0ytk6jli2cPqUnte):*
True dat. Good one!
___
*[`ui/src/components/common/RouteLineTableRow.tsx` line 47 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-NrdyKpUCYLVE2K5PhQS:-NrdyKpUCYLVE2K5PhQT:b3bsi7e) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/dd25dea1898531cb83e724d42f9fde785c439cad/ui/src/components/common/RouteLineTableRow.tsx#L47)):*
> ```XML
> 
> const yBorderClassnames = 'border-y border-y-light-grey';
> const rBorderClassNames = 'border-r border-r-light-grey';
> ```

Not quite part of this PR, but the yBorderClassnames above has different capitalization. The way you wrote it is the correct form, but could you fix the existing for consistency 
___
*[`ui/src/components/stop-registry/main/StopRegistryMainPage.tsx` line 18 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre1y-J6gDCIaEPpKNQ:-Nre1y-J6gDCIaEPpKNR:b-tj1c9g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/37445e696dc2cd32b0209f9c00efc15bade6793d/ui/src/components/stop-registry/main/StopRegistryMainPage.tsx#L18)):*
> ```XML
>         <h1>{t('stops.stops')}</h1>
>         <div className="space-x-4">
>           <SimpleButton
> ```

This could probably benefit from some kind of "not implemented yet" -comment etc. This time there were no problems seeing this, since there is not much code here and I already saw this in the demo you had, but otherwise I might have been confused that what is this.
___
*[`ui/src/components/stop-registry/search/StopSearchBar.tsx` line 71 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre3O3FETWU8C32Vf30:-Nre3O3FETWU8C32Vf31:bwv6a57) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/37445e696dc2cd32b0209f9c00efc15bade6793d/ui/src/components/stop-registry/search/StopSearchBar.tsx#L71)):*
> ```XML
>         </div>
>         <Row className="flex justify-end bg-background py-4">
>           <SimpleButton
> ```

FYI, This doesn't have testId (neither has the existing search-components in that regard), and not even sure if we need it since we do have the same functionality already on the Chevron.
___
*[`ui/src/components/stop-registry/search/StopTableRow.tsx` line 16 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre4wDp2EJF7EzFztUL:-Nre4wDp2EJF7EzFztUM:befleq6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/37445e696dc2cd32b0209f9c00efc15bade6793d/ui/src/components/stop-registry/search/StopTableRow.tsx#L16)):*
> ```XML
> 
>   const yBorderClassnames = 'border-y border-y-light-grey';
> ```

Same comment here with the capitalization. Also, would this benefit from the rBorderClassNames as well? 
___
*[`ui/src/components/stop-registry/stops/stop-details/StopDetailsPage.tsx` line 28 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre-zvtEUIYewY6V9rx:-Nre-zvtEUIYewY6V9ry:b-cno80x) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/81e743fb5bb69b11fd3ca161379694ab56b52ba2/ui/src/components/stop-registry/stops/stop-details/StopDetailsPage.tsx#L28)):*
> ```XML
>         <div
>           className="ml-4"
>           title={t('accessibility:stops.validityPeriod')}
> ```

Out of curiosity, do we have more of these titles that are on "normal elements" and not functionable buttons etc.? Its a good thing that we have, but just curious to know that are these defined somewhere?

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771)* status: all files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/common/RouteLineTableRow.tsx` line 47 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-NrdyKpUCYLVE2K5PhQS:-Nre7R1JC6w_NeBRFwVu:ba1q6s7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/dd25dea1898531cb83e724d42f9fde785c439cad/ui/src/components/common/RouteLineTableRow.tsx#L47)):*

Not quite part of this PR, but the yBorderClassnames above has different capitalization. The way you wrote it is the correct form, but could you fix the existing for consistency
</blockquote></details>

:scream: fixed!
___
*[`ui/src/components/stop-registry/main/StopRegistryMainPage.tsx` line 18 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre1y-J6gDCIaEPpKNQ:-Nre8DQx5T2nGneqJzSQ:b-b3673z) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/37445e696dc2cd32b0209f9c00efc15bade6793d/ui/src/components/stop-registry/main/StopRegistryMainPage.tsx#L18)):*

This could probably benefit from some kind of "not implemented yet" -comment etc. This time there were no problems seeing this, since there is not much code here and I already saw this in the demo you had, but otherwise I might have been confused that what is this.
</blockquote></details>

True. Slapped a TODO here.
___
*[`ui/src/components/stop-registry/search/StopSearchBar.tsx` line 71 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre3O3FETWU8C32Vf30:-Nre8ft28NDIXfDUqmqV:btpnyg4) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/37445e696dc2cd32b0209f9c00efc15bade6793d/ui/src/components/stop-registry/search/StopSearchBar.tsx#L71)):*

FYI, This doesn't have testId (neither has the existing search-components in that regard), and not even sure if we need it since we do have the same functionality already on the Chevron.
</blockquote></details>

It's in the designs so here it stays. Also, I think it's fine that it doesn't have a test-id since it's not actually used in any tests.
___
*[`ui/src/components/stop-registry/search/StopTableRow.tsx` line 16 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre4wDp2EJF7EzFztUL:-Nre92PV8alJo3IcnV0T:bujfkcw) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/37445e696dc2cd32b0209f9c00efc15bade6793d/ui/src/components/stop-registry/search/StopTableRow.tsx#L16)):*

Same comment here with the capitalization. Also, would this benefit from the rBorderClassNames as well?
</blockquote></details>

Fixed.

Thought bout the rborder (lborder here actually) but chose not to add a variable for it here. The case is slightly different here than in the other component. Here the different borders aren't even same color like in the other component, so defining them next to each other doesn't help as much.
___
*[`ui/src/components/stop-registry/stops/stop-details/StopDetailsPage.tsx` line 28 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nre-zvtEUIYewY6V9rx:-NreAU3Y6GhA1oHGAZPY:b15j4ai) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/81e743fb5bb69b11fd3ca161379694ab56b52ba2/ui/src/components/stop-registry/stops/stop-details/StopDetailsPage.tsx#L28)):*

Out of curiosity, do we have more of these titles that are on "normal elements" and not functionable buttons etc.? Its a good thing that we have, but just curious to know that are these defined somewhere?
</blockquote></details>

Not that I know. I'm just adding these to elements I think might need one, when I remember.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771)* status: 34 of 38 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/e2e/stop-registry/stopSearch.cy.ts` line 180 at r22](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nrf57de7HbWl87kBy3W:-NriOf8XBU2mGAE9cnPG:b-w7vyfl) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c58743338225d5d9082c9f36836b54abcc286d10/cypress/e2e/stop-registry/stopSearch.cy.ts#L180)):*

I think we went through this discussion already half a year ago, and I'm not requesting changes due to that, but I still think that we shouldn't listen to newtork requests in E2E tests and we should only use the visuals (loading indicators) for this kind of waiting to simulate the user actions as precisely as possible.
</blockquote></details>

Had forgotten about that. I basically copypasted this from the routes/lines search suite. On one hand I like this since it feels quite robust, even if we are poking code internals.
___
*[`cypress/pageObjects/stop-registry/StopSearchResultsPage.ts` line 15 at r22](https://reviewable.io/reviews/HSLdevcom/jore4-ui/771#-Nrf-_Ni3-vPN2Dqyn6A:-NriOp2-8E7TEpOE9FVa:b-yv26gb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/c58743338225d5d9082c9f36836b54abcc286d10/cypress/pageObjects/stop-registry/StopSearchResultsPage.ts#L15)):*

I think this doesn't work as you have intended it. the cy.get() always starts a new chain (so searches from the whole document) so currently the last `.get('tbody tr')` is the only part that does anything. I need to refresh my memory little bit since haven't checked the e2e tests in a while, but I think .find() might be the one you are looking for, or `.within()`-method which narrows down the scope of `cy.get`to the component.

So probably  either 
`return this.getResultsTable().first().find('tbody tr');`

Or
`return this.getResultsTable().first().within(() => { cy.get(`tbody tr`)})`

But as said, can't remember precisely if these are the correct ones, so you should definitely test that these actually work like intended.


Was also about to comment that it would be nice to have testId instead of `tbody tr`, but then I saw that you already are searching inside the `StopSearchResultList::table` so that part is okay :+1:
</blockquote></details>

Good catch. Was trying to get `within` to work earlier but didn't managed, and this _seemed to_ work. I had tried exactly your second suggestion but that doesn't work correctly. `.find` seems to be the answer here.

PR #786, opened 2024-03-08
Fix e2e occasional substitute day test

HenrikHartiala:
We saved an occasional substitute day "in the future" so that the observation period should change due to that. We made an assertion that the observation period is changed and that we get a toast message for it. But since the test did not explicitly set the observation date and just used the default period, which is 1 year from "today". And since the occasional date was saved as 8.3.2025 this date would be inside the observation period when we hit 8.3.2024. That day was today and the test broke. So by explicitly setting the observation period, this test should always work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/786)

### DIFF
--- a/cypress/e2e/occasionalSubstituteDays.cy.ts
+++ b/cypress/e2e/occasionalSubstituteDays.cy.ts
@@ -453,6 +453,12 @@ describe('Occasional substitute operating periods', () => {
     'Should create and delete an occasional substitute operating period successfully',
     { tags: [Tag.Timetables, Tag.Smoke] },
     () => {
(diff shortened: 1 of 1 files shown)


culka (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/786#-:-NsSfNaz1LTVxro0Ml_S:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #796, opened 2024-03-26
Add e2e tests for viewing stop place basic and location details

Leitsi:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/796)

### DIFF
--- a/cypress/e2e/stop-registry/stopDetails.cy.ts
+++ b/cypress/e2e/stop-registry/stopDetails.cy.ts
@@ -177,4 +177,28 @@ describe('Stop details', () => {
     details.getStopState().should('have.text', 'Pois käytöstä');
     details.getElyNumber().should('have.text', '1234567');
   });
+
+  it('should view location details', { tags: [Tag.StopRegistry] }, () => {
+    stopDetailsPage.visit(dbResources.stops[1].scheduled_stop_point_id);
+    stopDetailsPage.page().should('be.visible');
+
+    const details = stopDetailsPage.locationDetailsViewCard;
+    details.getContainer().should('be.visible');
+    details.getStopAddress().should('have.text', 'Mannerheimintie 22-24');
+    details.getPostalCode().should('have.text', '00100');
+    details.getMunicipality().should('have.text', '-');
+    details.getTariffZone().should('have.text', '-');
+    details.getLatitude().should('have.text', '60.166003223527824');
+    details.getLongitude().should('have.text', '24.932072417514647');
+    details.getAltitude().should('have.text', '0');
+    details.getFunctionalArea().should('have.text', '20 m');
+    details.getStopArea().should('have.text', '-');
+    details.getStopAreaName().should('have.text', '-');
+    details.getStopAreaStops().should('have.text', '-');
+    details.getQuay().should('have.text', '-');
+    details.getStopAreaQuays().should('have.text', '-');
+    details.getTerminal().should('have.text', '-');
+    details.getTerminalName().should('have.text', '-');
+    details.getTerminalStops().should('have.text', '-');
+  });
 });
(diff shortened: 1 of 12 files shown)


HenrikHartiala (approved):
> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 8 of 8 files at r3, 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/796)* status: all files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`ui/src/components/stop-registry/stops/stop-details/LocationDetailsViewCard.tsx` line 10 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/796#-Nu2us3XBF97jZOz2-VG:-Nu2us3XBF97jZOz2-VH:bf6nwp1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9264da2ba70e58638a1b60e41d221bed61650dfd/ui/src/components/stop-registry/stops/stop-details/LocationDetailsViewCard.tsx#L10)):*
> ```XML
> const testIds = {
>   container: 'LocationDetailsViewCard::container',
>   detailId: (fieldName: string) => `LocationDetailsViewCard::${fieldName}`,
> ```

This pattern does reduce repetition and the long testId objects, but the downside to this pattern, is that we can't find these by searching for the whole testId (If a test fails and says that "LocationDetailsViewCard::stopAddress was not found" and you try to find this from the code with global search, you cant find it) . Thats probably not a BIG problem, since we do have the components name imprinted on the testId.

We do use this kind of pattern in places where the testId is dynamically generated, but I was wondering if we should add some mark for the dynamically generated testIds.
For example If we have FooComponent that takes in some dynamic label as well, we could use something like underscore to indicate dynamic part: `FooComponent::_H1234`. So now we would have an indication that "do not try to use the global search with the complete testId".

No need to make any changes here, but if this is something we want to continue with, the we could adjust our guidelines and change all dynamic testIds at once.

Any thoughts on this? We can also continue the discussion in slack, or when we get to take e2e tests on the fixing table next sprint.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/796)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Leitsi)

___
*[`ui/src/components/stop-registry/stops/stop-details/LocationDetailsViewCard.tsx` line 10 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/796#-Nu2us3XBF97jZOz2-VG:-Nu3ASzx4RSA-mZNLH50:b-4l6fx0) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9264da2ba70e58638a1b60e41d221bed61650dfd/ui/src/components/stop-registry/stops/stop-details/LocationDetailsViewCard.tsx#L10)):*

This pattern does reduce repetition and the long testId objects, but the downside to this pattern, is that we can't find these by searching for the whole testId (If a test fails and says that "LocationDetailsViewCard::stopAddress was not found" and you try to find this from the code with global search, you cant find it) . Thats probably not a BIG problem, since we do have the components name imprinted on the testId.

We do use this kind of pattern in places where the testId is dynamically generated, but I was wondering if we should add some mark for the dynamically generated testIds.
For example If we have FooComponent that takes in some dynamic label as well, we could use something like underscore to indicate dynamic part: `FooComponent::_H1234`. So now we would have an indication that "do not try to use the global search with the complete testId".

No need to make any changes here, but if this is something we want to continue with, the we could adjust our guidelines and change all dynamic testIds at once.

Any thoughts on this? We can also continue the discussion in slack, or when we get to take e2e tests on the fixing table next sprint.
</blockquote></details>

Discussed in Slack, removing this and just declaring the test ids in the "traditional" way. This didn't really save that much code to begin with.

HenrikHartiala (approved):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/796)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Leitsi)

PR #800, opened 2024-04-08
Fix inconsistent Hasura metadata by reloading when starting dev/e2e environments

Leitsi:
The metadata sometimes ends up in an inconsistent state, preventing normal behavior of the GraphQL API.
One example of this is when one restarts their computer. The depends_on settings in docker compose file take care of proper start up order when initially creating the containers, but this does not happen when they are restarted after computer reboot. So what happens is, Hasura is started before Tiamat and ends up with inconsistent metadata.

This can be fixed manually by triggering metadata reload from Hasura console. Same thing can be done via the Hasura API.
Similar thing is done on CI, see https://github.com/HSLdevcom/jore4-tools/commit/af15d8f2612c3def63f8097d87eec051613da607 This script tries to automate that for local environments.

For dev and local e2e environments, we now check the metadata state when we _think_ there _could_ be metadata inconsistencies, and then reload if needed.
The reload can still be done explicitly with  a --force flag as well. The dev and local e2e environments live in different containers and ports so they need to be dealt with separately, thus the --e2e flag for toggling this.

Added this check to yarn commands that typically are run first when restarting the dev / local e2e environments, that is, "yarn dev" and eg. "yarn test:e2e:open".

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/800)


culka (commented):
:lgtm: 

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/800#-:-Nv0M8Rv1rB1MRum8uFe:bg7igkv)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @Leitsi)

PR #817, opened 2024-05-02
E2e test renovation

HenrikHartiala:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)


jarkkoka (commented):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/datasets/basedataset.ts` line 28 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtJ__VA_T8xvuSLd5g:-NwtJ__VA_T8xvuSLd5h:bl561rp) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/datasets/basedataset.ts#L28)):*
> ```XML
> const timingPlaces = [
>   buildTimingPlace('f7fd2b8c-380b-48da-b87c-78bfa1690aa3', '1AACKT'),
>   buildTimingPlace('3faa5ec1-aa5c-423e-9064-1523c460299e', '1AURLA'),
> ```

In Hastus exports, the first and last stop point of each journey pattern must be linked to a Hastus place. However, we should also test for cases where an interim stop point (not the first or last) is related to (1) a Hastus place and (2) not related to a Hastus place. That's why I think we need at least three Hastus places in the base dataset.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/datasets/basedataset.ts` line 28 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtJ__VA_T8xvuSLd5g:-NwtKYwsBMZOGIRLY-f5:b-9vmh8l) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/datasets/basedataset.ts#L28)):*

In Hastus exports, the first and last stop point of each journey pattern must be linked to a Hastus place. However, we should also test for cases where an interim stop point (not the first or last) is related to (1) a Hastus place and (2) not related to a Hastus place. That's why I think we need at least three Hastus places in the base dataset.
</blockquote></details>

...to be more specific: there are already such cases in the e2e tests.

Leitsi (commented):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 5 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @jarkkoka)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 7 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtEnUt9TYPM8kqJOD2:-NwtEnUt9TYPM8kqJOD3:b-9uspwc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/e2e/editRouteShape.cy.ts#L7)):*
> ```XML
>   buildTestStopsOnTestInfraLinks,
>   testInfraLinkExternalIds,
> } from '../datasets/basedataset';
> ```

Could just as well be `/datasets/base` I think, or at least eg. `/datasets/base-dataset` 
___
*[`cypress/e2e/editRouteShape.cy.ts` line 42 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtHG4uDS8t4Rlr3kUL:-NwtHG4uDS8t4Rlr3kUM:b3vfnjg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/e2e/editRouteShape.cy.ts#L42)):*
> ```XML
>       testInfraLinkExternalIds,
>     ).then((infraLinkIds) => {
>       const stops = buildTestStopsOnTestInfraLinks(infraLinkIds);
> ```

I think *build~~Test~~StopsOn~~Test~~InfraLinks* would be fine too, no need to lengthen the method names.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 55 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtLBk6B--D_zxVu0t5:-NwtLBk6B--D_zxVu0t6:bgrj245) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/e2e/editRouteShape.cy.ts#L55)):*
> ```XML
> 
>   beforeEach(() => {
>     deleteRoutesByLabel(['Test route 2']);
> ```

Not to do with this PR, but just wondering is this delete really necessary in before :thinking: Since the route is only created in tests and we already delete it in after.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 148 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtIGio86MLtvDKyHXV:-NwtIGio86MLtvDKyHXW:b-b56uax) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/e2e/editRouteShape.cy.ts#L148)):*
> ```XML
>             templateRouteSelectorInfo: {
>               priority: Priority.Standard,
>               label: 'Test route 1',
> ```

With this change it's no longer clear that the created route uses existing line/route data as a base. Possibly not super important for this test case (= test would work as intended even if we changed all the values here, not to match base dataset entities?), but in some others might be. Just to keep in mind.
___
*[`cypress/e2e/utils/tasks.ts` line 67 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nwt9DaE9X48DjE3W9ln:-Nwt9DaE9X48DjE3W9lo:undefined) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/8ce39f8ddd13bc0d5597b08521a333441827a3d7/cypress/e2e/utils/tasks.ts#L67)):*
> ```XML
> /**
>  * Gets infrastucturelink id's that match given external_link_ids
>  * Also returns them in the same order as the external id's were given
> ```





_Suggestion:_
```XML
 * Gets infrastuctureLink ids that match given external_link_ids
 * Also returns them in the same order as the external ids were given
```
___
*[`cypress/e2e/utils/tasks.ts` line 79 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nwt9tvp8-KXFzWRBQ0f:-Nwt9tvp8-KXFzWRBQ0g:b-cc9gcg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/8ce39f8ddd13bc0d5597b08521a333441827a3d7/cypress/e2e/utils/tasks.ts#L79)):*
> ```XML
>         (link) => link.external_link_id === id,
>       ).infrastructure_link_id;
>     });
> ```

I'm slightly worried about what kind of errors this will throw if (when) someone uses this method for external ids that actually don't exist. I'm guessing something like `cannot read properties of undefined`. So, maybe we could try to preemptively clarify those, something like:
```
  ).then((res) => {
    return infrastructureLinkExternalIds.map((id) => {
      const matchingLink = res.data.infrastructure_network_infrastructure_link.find(
        (link) => link.external_link_id === id,
      );
      if (!matchingLink) {
        throw new Error(`No matching infrastructure link found for external id ${id}`);
      }
      return matchingLink.infrastructure_link_id;
    });
  });
```
Would have to test this by actually running the tests to figure out what approach would be good. I'm guessing console logs get thrown away by cypress, but maybe error messages would remain.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/datasets/basedataset.ts` line 128 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtNOgU4z2q_c48EaM3:-NwtNOgU4z2q_c48EaM4:b74kw6v) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/datasets/basedataset.ts#L128)):*
> ```XML
>     on_line_id: lines[0].line_id,
>     validity_start: DateTime.fromISO('2022-08-11T13:08:43.315+03:00'),
>     validity_end: DateTime.fromISO('2032-08-11T13:08:43.315+03:00'),
> ```

The `validity_start` and `validity_end` columns of the `route` table are actually of the `date` type in the database. So the clock times are unused information here.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 5 files reviewed, 6 unresolved discussions (waiting on @HenrikHartiala and @jarkkoka)

___
*[`cypress/datasets/basedataset.ts` line 124 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NwtOPFT4lsSE3Znldpo:-NwxL0saDdcO2rXT0WPv:b-kdsddr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/910461bdb5b1f09319ecb59992317ff3070918cc/cypress/datasets/basedataset.ts#L124)):*

The route label MUST start with the label of the associated line. At least, Hastus import will break with this...

In addition, I'm not sure whether `jore4-hastus` microservice works with this non-standard form.
</blockquote></details>

We don't have such validation in UI/Hasura either, so users could create route with label such as this if they wanted. Should probably add some mechanism to prevent that if it breaks things. Or remove the limitation from Hastus import if that is the only place.

Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 1 of 20 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @jarkkoka)

___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 163 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyUbaGh4HJtcSpHd-mR:-NyUbaGh4HJtcSpHd-mS:b-vjynxe) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/ui/src/components/map/routes/DrawRouteLayer.tsx#L163)):*
> ```XML
>       }
> 
>       if (!infraLinksWithStops) return;
> ```

Isn't `isLoading` left `true` in this early return case?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 1 of 20 files reviewed, 4 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`ui/src/components/map/routes/DrawRouteLayer.tsx` line 163 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyUbaGh4HJtcSpHd-mR:-NyV4KSL158ryDeJMse9:br09od2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/ui/src/components/map/routes/DrawRouteLayer.tsx#L163)):*

Isn't `isLoading` left `true` in this early return case?
</blockquote></details>

Ah, probably yes. Fixed.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 1 of 23 files reviewed, 7 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`cypress/e2e/editRoute.cy.ts` line 96 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV3gbj134FHKieE_ew:-NyZ9FGbCn8HnBHhgt3t:b-spn4au) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/e2e/editRoute.cy.ts#L96)):*

Am I missing something or does this not do anything? Just a getByTestId without any assertions.
</blockquote></details>

The test fails if the element is not found (and if it is not asserted to not be found). So to answer the question, it does assert already for the existence of the element, but its another question if it is redundant, since we get it in the next line anyway. 

Not sure if the possible error is easier to spot if these are done in two steps. But this can be checked in a later iteration as well.
___
*[`cypress/e2e/editRoute.cy.ts` line 165 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV6lRN0ywqPdmmmQ5H:-NyZC5g25gxR2CP3zskv:b-d9ft0z) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/e2e/editRoute.cy.ts#L165)):*

Doesn't this modify the base dataset? So this would affect all suites that are ran after this, since the dataset is not copied for each test (probably should, these errors are hard to debug).
</blockquote></details>

Oh crap, good one! I actually already started resolve this issue during implementation but then something happened and I completely forgot.

Changed this so that the base.ts does not export baseDbResources anymore, but instead there is a getClonedBaseDbResources function which returns a clone.
___
*[`cypress/e2e/editRoute.cy.ts` line 177 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV4Zwl4CK7eHrBo3yf:-NyZwd8z-jcSfReV5Ogv:be08ox5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/e2e/editRoute.cy.ts#L177)):*

This looks like we do db setup twice, since outer describe does one set of inserts and we delete them here?

I would maybe rather structure the describes differently to avoid this. Something like:
```
describe('Route editing', () => {
  describe('basic editing', () => {
    // data setup befores / afters for these tests
  });

  describe('draft route', () => {
    // data setup befores / afters for these tests
  });
});
```
</blockquote></details>

Yeah, good one. I'll add another describe for the basics :+1: 
___
*[`cypress/e2e/hastusExport.cy.ts` line 123 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV9TC-9jC8-p0FoV-b:-NyZf8U-56ORBPecxeNi:bbg1fzi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/e2e/hastusExport.cy.ts#L123)):*

Isn't this the same exact data setup as on the success case describe? Should we just move it to outer describe and remove duplication?
</blockquote></details>

Yeah, Good one!  I'll refactor this as well :+1: 
___
*[`cypress/pageObjects/Navbar.ts` line 6 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV81VA0Bf5IgOXnDep:-NyZvorHDEd_TtuKOw5F:b391frs) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/pageObjects/Navbar.ts#L6)):*

This doesn't seem to be used, just added for completeness I suppose? `Stops` tab is missing as well though.
</blockquote></details>

This was used for a little while, but then refactored stuff and did not need this anymore so it was left unused. I could probably just remove this until used again.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 1 of 23 files reviewed, 7 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`cypress/pageObjects/Navbar.ts` line 6 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV81VA0Bf5IgOXnDep:-Ny_MdmZC-nQ1ZiNiYd5:b2ufd23) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/pageObjects/Navbar.ts#L6)):*

This was used for a little while, but then refactored stuff and did not need this anymore so it was left unused. I could probably just remove this until used again.
</blockquote></details>

WAit a second. it was used? :Dd in hastusExport.ts

I remember that there was a part where I did not use it, but apparently I did in the end :D

Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 1 of 23 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @jarkkoka)

___
*[`cypress/e2e/editRoute.cy.ts` line 96 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV3gbj134FHKieE_ew:-Ny_OLTw5355Ac1di3QD:b67wmgo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/e2e/editRoute.cy.ts#L96)):*

The test fails if the element is not found (and if it is not asserted to not be found). So to answer the question, it does assert already for the existence of the element, but its another question if it is redundant, since we get it in the next line anyway. 

Not sure if the possible error is easier to spot if these are done in two steps. But this can be checked in a later iteration as well.
</blockquote></details>

Slapping a `.should('exist');` or rather a `.should('be.visible');` here might clarify the purpose.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 1 of 23 files reviewed, 4 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`cypress/e2e/editRoute.cy.ts` line 96 at r7](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyV3gbj134FHKieE_ew:-Ny_SIaqAaiaIkudmAuV:b-q58py8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4db7ad5ab1dd466b9bff2bdba97ee41a473185f6/cypress/e2e/editRoute.cy.ts#L96)):*

Slapping a `.should('exist');` or rather a `.should('be.visible');` here might clarify the purpose.
</blockquote></details>

Yeah, thats true. Probably this is missing somewhere else as well in the existing code, but let's add them as we go :+1:

jarkkoka (commented):
> Reviewed 1 of 10 files at r9.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 2 of 23 files reviewed, 7 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/datasets/base.ts` line 20 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_W42eANzzMDEQjV6S:-Ny_W42eANzzMDEQjV6T:b-tpkcuy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L20)):*
> ```XML
> import { UUID } from '../types';
> 
> export const testInfraLinkExternalIds = [
> ```

Great to have these hardcoded IDs in one place as we will need to change all of these when we roll out a newer Digiroad data release! :+1: 
___
*[`cypress/datasets/base.ts` line 208 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_XwtAAzxErXrv8ZBy:-Ny_XwtAAzxErXrv8ZBz:b-y2nagz) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L208)):*
> ```XML
>     journeyPatternId: journeyPatterns[0].journey_pattern_id,
>     stopLabel: 'E2E001',
>     scheduledStopPointSequence: 0,
> ```

An improvement idea: utilise a function to reduce boilerplate by introducing a function.

```ts
 const buildStopsInJourneyPattern = (journeyPatternId: UUID, stopLabels: string[], isUsedAsTimingPoint: boolean[]) => ...
```
___
*[`cypress/datasets/base.ts` line 269 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_YnMlBV1lGIRKX0-e:-Ny_YnMlBV1lGIRKX0-f:bg9wc6) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L269)):*
> ```XML
> 
> export const buildInfraLinksAlongRoute = (
>   infrastructureLinkIds: UUID[],
> ```

An improvement idea: introduce another function to reduce boilerplate.

```ts
 const buildInfraLinksAlongRoute = (routeId: UUID, linksAlongRoute: UUID[], isTraversalForwards: boolean[]) => ...
```

jarkkoka (commented):
> Reviewed 2 of 10 files at r9.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 23 files reviewed, 8 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/package.json` line 31 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_bgx-6K6drEJzZIM-:-Ny_bgx-6K6drEJzZIM0:byea6oa) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/package.json#L31)):*
> ```JSON
>     "eslint-plugin-cypress": "^2.12.1",
>     "find-test-names": "^1.28.3",
>     "@types/lodash": "^4.14.191"
> ```

Hmmm...no alphabetical order in `package.json`. I would like one, if there is no reasonable obstacle to it

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 23 files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 63 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_daF0-1DQyqXSCWnN:-Ny_daF0-1DQyqXSCWnO:bulk4pu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L63)):*
> ```XML
> 
>     map = new Map();
>     routeStopsOverlay = new RouteStopsOverlay();
> ```

To be more precise about the terminology, variable should be renamed:

`routeStopsOverlay` -> `journeyPatternStopsOverlay`
___
*[`cypress/e2e/editRouteShape.cy.ts` line 95 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_cC0KDwKq2IPRUu8u:-Ny_cC0KDwKq2IPRUu8v:b-pqpqwn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L95)):*
> ```XML
>       routeStopsOverlay.routeShouldBeSelected('901');
> 
>       routeStopsOverlay.stopsShouldBeIncludedInRoute([
> ```

Had to look at `base.ts` to see if this is an `outbound` or `inbound` route (since `901` is basically just a line identifier). I would like to mention about it here.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 131 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_cyheBfjo5PgahrG5:-Ny_cyheBfjo5PgahrG6:b-pgo64n) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L131)):*
> ```XML
>       routeStopsOverlay.routeShouldBeSelected('901');
> 
>       routeStopsOverlay.stopsShouldBeIncludedInRoute([
> ```

Add a comment here:

```ts
The stop point labeled `E2E002` is not along the changed route geometry. That's why it must no longer be a part of the journey pattern.
```

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 23 files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 220 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_e2xN6WUy4oA3q026:-Ny_e2xN6WUy4oA3q027:b-lmb8be) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L220)):*
> ```XML
>       routeStopsOverlay.routeShouldBeSelected('902');
> 
>       // Verify that the edited route shape excludes the second stop
> ```

Add a more descriptive comment here:

```ts
// The stop point labeled `E2E002` is not along the changed route geometry. That's why it must no longer be a part of the journey // pattern.
```

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 23 files reviewed, 15 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/hastusExport.cy.ts` line 142 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_gObc1i-fQ-nPmeyr:-Ny_gObc1i-fQ-nPmeys:b-ccymvm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L142)):*
> ```XML
>         );
> 
>         // Set route 901 first stop to not be used as timing point
> ```

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.
___
*[`cypress/e2e/hastusExport.cy.ts` line 174 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_ggdPCLnvfT6Q28_v:-Ny_ggdPCLnvfT6Q28_w:b-ccymvm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L174)):*
> ```XML
>         );
> 
>         // Set route 901 last stop to not be used as timing point
> ```

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.
___
*[`cypress/e2e/hastusExport.cy.ts` line 209 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_gjAU8I5sGBltOPVm:-Ny_gjAU8I5sGBltOPVn:b-ccymvm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L209)):*
> ```XML
>         );
> 
>         // Set route 901 first stop to not be used as timing point
> ```

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.
___
*[`cypress/e2e/hastusExport.cy.ts` line 219 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_gkWk4Cqi9gbvcmdV:-Ny_gkWk4Cqi9gbvcmdW:b-ccymvm) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L219)):*
> ```XML
>         timingSettingsForm.getSavebutton().click();
> 
>         // Set route 901 last stop to not be used as timing point
> ```

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 23 files reviewed, 16 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/routeStops.cy.ts` line 30 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_hHpA3q2cytKO9Gc2:-Ny_hHpB-fWURBNgQZO1:b-9by57i) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/routeStops.cy.ts#L30)):*
> ```XML
> } from '../utils';
> 
> const infrastructureLinkExternalIds = ['445156', '442424', '442325'];
> ```

I'd prefer to index these external ID values from the `testInfraLinkExternalIds` array since these are subject to change in the near future.

I'd like to limit the places where we refer to Digiroad IDs.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 3 of 23 files reviewed, 17 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/routeStops.cy.ts` line 89 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_hopPBR_DNbgMS6gq:-Ny_hopPBR_DNbgMS6gr:btov0p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/routeStops.cy.ts#L89)):*
> ```XML
>     route_id: '61bef596-84a0-40ea-b818-423d6b9b1fcf',
>     on_line_id: lines[0].line_id,
>     validity_start: DateTime.fromISO('2022-08-11T13:08:43.315+03:00'),
> ```

Route validity clock times are not stored in the database. Maybe these could removed as was done in another file?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: 2 of 26 files reviewed, 17 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`cypress/package.json` line 31 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_bgx-6K6drEJzZIM-:-NydUlMKEZMjj7x5yaY_:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/package.json#L31)):*

Hmmm...no alphabetical order in `package.json`. I would like one, if there is no reasonable obstacle to it
</blockquote></details>

:+1: 
___
*[`cypress/e2e/editRouteShape.cy.ts` line 63 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_daF0-1DQyqXSCWnN:-Ny_edpM3z8FkleG2x3W:bljc8qv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L63)):*

To be more precise about the terminology, variable should be renamed:

`routeStopsOverlay` -> `journeyPatternStopsOverlay`
</blockquote></details>

The naming of the PageObjects comes 1:1 from UI component names. So if this is misnamed, then the actual UI component is also misnamed, and thats out from the scope of this PR.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 95 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_cC0KDwKq2IPRUu8u:-Ny_gZqqDVuGkH_VOGqn:b-diypph) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L95)):*

Had to look at `base.ts` to see if this is an `outbound` or `inbound` route (since `901` is basically just a line identifier). I would like to mention about it here.
</blockquote></details>

Yeah, good one. Need to add a check that we are looking for the right route's stops. 

Added a commit which removes the routeShouldBeSelected() method, since it did not check direction AND because it did not practically do anything.
The solution I added might not be the final(as also said in the commit description), but at least it now checks correct things and checks the direction as well.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 131 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_cyheBfjo5PgahrG5:-Ny_g8ZCAHvkKd5y3nSI:b-uyj4zj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L131)):*

Add a comment here:

```ts
The stop point labeled `E2E002` is not along the changed route geometry. That's why it must no longer be a part of the journey pattern.
```
</blockquote></details>

Ah, right. it was missing in this test when changing the geometry. I'll add the same explanation there. (Check the answer below)
___
*[`cypress/e2e/editRouteShape.cy.ts` line 220 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_e2xN6WUy4oA3q026:-Ny_gGheBkoUpE5-hvbU:bmw9jfr) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L220)):*

Add a more descriptive comment here:

```ts
// The stop point labeled `E2E002` is not along the changed route geometry. That's why it must no longer be a part of the journey // pattern.
```
</blockquote></details>

There is already a comment above about this, when we do the change. "Move the two handles so that the second stop is not included in the route" Although based on this I changed the comment to include "E2E002" instead of "second stop". 
___
*[`cypress/e2e/hastusExport.cy.ts` line 142 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_gObc1i-fQ-nPmeyr:-NydT8Fm8vCP-F1a9eeY:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L142)):*

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.
</blockquote></details>

:+1: 
___
*[`cypress/e2e/hastusExport.cy.ts` line 174 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_ggdPCLnvfT6Q28_v:-NydT98IAETxfv3JHHEY:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L174)):*

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.
</blockquote></details>

:+1: 
___
*[`cypress/e2e/hastusExport.cy.ts` line 209 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_gjAU8I5sGBltOPVm:-NydTB9y3B4T2YAYqXrP:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L209)):*

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.
</blockquote></details>

:+1: 
___
*[`cypress/e2e/hastusExport.cy.ts` line 219 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_gkWk4Cqi9gbvcmdV:-NydTC-uAqJSThDogR5X:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/hastusExport.cy.ts#L219)):*

`901` -> `901 (outbound)`

...since there exists two `901`- labeled routes in the base data.
</blockquote></details>

:+1: 
___
*[`cypress/e2e/routeStops.cy.ts` line 30 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_hHpA3q2cytKO9Gc2:-NydTMwQ7aAX5bH7CEJJ:bnmy01u) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/routeStops.cy.ts#L30)):*

I'd prefer to index these external ID values from the `testInfraLinkExternalIds` array since these are subject to change in the near future.

I'd like to limit the places where we refer to Digiroad IDs.
</blockquote></details>

The base dataset has been taken only in hastusExport, editRoute and editRouteShape. routeStops comes later.
___
*[`cypress/e2e/routeStops.cy.ts` line 89 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_hopPBR_DNbgMS6gq:-NydTgH1EtycO_mhb5Ca:benpsw9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/routeStops.cy.ts#L89)):*

Route validity clock times are not stored in the database. Maybe these could removed as was done in another file?
</blockquote></details>

Same comment as above.
___
*[`cypress/datasets/base.ts` line 20 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_W42eANzzMDEQjV6S:-NydTqEy6AQXnhVZVqcF:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L20)):*

Great to have these hardcoded IDs in one place as we will need to change all of these when we roll out a newer Digiroad data release! :+1:
</blockquote></details>

:+1: 
___
*[`cypress/datasets/base.ts` line 208 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_XwtAAzxErXrv8ZBy:-Ny_acprCaApRv8Sgc-L:bicmh1j) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L208)):*

An improvement idea: utilise a function to reduce boilerplate by introducing a function.

```ts
 const buildStopsInJourneyPattern = (journeyPatternId: UUID, stopLabels: string[], isUsedAsTimingPoint: boolean[]) => ...
```
</blockquote></details>

Yeah, something like that could help. Not that exact one, since it would pretty much mask the data to separate arrays, but a function that removes the need to put the stops in order and only need to give the journeyPatternId once. 
I checked and there actually already is one builder function that takes labels and journeyPatternId, but that one is hardcoded to not take in any timingSettings. 

This is a good idea to improve this in a later iteration, but now it would be a minor rabbit hole and also these can be easily refactored later and it won't affect the tests so I'll leave it for now.
___
*[`cypress/datasets/base.ts` line 269 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_YnMlBV1lGIRKX0-e:-Ny_d_wuDdOI5FAKENX_:brg5mh8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L269)):*

An improvement idea: introduce another function to reduce boilerplate.

```ts
 const buildInfraLinksAlongRoute = (routeId: UUID, linksAlongRoute: UUID[], isTraversalForwards: boolean[]) => ...
```
</blockquote></details>

Same thing here good idea to check for a better builder function in a later iteration :+1:

jarkkoka (commented):
> Reviewed 3 of 13 files at r3, 2 of 3 files at r5, 4 of 9 files at r7, 2 of 3 files at r8, 3 of 10 files at r9, 2 of 2 files at r10, 2 of 2 files at r11, 6 of 6 files at r12.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: all files reviewed (commit messages unreviewed), 9 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/editRoute.cy.ts` line 130 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyeLw691wtj5kua_j9-:-NyeLw691wtj5kua_j90:beqhzj1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRoute.cy.ts#L130)):*
> ```XML
> 
>     it('Should delete a route', { tags: Tag.Routes }, () => {
>       lineDetailsPage.visit(baseDbResources.lines[0].line_id);
> ```

IMO the line ID (UUID) should be assigned to a variable since it is accessed at least twice in this code with a moderately long access path: `baseDbResources.lines[0].line_id`
___
*[`cypress/e2e/editRouteShape.cy.ts` line 63 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_daF0-1DQyqXSCWnN:-NyeMvA9ApJalG9G6OOh:byv294j) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L63)):*

The naming of the PageObjects comes 1:1 from UI component names. So if this is misnamed, then the actual UI component is also misnamed, and thats out from the scope of this PR.
</blockquote></details>

Ok.

In cypress test code there are a lot of `stopsInJourneyPattern` variables. In that light, `routeStops` sounds a bit off. Personally, I don't see a problem with giving a variable a name that doesn't match the variable's type if it makes the e2e test code otherwise more consistent (readable).

However, I respect your perspective on this, which feels right since you've been more involved with this codebase.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 95 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_cC0KDwKq2IPRUu8u:-NyeOTJJ0FmfOsNb1OK-:bnyxcia) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L95)):*

Yeah, good one. Need to add a check that we are looking for the right route's stops. 

Added a commit which removes the routeShouldBeSelected() method, since it did not check direction AND because it did not practically do anything.
The solution I added might not be the final(as also said in the commit description), but at least it now checks correct things and checks the direction as well.
</blockquote></details>

Nice. :+1: 
___
*[`cypress/e2e/editRouteShape.cy.ts` line 75 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nyef3Oi8wO43O2jndf-:-Nyef3Oi8wO43O2jndf0:bqj9wsq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRouteShape.cy.ts#L75)):*
> ```XML
> 
>   afterEach(() => {
>     deleteRoutesByLabel(['902']);
> ```

Since this is also done in the `beforeEach` function, I wonder if this is really needed here?
___
*[`cypress/e2e/editRouteShape.cy.ts` line 151 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nyefai34mJJ4iMLx5sC:-Nyefai34mJJ4iMLx5sD:b-gzjz5g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRouteShape.cy.ts#L151)):*
> ```XML
>     () => {
>       // Location where all test stops and routes are visible.
>       const mapLocation = { lng: 24.929689228090112, lat: 60.16495016651525 };
> ```

Nothing needs to be changed, but 10 decimal places (in coordinates) already gives 1 mm accuracy on map, and here are more decimal places. :)

IMO there is no reason to use more decimals unless you want to place a stop point exactly on the center line of the road.
___
*[`cypress/e2e/routeStops.cy.ts` line 30 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_hHpA3q2cytKO9Gc2:-NydZ2nyDEOpmsgwrOra:b2nhll9) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/routeStops.cy.ts#L30)):*

The base dataset has been taken only in hastusExport, editRoute and editRouteShape. routeStops comes later.
</blockquote></details>

Ok.
___
*[`cypress/datasets/base.ts` line 208 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_XwtAAzxErXrv8ZBy:-Nyea6fYAhyceDrk_IOy:b-lb6cr1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L208)):*

Yeah, something like that could help. Not that exact one, since it would pretty much mask the data to separate arrays, but a function that removes the need to put the stops in order and only need to give the journeyPatternId once. 
I checked and there actually already is one builder function that takes labels and journeyPatternId, but that one is hardcoded to not take in any timingSettings. 

This is a good idea to improve this in a later iteration, but now it would be a minor rabbit hole and also these can be easily refactored later and it won't affect the tests so I'll leave it for now.
</blockquote></details>

Ok. :+1: 
___
*[`cypress/datasets/base.ts` line 269 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_YnMlBV1lGIRKX0-e:-NyeaA979Ol_tDCUtWZV:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/datasets/base.ts#L269)):*

Same thing here good idea to check for a better builder function in a later iteration :+1:
</blockquote></details>

:+1:

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: all files reviewed (commit messages unreviewed), 6 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 151 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nyefai34mJJ4iMLx5sC:-NyehWTPBKje8ee3fsLd:bkw1a11) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRouteShape.cy.ts#L151)):*

Nothing needs to be changed, but 10 decimal places (in coordinates) already gives 1 mm accuracy on map, and here are more decimal places. :)

IMO there is no reason to use more decimals unless you want to place a stop point exactly on the center line of the road.
</blockquote></details>

Extreme precision could be useful at two-way stop points.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`cypress/e2e/editRoute.cy.ts` line 130 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-NyeLw691wtj5kua_j9-:-Nysj3oe5RbmkXObtmVl:b-9vwivh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRoute.cy.ts#L130)):*

IMO the line ID (UUID) should be assigned to a variable since it is accessed at least twice in this code with a moderately long access path: `baseDbResources.lines[0].line_id`
</blockquote></details>

In tests we want to have as few steps as possible to find out the used values. If we assign this to another variable, we add another step for the developer trying to find out what values are used in the test. And that will reduce the debuggability of the tests.

When developing tests, the same principles do not apply as what applies for "normal code" and we actually want to reduce the amount of "helper functions" / abstraction levels etc.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 63 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Ny_daF0-1DQyqXSCWnN:-NyslNlDEbL0M7RBO9S4:b-4i8b12) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/a5648b75c3da767078a5986c0c47d49cf4640482/cypress/e2e/editRouteShape.cy.ts#L63)):*

Ok.

In cypress test code there are a lot of `stopsInJourneyPattern` variables. In that light, `routeStops` sounds a bit off. Personally, I don't see a problem with giving a variable a name that doesn't match the variable's type if it makes the e2e test code otherwise more consistent (readable).

However, I respect your perspective on this, which feels right since you've been more involved with this codebase.
</blockquote></details>

I mean I don't argue with the naming being wrong, but the naming comes from the UI and renaming it would require it's own PR. And yeah probably the variable could be changed without changing the class, but that would create a minor inconsistency between the namings of these page object variables. 

Maybe we can come back to this later :+1: 
___
*[`cypress/e2e/editRouteShape.cy.ts` line 75 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nyef3Oi8wO43O2jndf-:-NysmPW53beOY-N56X_N:bmdq4cy) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRouteShape.cy.ts#L75)):*
Tommi already asked the same but pointing out the delete in beforeEach. and I'll copypaste the answer:
> Might have some history on it, practically this makes sure we don't have this already on the database when starting tests. I also know that sometimes there is a stupid possibility that if you change the _currently running test_ on cypress to another, it might not run the "after"- functions and the other test will fail because "duplicate unique key" etc.  
I believe that is why we have these dbResource removals also in the before.

So in beforeEach the reason is described above and could be tested at some point if there was a better solution, but this afterEach is the essential one, since we do not want the route to remain in the database after this test is done. So we want to clean up everything after the test.
___
*[`cypress/e2e/editRouteShape.cy.ts` line 151 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nyefai34mJJ4iMLx5sC:-Nyso-fTEWvwd0D5b5HU:by6itd2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRouteShape.cy.ts#L151)):*

Extreme precision could be useful at two-way stop points.
</blockquote></details>

Yeah that is true. I mean these are practically just copypasted values which should be enough here, but probably at some point we can go through all the coordinates and reduce the precision :+1:

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 75 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-Nyef3Oi8wO43O2jndf-:-Nyss3ji0EgEMCYSwKj8:bnralut) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2c8ef975677e793dc7a6513395d540d9822812b8/cypress/e2e/editRouteShape.cy.ts#L75)):*

Tommi already asked the same but pointing out the delete in beforeEach. and I'll copypaste the answer:
> Might have some history on it, practically this makes sure we don't have this already on the database when starting tests. I also know that sometimes there is a stupid possibility that if you change the _currently running test_ on cypress to another, it might not run the "after"- functions and the other test will fail because "duplicate unique key" etc.  
I believe that is why we have these dbResource removals also in the before.

So in beforeEach the reason is described above and could be tested at some point if there was a better solution, but this afterEach is the essential one, since we do not want the route to remain in the database after this test is done. So we want to clean up everything after the test.
</blockquote></details>

Alright. Makes sense. But in that case, I would consider it a good practice to add a comment to the code that explains it (not in this PR), because then this should be a common practice.

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/817#-:-NytB-rRBjuEQyIrYx69:bnfp4nl)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @HenrikHartiala)

PR #840, opened 2024-06-13
Add observationDate to stopDetails and show highest priority

HenrikHartiala:
To show the highest priority stop details on given observation date we also had to do couple of other changes. Mainly the URL is now `/stops/:label` instead of `/stops/:id` so that we can use the label and observation date to show the correct details. Also updated the stopDetails links to use the label instead of id.

There is also minor changes to StopTitleRow etc. so that it accepts null and undefined values when the observationDate + label doesn't find any results (stop is not valid on chosen date)

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/840)

### DIFF
--- a/cypress/e2e/stop-registry/stopDetails.cy.ts
+++ b/cypress/e2e/stop-registry/stopDetails.cy.ts
@@ -269,7 +269,7 @@ describe('Stop details', () => {
     'should view details for a stop',
     { tags: [Tag.StopRegistry, Tag.Smoke] },
     () => {
-      stopDetailsPage.visit(dbResources.stops[1].scheduled_stop_point_id);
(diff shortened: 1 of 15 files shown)


Perholehto (approved):
Seems like all the changes are made that was discussed at the office :lgtm: 

> Reviewed 12 of 15 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/840#-:-O-K0SjC4rugnN4ngqan:b-qqu9zi)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #849, opened 2024-06-28
Fix timetable version sorting and refactor pageobjects

HenrikHartiala:
Fix a minor bug where timetable versions got sorted randomly if the current sorting clauses ended up being equal. So it was practically just missing this daytype sorting. Now the sorting should be sufficient.

Add some testIds and remove redundant prefixing of some testids.

Refactor pageObjects so that the hierarchy and responsibilities are more clear.


Leitsi (changes requested):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 6 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`-- commits` line 14 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849#-O0UBj9a3-dMgbi07tdP):*
This is true, but it's a shame you can't really do much with the return value since it's not a full page object, so I don't think this is a silver bullet either. But no need to try any other solutions in this PR, just thinking aloud here.
___
*[`cypress/pageObjects/timetables/versions/TimetableVersionTable.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849#-O0U9Ki77rlKjgYXydV9:-O0U9Ki77rlKjgYXydVA:b6ir4tj) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/99e2561571336a05494d3ff73269a7074a1b8a37/cypress/pageObjects/timetables/versions/TimetableVersionTable.ts#L20)):*
> ```XML
>       .click();
> 
>     cy.getByTestId('TimetableVersionTableRow::versionPanelMenuItem').click();
> ```

I'd rather not move this here. We're dealing with the row component internals, so would rather just have this in row page object. Also this currently duplicates some selectors that already exist in the row page component.

Some test already has a structure that I feel is better suited, use `within` to handle accessing correct row:
```
 timetableVersionTable
      .getRows()
      .eq(0)
      .within(() => {
        timetableVersionTableRow.getDayType().shouldHaveText('Lauantai');

        timetableVersionTableRow.getActionsButton().click();
        timetableVersionTableRow.getVersionPanelMenuItemButton().click();
      });
```
With this you don't need the `nth` parameter for the page object. Can still add the `openVersionDetailsPanel()` helper method to the row page object if it feels useful.

Though I don't think this pattern is completely problematic either, since the page object method does not work correctly if not used with `within` (or if there's more than 1 such element), but that fact is not in any way visible from the method. But then, we've used this pattern quite often now so it's familiar for the developers at least.
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 119 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849#-O0U8V8843dY9oZzO6a1:-O0U8V8843dY9oZzO6a2:b-2wteya) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/0d2ef8f300c3e40466532d4f20a17fa97e7df7ec/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L119)):*
> ```XML
>     >
>       <td className={dayTypeClassName} data-testid={testIds.dayType}>
> ```

This element with dayType testid now also contains other texts. Don't see yet if that's a problem though.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`-- commits` line 14 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849#-O0UBj9a3-dMgbi07tdP):*

This is true, but it's a shame you can't really do much with the return value since it's not a full page object, so I don't think this is a silver bullet either. But no need to try any other solutions in this PR, just thinking aloud here.
</blockquote></details>

YEp. I think the page object model get complicated instantly when there is more than one of the same element existing, with a corresponding page object. I've looked though some other patterns as well but all of them seem to have different problems. 
___
*[`cypress/pageObjects/timetables/versions/TimetableVersionTable.ts` line 20 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849#-O0U9Ki77rlKjgYXydV9:-O0UKbUHEHItHjJgfEsu:bz768gg) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/99e2561571336a05494d3ff73269a7074a1b8a37/cypress/pageObjects/timetables/versions/TimetableVersionTable.ts#L20)):*

I'd rather not move this here. We're dealing with the row component internals, so would rather just have this in row page object. Also this currently duplicates some selectors that already exist in the row page component.

Some test already has a structure that I feel is better suited, use `within` to handle accessing correct row:
```
 timetableVersionTable
      .getRows()
      .eq(0)
      .within(() => {
        timetableVersionTableRow.getDayType().shouldHaveText('Lauantai');

        timetableVersionTableRow.getActionsButton().click();
        timetableVersionTableRow.getVersionPanelMenuItemButton().click();
      });
```
With this you don't need the `nth` parameter for the page object. Can still add the `openVersionDetailsPanel()` helper method to the row page object if it feels useful.

Though I don't think this pattern is completely problematic either, since the page object method does not work correctly if not used with `within` (or if there's more than 1 such element), but that fact is not in any way visible from the method. But then, we've used this pattern quite often now so it's familiar for the developers at least.
</blockquote></details>

Oh good one! I just moved this from the Row. I'll refactor this as suggested and as done in the other places.

And yes, I do agree it being the problem about the page objects not working correctly if not used in correct `within` blocks and so on. That has been a pickle pretty much since the start. But the aim was to go towards a guideline in which we create the page objects so that they should always assume they are the only "instance" of that class in the scope used. 
___
*[`ui/src/components/timetables/versions/TimetableVersionTableRow.tsx` line 119 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849#-O0U8V8843dY9oZzO6a1:-O0h9pchCTgmZimg6RAs:bla6kut) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/0d2ef8f300c3e40466532d4f20a17fa97e7df7ec/ui/src/components/timetables/versions/TimetableVersionTableRow.tsx#L119)):*

This element with dayType testid now also contains other texts. Don't see yet if that's a problem though.
</blockquote></details>

The idea is to return the cell here so should not be a problem, but if the naming implies that we only return the text, then not sure if we should include the Cell or datacell etc. in the naming. No changes for now.

Leitsi (approved):
:lgtm:

> Reviewed 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/849#-:-O0hH_9v5H8HOLDivE4U:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #850, opened 2024-06-28
Show different shelter section header when there are no shelters

Leitsi:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/850)

### DIFF
--- a/cypress/e2e/stop-registry/stopDetails.cy.ts
+++ b/cypress/e2e/stop-registry/stopDetails.cy.ts
@@ -837,6 +837,7 @@ describe('Stop details', () => {
         'should be able to delete all shelters',
         { tags: [Tag.StopRegistry] },
         () => {
+          stopDetailsPage.shelters.getAddShelterButton().should('not.exist');
           stopDetailsPage.shelters.getEditButton().click();
           view.getContainers().should('not.exist');
 
@@ -858,6 +859,15 @@ describe('Stop details', () => {
           stopDetailsPage.shelters
             .getTitle()
             .should('have.text', 'Ei pysäkkikatosta');
+
+          // No shelters left: edit mode will be started with one new shelter.
+          stopDetailsPage.shelters.getAddShelterButton().should('be.visible');
+          stopDetailsPage.shelters.getAddShelterButton().click();
+          view.getContainers().should('not.exist');
+          form.getShelters().should('have.length', 1);
+          stopDetailsPage.shelters
+            .getTitle()
+            .should('have.text', 'Pysäkkikatos (1)');
         },
       );
     });
--- a/cypress/pageObjects/stop-registry/stop-details/ShelterSection.ts
+++ b/cypress/pageObjects/stop-registry/stop-details/ShelterSection.ts
@@ -14,6 +14,10 @@ export class SheltersSection {
     return cy.getByTestId('SheltersSection::editButton');
   }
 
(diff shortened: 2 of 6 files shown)


HenrikHartiala (changes requested):
> Reviewed 2 of 2 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/850)* status: all files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`ui/src/components/stop-registry/stops/stop-details/layout/EditableContainerHeaderButtons.tsx` line 25 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/850#-O0hivfy8_cXQ7_fkohD:-O0hivfy8_cXQ7_fkohE:b-b2ts94) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4745fb4f8f95bac8e99fe1175520b128ee6495e3/ui/src/components/stop-registry/stops/stop-details/layout/EditableContainerHeaderButtons.tsx#L25)):*
> ```XML
>   testIdPrefix = '',
> }: Props) => {
>   // When toggling edit mode, if the card is not expanded: expand it.
> ```

This functionality should probably be `<ExpandableInfoContainer>` responsibility instead of the buttons component.
Noticed that there are now toggling here and there and some special rules for the toggles in different situations, in which I think we should/could now use setters instead.

It seems that there is one minor bug as well. If the stop has no shelters and you add a new shelter and save, it collapses the card. Probably due to the toggling chains (?)

So what if we tried to simplify this a bit and go for a logic in which
When pressing edit mode -> always set isExpanded to true.
Pressing the chevron -> setIsExpanded(!isExpanded) (the toggle behaviour)

Maybe also the editMode should/could be a setter instead of toggle.

What do you think? I am also fine if we go with this solution and refactor later.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/850)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`ui/src/components/stop-registry/stops/stop-details/layout/EditableContainerHeaderButtons.tsx` line 25 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/850#-O0hivfy8_cXQ7_fkohD:-O0iCmkbCUxXXaYXOCyG:b-aiche1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4745fb4f8f95bac8e99fe1175520b128ee6495e3/ui/src/components/stop-registry/stops/stop-details/layout/EditableContainerHeaderButtons.tsx#L25)):*
> This functionality should probably be `<ExpandableInfoContainer>` responsibility instead of the buttons component.
Good point, moved.

Changed the toggles on ShelterSection to states and that seemed to simplify things.
Didn't do any bigger changes with edit/expand modes though.

Managed to fix the bug with the special case in ShelterSection with some refactoring.

HenrikHartiala (approved):
> Reviewed 6 of 6 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/850)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Leitsi)

PR #853, opened 2024-07-03
Take e2e dataset in to use in commonSubstituteDays

HenrikHartiala:
- Now testing the schedules of both route directions
- Added pageObject usage instead of cy.getByTestIds calls inside tests
- Added observationDate change to boxing day in the "no traffic" tests, because it was missing completely in the end, so the test did not actually check the correct dates schedules earlier
- Moved testlogic from "before each" in "overlapping dates" to inside the test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/853)

### DIFF
--- a/cypress/e2e/commonSubstituteDays.cy.ts
+++ b/cypress/e2e/commonSubstituteDays.cy.ts
@@ -1,31 +1,21 @@
+import { RouteDirectionEnum } from '@hsl/jore4-test-db-manager';
 import {
-  GetInfrastructureLinksByExternalIdsResult,
-  InfraLinkAlongRouteInsertInput,
-  JourneyPatternInsertInput,
-  LineInsertInput,
-  RouteInsertInput,
-  RouteTypeOfLineEnum,
-  StopInJourneyPatternInsertInput,
-  StopInsertInput,
-  buildLine,
-  buildRoute,
-  buildStop,
-  buildStopInJourneyPattern,
-  buildTimingPlace,
-  extractInfrastructureLinkIdsFromResponse,
-  mapToGetInfrastructureLinksByExternalIdsQuery,
-} from '@hsl/jore4-test-db-manager';
-import { defaultDayTypeIds } from '@hsl/timetables-data-inserter';
-import { DateTime, Duration } from 'luxon';
+  buildInfraLinksAlongRoute,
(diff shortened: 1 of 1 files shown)


Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/853#-:-O0y4MTN0okseYZIiuJE:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

___
*[`cypress/e2e/commonSubstituteDays.cy.ts` line 197 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/853#-O0y3SeK5qW5JSVIbTr0:-O0y3SeK5qW5JSVIbTr1:bxk421i) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/84a578f3148196405bd0a6a0b63df29061579547/cypress/e2e/commonSubstituteDays.cy.ts#L197)):*
> ```XML
>         .getRouteSectionByLabelAndDirection('901', RouteDirectionEnum.Outbound)
>         .within(() => {
>           routeTimetablesSection
> ```

The deleted `// Check that the Boxing day schedule and Sunday timetables are not the same` could've been kept here I think (adjusted for new data of course).

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/853)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @Leitsi)

___
*[`cypress/e2e/commonSubstituteDays.cy.ts` line 197 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/853#-O0y3SeK5qW5JSVIbTr0:-O10g0wb3-lBWa4UFe92:b-ls6fay) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/84a578f3148196405bd0a6a0b63df29061579547/cypress/e2e/commonSubstituteDays.cy.ts#L197)):*

The deleted `// Check that the Boxing day schedule and Sunday timetables are not the same` could've been kept here I think (adjusted for new data of course).
</blockquote></details>

True, added comment but this also made me realise that we could also check that the correct timetable is in effect on boxing day, so added those checks as well. (When the "show all valid" is toggled off

PR #855, opened 2024-07-03
Change cypress intercept config to speed up the tests

HenrikHartiala:
When running cypress on local machine it takes 10-15s per test (.visit() command) to load up before the test actually starts. This seems to be mostly because of cypress was intercepting every single call made. There was a comment about "we should only intercept the hasura calls, but it doesnt work". But at least now it seems to work just fine. And also this sped up the tests majorly which feels amazing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/855)

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -55,11 +55,16 @@ Cypress.Commands.add('mockLogin', () => {
       body: userInfo,
     }).as('userInfo');
 
-    // intercept calls to hasura and log in as admin (to avoid it going to auth backend)
-    // TODO: we should match only '/api/graphql' requests, but for some
-    // reason that doesn't seem to work. (Could be because our graphql
-    // requests use ws:// protocol?)
-    cy.intercept('**', (req) => {
+    // intercept calls to hastus service
+    // and log in as admin (to avoid it going to auth backend)
+    cy.intercept('/api/hastus/**', (req) => {
+      // eslint-disable-next-line no-param-reassign
+      req.headers['x-hasura-admin-secret'] = 'hasura';
+    }).as('hastusAuth');
+
+    // intercept calls to hasura
+    // and log in as admin (to avoid it going to auth backend)
+    cy.intercept('/api/graphql/**', (req) => {
       // eslint-disable-next-line no-param-reassign
       req.headers['x-hasura-admin-secret'] = 'hasura';
     }).as('hasuraAuth');


Leitsi (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/855)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @HenrikHartiala)

___
*[`cypress/support/commands.ts` line 58 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/855#-O0xz2Nu8uk-PPnem774:-O0xz2Nu8uk-PPnem775:btbf696) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/406dfe22f298d306a0324cda7085e00512297031/cypress/support/commands.ts#L58)):*
> ```XML
>     }).as('userInfo');
> 
>     // intercept calls to hastus-converter
> ```

Don't think we've called this hastus-converter anywhere :thinking:  Maybe just "hastus service"?
___
*[`cypress/support/commands.ts` line 63 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/855#-O0xyQco2Zr3p5qheuqu:-O0xyQco2Zr3p5qheuqv:b9mtsxi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/406dfe22f298d306a0324cda7085e00512297031/cypress/support/commands.ts#L63)):*
> ```XML
>       // eslint-disable-next-line no-param-reassign
>       req.headers['x-hasura-admin-secret'] = 'hasura';
>     }).as('hasuraAuth');
> ```

These two intercepts now use the same alias. Maybe rename this hastus one?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/855)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`cypress/support/commands.ts` line 58 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/855#-O0xz2Nu8uk-PPnem774:-O10mOKbEIZ46WaagNAp:b-8pl7oo) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/406dfe22f298d306a0324cda7085e00512297031/cypress/support/commands.ts#L58)):*

Don't think we've called this hastus-converter anywhere :thinking:  Maybe just "hastus service"?
</blockquote></details>

Yeah, I think we at some point decided to change the name to this converter to avoid any misunderstandings (since it is not hastus, even though it is named hastus). But you are right that this change is nowhere yet. I'll change this as suggested.
___
*[`cypress/support/commands.ts` line 63 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/855#-O0xyQco2Zr3p5qheuqu:-O10mtIb64aAdmfGRTz9:b93sach) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/406dfe22f298d306a0324cda7085e00512297031/cypress/support/commands.ts#L63)):*

These two intercepts now use the same alias. Maybe rename this hastus one?
</blockquote></details>

Good one! I missed completely this alias. Changed

PR #860, opened 2024-07-08
Add min value to route properties variant-field

suvikankare:
Hidden variant -field allowed negative numbers which caused an error when saving a route. Added min value 0 to the field.

In addition, add nullablePositiveNumber to zod methods and change it to all nullableNumbers that had min value 0.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860)


HenrikHartiala (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @suvikankare)

___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 105 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860#-O1GENDDEl62x8OJTzQ0:-O1GENDDEl62x8OJTzQ1:b-6q25qt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/013f8e21ac4e6db3035ef29dcbbf5660ebf46fc2/ui/src/components/forms/route/RoutePropertiesForm.tsx#L105)):*
> ```XML
>               testId={testIds.variant}
>               className="col-span-1"
>               min="0"
> ```

NIT, the type is number, so would rather use number instead of string.



_Suggestion:_
```XML
              min={0}
```
___
*[`ui/src/components/forms/route/RoutePropertiesForm.tsx` line 105 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860#-O1GSZg5CpzUi0bEMI-c:-O1GSZg6DpJ4fPzoLKL1:b-jbv41h) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/013f8e21ac4e6db3035ef29dcbbf5660ebf46fc2/ui/src/components/forms/route/RoutePropertiesForm.tsx#L105)):*
> ```XML
>               testId={testIds.variant}
>               className="col-span-1"
>               min="0"
> ```

There is also another thing here. So this change is kind of half the solution: while this prevents negative values when using the input arrows, we can still input negative value by manually typing, because HTML input "min" attribute just works that way :D 

So we still need to prevent the manual input, or in this case we could probably just restrict the form from being submitted if the value is negative and give the user an error. The form validation is done in our zod schemas. I think either way is fine.

For the zod schemas the number and nullable number is already a bit tricky and we are currently using the custom method `nullableNumber` in `src/components/forms/common/customZodSchemas` because otherwise the form will throw an error about null value (which it should not, since it should be nullable).

I also realised that we have this problem already with some other parts, so I'll transfer this discussion to slack ->

HenrikHartiala (changes requested):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860)* status: all files reviewed, 4 unresolved discussions (waiting on @suvikankare)

___
*[`-- commits` line 6 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860#-O1QGbf750uYFMMcjHr5):*
We have rebase flow, which means that each commit will be on main branch as they are, so no merge commit. Because of this flow, we avoid having these fixup commits separately, and they should be squashed to the original commits.

So fixup commits and then interactive rebase to squash those fixup commits to their original commits.

Here the `Refactor min value from string to number` should be squashed with `Add min value to route properties variant-field`
and the last commit `Fix prettier errors` should be squashed in to the original `Add nullablePositiveNumber to zod methods and change it to all nullableNumbers that had min value 0` commit.
___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860#-O1QJbv46xtVWe5ckDwH):*
This commit subject here is a bit lengthy. If you can simplify it a bit, would be nice. All description and extra information can be included in the commit body.

Here is some great guidelines and explanations for the commit messages. Strong suggestion to read:
https://cbea.ms/git-commit/ 

So we should aim for shorter messages in the subject line, and then we can use the commit body as much as needed after that. Also sometimes we can slip from the 72 characters or 80 character limit, but if we mainly keep the subject line within those limits, it should be fine.

HenrikHartiala (approved):
great :ok_hand: 

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/860#-:-O1RP0E_73k-w_-2MfhA:bilkb9y)* status: all files reviewed, 4 unresolved discussions (waiting on @suvikankare)

PR #863, opened 2024-07-09
Fix local timetables-api-e2e database connection

HenrikHartiala:
For some reason (I was unable to figure out why) using the HOSTNAME, DATABASE and PORT environment variables did not work, and timetables api always tried to do imports to dev environment database instead of e2e. By reading the README from timetables api and quick skim through code, those should have done the same thing, but just did not work. But setting the exact database url JORE4_DB_URL works. So for now we can use that and with that our e2e import tests will also work in local environment. (This has no effect to CI)

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/863)

### DIFF
--- a/docker/docker-compose.e2e.yml
+++ b/docker/docker-compose.e2e.yml
@@ -30,7 +30,9 @@ services:
     container_name: "timetablesapi-e2e"
     environment:
       - SKIP_SET_VARIABLE_SECRET_OVERRIDE=true
-      - JORE4_DB_HOSTNAME=jore4-testdb-e2e
+      # Set the URL to e2e database, because for some reason giving the JORE4_DB_HOSTNAME and
+      # JORE4_DB_PORT and JORE4_DB_DATABASE does not work properly and using them points to dev database.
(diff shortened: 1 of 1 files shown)


HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/863)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @Leitsi)

___
*[`docker/docker-compose.e2e.yml` line 33 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/863#-O1MYsdb2b_19KrnEobu:-O1MZfmq2rFuXOIYi9-Y:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/0058f7afe9b5889b1729dc0778387cb72a0dfa29/docker/docker-compose.e2e.yml#L33)):*

Here you mention `JORE4_DB_HOST`  but the actual variable was ` JORE4_DB_HOSTNAME` (apparently).
</blockquote></details>

:+1:

PR #869, opened 2024-07-12
DRAFT: Update UI deps

Huulivoide:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/869)


HenrikHartiala (commented):
> Reviewed 9 of 326 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/869)* status: 9 of 326 files reviewed, 1 unresolved discussion (waiting on @Huulivoide)

___
*[`cypress/e2e/editRouteShape.cy.ts` line 199 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/869#-O3qvVIJ0cebsRQWoH8b:-O3qvVIJ0cebsRQWoH8c:btghanc) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d14e54d852453750f91b71540523db835196e1a9/cypress/e2e/editRouteShape.cy.ts#L199)):*
> ```XML
> 
>       // Wait until the unwanted stops have actually disapeared from the listing.
>       routeStopsOverlay.stopsShouldNotBeIncludedInRoute(['E2E002', 'E2E007']);
> ```

Does the lack of this check cause problems? I was just thinking that we already have the loader checks above. If that is not enough and causes problems, then I would probably research why, because that means that actual user would have the same problem and we need to address that.

HenrikHartiala (approved):
> Reviewed 278 of 326 files at r1, 52 of 52 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/869)* status: all files reviewed, 1 unresolved discussion (waiting on @Huulivoide)

PR #871, opened 2024-07-19
Use dataset in replace and combine e2e

HenrikHartiala:
Commits: 

**Add pageobject ConfirmTimetablesImportModal**
For some reason the confirmTimetableImportForm was inside PreviewTimetablesPage pageObject.
Added the pageObject for the modal and changed all usage to go through this pageObject instead of the
PreviewTimetablesPage pageObject (they are not related).

**Use dataset in timetableReplaceAndCombine**
- Enhanced the tests to now cover more cases.
- Removed all extra visits and extra network waits from tests
- Removed the now unused importfiles
- Removed the now unused Legacy classes

**Add error case e2e tests to timetablesReplaceAndCombine**

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/871)


Leitsi (changes requested):
> Reviewed 6 of 6 files at r1, 7 of 7 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/871)* status: all files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`cypress/e2e/timetablesReplaceAndCombine.cy.ts` line 367 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/871#-O2ONRKUExRSFFAmMnnx:-O2ONRKUExRSFFAmMnny:bk10gpi) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d822d1f3aeb863621bb12cff02cbe61900a11ae3/cypress/e2e/timetablesReplaceAndCombine.cy.ts#L367)):*
> ```XML
>             const { passingTime } = row;
>             // E2E001
>             passingTimesByStopTable.getStopRow('E2E001').within(() => {
> ```

I'm thinking if it would be possible to simplify these expects. This is rather long now and it's quite impossible to tell what is what. Could we maybe just check some subset of this, eg. check that some passing time from old and new timetables exist (and maybe total counts or something). It is not really e2e tests responsibility to verify all the intricacies of the combine logic anyways, these are tested in timetables-api already which does the actual combining.
___
*[`cypress/e2e/timetablesReplaceAndCombine.cy.ts` line 740 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/871#-O2OOuCr4W-8P27HjD9a:-O2OOuCr4W-8P27HjD9b:b-43bnqx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/fb475e712f192d44ce7f783c05a3d619191c05f6/cypress/e2e/timetablesReplaceAndCombine.cy.ts#L740)):*
> ```XML
>       });
> 
>       it('should throw an error when the target timetable has different validity than import', () => {
> ```

I would maybe name these a bit differently, `throw an error` sounds more like backend functionality which maybe isn't the main focus here.

Same with the next test.



_Suggestion:_
```XML
it('should show an error message when the target timetable has different validity than import', () => {
```

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/871#-:-O2PRU_j22lsf_FZDgte:bnfp4nl)* status: 5 of 12 files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #872, opened 2024-07-23
Use dataset in create route

HenrikHartiala:
The first commit was supposed to be in the earlier PR already, but forgot so adding it here.

**Remove temporary test id from VehicleJourneyRow**
After we removed the `getVehicleJourneyRow(routeLabel: string, routeDirection: string)` usage from e2e tests,
we can remove it from the component test and remove it completely and replace it with the data-testidtemp, which
doesn't have any identifying parameters. For context: the route label and direction was not enough to identify one row anyway,
because there could be many rows with the same label and direction. Also we have been refactoring these usages towards
a guideline that we get all rows from a list and then pick the nth row. With this approach we also can be sure that the lists are
in correct order

**Use dataset in createRoute and refactor**
- Replaced dynamic testIds with static ones (dynamic testIds are a bit problematic and e.g. using dynamic testIds in lists
ignore the order of the list completely)
- Added pageObject RouteStopsOverlayRow
- Renamed `addToJourneyPatternButton` testId to `toggleInJourneyPatternButton` since the button has two different functionalities depending
on whether the stop belongs to the journey pattern or not.
- Removed mapModal pageObject's `createRoute` method completely. It supported very narrow usage and we should not hide that much UI test logic
behind very complex and long methods. Also removed `createStopNextToAnotherStop` method as unused and also if in the future we merge the stopMarkers
as the map's `feature`, we can't use the mapMarkers in the tests at all.
- Removed removeStopsFromRoute method from RouteStopsOverlay for many reasons. We should have this functionality in the tests, not in method. Also this
was named `removeStopsFromRoute`, even though it would just as well add a stop to a route if the chosen stop was not used on the route.
- Removed some unused methods from RouteEditor, some of these wouldn't even work anymore since the map update (snapping line & edit handles
are not working the same way) and also
- End note: As iterative approach I did not change/fix everything at once, so there are still things to enhance here, probably some methods might be unused, but
small parts at a time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872)

### DIFF
--- a/cypress/e2e/createRoute.cy.ts
+++ b/cypress/e2e/createRoute.cy.ts
@@ -1,224 +1,48 @@
 import {
-  GetInfrastructureLinksByExternalIdsResult,
-  InfraLinkAlongRouteInsertInput,
-  JourneyPatternInsertInput,
-  LineInsertInput,
   Priority,
   ReusableComponentsVehicleModeEnum,
   RouteDirectionEnum,
-  RouteInsertInput,
-  RouteTypeOfLineEnum,
-  StopInJourneyPatternInsertInput,
-  StopInsertInput,
-  buildLine,
-  buildRoute,
-  buildStop,
-  buildStopInJourneyPattern,
-  buildTimingPlace,
-  extractInfrastructureLinkIdsFromResponse,
-  mapToGetInfrastructureLinksByExternalIdsQuery,
 } from '@hsl/jore4-test-db-manager';
-import { DateTime } from 'luxon';
+import {
+  buildInfraLinksAlongRoute,
+  buildStopsOnInfraLinks,
+  getClonedBaseDbResources,
+  testInfraLinkExternalIds,
+} from '../datasets/base';
 import { Tag } from '../enums';
-import { MapModal, RouteEditor } from '../pageObjects';
+import { MapFooter, MapModal, Toast } from '../pageObjects';
 import { FilterPanel } from '../pageObjects/FilterPanel';
 import { RouteStopsOverlay } from '../pageObjects/RouteStopsOverlay';
 import { UUID } from '../types';
 import { SupportedResources, insertToDbHelper } from '../utils';
 
-const testRouteLabels = {
-  label1: 'T-reitti 1',
-  label2: 'T-reitti 2',
-  label3: 'T-reitti 3',
-  label4: 'Indefinite end time route',
-  label5: 'Template route',
-};
-
-// These infralink IDs exist in the 'infraLinks.sql' test data file.
-// These form a straight line on Eerikinkatu in Helsinki.
-// Coordinates are partial since they are needed only for the stop creation.
-
-const testInfraLinks = [
-  {
-    externalId: '445156',
-    coordinates: [24.926699622176628, 60.164181083308065, 10.0969999999943],
-  },
-  {
-    externalId: '442424',
-    coordinates: [24.92904198486008, 60.16490775039894, 0],
-  },
-  {
-    externalId: '442325',
-    coordinates: [24.932072417514647, 60.166003223527824, 0],
-  },
-];
-
-const stopLabels = ['Test stop 1', 'Test stop 2', 'Test stop 3'];
-
-const lines: LineInsertInput[] = [
-  {
-    ...buildLine({ label: '1 Test line 1' }),
-    line_id: '88f8f9fe-058b-49a2-ac8d-42d13488c7fb',
-    type_of_line: RouteTypeOfLineEnum.StoppingBusService,
-  },
-  {
-    ...buildLine({ label: '1 Test line 2' }),
-    line_id: 'c0e4e702-60b6-4b90-9313-e463814a9422',
-    type_of_line: RouteTypeOfLineEnum.StoppingBusService,
-  },
-  {
-    ...buildLine({ label: '1 Test line 3' }),
-    line_id: '71e19f0a-6eb3-40f2-9818-fb8ea5be135e',
-    type_of_line: RouteTypeOfLineEnum.StoppingBusService,
-  },
-  {
-    ...buildLine({ label: '1 Line with indefinite end time' }),
-    line_id: 'ecbd895b-a720-4211-849f-ca380465c838',
-    type_of_line: RouteTypeOfLineEnum.StoppingBusService,
-  },
-];
-
-const timingPlaces = [
-  buildTimingPlace('5bd61815-4c60-4f10-9a22-df954c516f5f', '1AACKT'),
-  buildTimingPlace('732eec2e-5728-4494-bec4-3a9718dfdde5', '1AURLA'),
-];
-
-const buildStopsOnInfrastrucureLinks = (
-  infrastructureLinkIds: UUID[],
-): StopInsertInput[] => [
-  {
-    ...buildStop({
-      label: stopLabels[0],
-      located_on_infrastructure_link_id: infrastructureLinkIds[0],
-    }),
-    validity_start: DateTime.fromISO('2020-03-20T22:00:00+00:00'),
-    scheduled_stop_point_id: '3f23a4c5-f527-4395-bd9f-bbc398f837df',
-    timing_place_id: timingPlaces[0].timing_place_id,
-    measured_location: {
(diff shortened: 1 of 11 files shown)


Leitsi (changes requested):
> Reviewed 4 of 4 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872)* status: all files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 25 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZQAHz5jI47SbFlccw):*
`and also` ?
___
*[`cypress/e2e/createRoute.cy.ts` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZQtYe81gpWZvhPeZT:-O2ZQtYe81gpWZvhPeZU:b2621fn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L29)):*
> ```XML
>   const baseDbResources = getClonedBaseDbResources();
>   // Location where all test stops and routes are visible.
>   const mapLocation = { lng: 24.929689228090112, lat: 60.16495016651525 };
> ```

Thinking if the `zoom` should be here as well, since it affects on visible area as well, and seems to be the same whenever `mapLocation` is used :thinking: Wnat do you think?
___
*[`cypress/e2e/createRoute.cy.ts` line 112 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZRqAm9Rz8QszhluEP:-O2ZRqAn56WQ7duZ5M21:bi73hii) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L112)):*
> ```XML
>       // Create a geometry for route that includes dataset stops E2E001,
>       // (exclude E2E002) E2E003, E2E004 and E2E005
>       mapModal.map.clickAtPosition(1395, 488);
> ```

How does one get these offsets for the position?

To be fair, the previous approach by clicking with marker test id felt clearer. Now it is quite hidden what is actually happening here, these positions just feel like magic numbers and their connection to the actual markers is in no way visible from test code.

(not that the previous approach was completely unproblematic either, I have no idea why eg. `rightOffset: 35,` and `downOffset: -20,` are what they are and there's no explanation, :shrug: )
___
*[`cypress/e2e/createRoute.cy.ts` line 119 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZTLEeAWn-AkR9LTnr:-O2ZTLEeAWn-AkR9LTns:bavtl3x) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L119)):*
> ```XML
>       mapModal.map.clickAtPosition(1095, 592);
>       mapModal.map.clickAtPosition(991, 657);
>       mapModal.map.clickAtPosition(991, 657);
> ```

Why is this last click twice? Same in all other tests. :thinking: 
___
*[`cypress/e2e/createRoute.cy.ts` line 406 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZTvs55u4SF7ET9Xta:-O2ZTvs55u4SF7ET9Xtb:bgt7e10) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L406)):*
> ```XML
>         .getHeader()
>         .should('contain', '901T')
>         .and('contain', 'Template test route');
> ```

Could we maybe change the name of the of the route? Now it looks like we're checking the template route, and not the one we created based on the template.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872)* status: 4 of 11 files reviewed, 4 unresolved discussions (waiting on @Leitsi)

___
*[`-- commits` line 25 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZQAHz5jI47SbFlccw):*

`and also` ?
</blockquote></details>

No idea, I'll remove this.
___
*[`cypress/e2e/createRoute.cy.ts` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZQtYe81gpWZvhPeZT:-O36UIuO6hrH5qYmP-eH:b-4hz0iu) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L29)):*

Thinking if the `zoom` should be here as well, since it affects on visible area as well, and seems to be the same whenever `mapLocation` is used :thinking: Wnat do you think?
</blockquote></details>

:+1: good one, I'll add it.
___
*[`cypress/e2e/createRoute.cy.ts` line 112 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZRqAm9Rz8QszhluEP:-O36UqZ-BuYBtLlZroPw:b-br98w1) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L112)):*

How does one get these offsets for the position?

To be fair, the previous approach by clicking with marker test id felt clearer. Now it is quite hidden what is actually happening here, these positions just feel like magic numbers and their connection to the actual markers is in no way visible from test code.

(not that the previous approach was completely unproblematic either, I have no idea why eg. `rightOffset: 35,` and `downOffset: -20,` are what they are and there's no explanation, :shrug: )
</blockquote></details>

Yeah, like explained in slack, this is unfortunately the sustainable way to just use viewport coordinates for the clicks. I agree that stopMarker-relative-clicking is a bit more understandable, but if we proceed with our plan to merge the stopMarkers "in to the map" as map features, then clicking relatively with stopMarkers way won't work anymore, but this will. So this is a pre-emptive change. 

The same thing already applied when we updated the map library and we "lost" the snapping line, which is now part of the map. There we already have to use viewport coordinates. But it is pretty effortless to generate the correct coordinates and I will add the instructions to documentation at some point.
___
*[`cypress/e2e/createRoute.cy.ts` line 119 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZTLEeAWn-AkR9LTnr:-O2Zf-lf3hgfZ0sQ7d9G:b-9w44xd) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L119)):*

Why is this last click twice? Same in all other tests. :thinking:
</blockquote></details>

The last node needs to be clicked to finish the route draw. But good question, I will add a comment about this :+1: 
___
*[`cypress/e2e/createRoute.cy.ts` line 406 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-O2ZTvs55u4SF7ET9Xta:-O36WzVa0iTiV-EKsinc:b-9hnu84) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/bb22f4a3128759e63b7fa8eb6933b2ca9bbb78b9/cypress/e2e/createRoute.cy.ts#L406)):*

Could we maybe change the name of the of the route? Now it looks like we're checking the template route, and not the one we created based on the template.
</blockquote></details>

Yeah good point. I changed this to "Based on template test route". Im not sure if we would need these kind of "describing route names" but maybe there is no harm either.

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/872#-:-O36c2PV59pRf_618SaC:bnfp4nl)* status: 4 of 11 files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #875, opened 2024-07-30
Add menu for creating new stop area

Leitsi:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875)


HenrikHartiala (changes requested):
> Reviewed 1 of 1 files at r1, 3 of 3 files at r2, 4 of 4 files at r3, 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875)* status: all files reviewed, 3 unresolved discussions (waiting on @Leitsi)

___
*[`ui/src/components/map/MapFooterActionsDropdown.tsx` line 36 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O335vQQ24oXrhxp88ML:-O335vQQ24oXrhxp88MM:bnj6306) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9cf1d84cfe3ea961a50bebce5c8cf3f130cc6dc3/ui/src/components/map/MapFooterActionsDropdown.tsx#L36)):*
> ```XML
>             data-testid={testIds.menu}
>             as="div"
>             className="flex w-12 justify-around"
> ```

I don't think this precise width is needed (?)
___
*[`ui/src/components/timetables/main/TimetablesMainPage.tsx` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O32qJ4a7KCGNEx1KzOT:-O32qJ4a7KCGNEx1KzOU:b-bqbqy8) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/24077bdea14622f1addb068115ef7499b6c07048/ui/src/components/timetables/main/TimetablesMainPage.tsx#L21)):*
> ```XML
>   return (
>     <Container>
>       <Row>
> ```

I would much rather change the others towards this style.
In my opinion it is better to have the placing responsibility on the parent. 
Also even though using `ml-auto` is a fast way to push everything to the end of the page (even though I don't personally like it at all), I think its not good that we have an element which affects a siblings position.

So hypothetically if we wanted to completely remove the `settingsButton` and you simply delete the code for the settings button -> the ml-auto gets removed and the import button would change place from the right side of the page to the left side of the page.
In the other structure, if you remove `settings button`, the `import button` would stay completely still.

And lastly, with the other style you can add as many buttons there as you like, and they will always have the same space between them, without having to give them any margin whatsoever.
___
*[`ui/src/uiComponents/SimpleDropdownMenu.tsx` line 34 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O32vEaXDDuOyGzGGiMr:-O32vEaXDDuOyGzGGiMs:bqeqeui) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/74265836809cde827acc2ace5b1177c6c72a7d9e/ui/src/uiComponents/SimpleDropdownMenu.tsx#L34)):*
> ```XML
>             <MdMoreVert className="aria-hidden text-3xl" />
>           </Menu.Button>
>           <SimpleDropdownMenuItems open={open} alignItems={alignItems}>
> ```

open sounds a bit like an action rather than boolean, so would use `isOpen`

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875)* status: all files reviewed, 2 unresolved discussions (waiting on @HenrikHartiala)

___
*[`ui/src/components/map/MapFooterActionsDropdown.tsx` line 36 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O335vQQ24oXrhxp88ML:-O36Y6ce73kKogw_Bcln:brrog27) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/9cf1d84cfe3ea961a50bebce5c8cf3f130cc6dc3/ui/src/components/map/MapFooterActionsDropdown.tsx#L36)):*

I don't think this precise width is needed (?)
</blockquote></details>

Hmm, seems to be so. Means this component is actually less buggy than I feared. Will remove and update commit message accordingly. :pray: 
___
*[`ui/src/components/timetables/main/TimetablesMainPage.tsx` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O32qJ4a7KCGNEx1KzOT:-O36a5F64TKuI_5EL5aO:b9sliij) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/24077bdea14622f1addb068115ef7499b6c07048/ui/src/components/timetables/main/TimetablesMainPage.tsx#L21)):*

I would much rather change the others towards this style.
In my opinion it is better to have the placing responsibility on the parent. 
Also even though using `ml-auto` is a fast way to push everything to the end of the page (even though I don't personally like it at all), I think its not good that we have an element which affects a siblings position.

So hypothetically if we wanted to completely remove the `settingsButton` and you simply delete the code for the settings button -> the ml-auto gets removed and the import button would change place from the right side of the page to the left side of the page.
In the other structure, if you remove `settings button`, the `import button` would stay completely still.

And lastly, with the other style you can add as many buttons there as you like, and they will always have the same space between them, without having to give them any margin whatsoever.
</blockquote></details>

True, I agree, didn't think this through. I'll just remove the whole commit, not part of this PR.
___
*[`ui/src/uiComponents/SimpleDropdownMenu.tsx` line 34 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O32vEaXDDuOyGzGGiMr:-O36Z9Oa0YYmig9lPVPH:b-u3z26q) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/74265836809cde827acc2ace5b1177c6c72a7d9e/ui/src/uiComponents/SimpleDropdownMenu.tsx#L34)):*

open sounds a bit like an action rather than boolean, so would use `isOpen`
</blockquote></details>

Thought the same but went with the same naming as in the headlessui component. Will change this to `SimpleDropdownMenuItems` at least :ok_hand:

HenrikHartiala (approved):
> Reviewed 12 of 12 files at r5, 5 of 5 files at r6, 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Leitsi)

___
*[`ui/src/uiComponents/SimpleDropdownMenu.tsx` line 34 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O32vEaXDDuOyGzGGiMr:-O36d7KP2Qovg9SG3RW3:bc5fw0q) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/74265836809cde827acc2ace5b1177c6c72a7d9e/ui/src/uiComponents/SimpleDropdownMenu.tsx#L34)):*

Thought the same but went with the same naming as in the headlessui component. Will change this to `SimpleDropdownMenuItems` at least :ok_hand:
</blockquote></details>

Ah right I understand. Yeah I think it's better if we use the more describing naming as deep as we can. Thanks!
___
*[`ui/src/components/timetables/main/TimetablesMainPage.tsx` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/875#-O32qJ4a7KCGNEx1KzOT:-O36dcvT3Qgp4e0nzRyp:b-ucfknh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/24077bdea14622f1addb068115ef7499b6c07048/ui/src/components/timetables/main/TimetablesMainPage.tsx#L21)):*

True, I agree, didn't think this through. I'll just remove the whole commit, not part of this PR.
</blockquote></details>

Thanks!

PR #877, opened 2024-08-01
Refactor line route table

HenrikHartiala:
**Commits**

**Add className parameter to EditButton and remove margin**
Margin shouldn't be inside the component, it should always come from outside, beause
the margin is not wanted everywhere.

**Refactor StopActionsDropdown responsibility**
The responsibilities were a bit scattered around. So moved the onAddToRoute
and onRemoveFromRoute handlers to StopActionsDropdown component, instead
of passing those handlers through two components. Instead we now just pass
the route through the components, which is in my opinion a bit more clear.
Also this decouples StopActionsDropdown and RouteStopsSection.

**Refactor Line routes from table to list and rename**
- From accessibility point of view things that are not
tables, should not be created by table elements. Also tables add unnecessary complexity to
the structure and readability compared to a simple list and list items.
- Naming was a bit off in the old structure: we had RouteStopsTable (now LineRouteList) then a single row of that table
was RouteStopsSection (now LineRouteListItem). RouteStopsSection had ExpandableRouteRow (now just RouteRow)
component and the RouteStopsRow (now RouteStopListItem).
- The element structure was also semantically off, since the RouteStopsSection returned a <tbody> and not a single <tr>
and we actually populated the same table with two different data: ExpandableRouteRow and RouteStopsRow.
- The list item rows are now structured with CSS grid, which might not be the absolute best here, but has some nice benefits
with smaller screens. But this could also be achieved using percentage widths.
- The white lines and gray lines between list items (border) is now created with a pseudo element to achieve the designed visuals for the
border. This can be changed if we implement the left "colour coded" alerts with some other way. By using different coloured border for
top and left will create a indirect line to the corner of the element and at least I couldn't work my way around it.
- All the tests have been matched with the new names and new testIds, so nothing more has been done there.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)


Leitsi (commented):
> Reviewed 4 of 4 files at r1, 4 of 4 files at r2, 17 of 17 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)* status: all files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`-- commits` line 30 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DN-4GANJCsMnhdIW6):*
> This can be changed if we implement the left "colour coded" alerts with some other way.

I'm thinking if it would make more sense to implement these alerts as pseudo elements, and use borders for borders. But probably not worth the trouble.
___
*[`cypress/pageObjects/index.ts` line 22 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DFrUt7qb9rivyQbqa:-O3DFrUt7qb9rivyQbqb:bwlu0yn) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/cypress/pageObjects/index.ts#L22)):*
> ```XML
> export * from './RoutePropertiesForm';
> export * from './RouteStopsOverlayRow';
> export * from './LineRouteList';
> ```

Not in alphabetical order.
___
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMS444mFo6qdnuJum:-O3DMS444mFo6qdnuJun:b-derspb) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L50)):*
> ```XML
> }
> 
> export const RouteRow: FC<PropsWithChildren<Props>> = ({
> ```

The unused `ExpandableRouteRow` component still exists. Maybe git would've recognized this as renamed if it was deleted.

Same for the page object.
___
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 116 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMFch8OWTApcZwkjL:-O3DMFci6_Twfx35_Bq8:b-5g4tvv) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L116)):*
> ```XML
>         />
>       </div>
>       <span className="col-span-4 text-center" data-testid={testIds.lastEdited}>
> ```

This used to have `tooltip={t('accessibility:lines.versionHistory')}`
___
*[`ui/src/uiComponents/EditButton.tsx` line 30 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DCCAt5SRZ-BTwDT-r:-O3DCCAt5SRZ-BTwDT-s:b722166) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4b3b61f5aea053c5810f7458bd4eeabfa7ad796a/ui/src/uiComponents/EditButton.tsx#L30)):*
> ```XML
> 
> export const EditButton: React.FC<Props> = (props) => {
>   const { testId, tooltip, className = '' } = props;
> ```

Seems like calling this with `ml-5` is a typical use case, so I think I would've used that as a default here and then `twMerge` with `className` parameter. Like we often do elsewhere. That also still fulfills the requirement of adjusting the margin if needed.
___
*[`ui/src/components/routes-and-lines/line-details/LineRouteList.tsx` line 41 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DNpcnCTJYqItWKnUX:-O3DNpco7JlfQENLRYf6:bslwy1p) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/LineRouteList.tsx#L41)):*
> ```XML
>             <LineRouteListItem
>               showUnusedStops={showUnusedStops}
>               key={item.label + item.direction + item.variant}
> ```

I suppose we know that label+direction+variant is unique, yes, but wasn't route_id already doing the same job? It felt safer / more explicit to me.
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopListItem.tsx` line 171 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DJX8KBB-XqSTh_yJD:-O3DJX8KBB-XqSTh_yJE:b-6buxiq) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteStopListItem.tsx#L171)):*
> ```XML
>         scheduledStopPointSequence={scheduledStopPointSequence}
>       />
>       {/* gray line between list items as pseudo element */}
> ```

A `div` isn't really a pseudo element though. Same in RouteRow and commit message. I was initally trying to look for `::before` or `::after` here after just reading the commit message.

I'm thinking if these would be better defined at the start of the element :thinking: Makes no difference since they're absolute. Then it would be closer to other styles of the root element.
___
*[`ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap` line 84 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DDMCH7lNlWV9jz0dB:-O3DDMCH7lNlWV9jz0dC:b-d25wwh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/e5982803cc84089e2a0ff4e3c467033099238c5a/ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap#L84)):*
> ```
>               </span>
>               <a
>                 class="rounded-full ml-5"
> ```

These are in wrong commit I suppose.

niknii-vincit:
*[`ui/src/components/routes-and-lines/line-details/LineRouteList.tsx` line 39 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3HWgb5AGzVdBs7IS4U:-O3HWgb5AGzVdBs7IS4V:b-fr3vvt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2fa9c68cd1d4529305938f55ec7a1fdcb3b45e5e/ui/src/components/routes-and-lines/line-details/LineRouteList.tsx#L39)):*
> ```XML
>         </HuiSwitch.Group>
>       </div>
>       {/* setting a fake "height: 1px" so that "height: 100%" would work for the table cells */}
> ```

Nice. Always good to get away from hacky things like this :fire:

niknii-vincit:
*[`ui/src/components/routes-and-lines/line-details/LineRouteList.tsx` line 41 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DNpcnCTJYqItWKnUX:-O3HYfkz-n462vcbjQRC:bt0pij2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/LineRouteList.tsx#L41)):*

I suppose we know that label+direction+variant is unique, yes, but wasn't route_id already doing the same job? It felt safer / more explicit to me.
</blockquote></details>

I agree with Tommi, especially if the missing variant creates a `null` there. 

I like the idea otherwise though, and might be worth keeping if the element itself did something with it, for example some accessibility labelled by or id with the same semantic "key". 

Isn't the key here used just for React to differentiate different elements from one-another?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)* status: all files reviewed, 5 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`-- commits` line 30 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DN-4GANJCsMnhdIW6):*

> This can be changed if we implement the left "colour coded" alerts with some other way.

I'm thinking if it would make more sense to implement these alerts as pseudo elements, and use borders for borders. But probably not worth the trouble.
</blockquote></details>

Thats actually not a bad idea at all. Because I was already thinking when implementing that it feels bad to create the border between rows with this extra div.. But as I struggled to get the borders to work together on the same element so I ended up with this one. So I can check if I can change this easily.
___
*[`cypress/pageObjects/index.ts` line 22 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DFrUt7qb9rivyQbqa:-O3H625RDumjbIxyAT7Y:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/cypress/pageObjects/index.ts#L22)):*

Not in alphabetical order.
</blockquote></details>

:+1: 
___
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 50 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMS444mFo6qdnuJum:-O3Gxx9T-wRVg7VSK857:bbgqk8v) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L50)):*

The unused `ExpandableRouteRow` component still exists. Maybe git would've recognized this as renamed if it was deleted.

Same for the page object.
</blockquote></details>

I deleted and added them back a couple of times when I was checking stuff and it seems that I missed this the last time, so good spot!
___
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 116 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMFch8OWTApcZwkjL:-O3GyHxlAEhwmp_fiqlu:b-od1dm5) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L116)):*

This used to have `tooltip={t('accessibility:lines.versionHistory')}`
</blockquote></details>

Yeah, good one, forgot about this one. So the reason why I removed it was, because I doubted it had the correct text, since this is the *lastEdited* field. @niknii-vincit Do you know more about this?
___
*[`ui/src/uiComponents/EditButton.tsx` line 30 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DCCAt5SRZ-BTwDT-r:-O3Gzctv1qT9w0e2cnl5:bsfkl91) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/4b3b61f5aea053c5810f7458bd4eeabfa7ad796a/ui/src/uiComponents/EditButton.tsx#L30)):*

Seems like calling this with `ml-5` is a typical use case, so I think I would've used that as a default here and then `twMerge` with `className` parameter. Like we often do elsewhere. That also still fulfills the requirement of adjusting the margin if needed.
</blockquote></details>

Well yeah, true, although now I need to use "ml-0" if I don't want the margin. I probably personally would still give the margin where it is needed instead of a default, even though it is a typical use case. I still think that margins should not be given inside the component itself, but I can live with this solution as well. 
___
*[`ui/src/components/routes-and-lines/line-details/LineRouteList.tsx` line 41 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DNpcnCTJYqItWKnUX:-O3GxXPZBpqC8ntukBZD:bsw7hrx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/LineRouteList.tsx#L41)):*

I suppose we know that label+direction+variant is unique, yes, but wasn't route_id already doing the same job? It felt safer / more explicit to me.
</blockquote></details>

Yes of course, good spot! I don't know why I've changed this. Probably some brainfart.
___
*[`ui/src/components/routes-and-lines/line-details/RouteStopListItem.tsx` line 171 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DJX8KBB-XqSTh_yJD:-O3H6JMQCqO-07wiSaXr:bw1g3mh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteStopListItem.tsx#L171)):*

A `div` isn't really a pseudo element though. Same in RouteRow and commit message. I was initally trying to look for `::before` or `::after` here after just reading the commit message.

I'm thinking if these would be better defined at the start of the element :thinking: Makes no difference since they're absolute. Then it would be closer to other styles of the root element.
</blockquote></details>

Right yeah, had the wrong word for it. I can put this on top (where it was at first, but then moved it here) But I'll first check if I can actually use the borders for this and implement the alerts with another solution, if that doesn't create too much of a hassle. It would make much more sense.
___
*[`ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap` line 84 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DDMCH7lNlWV9jz0dB:-O3H-QJv61uibeFFqVGZ:b21gy77) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/e5982803cc84089e2a0ff4e3c467033099238c5a/ui/src/components/routes-and-lines/line-details/__snapshots__/RouteStopsTable.spec.tsx.snap#L84)):*

These are in wrong commit I suppose.
</blockquote></details>

No, I think these are correct. 
"Refactor Line routes from table to list and rename" has this file created, since we create the spec there, and in the "Add accessibility attributes to route list components" we add the aria-controls and id attributes, so they are updated to the snap there.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)* status: 15 of 23 files reviewed, 4 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/LineRouteList.tsx` line 39 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3HWgb5AGzVdBs7IS4U:-O3H_G6F6E6zQ9aE2Pny:b-r0r2jt) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/2fa9c68cd1d4529305938f55ec7a1fdcb3b45e5e/ui/src/components/routes-and-lines/line-details/LineRouteList.tsx#L39)):*

Nice. Always good to get away from hacky things like this :fire:
</blockquote></details>

Exactly :ok_hand:

niknii-vincit:
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 116 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMFch8OWTApcZwkjL:-O3HqaGu5dG2aS-yknsL:b-hj0sm2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L116)):*

Yeah, good one, forgot about this one. So the reason why I removed it was, because I doubted it had the correct text, since this is the *lastEdited* field. @niknii-vincit Do you know more about this?
</blockquote></details>

Hmm, I think there was some issue that the icon has `aria-hidden` :thinking: So that way it was the text AND the icon.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)* status: 14 of 23 files reviewed, 1 unresolved discussion (waiting on @Leitsi and @niknii-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 116 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMFch8OWTApcZwkjL:-O3HxexI3aHFskOiByDW:b6led4a) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L116)):*

Hmm, I think there was some issue that the icon has `aria-hidden` :thinking: So that way it was the text AND the icon.
</blockquote></details>

Ahh, so it was for the icon? But what I meant was that the tooltip text was probably wrong because this field is the lastEdited field, so shouldn't the tooltip say "last edited" instead of "version history"? Or at least I don't understand how I should interpret the "version history" tooltip here.

niknii-vincit:
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 116 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMFch8OWTApcZwkjL:-O3I6V_oB69YeENj4NuP:bpb4va2) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L116)):*

Ahh, so it was for the icon? But what I meant was that the tooltip text was probably wrong because this field is the lastEdited field, so shouldn't the tooltip say "last edited" instead of "version history"? Or at least I don't understand how I should interpret the "version history" tooltip here.
</blockquote></details>

Yeah good point. Wasn't the icon supposed to become a button at some point? Now I remembered there might have been some issue with the MDHistory element, that it somehow messed with the title attribute. Can't remember for certain. 

Anyway, I agree with you. Maybe the dates could have "last edited" and the icon the "version history", but could you try to put it, if it works in the MdHistory component?

niknii-vincit (commented):
> Reviewed 8 of 8 files at r5, 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)* status: all files reviewed, 2 unresolved discussions (waiting on @Leitsi)

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)* status: all files reviewed, 2 unresolved discussions (waiting on @Leitsi and @niknii-vincit)

___
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 116 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMFch8OWTApcZwkjL:-O3WGS288DqyUaL6VaX5:b3ur4vh) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L116)):*

Yeah good point. Wasn't the icon supposed to become a button at some point? Now I remembered there might have been some issue with the MDHistory element, that it somehow messed with the title attribute. Can't remember for certain. 

Anyway, I agree with you. Maybe the dates could have "last edited" and the icon the "version history", but could you try to put it, if it works in the MdHistory component?
</blockquote></details>

Ah okay the button makes sense. So if it's supposed to have a functionality then I think its best if we add the tooltip at the same time we add the functionality, otherwise it is just confusing.

Leitsi (approved):
:lgtm:

> Reviewed 7 of 8 files at r5, 2 of 2 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-:-O3fuDi11LUtna_-tSXQ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @niknii-vincit)

niknii-vincit (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

___
*[`ui/src/components/routes-and-lines/line-details/RouteRow.tsx` line 116 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-O3DMFch8OWTApcZwkjL:-O3gVYRF46U2VCcLsMWq:b-6ei8hx) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/d11db3d9a877060fbb8cdb9836b6678511e10f6e/ui/src/components/routes-and-lines/line-details/RouteRow.tsx#L116)):*

Ah okay the button makes sense. So if it's supposed to have a functionality then I think its best if we add the tooltip at the same time we add the functionality, otherwise it is just confusing.
</blockquote></details>

Well it also works as a alt text for the icon, so it kinda needs to be there, but kinda not you know :sweat: That icon is meaningful, IMO, because it implicates the date before it is "last edited".

It either caused some weird issue, or perhaps I didn't want to create a `<CustomMdHistory>` component, just to give it a title, or something like that :sweat_smile:

niknii-vincit (approved):
:shipit: 

> Reviewed 5 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/877#-:-O3mAX5-1lg6284hQpAq:bizlbj6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #896, opened 2024-08-30
Fix unauthorized map login loop

HenrikHartiala:
The login loop issue is actually a two fold:
We have protected routes from unauthorized users by redirecting them to front page for login,
in case the userInfo can't be retrieved. But we also have a "temporary quality of life" addition so
that if a GQL request fails, we reload the page, which is added so that when developers are stopping
and starting the dependencies, they lose the authorisation from the auth service, but the userInfo
is still there, so the pages are accessible, but the GQL's return nothing, and with the reload it actually
fetches the userInfo again -> not authorized -> redirect to front page.

But since mapModal is not an individual page and is in fact a modal and visible anywhere if the queryString
'mapOpen=true' is present, and if the map is open on top of the root page so localhost:3300?mapOpen=true
we were not redirecting the user to front page, because front page does not require auth. But the temp QoL
causes the page to reload -> So in the end the page reloads -> sends GQLs -> gets unauthed -> reloads..

The second part is that even if we removed the "reload QoL addition" in client.ts which would fix the loop,
it would still mean that the map can be opened on top of front page, it would just have an infinite loader spinning
and no data. So probably best to "protect" it as well for now, before we get some better solutions.

- This change broke some e2e tests, because we had some random network waits.
  - Removed those waits since they are not good practice anways.
- Adjusted the click points on tests to be more far away from the loading indicator.
  - The original click now clicked on top of the loading indicator which broke the tests. Optimally the
  loading indicator could be waited here, but there is still some inconsitencies with the loading indicators and
  not part of this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/896)

### DIFF
--- a/cypress/e2e/createStop.cy.ts
+++ b/cypress/e2e/createStop.cy.ts
@@ -51,7 +51,7 @@ describe('Stop creation tests', () => {
     cy.mockLogin();
 
     mapModal.map.visit({
-      zoom: 13,
+      zoom: 16,
       lat: 60.164074274478054,
       lng: 24.93021804533524,
     });
@@ -68,8 +68,8 @@ describe('Stop creation tests', () => {
           priority: Priority.Standard,
         },
         clickRelativePoint: {
-          xPercentage: 43.5,
-          yPercentage: 53,
+          xPercentage: 40,
+          yPercentage: 55,
         },
       });
 
@@ -100,8 +100,8 @@ describe('Stop creation tests', () => {
           priority: Priority.Standard,
         },
         clickRelativePoint: {
-          xPercentage: 50,
-          yPercentage: 45,
+          xPercentage: 40,
+          yPercentage: 55,
         },
       });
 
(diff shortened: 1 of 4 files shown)


Huulivoide (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/896)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

Leitsi (approved):
:lgtm:

> Reviewed 1 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/896#-:-O5mjO567f8FvbIw67Vg:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #897, opened 2024-08-30
Move MapLoader to MapModal component

HenrikHartiala:
The map is not it's own page and is in fact a modal/dialog. So if you "click outside of the dialog" it will close, and because the map loader was not inside the map modal component, it is considered as "outside of dialog" and if you click the loader, the map will close.

Don't know or can't remember why map loader was outside of map modal component in the first place, but it probably should be inside anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/897)

### DIFF
--- a/ui/src/components/map/MapModal.tsx
+++ b/ui/src/components/map/MapModal.tsx
@@ -4,6 +4,7 @@ import { useMapQueryParams } from '../../hooks';
 import { Map } from './Map';
 import { MapFooter } from './MapFooter';
 import { MapHeader } from './MapHeader';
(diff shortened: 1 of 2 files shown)


Huulivoide (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/897)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #898, opened 2024-08-30
Fix url pathname from disappearing when opening the map

HenrikHartiala:
useUrlQuery's setMultipleParametersToUrlQuery's pathname was defaulted to a an empty string when it should've been undefined. Under the hood it uses useNavigate hook and an empty value is a valid value for the navigate function and it will empty the pathname.
So we change the default value to undefined, which results in no changes to pathname.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/898)

### DIFF
--- a/ui/src/hooks/urlQuery/useUrlQuery.ts
+++ b/ui/src/hooks/urlQuery/useUrlQuery.ts
@@ -141,7 +141,7 @@ export const useUrlQuery = () => {
     ({
       parameters,
       replace = false,
-      pathname = '',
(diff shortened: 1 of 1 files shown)


Huulivoide (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/898)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @HenrikHartiala)

PR #913, opened 2024-09-20
E2E: Properly check GGL network requests succeed

Huulivoide:
Unless there is something wrong with he GraphQL server or an error occurs on a proxy/gateway level, the network response's status code will always be 200.

So:
```
  cy.wait('@gqlSaveMyThing')
    .its('response.statusCode)'
    .should('equal', 200)
```
Does not actually check that `myThing` was properly saved, the saving could have actually failed. Instead the correct thing to do, is to check that the Body of the Response does not contain any values in the `errors` field.

Also, the example above has the added problem, that `should` simply implies a retry check. So even if a lower level network error was to occur and `statusCode` was set to 500, cypress would just continue to monitor for another call for the same Query, until it finds one that succeeds. Normally there will ever be only a single network call, and thus Cypress will simply timeout and say that it newer saw any network requests with status of 200.

With this change Cypress will capture the 1st instance of the requested network call, and then continues to check that it was a truly successful call, without any low level errors nor any GraphQL level errors in it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/913)

### DIFF
--- a/cypress/e2e/editRoute.cy.ts
+++ b/cypress/e2e/editRoute.cy.ts
@@ -14,6 +14,7 @@ import {
 } from '../pageObjects';
 import { UUID } from '../types';
 import { SupportedResources, insertToDbHelper } from '../utils';
+import { expectGraphQLCallToSucceed } from '../utils/assertions';
 
 describe('Route editing', () => {
   let editRoutePage: EditRoutePage;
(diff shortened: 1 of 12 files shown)


culka (approved):
:lgtm:

> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/913#-:-O7TGH8X8irqMg_pen6B:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Huulivoide)

PR #921, opened 2024-10-01
Html lang attr

Huulivoide:
Add react-helmet-async to handler lang attr in <html>
Also handles <title> and <link rel="icon">

This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/921)


niknii-vincit:
What is the benefit of adding a dependency for this?

I've thought that the `lang` attribute and `title` are pretty simple things to update without extra dependencies 🤔

Huulivoide:
For the lang attr: not needed as such. But it does also allow us to replace next/head with the helmet in managing <title>, thus allowig us to drop next off as a useless complex piece of :apple: in the future. As we don't use it for anything else, as I said earlier when I joined the projects, it would make sense to dich Next in favour of a standard simple Vite config.

niknii-vincit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/921#-:-O8B5c2pDqHnJWgLzOo0:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved

PR #932, opened 2024-10-24
Utility updates

Huulivoide:
### Add helper utils to deal with numbered enums.
Number based Typescript enum definitions contain a reverse mapping of
the enum on the runtime object, making regular Object methods hard to
use on them for iteration.

Given enum definition: `enum MyEnum { A = 1, B, C }`
* `Object.entries` would produce:
  `[['A',1], ['B',2], ['C',3], ['1', 'A'], ['2', 'B'], ['3', 'C']]`
* `Object.keys: `['A', 'B', 'C', '1', '2', '3']`
* `Object.values`: `['1', '2', '3', 'A', 'B', 'C]`

This commit introduces helper functions that exclude the reverse
mapping from the entries/values/keys methods.

* numberEnumEntries: Get properly typed array `Array[EnumKey, EnumValue]>`
                     of enum entries.
* numberEnumValues: Get array of the actual enum values `Array<EnumValue>`
* numberEnumKeys: Get array of the actual enum keys `Array<EnumKey>`

### Add `areEqual` utility - dispatches to obj's own equals methods
Lodash's `isEqual` which performs a simple deep comparison of objects
based on their own enumerable properties. But some times classes
provide their own semantics on equality. For example Luxons DateTime
does so and in some cases `a.equals(b)` !== `isEqual(a, b)`.

This helper cheks the objects being compared for custom
implementations of `equals` funktion and dispatched the compartion to
that function instead.

### Add useTypedUrlState hook for better URL state management

Instead of random hooks nested in multiple components within the
same page, all poking, reading and parsing the URL on each render
cycle, provide a single unified useState -like hook.

### Problems with the existing useUrlQuery hook:

* Not typed, on multiple levels:
  - No typing on info on the over shape of the collection of params
    present in the URL. Individual fields are referenced to with
    just random strings.
  - Individual fields are not typed, instead every read to the param
    can happen under different type of "fetcher" get-function.
* Does not provide an elegant API. Instead of providing all of the
  params as plain values out of the hook, each field needs an explicit
  read call to be made.
* Has limited set of supported data types. Could be expanded, but
  each addition changes the API.
* Can cause duplicate work: Hook recommends a coding pattern where the
  hook is always called at the use site, in each individual sub
  component. This can cause the same URL params to be read and parsed
  multiple times within one render cycle of a single component.
  - This is a bigger problem overall on the code base as whole. Same
    complex hook chains that might perform network calls and complex
    data transformation are called from multiple levels on a single
    page. Instead it would be better to propagate the data down trough
    component props or with the help of Contexts. Also the hooks
    should be split into smaller pieces so that functionality A can be
    used where needed, without having to also pull in B, C, D, and E.
* Can cause problems with synchronization of state between
  components. As each field requires an explicit read, if Component A
  writes to field F, and the same field is used in Component B, B must
  contain proper logic to be able to correctly react to the
  change. This can especially cause a bug if the value or some
  derivative of it is stored in useState slot.
* Also based on above point, even local updates might require an
  explicit reread of the URL params after an update. As the component
  does not necessarily automatically react to the change of the fields
  value.
* Does not provide referential stability. Each read of a complex
  object (DateTime, Array) outputs a new instance, rendering the use
  of memoization null.
* Does not allow for "Navigate to other page with URL params" type of
  navigation. Instead that needs to be handled manually

### Pros of the existing useUrlQuery hook:

* Allows for very easy updates to be made to the URL params, as no
  state needs to be passed down from parent components.
* Allows for very easy read of the URL Params, as no state needs to be
  passed down from parent components.

### Points on the new useTypedUrlState hook:

* Single source of truth:
  - Multiple calls to the hook are not supported within a single page.
    Instead the state, full or partial, must be passed down to
    children trough the use of component props or React contexts.
    It is more than likely that the children need only a part of the
    complete URL Params, and even then it is likely to be a derived
    version of the data they need, and not the actual raw value, thus
    the data can be refined as it being passed down the component
    tree.
  - While not initially provided, it is trivial to construct a
    `Context<[StateT, Dispatch<SetStateAction<StateT>>]>` context,
    that is provided and the top level of the page, and which then
    allows for full easy access to the URL state with
    `useContext(pageAUrlState)` hook call.
  - Provides proper synchronization: Even if the URL state gets somehow
    changed externally, a call through the old useUrlQuery for example,
    the state gets properly synchronized.
* Mandates the field names to be declared beforehand, only providing
  access to known fields. Thus providing typings to the shape of the
  collection of URL params.
* Mandates each field to have a predefined and fixed type. Allowing
  for custom serializers and deserializers for each field.
* Provides a simple useState -like API:
  ```typescript
  const [urlState, setUrlState] = useTypedUrlState(...);
  const onReset = () => setUrlState(defaultValues);
  const onToggleA = () => setUrlState(p => ({ ...p, a: !p.a }));
  ```
* Default values:
  - Prefers to keep the URL clean. By default the search string should
    always be empty, but we should still have a proper state available
    in the code.
  - Just like useState takes in an initiaValue -param, useTypedUrlState
    takes in defaultValues -param.
  - Any field not present in the URL Params, gets its value from the
    default values object.
  - Unlike useState, default values must be a static object and cannot
    contain dynamic data. useTypedUrlState should always be called on
    the top level of the page's component tree, thus there should not
    event be any other data to derive the state from.
  - If a "dynamic initial" value is needed, that should be provided
    on initial URL during a page navigation.
  - Can provide a single source of truth for handling "unset" fields.
* Diffs the state, only including those fields in the URL that have
  values different from the default values.
* Prefers replacing the URL state, instead of pushing an state change
  into the navigation history. Changing the values on a form input,
  should not cause the browser history be polluted by million state
  changes. If a new entry is needed in the back navigation history,
  then an explicit page navigation needs to be made.
* Can, but should not, coexist with useUrlQuery, as long as data
  injected into the URL params is compatible.

### Use Tailwind classes for ListboxButton
* Drop open prop. Replaced with tailwind open-state classes.
* Drop disabled. Handled internally by HeadlessUI. Replaced with
  disabled-state classes.
* General cleanup of CSS class declarations.

### Allow SimpleButton to be used as raw form submit/reset button
Both href and onClick are optional.

### Allow InputLabel to be used as Listbox.Label implementation
Simply pass in ForwardedRef.

### Split buildSearchStopByLabelOrNameFilter from buildSearchStopsGqlQueryVariables
Allows for better reusability, without having to specify full set of
filters.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/932)

### DIFF
--- a/ui/src/components/forms/common/InputLabel.tsx
+++ b/ui/src/components/forms/common/InputLabel.tsx
@@ -1,4 +1,5 @@
 import get from 'lodash/get';
+import { ForwardedRef, forwardRef } from 'react';
 import { FieldValues, Path, useFormContext } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { TranslationKey } from '../../../i18n';
@@ -9,11 +10,10 @@ interface Props<FormState extends FieldValues> {
   translationPrefix: TranslationKey;
 }
 
-export const InputLabel = <FormState extends FieldValues>({
-  className = '',
-  fieldPath,
-  translationPrefix,
-}: Props<FormState>): React.ReactElement => {
+const InputLabelImpl = <FormState extends FieldValues>(
+  { className = '', fieldPath, translationPrefix }: Props<FormState>,
+  ref: ForwardedRef<HTMLLabelElement>,
+): React.ReactElement => {
   const { t } = useTranslation();
   const {
     formState: { errors },
@@ -22,7 +22,11 @@ export const InputLabel = <FormState extends FieldValues>({
   const hasError = !!get(errors, fieldPath);
 
   return (
-    <label className={className} htmlFor={`${translationPrefix}.${fieldPath}`}>
+    <label
+      className={className}
+      htmlFor={`${translationPrefix}.${fieldPath}`}
+      ref={ref}
+    >
       {/* Regex removes dot and series of numbers eg. ".10"
           This is needed for translation to work with fields that are
           created with React Hook Form useFieldArray */}
@@ -31,3 +35,9 @@ export const InputLabel = <FormState extends FieldValues>({
     </label>
   );
 };
+
+export const InputLabel = forwardRef(InputLabelImpl) as <
+  FormState extends FieldValues,
+>(
+  props: Props<FormState> & { readonly ref?: ForwardedRef<HTMLLabelElement> },
+) => ReturnType<typeof InputLabelImpl>;
--- a/ui/src/components/forms/stop-area/SelectMemberStopsDropdown/useFindStopPlaceByQuery.ts
+++ b/ui/src/components/forms/stop-area/SelectMemberStopsDropdown/useFindStopPlaceByQuery.ts
@@ -4,10 +4,7 @@ import {
   StopsDatabaseStopPlaceNewestVersionBoolExp,
   useFindStopPlacesByQueryAndGroupQuery,
 } from '../../../../generated/graphql';
-import {
-  AllOptionEnum,
-  buildSearchStopsGqlQueryVariables,
-} from '../../../../utils';
+import { buildSearchStopByLabelOrNameFilter } from '../../../../utils';
 import { stopAreaMemberStopSchema } from '../stopAreaFormSchema';
 
 const LIMIT = 20;
@@ -44,12 +41,7 @@ function buildMemberStopGqlFilter(
   query: string,
   editedStopAreaId: string | null | undefined,
 ): StopsDatabaseStopPlaceNewestVersionBoolExp {
-  const baseWhere = buildSearchStopsGqlQueryVariables({
-    elyNumber: '',
-    municipalities: AllOptionEnum.All,
-    // Automatic wild card as this is an "autocomplete" component.
-    labelOrName: `${query}%`,
-  });
+  const baseWhere = buildSearchStopByLabelOrNameFilter(`${query}%`);
 
   const notInGroup: StopsDatabaseStopPlaceNewestVersionBoolExp = {
     _not: { group_of_stop_places_members: {} },
--- a/ui/src/hooks/index.ts
+++ b/ui/src/hooks/index.ts
@@ -21,6 +21,7 @@ export * from './usePagination';
 export * from './useRequiredParams';
 export * from './useShowRoutesOnModal';
 export * from './useToggle';
+export * from './useTypedUrlState';
 export * from './vehicle-schedule-frame';
 export * from './vehicle-service';
 export * from './via';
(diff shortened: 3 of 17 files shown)


culka (approved):
> Reviewed 3 of 3 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/932)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Huulivoide)

niknii-vincit:
*[`ui/src/hooks/useTypedUrlState.ts` line 54 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/932#-OAHOEoO5Zjj1kviZc7b:-OAHOEoO5Zjj1kviZc7c:b-onmq4g) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/e72ff8f3b4e0ad025968428b44282c7bbb1bd85d/ui/src/hooks/useTypedUrlState.ts#L54)):*
> ```XML
> };
> 
> // Throw on dev mode and log warning on prod, with the expectation that the
> ```

Nit I would reword this as "On dev mode throws error, otherwise (in prod) logs a warning".

niknii-vincit:
*[`ui/src/hooks/useTypedUrlState.ts` line 145 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/932#-OAHTDLd3NTYBShe7Gmt:-OAHTDLd3NTYBShe7Gmu:b-qa479s) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/e72ff8f3b4e0ad025968428b44282c7bbb1bd85d/ui/src/hooks/useTypedUrlState.ts#L145)):*
> ```XML
>  * or fallback on to the default value of the field.
>  *
>  * In Dev mode requires the known params to be valid ans deserializable.
> ```

I hate being "that guy", but typo here



_Code quote:_
```XML
ans
```

Huulivoide (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/932)* status: 5 of 17 files reviewed, 1 unresolved discussion (waiting on @culka and @niknii-vincit)

___
*[`ui/src/hooks/useTypedUrlState.ts` line 145 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-ui/932#-OAHTDLd3NTYBShe7Gmt:-OAHYCdHB3bqIpWVaZHM:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-ui/blob/e72ff8f3b4e0ad025968428b44282c7bbb1bd85d/ui/src/hooks/useTypedUrlState.ts#L145)):*

I hate being "that guy", but typo here
</blockquote></details>

Done.

Huulivoide:
E2E testi vaan jääneet jumiin tunnettuun "Info spotteja, ei sortatat → Tulee satuunaisessa järjestykessä ruudulle" ongelmaan. EN aja testejä uudelleen.

PR #939, opened 2024-11-06
Bug fixes and improvements

Huulivoide:
### Cleanup ui/src/types/index.ts
Move actual definitions to sub files, as the index file should not
have any definitions in itself.

### useTypedUrlState: Fix URL synching and Luxon comparison
* Fix expectedSearchRef to include the leading `?` char, so that it
  matches the `search` provided by react-routers `useLocation`.
* areEquals method was originally added in commit [a3c3823], so that
  it could then be used here, and elsewhere. But I had forgotten to
  actually make use of it here. -> Replace `isEqual` with `areEquals`
  in the serialization function, to properly support Luxon DateTimes.

### Fix openOnMap -hooks to not mutate URL on call
`useOpenStopOnMap` and `useShowStopAreaOnMap` -hooks rely on
`useObservationDateQueryParam` -hook to provide the currently active
observationDate for `openMapWithParameters` function. By default
`useObservationDateQueryParam` performs a page navigation to set the
observationDate to the default value if it not already present in the
current page's search string.

`useOpenStopOnMap` is also present on the StopSearch StopTableRow
component, which results in the URL being manipulated in loop, each
loop rewriting the previous iterations query param with new version,
once for each row present in a initial page render of an stop search
result table.

Fixed the hooks, to only read in the current observationDate, never to
set it during render.

### Refactor Pagination component
A more generic externally controlled Pagination component has been
extracted from the previous Pagination component. A new CompatPagination
component has been added in, that wraps around the new Pagination
component, providing it with usePagination-hook based state.

Some functionality has been extracted out of usePagination hook into
another file, so that it can be used in the new Pagination component.
usePagination API has not been changed, it just re-exports some of the
functions previously defined within it.


suvikankare (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/939#-:-OB4b6noCDhpxxr62QCu:bnfp4nl)* status: 0 of 24 files reviewed, all discussions resolved

PR #965, opened 2024-12-13
Fix KeyHelpers.NonNullableKeys typings

Huulivoide:
* Changed all RequiredKeys, OptionalKeys and NonNullableKeys typings to omit the requested keys 1st. No effect on the type, but makes logically more sense to drop the fields (mark then with type never) and then override those fields, than to 1st make the changes and then merger them with never types.

* NnnNullableKeys did not do what it's docstring said it would. Previously any optional fields were passed trough as optionals, essentially typing the field as having `undefined` as one of the field value types. Now the optional marker is removed and the value in question has to truly be a nonnullable value. `value !== null && value !== undefined`

* As a result of fixing the NonNullabelKeys type, some other types derived with it started breaking code. Stop UPDATE-operation related code has been fixed to allow partial updates, and to not require all of the non-null DB fields. INSERT-operations still use the proper no-nulls type.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/965)

### DIFF
--- a/ui/src/components/forms/stop/StopForm.tsx
+++ b/ui/src/components/forms/stop/StopForm.tsx
@@ -11,7 +11,7 @@ import {
   ReusableComponentsVehicleModeEnum,
   ServicePatternScheduledStopPoint,
 } from '../../../generated/graphql';
-import { ScheduledStopPointSetInput } from '../../../graphql';
+import { PartialScheduledStopPointSetInput } from '../../../graphql';
 import {
   CreateChanges,
   EditChanges,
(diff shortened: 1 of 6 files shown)


suvikankare (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/965)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Huulivoide)

PR #984, opened 2025-01-27
StopPoint StopPlaceRef changes

villepynttari:
Modified saving stop point to set stop place ref from quay instead of stop place (only ref change, no other query changes in this PR)

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/984)


villepynttari:
Wrong brance

PR #1058, opened 2025-04-10
Small improvements and changes made as part of Stop Form revamp

Huulivoide:
### Handle programmatic map view state transitions
* MapLibre: Subscribe to MoveEnd event. Non animated state changes do not trigger a Move event, only MoveStart and MoveEnd.
* E2E: Launch the browser with Prefer reduced motion option enabled. This disables map animations and simply jumps the map straight to the new view state. Should reduce map related test flakyness.

### Update global CSS styles
* Set textarea to have min-height of 44px instead of forcing the height prop to 44px. input elements still have hard height of 44px.

* Make legend element have identical style as label.

### Preemptively add StopRegistryNamedEntitity form bits.
These were still needed in the initial iteration of for the new Stop creation modal, but were later dropped from the design. But these will be needed when we update the StopArea cration modal in the future. So as to not lose this already written valid code to /dev/null, the files are here now.

### Update gis utility functions
* Allow limiting precision in mapLngLatToPoint.
* Add function overload to getGeometryPoint for better typings.

### Add RequiredNonNullableKeys helper type
Makes the requested keys of the type truly required:
  - Not optional
  - Not null
  - Not undefined

### E2E: Improve map/createRoute tests
Previously the form was filled in pieces: base details and then the priority and validity period were filled in manually afterwards. But the code in charge of filling in the base details also tries to fill in the priority and validity periods. This split resulted in the endDate/indefinite filds to flicker between states, sometimes resulting in test failures.

Now the form is filled-in in one swell swoop.

### Improve the generic Modal component on the Map view
* Allow injecting custom CSS classes to the Header, Body and Footer sections of the modal.
* Allow Heading to receive a complex ReatNode for content as the title "string".
* Instead of having the full content of the modal be scrollable, only apply scrolling to the Body section and keep the Header and Footer always visible.

### Improve <LineRenderLayer> component
* Add support for more paint properties.
* Make types readonly.
* Hoist const data to top level.

### Improve Stop copying
Coder's previous assumption was that one could alter the Quays included in a StopPlace by simply altering the contents of quays field in the mutation payload. But that is not case. Instead on Tiamat's end the quays field is always populated from the DB, and then any new quays are added from the mutation payload or existing ones changes, if there is a QuayInput fragment with a matching ID in the payload. Thus, omitting a Quay from the update payload, does not result in it getting deleted from StopPlace.

Instead to delete a Quay from a StopPlace a special `deleteQuay`
mutation must be called.

### DIFF
--- a/cypress/cypress.config.ts
+++ b/cypress/cypress.config.ts
@@ -1,6 +1,6 @@
 import { defineConfig } from 'cypress';
 import * as tasks from './e2e/utils/tasks';
-import { onLaunchBrowser } from './support/setHeadlessScreenSize';
+import { onLaunchBrowser } from './support/launchBrowser';
 
 // eslint-disable-next-line import/no-default-export
 export default defineConfig({
--- a/cypress/e2e/map/createRoute.cy.ts
+++ b/cypress/e2e/map/createRoute.cy.ts
@@ -96,18 +96,11 @@ describe('Route creation', mapViewport, () => {
           swedishName: 'Test destination SWE',
           swedishShortName: 'Test destination SWE shortName',
         },
+        priority: Priority.Standard,
+        validityStartISODate: '2022-01-01',
+        validityEndISODate: '2025-12-01',
       });
 
-      mapModal.routePropertiesForm.changeValidityForm.setPriority(
-        Priority.Standard,
-      );
-
-      mapModal.routePropertiesForm.changeValidityForm.validityPeriodForm.setStartDate(
-        '2022-01-01',
-      );
-      mapModal.routePropertiesForm.changeValidityForm.validityPeriodForm.setEndDate(
-        '2025-12-01',
-      );
       mapModal.editRouteModal.save();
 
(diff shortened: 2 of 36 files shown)


suvikankare (approved):
> Reviewed 3 of 3 files at r1, 6 of 6 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 4 of 4 files at r5, 3 of 3 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, 1 of 1 files at r11, 1 of 1 files at r12, 1 of 1 files at r13, 3 of 3 files at r14, 1 of 1 files at r15, 2 of 2 files at r16, 1 of 1 files at r17, 1 of 1 files at r18, 6 of 6 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1058)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Huulivoide)

PR #1168, opened 2025-08-22
Clean and merge Airbnb ESLint config with our own + Prepare our conf for FlatConfig

Huulivoide:
### ESLint: Replace Airbnb config with local copy
Rule loading ordering should be mostly the same with the new version, but one rule eluded this coder: import/no-named-as-default: Aka, don't import default and call functions trough that if the function is export on it's own too.

Wrong: `import qs from 'qs'; qs.stringify();`
Right: `import { stringify } from 'qs'; stringify();`

Few eslint-ignore comments have been added to deal with that.

### ESLint: Prune and clean Airbnb config
Removed rules for:
- Rules set to 'off', but which aren't even enabled in the "recommended" rulesets of the other ESLint plugins.
- Rules set to 'error', but which are already enabled by the base "recommended" rulesets of the other plugins.
- Rules already removed or deprecated versions of rules in previous versions of ESLint or one of the related plugins (React).
- Style based rules that would get disabled by ESLint Prettier config.

React-a11y:
- Removed unneeded rule definitions, that duplicated the "recommended" config ruleset.
- Simplified rule configurations → Removed parameters that were set to the default value of said parameter.
- Dropped couple of minor rules, that had ever so subtle differences with the recommended rule set, likely these were copypasta🍝 from a previous version of the plugins recommended ruleset.

Our own custom rule definitions have not been taken into account yet. A merge of those will come later.

### ESLint: Cleanup & create a single unified config
* Prepare for Flat config:
  - All ESLint deps moved to the top level package.json dev-deps section.
  - Dropped eslint-config-next package, as that was not used.
  - Sub configs merged to the single top level config.
  - Single catch all setup block replaced with override blocks targeting specific subset of files needing different configs:
    * UI (ts + React)
    * Jest based unit/integration tests (ts + React + Jest + node)
    * Localization files
    * Cypress (ts + node + cypress)
    * Test DB Manager (ts + node)

* Our own rule overrides previously found in the top level .eslintrc.js file have been merged into the "airbnb" config and the overridden rule definitions have been dropped.

* Renamed the "airbnb" config simply to rules, as it now represents our enabled rule selection/configuration.

* Minor Cleanup to cypress codes due to improved ESLint config.

* Reverted some eslint-disable comments which were added in the original [87bcf8c](Replace Airbnb config with local copy) commit. Rule ordering is fixed again.

* Re-enabled rule react/jsx-no-useless-fragment as per react's recommended eslint ruleset. That rule had previously been turned off, because previous developers did not know how to type React components properly.

[87bcf8c]: https://github.com/HSLdevcom/jore4-ui/commit/87bcf8caf03343e03034c7f15dd741d3f410d4d0

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1168)


suvikankare (approved):
> @suvikankare reviewed 35 of 35 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1168)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Huulivoide)

PR #1418, opened 2026-04-29
Assorted fixes

Huulivoide:
### Reacy: Fix improper/missing key props

- ExternalLinks: Fix key prop to use proper id in edit form.
- StopsByLineSearchGroupedStopsResults: Add missing key prop.

### Apollo: Drop manual cache item eviction code

There were several problems with the code sections removed:
* Some trying to remove wrong type of an cache object: Made changes to DB entity type A, but dropped item of type B from cache.
* Some were giving in undefined ids, thus causing errors to be printed into the dev console.
* Some were just outright illogical, patching an object, returning and caching the updated data, and then removing the freshly updated item.
* There were couple of proper use cases (DeleteStop, DeleteRoute) where the eviction call was valid, but those have also been removed to let automatic garbage collection then handle that at a later point.

BUT!: Then there was a bug in the eviction code it self. `cache.evict` takes in an options parameter in the form of `{ id: thing to remove }`, but the `removeFromApolloCache` function incorrectly passed in the id string straight to the evict call. Type script did warn the original developer about this, but said developer (T. K.) ignored the warning and thought he knew better and suppressed said error with a handy dandy `@ts-expect-error` comment.

What that did do, was that it instead removing a single item from the cache, it reset the whole Apollo cache to an empty state, which in turn resulted in active queries being refetched from the database. Line details page was relying on this unconventional data refresh strategy to function correctly, so a proper data refetch has been added for that as a fix.

This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1418)

### DIFF
--- a/ui/src/components/map/queries/useResolveStopHoverTitle.ts
+++ b/ui/src/components/map/queries/useResolveStopHoverTitle.ts
@@ -24,7 +24,9 @@ type ResolveStopNameFn = (stopPlaceNetexId: string) => Promise<string | null>;
 function useResolveStopName(
   areas: ReadonlyArray<MapStopArea>,
 ): ResolveStopNameFn {
-  const [resolveStopNameFromDB] = useResolveStopNameLazyQuery();
+  const [resolveStopNameFromDB] = useResolveStopNameLazyQuery({
+    fetchPolicy: 'cache-first',
+  });
 
   return useCallback(
     async (stopPlaceNetexId: string) => {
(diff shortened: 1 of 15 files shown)


suvikankare (approved):
> @suvikankare reviewed 15 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1418)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Huulivoide).

PR #1419, opened 2026-04-30
Improve time parsing handling

Huulivoide:
### SubstitutePeriod: Fix types and interval parsing

Due to an oversight in the implementation of `extendHook` function, it was returning an function that took in `any` as it's parameters, thus loosing type info in between.

Then later in the code there ware expectations made about the input Substitution periods' shape, namely they were all expected to be occasional ones with custom begin and end times defined. But in reality they could have also been common periods as well, which default to the full time period (beginTime = null, endTime = null) in the db. As all periods were expected by the code to contain a valid time periods, those fields were directly fed into the Luxon.Duration parsing functions.

That can work, when the Luxon settings are set to not throw on invalid input and instead just invalid return garbage data, that eventually mapped to null when serializing the data on the transport layer.

### Improve time parsing handling
* Improve timezone handling when parsing date(time) strings:
  - Correctly set Date only values to Helsinki midnight DateTime objects.
  - Assume date time strings without an explicit timezone or offset info are in UTC time, and convert them to Helsinki time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1419)

### DIFF
--- a/cypress/e2e/stop-registry/stopAreaDetails.cy.ts
+++ b/cypress/e2e/stop-registry/stopAreaDetails.cy.ts
@@ -209,8 +209,7 @@ describe('Stop area details', { tags: Tag.StopRegistry }, () => {
     StopAreaDetailsPage.versioningRow
       .getChangeHistoryLink()
       .shouldBeVisible()
-      .invoke('text')
-      .should('match', /\d{2}\.\d{2}\.\d{4}\s+\d{2}:\d{2}/); // Matches format: DD.MM.YYYY HH:mm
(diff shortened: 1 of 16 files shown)


suvikankare (approved):
> @suvikankare reviewed 20 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1419)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Huulivoide).

PR #1430, opened 2026-05-26
Random changes

Huulivoide:
### Codegen: Improve Tiamat entity plugin

Instead of an JSON file, create a Typescript file instead, so we can easily extract the static type info from the data. Use said type info to replace a hard coded list in `ui/src/graphql/types.ts` file.

### Fix isArray handling for ReadonlyArrays.
A prior commit where interfaces were replaced with types, a single instance where interface is needed was also erroneously converted into a type, thus breaking it's effect.

### E2E: Improve StopRegistry DB reset
* Drop cleaning of GroupsOfStopPalaces, not used for Stop Areas or Terminals any more.
* Move Organisation cleanup to happen with the table truncate.
* Add cleanup for info spots.
* Hopefully fixes 'should allow entering manual spot & poster sizes' test flakyness.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1430)


suvikankare (approved):
> @suvikankare partially reviewed 41 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1430)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Huulivoide).

PR #1465, opened 2026-07-01
Map route creation code fixes

Huulivoide:
### Map: Fix DrawControl event handler references

`useControl` does not update the state in any manner. If the input functions for the event handlers were to change, those updates would not have been reflected on the control, and instead it would have simply called the version of the function it was initially given.

In addition an a teardown function has been added to ensure the event listeners get cleaned up on unmount.

### Map: Fix creating an Route with no end date.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1465)

### DIFF
--- a/ui/src/components/map/DrawControl.tsx
+++ b/ui/src/components/map/DrawControl.tsx
@@ -1,10 +1,11 @@
 import MapboxDraw from '@mapbox/mapbox-gl-draw';
-import noop from 'lodash/noop';
 import type { Map as MapBoxMap } from 'mapbox-gl';
 import {
   ForwardRefRenderFunction,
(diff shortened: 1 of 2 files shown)


patricvincit (approved):
:+1: 

> @patricvincit reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-ui/1465#-:-OwS1GAp-fJ9N5iXgnSQ:b-8fx2u7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Huulivoide).

